PR #10, opened 2024-05-15
Update advanced settings

PeterJhongLinksys:
1. Fix state not updated after save setting on changed connection type
2. Modify the save setting functions
3. Update the widgets.diff file


AustinChangLinksys (approved):
Looks good to me!

PR #12, opened 2024-05-16
Minor issues fixing

AustinChangLinksys:
- Fixed WiFi list doesn't update state after changes
- Fixed Ipv6 port service add/delete rules not working
- Update router password and hint UX
- Fixed internet to device speed check always disabled
- Fixed WiFi share view doesn't include guest WiFi


HankYuLinksys (approved):
Looks good!

PR #13, opened 2024-05-16
Update testcase because remove an item on the UI

AustinChangLinksys:
Update testcase because remove an item on the UI


PeterJhongLinksys (approved):
Looks good.

PR #16, opened 2024-05-16
save admin password after fetch device info and go to pnp setup view after check admin password passed

AustinChangLinksys:
save admin password after fetch device info and go to pnp setup view after check admin password passed


PeterJhongLinksys (approved):
Looks good.

PR #22, opened 2024-05-22
Fix router password hint display incorrect after set a new router password

AustinChangLinksys:
Fix the router password hint display incorrect after set a new router password hint.
Force fetch password hint from the router
Tested with OAK on local build


HankYuLinksys (approved):
Looks good!

PeterJhongLinksys (approved):
Looks good.

PR #23, opened 2024-05-22
Add show full screen spinner overlay function on shortcuts.

AustinChangLinksys:
Add show full screen spinner overlay function on shortcuts.
Add JNAP side effect error to handle unsuccess side effect case.
Update spinner when restarting router
Update dialog when confirm to restart router


HankYuLinksys (approved):
Looks good!

PR #31, opened 2024-05-27
Implement Firewall settings and locs

AustinChangLinksys:
Implement firewall settings and import locs.

Tested getting and setting scenarios in the local environment.


PeterJhongLinksys (approved):
Looks good to me.

PR #38, opened 2024-06-03
Feature manual fw update fixes

AustinChangLinksys:
- Add redirection provider
- Fixed setting bridge mode not redirect to propor url
- Fix modify local ip address not redirect to propor url
- Update side effect response and error handling
- Revise firmware update UI


PeterJhongLinksys (approved):
Looks good!

PR #40, opened 2024-06-03
[Fix] Bug fix

PeterJhongLinksys:
1. Fix region button disable on callback request page
2. Set night mode default to false


AustinChangLinksys (approved):
Looks good to me

PR #55, opened 2024-06-24
[Speed Test] Add background for speed test on running and complete views. Modify the animation frequency for smoothness.

PeterJhongLinksys:
Add background for speed test on running and complete views. Modify the animation frequency for smoothness.


AustinChangLinksys (approved):
Looks good to me

PR #59, opened 2024-06-25
UI apply column grid

AustinChangLinksys:
- Updated Spacing and Gap
- Applied column grid system
- Updated all pages' UI
- Added display column overlay shortcut
- Updated screenshot script to support display column overlay


PeterJhongLinksys (approved):
Looks good!!

PR #66, opened 2024-06-28
Fixed device filter display white block when node name is too long

AustinChangLinksys:
- Fixed device filter display white block when node name is too long
- Fixed unexcepted error when given a specific device size but the test case not supported


PeterJhongLinksys (approved):
Looks good.

PR #70, opened 2024-07-01
Fixed some locales can't be executed

AustinChangLinksys:
- Fixed some locales can't be executed
- Generate html testing report with localses and screens naming.


PeterJhongLinksys (approved):
Looks good.

PR #71, opened 2024-07-02
[SpeedTest] Fix broken column system on speedtest page

PeterJhongLinksys:
1. Fix broken column system on speedtest page
2. Add error view for speed test failed
3. Add test case of error view on localization testing for speed test view


AustinChangLinksys (approved):
Looks good to me

PR #78, opened 2024-07-12
[Add Node] Add refresh function on add node result view

PeterJhongLinksys:
Add refresh function on add node result view


AustinChangLinksys (approved):
Looks good to me

PR #88, opened 2024-07-22
[Testing] Localization testing for ineternet settings. Fix issue on subnet mask input in internet settings.

PeterJhongLinksys:
Localization testing for ineternet settings. Fix issue on subnet mask input in internet settings.


AustinChangLinksys (approved):
Looks good to me

PR #94, opened 2024-08-02
Fixed forcing logout when restarting router

AustinChangLinksys:
- Fixed forcing logout when restarting router
- Navigate support page to faq list


PeterJhongLinksys (approved):
Looks good!

PR #96, opened 2024-08-02
[Port Setting] Fix add/delete rules state sync issue.

PeterJhongLinksys:
Fix add/delete rules state sync issue.


AustinChangLinksys (approved):
looks good to me

PR #100, opened 2024-08-06
-fixed device/node details display signal strength if is wired connection

AustinChangLinksys:
- fixed device/node details display signal strength if is wired connection
- sorting orders by timezone value in timezone selection page


PeterJhongLinksys (approved):
Looks good.

PR #117, opened 2024-08-29
[Semantic Label] Add semantic label for icon button, icon, image, and ip/mac input field

PeterJhongLinksys:
Add semantic label for icon button, icon, image, and ip/mac input field


AustinChangLinksys (approved):
looks good to me

PR #120, opened 2024-09-06
Instant topology, instant verify rework

AustinChangLinksys:
- Rename key features
- Instant Topology rework
- Instant Verify rework
- Update test cases
- Add GetIt plugin
- Add ServiceHelper for checking service is support
- Node detail information rework
- Device detail information rework
- Printing function


PeterJhongLinksys (approved):
Looks good.

PR #121, opened 2024-09-07
[Instant Privacy] Apply the new design for Instant privacy.

PeterJhongLinksys:
1. Rename Mac filtering and apply the new design.
2. Remove the test case for mac filtering on wifi settings


AustinChangLinksys (approved):
look goods to me

PR #128, opened 2024-09-13
[NodeDetail] Node detail page rework

PeterJhongLinksys:
1. Node detail page rework
2. Merge the device list widget which used in node detail and device list page


AustinChangLinksys (approved):
looks good to me

PR #131, opened 2024-09-18
[Semantic Label] 1. Add semantic label for back button, close button, switch and checkbox

PeterJhongLinksys:
1. Add semantic label for back button, close button, switch and checkbox
2. Fix select device page issue


AustinChangLinksys (approved):
Looks good to me

PR #135, opened 2024-09-23
Fix admin password limitation

PeterJhongLinksys:
1. Not start with space or end with space
 2. Not have the same character twice in a row
 3. Not use unsupport special character
 4. Hint can not contain password


AustinChangLinksys (approved):
Looks good to me

PR #141, opened 2024-09-30
[NOW-104]: Instant-Admin: After changing, the time zone still displays the previous setting

PeterJhongLinksys:
[Root cause]
Fetch cache data not the latest.

[Soluiton]
Fetch the latest data after save changes.


AustinChangLinksys (approved):
looks good to me

PR #143, opened 2024-09-30
[Now-111] Reset password: Password tips collapse once leave password column.

PeterJhongLinksys:
[Solution]
Always display the tips on the reset password page


AustinChangLinksys (approved):
looks good to me

PR #144, opened 2024-09-30
[NOW-77]: Log in: Unknown error displayed when logging in with wrong password and then refreshing the page and logging in with correct password

PeterJhongLinksys:
[Root cause]
Did not handle the error response: errorPasswordCheckDelayed
    
[Solution]
Handle the error errorPasswordCheckDelayed, so that will not show unknown error to user.
Add getAdminPasswordAuthStatus to check the auth status


AustinChangLinksys (approved):
looks good to me

PR #149, opened 2024-10-07
[Deauth] Add disconnect client function to device page and node detail page

PeterJhongLinksys:
Add disconnect client function to device page and node detail page


AustinChangLinksys (approved):
looks good to me

PR #151, opened 2024-10-08
Fixing manual fw update

AustinChangLinksys:
- fixed restart poll data after router back
- UI update, to hide menu during updating and restarting
- Remove start/stop polling when tab lose focus
- Modify manual fw update flow
- refactor fetch ui version
- fixed manual fw update not align in middle and animations


PeterJhongLinksys (approved):
Looks good

PR #156, opened 2024-10-14
v1.1.1 bug fixing

AustinChangLinksys:
- NOW-144: Instant-Topology: The confirmation dialog should not show "Parent" text when reboot/factory reset the child node
- NOW-147: [Instant-Topology] "Blink Device Light" should be hidden when the node doesn't support the function.
- NOW-150: [Instant-Topology] "Node light" should be hidden/disable once the node doesn't support the function.
- NOW-145: [Instant-Admin] Click on Timezone, redirect to manual  firmware update page.
- NOW-143: Incredible-WiFi: Error message displays when turning off the WiFi band
- NOW-142: Incredible-WiFi: It is recommended to hide the password field when the security mode is Open/Enhanced Open Only/Open and Enhanced Open


PeterJhongLinksys (approved):
Looks good.

PR #157, opened 2024-10-14
Fixed NOW-147

AustinChangLinksys:
- NOW-147: [Instant-Topology] "Blink Device Light" should be hidden when the node doesn't support the function.


PeterJhongLinksys (approved):
Looks good.

PR #158, opened 2024-10-14
NOW-152: Instant-Safety: Blank page appears when updating safe browsing settings

AustinChangLinksys:
- auto stop polling when starting handling side effect
- fixed not showing router not found when router is disconnected


PeterJhongLinksys (approved):
Looks good.

PR #159, opened 2024-10-17
[NOW-161]: Menu page: The Instant-Safety status is not updated after changing DNS settings

PeterJhongLinksys:
[Root cause]
Didn't update instantSafety state

[Solution]
Update instatntSafety state after set lan setting


AustinChangLinksys (approved):
Looks good to me

PR #160, opened 2024-10-17
[NOW-160]: Instant-Safety: The feature name should display "Instant-Safety" instead of "Safe Browsing"

PeterJhongLinksys:
[Root cause]
Didn't update the string

[Solution]
Update the string.


AustinChangLinksys (approved):
Looks good to me

PR #161, opened 2024-10-17
[NOW-159]: Local Network: Clearing DNS server settings should not prompt the user for unsaved changes

PeterJhongLinksys:
[Root cause]
Did not call the setState funciton when update the latest state.
    
[Solution]
Update the state in the setState function.


AustinChangLinksys (approved):
Looks good to me

PR #164, opened 2024-10-21
- NOW-141: [Cherry 7_Corner case] Disable WiFi settings, user can't access WebGUI.

AustinChangLinksys:
Fixed checking LAN ports, if no LAN ports then the Main WiFi can not be disabled. The wireless local management option applied the same logic


PeterJhongLinksys (approved):
Looks good.

PR #165, opened 2024-10-22
Fixed a part of advanced settings issues

AustinChangLinksys:
- [NOW-178] DMZ: Error message displays when selecting destination target from DHCP client table then saving
- [NOW-179] DMZ: Unable to change Source IP Address setting
- [NOW-180] DMZ: Error message displays when changing destination IP address setting
- [NOW-181] Port Range Triggering: Should not display error message if the forwarded range's end port is equal to the start port
- [NOW-183] Port Range Forwarding/Port Range Triggering: Should not display error message when changing port setting


PeterJhongLinksys (approved):
Looks good.

PR #167, opened 2024-10-22
Fixed polling timeout issue

AustinChangLinksys:
- Extract GetWANExternal as along provider to avoid affecting core transaction.


PeterJhongLinksys (approved):
Looks good.

PR #168, opened 2024-10-22
[Fix] Fix after adding node, didn't show parent node oin list.

PeterJhongLinksys:
Fix after adding node, didn't show parent node on list.


AustinChangLinksys (approved):
Looks good to me

PR #170, opened 2024-10-25
Disable semantic label on Android and iOS os

AustinChangLinksys:
Add defaultPlatformTarget to the log


PeterJhongLinksys (approved):
Looks good.

PR #171, opened 2024-10-25
[Fix] issues

PeterJhongLinksys:
NOW-154: [Node details] Signal strength is not aligned.
NOW-88: [Instant-Verify] Pinging IP no response, no logs are printed.
NOW-175: [Instant-Verify] Traceroute details doesn't display completely.
NOW-155: [Mobile]Dashboard: Once the topology has more than 4 child nodes, the scroll bar is difficult to scroll.
NOW-173: [Advanced Settings] Cancel on connection type still keep unsaved change


AustinChangLinksys (approved):
Looks good to me!

PR #173, opened 2024-10-30
Fixed now-198 now-199

AustinChangLinksys:
- Fixed display error on ipv6 port service and port forwarding
- Add new method for check internet connection on Node


PeterJhongLinksys (approved):
Looks good.

PR #175, opened 2024-10-30
[NOW-194]: [Instant-Setup]: Static IP - DNS2 can't be deleted.

PeterJhongLinksys:
Fix state error.


AustinChangLinksys (approved):
looks good to me

PR #182, opened 2024-11-05
Fixed NOW-207

AustinChangLinksys:
- Revert check internet connection on PnP
- Add MAC and IP check on dhcp reservations
- Fixed error response when set wan type as PPTP or L2TP


PeterJhongLinksys (approved):
Looks good.

PR #191, opened 2024-11-22
Fix issues on local network and internet settings

PeterJhongLinksys:
[NOW-274]: Local Network: When changing the host name and adding a DH…CP reservation at the same time, the DHCP reservation is not saved

[NOW-224]: Local Network: It should prompte the user  that the router was not found after changing the DUT IP address

[NOW-278]: Local Network: The local IP address cannot be changed when there is a DHCP reservation

[NOW-275]: Local Network: Deleting DNS/WINS setting should not prompt the user for unsaved changes

[NOW-222]: CLONE - [UI]No show "http://LinksysXXXXX.local" for URL tip when WAN is under Bridge mode.

[NOW-293]: Internet Settings: Unable to edit MAC address field when MAC Address Clone feature is enabled

[InternetSettings] Fix issues on setting vlanId

### DIFF
--- a/lib/core/jnap/models/wan_settings.dart
+++ b/lib/core/jnap/models/wan_settings.dart
@@ -41,10 +41,12 @@ class RouterWANSettings extends Equatable {
     this.wanTaggingSettings,
   });
 
-  factory RouterWANSettings.dhcp({required int mtu}) {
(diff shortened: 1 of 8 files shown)


AustinChangLinksys (approved):
Looks good to me

PR #193, opened 2024-11-25
- import locs from legacy UI and Apps

AustinChangLinksys:
- import locs from legacy UI and Apps


PeterJhongLinksys (approved):
Looks good.

PR #203, opened 2024-11-29
[NOW-253] Fixed the incorrect number of displaying node status right after updating firmware versions

HankYuLinksys:


AustinChangLinksys (approved):
Looks good to me

PR #205, opened 2024-12-04
[NOW-253] more fix

HankYuLinksys:
Fixed: Only one master node status is displayed on the list view after updating the firmware version


AustinChangLinksys (approved):
Looks good to me

PR #209, opened 2024-12-05
Fix issues

PeterJhongLinksys:
[NOW-345]: Internet Settings: The domain name should not be editable when connection type is PPTP

[NOW-330]: [External speed test] Mobile user might can't click options on External Speed Test page

[NOW-240]: When adding a new node and a new node isn't found, do not display router.

[NOW-138]: [Instant-Devices] Filter criteria doesn't contain [By WiFi].


AustinChangLinksys (approved):
Looks good to me

PR #216, opened 2024-12-12
Now-361

PeterJhongLinksys:
1. [NOW-361]: [Advanced Settings_UX]Local Network_Change LAN IP to different subnet Ip from the DHCP server, [Save] button is disabled.

2. Show ipv6 address if there is no ipv4 address in instant device page

3. Hide the 'Reserve Ip' button if there is no ipv4 address in device detail page


AustinChangLinksys (approved):
Looks good to me

PR #221, opened 2024-12-18
NOW-349: [No Internet] Page with no functional banner after config

AustinChangLinksys:
- Mark all troubleshooting pages as noNaviRail = true


HankYuLinksys (approved):
Looks good!

PR #222, opened 2024-12-18
NOW-389: Instant-Admin: Cannot change time zone to a time zone that does not support daylight saving time

AustinChangLinksys:
- Set autoAdjustForDST as false if selected timeZone doesn't support it.
- Show snackbar for set timezone operation


HankYuLinksys (approved):
Looks good!

PR #225, opened 2024-12-19
[NOW-240]: When adding a new node and a new node isn't found, do not display router.

PeterJhongLinksys:
The added devices did not show in the result of getDevices. Modify the retry longer.


HankYuLinksys (approved):
Looks good!

PR #227, opened 2024-12-23
NOW-391: [PnP]The icon of connection method on child node is not correct.

AustinChangLinksys:
- Add check backhaul info after check nodes online
- Update wireless connection info on add nodes provide


PeterJhongLinksys (approved):
Looks good.

PR #232, opened 2024-12-24
[NOW-347] [NOW-390]

PeterJhongLinksys:
[[NOW-347]: [Enhance] Switch to Advanced tab under Incredible-WiFi doesn't show unsaved change dialog

[NOW-390]: Instant-Privacy: It's hard to scroll through the devices list on mobile phone


AustinChangLinksys (approved):
Looks good to me

PR #237, opened 2024-12-25
[NOW-253]

HankYuLinksys:
Fixed mismatches of firmware versions and updated status in displayed info after updating


AustinChangLinksys (approved):
Looks good to me

PR #261, opened 2025-02-05
NOW-434: GetWanSettings5 not being called

AustinChangLinksys:
- Implement sorting service function to sort service version
 - Add service sorting test cases


PeterJhongLinksys (approved):
Looks good.

PR #262, opened 2025-02-05
NOW-408: Instant-Admin: The "Router password hint" field goes out of the frame, and the "Cancel" and "Save" buttons overlap with the password requirements when the mobile virtual keyboard appears

AustinChangLinksys:
- Fixed minor UI issues
- Fixed MAC filtering state consists issue
- Fixed set router password dialog focused issue when virtual keyboard displayed


PeterJhongLinksys (approved):
Looks good.

PR #266, opened 2025-02-12
Fixed scanning QR code w/ password not working.

AustinChangLinksys:
- update url strategy to PathUrlStrategy
- Update wifi share icon
- add screenshot widget function
- add save bytes to file function
- add get full location path function


HankYuLinksys (approved):
Looks good!

PR #267, opened 2025-02-12
[NOW-436]

HankYuLinksys:
NodeType will be NULL but isAuthority will be True in factory settings. To avoid errors caused by empty node devices while entering the dashboard on an unconfigured router, add a check for isAuthority.

Deliberately changing passwords by executing specific JNAP API is not a normal use case. This is unlikely to happen to ordinary users.


AustinChangLinksys (approved):
Looks good to me

PR #272, opened 2025-02-21
NOW-450: [Incredible Wifi] Add device then disable WIFI MAC Filteringstill can click Save button; NOW-453: DMZ: If the end IP of the source IP address is equal to the start IP, the setting cannot be saved; NOW-454: Port range forwarding/Port range triggering: If the port/protocol of the rule does not overlap with other rules, the error message should disappear; NOW-455: [Enhancement] Too small QRCode can be hide.; NOW-457: Update MLO warning description

AustinChangLinksys:
- Improve horver QR code btn behavior.
- Fixs wrong error message display on DMZ, Port settings.
- Fixs behavior on MAC Filtering.
- Update MLO warning on Advanced WiFi Settings page.
- Fixs Instant Topology cannot enter if in bridge mode.


HankYuLinksys (approved):
Looks good!

PR #273, opened 2025-02-21
[NOW-445]

HankYuLinksys:
SuperTooltip widget will be shown unexpectedly even if the showTooltip function is not called. As a result, the tooltip will only be installed when the error exists


AustinChangLinksys (approved):
Looks good to me

PR #274, opened 2025-02-21
NOW-458: PNP setup: Enter the dashboard without doing FW update check and update if the WiFi name/password has changed

AustinChangLinksys:
- Fixes wrong steps when WiFi credentials changes on setup flow


PeterJhongLinksys (approved):
Looks good

PR #287, opened 2025-03-07
[NOW-468]

HankYuLinksys:
Added a new property pivotalIndex which indicates any changes on that cell will cause a recheck for all other cells.

https://github.com/user-attachments/assets/b3679db5-1abf-4b0d-8970-6986fa15a013


PeterJhongLinksys (approved):
Looks good.

PR #288, opened 2025-03-07
[NOW-477] Fixed the crashed screen on PNP setup pages

HankYuLinksys:
Root cause: The Lookbehind assertion in JS RegExp is not supported by Safari, below the iOS version 16.4
https://caniuse.com/js-regexp-lookbehind

As a result, the RegExp used for the SSID check is revised.


AustinChangLinksys (approved):
looks good to me

PR #291, opened 2025-03-12
NOW-472: Unused strings cleanup; NOW-469:DMZ: The Save button is still available after saving the setting that the end IP of the source IP address is equal to the start IP; ...

AustinChangLinksys:
- Bumped UI version to 1.1.7
- Cleanup unwanted strings
- NOW-479: [String update] under device actions, when selecting the last child in the chain for factory default, confirmation screen msg change from :" Yes Factory reset all" to "Yes, Factory reset"
- NOW-480: [String update] since AU and NZ regulation are aligned, replace AU with AU/NZ
- NOW-482: [String update] Update reboot popup message
- NOW-483: Remove WAN status on print function


PeterJhongLinksys (approved):
Looks good.

HankYuLinksys (approved):
Looks good!

PR #292, opened 2025-03-13
NOW-484: Cherry7_Feature request_longer screen timeout

AustinChangLinksys:
- Add whitelist for idle check
- Add add nodes page into idle check white list to prevent user been logged out when add nodes


HankYuLinksys (approved):
Looks good!

PR #297, opened 2025-03-17
Fixes incorrect JNAP call when router is supporting nodes firmware update

AustinChangLinksys:
Fixes incorrect JNAP call when router is supporting nodes firmware update


PeterJhongLinksys (approved):
Looks good.

PR #298, opened 2025-03-17
Modifies logic of get master node and slave nodes

AustinChangLinksys:
Modifies logic of get master node and slave nodes
Master node = device.isAuthority == true || device.nodeType == 'Master'
Slave nodes = device.isAuthority == false && device.nodeType == 'Slave'


PeterJhongLinksys (approved):
Looks good.

PR #301, opened 2025-03-20
NOW-483: Cleanup WAN status on print function

AustinChangLinksys:
- Cleanup WAN status on print format
- Cleanup Backhaul info on print function
- Add Device list on print function


PeterJhongLinksys (approved):
Looks good.

PR #304, opened 2025-03-24
NOW-482: [String update] Update reboot popup message

AustinChangLinksys:
- Update string for child node


HankYuLinksys (approved):
Looks good!

PR #312, opened 2025-03-27
NOW:506- Re-order JNAP commands in PnP save settings transaction

AustinChangLinksys:
- Put close command to the last order in the transaction


HankYuLinksys (approved):
Looks good!

PR #315, opened 2025-03-31
[NOW-502] Correct two of the Wi-Fi channel titles

HankYuLinksys:
Fixed the incorrectly displayed titles for channels 64 and 69


AustinChangLinksys (approved):
Looks good to me

PR #318, opened 2025-04-02
NOW-507: Region string [United States] replace to wrong [Australia / New Zealand]

AustinChangLinksys:
- Correct right strings


HankYuLinksys (approved):
Looks good!

PR #320, opened 2025-04-09
NOW-511: Add supports router14 service and SetWANSettings5

AustinChangLinksys:
- Add support Router14 service
- Add support SetWANSettings5


PeterJhongLinksys (approved):
Looks good.

PR #321, opened 2025-04-09
NOW-510: reposition and adjust text location and SSID config layout on page to have all SSIDs show in a row on screen

AustinChangLinksys:
- Move info card to the top of the page on Desktop layout


PeterJhongLinksys (approved):
Looks good.

PR #328, opened 2025-04-15
[NOW-518] Fixed: Several input fields in Advanced Settings need to be tapped twice to have them focused.

HankYuLinksys:
[Root cause]
Whenever the focus state changes, it will determine if an error has occurred and rebuild the widget tree, which will cause the input field that should have been focused to be replaced by the new field widget.


AustinChangLinksys (approved):
Looks good to me

PR #334, opened 2025-04-29
[NOW-523] LOC: Dashboard: Displays the time twice instead of the time and date

PeterJhongLinksys:
[Root cause] Import string file error

[Solution] Fix the error parameter


AustinChangLinksys (approved):
Looks good to me

PR #339, opened 2025-05-06
NOW-555: Internect connectivity check does not work with new model number

AustinChangLinksys:
- Remove isNode check on checkInternetConnection on PnP provider


HankYuLinksys (approved):
Looks good!

PR #340, opened 2025-05-07
[NOW-532]: CLONE - [GUI][UX][LAN] Since the maximum length of LAN Host Name is 15, suggest reminding user about this limitation, such as (1) add description/notification in GUI, (2) if len=16, disable the Save button in gray.

PeterJhongLinksys:
Use host name rule to show the error prompt.


AustinChangLinksys (approved):
Looksa good to me

PR #345, opened 2025-05-09
[NOW-525][NOW-522]

HankYuLinksys:
1. Added error handler for WiFi toggles on the grid of the dashboard to prevent infinite spinner issues
2. Remove the empty condition check in every build() of the advanced setting view to prevent the issue where languages cannot be switched

https://github.com/user-attachments/assets/d25f6b2c-3d58-4e41-b33f-3d010a1c59c6


PeterJhongLinksys (approved):
Looks good.

PR #346, opened 2025-05-09
[NOW-544]: Remove time estimation on FW update page. Use infinite linear progress bar instead.

PeterJhongLinksys:
Remove time estimation on FW update page. Use infinite linear progress bar instead.


AustinChangLinksys (approved):
Looks good to me

PR #348, opened 2025-05-12
NOW-538: CLONE - [GUI][Topo] Remove WAN connection, GUI does not update some WAN status as disconnected, e.g. <1> Topo\the line between the two icons, my node and Internet. <2> the info in my node is always "Connected to INTERNET".

AustinChangLinksys:
NOW-538: CLONE - [GUI][Topo] Remove WAN connection, GUI does not update some WAN status as disconnected, e.g. <1> Topo\the line between the two icons, my node and Internet. <2> the info in my node is always "Connected to INTERNET".
NOW-504: Device list_Improvement_Allow copy function for troubleshooting purposes in Device list
NOW-540: Improvement - Make sure wait for child online on Add node flow has node type attribute
NOW-541: LOC: FW update progress UI is not localized
CHERRY7-5
- Add dotted line on topology represent offline connection
- Remove INTERNET on Master if offline
- Make detail value selectable on node/device detail page
- Check nodeType during add nodes flow


PeterJhongLinksys (approved):
Looks good.

PR #357, opened 2025-05-14
[NOW-451]: Advanced Routing: The "Invalid IP Address" prompt should not be displayed when adding a LAN/Wireless static route with IP 192.168.1.x

PeterJhongLinksys:


AustinChangLinksys (approved):
Looks good to me

PR #368, opened 2025-05-28
[NOW-574] Added a length check for the application name of IPv6 Port Service

HankYuLinksys:


AustinChangLinksys (approved):
Looks good to me

PR #369, opened 2025-05-28
[NOW-573] Added a missing Select Devices function for the desktop layout of IPv6 port service

HankYuLinksys:


AustinChangLinksys (approved):
Looks good to me

PR #389, opened 2025-06-20
CLONE - [WEB]Verify DUT is able to filter the illegal Border Relay on 6rd Tunnel manul setting. CLONE - [WEB]Verify DUT is able to filter the illegal IPv6 address on 6rd Tunnel manul setting.

AustinChangLinksys:
- Add IPv4 with reserved Rule
- Add IPv6 with reserved Rule


PeterJhongLinksys (approved):
Looks good.

PR #401, opened 2025-07-01
NOW-514: Share WiFi: Should not display QR code when clicking on the specific area

AustinChangLinksys:
- disable tab on show tooltip


HankYuLinksys (approved):
Looks good!

PR #402, opened 2025-07-02
[NOW-604][NOW-237] Redirect the "Need Support" to Linksys Support instead of Call support regions

HankYuLinksys:


AustinChangLinksys (approved):
looks good to me

PR #405, opened 2025-07-04
[NOW-177] The node light function will be only available on the parent node

HankYuLinksys:


AustinChangLinksys (approved):
Looks good to me.

PR #407, opened 2025-07-08
NOW-609:Instant-Admin:  The "Save" button is still disabled after changing time zone's daylight saving setting

AustinChangLinksys:
- Fixes the preserve state issue.


HankYuLinksys (approved):
Looks good!

PR #421, opened 2025-07-24
[NOW-589] Fixed unexpected blank screen after updating FW versions

HankYuLinksys:
This is because some node objects that need to be updated are missing from the DeviceManager, and therefore the 'firstWhere' method will throw an error. (This may occur in the Pinnacle routers with CF production.)

Solution: Save the node objects and their corresponding status into memory as a cache, which can be used later if the node objects unexpectedly get lost during rebooting.


AustinChangLinksys (approved):
LGTM!

PeterJhongLinksys (approved):
Looks good.

PR #423, opened 2025-07-24
NOW-625: CLONE - GUI takes a long time to load when logging in

AustinChangLinksys:
- Update new widget lib
- Add woff2 fonts instead ttf fonts
- Screenshot testcases should load ttf fonts for now
- Update canvaskit.js and canvaskit.wasm to support woff2 fonts and lazy loadings


PeterJhongLinksys (approved):
Looks good.

PR #424, opened 2025-07-25
[NOW-629]: Advanced Routing - Static routing can be set but has no function

PeterJhongLinksys:
This change hides the RIP routing settings for specific Velop models.

- Added isHidingRipRouting property to the node configuration for several Velop models in lib/core/utils/nodes.dart.
- Introduced a new utility function, isHidingRipRouting, to determine whether to hide the routing feature based on the devices model and hardware version.
- Updated lib/page/advanced_settings/advanced_settings_view.dart to conditionally display the Advanced Routing section based on the output of the isHidingRipRouting function.


AustinChangLinksys (approved):
Looks good!

PR #431, opened 2025-07-30
NOW-618: Firewall: Unable to add IPv6 port service if selecting device from device list

AustinChangLinksys:
NOW-618: Firewall: Unable to add IPv6 port service if selecting device from device list - Revise editable_table_settings_view - Add controller for call validatation -Refactor interface for editCellBuilder


PeterJhongLinksys (approved):
Looks good.

PR #440, opened 2025-08-11
Peter/now 626

PeterJhongLinksys:
Merge remote assistance into 1.2.5


lib/l10n/app_en.arb
```diff
@@ -1027,5 +1028,63 @@
   "donePairing": "Done pairing",
```
AustinChangLinksys, line 1028: The new translations for the remote assistance feature are only available in English. Please ensure that the other languages are updated as well.

lib/core/cloud/model/guidan_remote_assistance.dart
AustinChangLinksys, line 1: There seems to be a typo in the file name and throughout the code. 'Guidan' should likely be 'Guardian'. Please correct this for consistency.

AustinChangLinksys (approved):
Overall, this is a great PR. I've pointed out a couple of minor things to address, but otherwise, the changes look good to go.

lib/l10n/app_en.arb
AustinChangLinksys: This will be provided on the next patch

PR #442, opened 2025-08-12
NOW-639: Instant-Topology: The wired child node info should not show the mesh health status

AustinChangLinksys:
- Hide mesh health if the node is wired connection.


HankYuLinksys (approved):
Looks good!

PR #447, opened 2025-08-19
[Instant Admin Testing] Updated the test case

HankYuLinksys:
Updated the integration testing of Instant Admin:
Reset the admin password to the original one during the testing rather than in the teardown


AustinChangLinksys (approved):
looks good to me

PR #690, opened 2026-03-17
feat: Instant Privacy (USP MAC whitelist)

HankYuLinksys:
## Summary

- Add **Instant Privacy** feature page under the USP stack — allows users to lock WiFi access to only currently connected devices with a single toggle
- New USP YAML definition (`mac_filter_access_points`) for reading and writing WiFi AP MAC filter settings via TR-181
- Implement `UspInstantPrivacyService` for MAC list transformation, device enrichment, and AP update descriptor building
- Implement `UspInstantPrivacyNotifier` with `enable`, `disable`, and `addMac` operations — all atomic across all APs (`allowPartial: false`)
- Toggle is disabled when no devices are connected (hard block per spec) and locked during in-progress operations
- Allowed devices shown with friendly hostnames; manually added devices fall back to "Unknown Device"
- Add "Add device manually" dialog with inline MAC format validation and duplicate detection
- Register `/uspInstantPrivacy` route and add menu entry in USP Menu
- Fix codegen filter issue where disabled APs (all-zero values) were incorrectly excluded from parsed results by adding `SSIDReference` as an always-present anchor field
- Regenerate all codegen files with updated usp-codegen version


AustinChangLinksys (approved):
Looks Good To Me

PR #693, opened 2026-03-17
chore: untrack .specify/ directory from version control

HankYuLinksys:
## Summary
- Add `.specify/` to `.gitignore` to exclude the entire speckit tooling directory
- Remove speckit scripts, templates, and constitution memory file from git tracking
- Local files are preserved for individual developer use; not intended to be shared via version control


AustinChangLinksys (approved):
Looks good to me

PR #716, opened 2026-03-20
fix: dashboard loading resilience, edit mode resize & SSE banner

AustinChangLinksys:
## Summary
- **Edit mode resize**: `AbsorbPointer` → `IgnorePointer` so `DashboardOverlay` receives pointer move/up events for resize handles
- **Dashboard skeleton stuck**: `devicesDataProvider` no longer hard-blocks on `wifiDataProvider.future`; uses 10s timeout + fallback empty data. WiFi listener rebuilds models when data arrives later
- **WiFi fetch timeout**: `Future.wait` in `wifiDataProvider` now has 15s timeout to prevent indefinite hang on bridge 503
- **Orchestrator retry**: 5s delayed check invalidates domain providers stuck in error state after bridge startup race
- **SSE banner grace period**: 3s debounce before showing connecting/reconnecting banner, preventing flicker during brief reconnection cycles. Severe states (suspended/disconnected) still show immediately

## Test plan
- [ ] Hot restart with bridge temporarily unavailable → dashboard cards load progressively instead of staying skeleton
- [ ] Edit mode → drag resize handle on card → card resizes correctly
- [ ] SSE brief disconnect/reconnect → banner does not flicker
- [ ] SSE suspended (bridge down >30s) → banner appears immediately with Reconnect button
- [ ] Pull-to-refresh → all cards reload

### DIFF
--- a/lib/demo/providers/demo_overrides.dart
+++ b/lib/demo/providers/demo_overrides.dart
@@ -92,12 +92,15 @@ class _DemoAuthNotifier extends AuthNotifier {
   @override
   Future<AuthState?> init() async {
     debugPrint('Demo: Auth init called - preserving local login');
-    final demoState = AuthState(
-      loginType: LoginType.local,
-      localPassword: 'demo-password',
-    );
-    state = AsyncValue.data(demoState);
-    return demoState;
+    // Use AsyncValue.guard to match base class pattern — the `await` ensures
+    // the state assignment happens after a microtask boundary, avoiding the
+    // "Tried to modify a provider while the widget tree was building" error
+    // when init() is called from go_router redirect during build phase.
+    state = await AsyncValue.guard(() async => AuthState(
+          loginType: LoginType.local,
+          localPassword: 'demo-password',
+        ));
+    return state.value;
   }
 
   @override
--- a/lib/page/_shared/components/sse_connection_banner.dart
+++ b/lib/page/_shared/components/sse_connection_banner.dart
@@ -1,3 +1,5 @@
+import 'dart:async';
+
 import 'package:flutter/material.dart';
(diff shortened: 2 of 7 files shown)


AustinChangLinksys:
## Qodo Review Fixes (e776f28)

Addressed all 4 items from the Qodo code review:

### ✅ #2 setState during build — Fixed
Moved `SseConnectionBanner._reconcile()` from `build()` inline `whenData()` to `initState()` via `ref.listenManual()`. This ensures `setState()` never fires during the widget build phase, preventing Flutter assertion errors during SSE reconnection cycles.

### ✅ #3 Timeout escapes to WiFi settings — Fixed
Added try-catch around `wifiDataProvider.future` in `UspWifiSettingsNotifier.performFetch()`. On timeout/error, returns `(null, WifiSettingsStatus(errorMessage: ...))` error tuple so the UI correctly exits loading and shows the error instead of hanging indefinitely.

### ✅ #4 Overlapping retry scheduling — Fixed
- `_scheduleProviderRetry()` now uses a `Timer?` field with cancel-before-reschedule
- Added `ref.onDispose(() => _retryTimer?.cancel())` to prevent callbacks after disposal
- Removed `wifiDataProvider` from the retry list — it is not eagerly triggered in `_buildImpl()`, so reading it in the retry callback would cause an unintended side-effect initialization

### ⏭️ #1 Provider location — Not applicable
False positive. The project uses feature-based organization where all domain providers reside in `lib/page/*/providers/`. This is the established architectural convention, not a violation.

---
Additionally, f412bfc fixes a demo-mode Riverpod build-phase error (`_DemoAuthNotifier.init()` switched to `AsyncValue.guard` to defer state assignment past the microtask boundary).

PeterJhongLinksys (approved):

PR #816, opened 2026-04-15
fix: use inline fill in brand text logo SVG for HTML web renderer compatibility

### LINKED ISSUE #817: fix: brand text logo displays incorrectly on HTML web renderer
## Description
Brand text logo (`brand_text_logo_tb.svg`) displays as black/invisible when building with `--web-renderer html`.

## Root cause
HTML web renderer's `colorFilter(BlendMode.srcIn)` is unreliable. In canvaskit, it correctly overrides the SVG's original `#8a8c8e` fill to white, but in HTML renderer it fails — causing the logo to appear black (CSS `<style>` not parsed) or gray (inline `#8a8c8e` leaking through).

## Solution
Replace CSS `<style>` with inline `fill="#fff"` on all `<path>` elements. The original `#8a8c8e` gray fill was never visible to users — it was always overridden to white by `colorFilter` in canvaskit. By setting the SVG fill directly to `#fff`, both renderers display the correct white color.

PR: #816 (branch: `peter/toob-logo-238`)

PeterJhongLinksys:
## Summary
- Replace CSS `<style>` class-based fill with inline `fill="#fff"` in `brand_text_logo_tb.svg`
- Fixes logo appearing black/invisible when building with `--web-renderer html`

## Root cause
HTML web renderer's `colorFilter(BlendMode.srcIn)` is unreliable. In canvaskit, it correctly overrides the SVG's original `#8a8c8e` fill to white, but in HTML renderer it fails — causing the logo to appear black (CSS `<style>` not parsed) or gray (inline `#8a8c8e` leaking through).

## Solution
Replace CSS `<style>` with inline `fill="#fff"` on all `<path>` elements. The original `#8a8c8e` gray fill was never visible to users — it was always overridden to white by `colorFilter` in canvaskit. By setting the SVG fill directly to `#fff`, both renderers display the correct white color: canvaskit still applies `colorFilter` (white on white), and HTML renderer falls back to the SVG's own fill which is now already white.

expected


AustinChangLinksys (approved):
Looks good to me

PR #891, opened 2026-05-21
fix(topology): correct device-to-node mapping with dataElementsId

AustinChangLinksys:
## Summary
- Match wireless clients to mesh extenders using **both** Hosts MAC (`deviceId`) and DataElements ID (`dataElementsId`), normalized (uppercase, no colons), since the two IDs may differ for the same node. Without this, clients whose `parentNodeId` came from DataElements `clientToNodeMap` were silently falling back to the gateway and the topology graph misrepresented the mesh.
- Add `dataElementsId` field on `NodeUIModel` and populate it from `slaveMeshInfo.deviceId` in `UspDevicesDataService`.
- Refactor `_rssiValueToLevel` to delegate to `getWifiSignalLevel` from `lib/core/utils/wifi.dart`, removing the duplicated RSSI thresholds (now a single source of truth).
- Update `_rssiToDistanceFactor` to use the wider -90..-50 range, aligned with the same thresholds.

## Test plan
- [x] `flutter test` (full functional suite): 2309 / 2309 passing
- [x] `dart format` clean
- [x] `flutter analyze` clean on changed files
- [x] New unit cases cover dataElementsId fallback and normalized parentNodeId matching
- [ ] Manual verification on multi-node mesh: clients render under correct extender
- [ ] Manual verification: ethernet clients still resolve to gateway

\U0001F916 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/lib/page/devices/services/usp_devices_data_service.dart
+++ b/lib/page/devices/services/usp_devices_data_service.dart
@@ -412,6 +412,7 @@ class UspDevicesDataService {
 
       nodes.add(NodeUIModel(
         deviceId: slave.mac,
+        dataElementsId: slaveMeshInfo?.deviceId,
         friendlyName: slave.friendlyName,
         hostName: slave.hostName,
         model: slaveMeshInfo?.model ?? slave.modelName ?? '',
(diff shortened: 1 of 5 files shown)


PeterJhongLinksys (approved):
Looks good.

Issue #857, opened 2026-05-07
[Speed Test] Not Supported via TR-181


AustinChangLinksys:
## Problem

No TR-181 support for network speed testing.

**Current behavior**: TR-181 data model does not expose speed test functionality.

**Expected behavior**: Should support:
- Initiate speed test (download/upload)
- Get speed test results (bandwidth, latency, jitter)
- Speed test history

## Impact

- 🟡 **P2 Secondary Feature** — Speed test diagnostic tool unavailable
- Users cannot verify their internet connection performance from the app

## Notes

- Part of diagnostic functionality alongside IPPing and TraceRoute
- May require vendor extension or integration with external speed test service
- Consider privacy implications of third-party speed test services

## TR-181 Paths (Expected)

```
Device.IP.Diagnostics.SpeedTest.
Device.IP.Diagnostics.SpeedTest.Download.
Device.IP.Diagnostics.SpeedTest.Upload.
```

## Related

- Parent tracker: #791
- Master tracker: #788

---

**Labels**: `usp:investigation`, `feature:monitoring`, `blocked:fw-support`

AustinChangLinksys:
State for this issue is now tracked in #1612, which re-measures it on FW `2.0.2.26091803` (2026-09-23). The premise still holds — the firmware has no source for this feature. Staying open; see the "No firmware source" table there for the evidence.

PR #943, opened 2026-06-16
feat: FW 1.2.1 alias update and Internet Settings enhancements (#942)

### LINKED ISSUE #852: [Internet Settings] IPv6rd Prefix Format Validation Failure
## Problem

Set operation on IPv6rd prefix field returns:
```
fault 9007: List patterns did not match
```

## Impact

Cannot set prefix field for 6rd tunnel configuration.

## TR-181 Path

`Device.IPv6rd.InterfaceSetting.{i}.IPv6Prefix`

## Expected Behavior

Should accept valid IPv6 prefix format and persist to UCI.

## Steps to Reproduce

1. Navigate to IPv6 settings
2. Configure 6rd tunnel
3. Attempt to set IPv6 prefix value
4. Observe fault 9007 error

---

**Parent Tracker**: #789  
**Priority**: P1 🔴

AustinChangLinksys:
## Summary
- Update codegen alias resolution for FW 1.2.1 (`cpe-wan`/`cpe-lan` → `wan`/`lan`)
- Add MTU validation with error messages (576-1500, PPPoE: 576-1492)
- Add 6rd tunnel field validation with error messages
- Fix Dashboard sync after Internet Settings save
- Optimize WiFi guest detection using alias suffix

### Codegen Alias Update
- Updated 16 generated files with new alias values per Architecture #110
- Added 3 new codegen files (ethernet_links, ip_interfaces, dhcpv4server_pools)

### MTU Validation
- Range: 576 (IPv4 RFC 791 min) to 1500 (1492 for PPPoE)
- Inline error message when value is out of range
- Auto toggle hidden pending FW support

### 6rd Tunnel Validation (Closes #852)
- IPv6 prefix: CIDR format validation using `IPv6WithReservedRule`
- Border relay: IPv4 format validation using `IpAddressRule`
- Prefix length: 0-32 range check
- Inline error messages for invalid input
- Rejects reserved addresses (loopback, multicast, unspecified)
- Verified on FW 1.0.16.26013014

### Dashboard Sync
- Added `ref.invalidate(wanDataProvider)` after save
- Ensures WAN card refreshes with updated values

### WiFi Guest Detection
- Changed from instance index to alias suffix (`-guest`)

## Test plan
- [x] All functional tests pass (2763+ tests)
- [ ] Verify Internet Settings page loads with updated alias paths
- [ ] Test MTU input validation shows error for out-of-range values
- [ ] Test 6rd tunnel validation shows error for invalid prefix/relay
- [ ] Confirm Dashboard WAN card refreshes after saving settings
- [ ] Check WiFi guest network detection works correctly

Closes #942, Closes #852


HankYuLinksys (commented):
## Review — FW 1.2.1 alias update & Internet Settings enhancements

Overall this is a solid change: good test coverage (MTU/6rd/guest all have added cases), and the codegen alias update is consistently mirrored across the service test mocks. I have **one blocking concern** plus a few items worth confirming.

### 🔴 Blocking — Bridge mode can never be saved (MTU regression)

The new validator removes the old `mtu == 0` (auto) exemption:

```dart
// before
if (form.mtu != 0 && (form.mtu < 576 || form.mtu > 1500)) return false;
// after
if (form.mtu < 576 || form.mtu > mtuMax) return false;   // 0 is now invalid
```

But `updateConnectionType()` still resets mtu to 0 when switching to bridge:

```dart
// usp_internet_settings_notifier.dart:235
if (type == UspWanConnectionType.bridge) {
  form = form.copyWith(mtu: 0);
}
```

`_validateOptionalFields` runs for **all** connection types (bridge passes `_validateIpv4Fields` but optional fields are still checked). Result: switching to bridge → `mtu = 0` → form is permanently invalid → **Save button stays locked**.

Same risk elsewhere: `UspInternetSettingsForm.mtu` defaults to `0`, and MTU is populated from firmware via `mtu: wan.mtu` (service:100). If firmware ever reports MTU as 0/auto, the user is stuck on an unsavable form even without touching the field.

**Suggestion:** bridge mode should not require MTU (either skip optional validation for bridge, or reset to 1500 instead of 0), and please confirm how a fetched `mtu == 0` is handled.

### 🟡 Worth confirming

1. **WiFi guest detection now uses alias suffix.** Changed from instance-index ordering to `alias.endsWith('-guest')`. The old logic worked on any firmware; the new one depends on FW 1.2.1+ auto-provisioning `-guest` aliases. On older firmware (alias is null) **every guest network would be classified as main**. Please confirm older FW is no longer supported, or document this compatibility boundary in the PR description.

2. **MTU input `onChanged` only updates when `parsed > 0`.** If the user clears the field or enters 0, the form keeps the stale value, so the inline error doesn't reflect the current input. Minor UX issue, non-blocking.

3. **`currentMacAddress` is hard-coded to `''`** (service:117). Reasonable while MAC Clone is disabled, but it means the current MAC is no longer surfaced in the UI. Confirm this is intended (permanent disable vs. temporary).

### 🟢 Nits

- `_accountLockedPattern` `(locked|lock)`: `lock` already subsumes `locked`; can simplify to `lock(ed)?`.
- Network-error detection via `errorStr.contains('HTTP 5')` is broad and could match unrelated messages containing that substring. Low risk in practice.
- `auth_provider._mapToViewError` flattens every `ServiceError` into `UnexpectedError` carrying an error code in `message`. Works, but the type information is lost here, so future per-error UI handling would have to inspect the message string. Fine for now.

### ✅ Nicely done

- 6rd validation reuses the existing `IPv6WithReservedRule` / `IpAddressRule` and adds loopback/multicast/unspecified test cases.
- The codegen alias change (`cpe-lan`→`lan`, etc.) is mirrored across all service test mocks with no gaps.
- speedTest disable keeps the code as comments + updates the test counts (19→18), tracked under #857.

**Conclusion:** I'd recommend addressing the bridge MTU issue before merge; the rest are confirms/nits.

AustinChangLinksys:
## Response to Review Comments

Thank you for the thorough review! Here are my responses:

---

### 🔴 Blocking — Bridge mode MTU

**Not an issue.** When `mtu: 0` is sent to the firmware, it auto-assigns an appropriate value (e.g., 1500) — it doesn't actually store 0. After `save()`, a subsequent `fetch()` retrieves the firmware-assigned actual value.

This is by design: TR-181 cannot represent an "auto" state, so 0 serves as a sentinel value that lets the firmware decide the appropriate MTU.

---

### 🟡 Worth confirming

**1. WiFi guest detection now uses alias suffix**

Confirmed: **backward compatibility is not required.** This PR targets FW 1.2.1+; older firmware versions are out of scope.

**2. MTU input `onChanged` only updates when `parsed > 0`**

This is intentional. 0 is not a valid MTU value (valid range: 576–1500). Rejecting invalid input and preserving the last valid value is a safer UX pattern than allowing the form to enter an invalid state where the displayed value differs from what will be saved.

**3. `currentMacAddress` is hard-coded to `''`**

Confirmed: **this feature has been removed.** MAC Clone functionality is disabled because the USP data model does not support write operations. The related UI code has been commented out (see `usp_optional_section.dart:108-114`).

---

### Summary

| Item | Status | Notes |
|------|--------|-------|
| Bridge MTU | ✅ Not an issue | FW auto-assigns value |
| Guest alias | ✅ Confirmed | No backward compatibility needed |
| MTU onChanged | ✅ Intentional | Rejecting invalid input is correct |
| MAC Clone | ✅ Confirmed | Feature removed |

All concerns addressed. Ready for re-review.

HankYuLinksys (commented):
Thanks! 2/3/4 are clear. But the bridge MTU point is about the **client-side validator gate**, not firmware semantics — `mtu: 0` never reaches the firmware.

- Save is gated by `isPositiveEnabled: state.isDirty && isValid` (`usp_internet_settings_view.dart:178`)
- `isValid` = `validateForm()`, which now runs `_validateOptionalFields` for all types and returns false when `mtu < 576`
- `updateConnectionType()` sets `mtu: 0` for bridge (`usp_internet_settings_notifier.dart:235`)

So switching to bridge makes the form dirty **and** invalid → Save button stays disabled → 0 can never be sent. Your point that firmware treats 0 as a valid sentinel actually argues the validator shouldn't reject 0.

Could you confirm the Save button is actually clickable after switching to bridge? A quick test should settle it.

HankYuLinksys (approved):
Looks good!

PR #952, opened 2026-06-17
fix(internet-settings): use SET for VLAN tagging and enable PPPoE ServiceName

### LINKED ISSUE #842: [FW-INTERNET] PPPoE ServiceName SET Rejected (fault 9001)
## PPPoE ServiceName — SET Rejected

> **Parent Tracker**: #790

### Problem
`Device.PPP.Interface.1.PPPoE.ServiceName` SET is rejected with fault 9001 despite schema marking it as writable (`data:"1"`).

The bbfdm `netmngr.so` handler's `set_ServiceName` rejects all SET operations unconditionally.

### Impact
ServiceName cannot be configured via USP. Most ISPs do not require ServiceName, so practical impact is low, but it blocks full PPPoE configuration parity with JNAP.

### Reproduction
```bash
# SSH on router (any WAN mode):
ubus call bbfdm set '{"path":"Device.PPP.Interface.1.PPPoE.ServiceName","value":"test-service"}'
# → fault 9001: "Request denied (no reason specified)"
```

### App-side workaround
ServiceName UI and save logic commented out. Search marker: `bbfdm rejects SET (fault 9001)`

### Firmware action required
Fix `set_ServiceName` handler in `netmngr.so` to write the value to UCI `network.wan.service`.

### Related
- #790 (Internet Settings tracker)
- Documentation: `doc/usp/issues/015-pppoe-servicename-and-mac-clone.md`

PeterJhongLinksys:
## Summary
- Replace VLAN lifecycle (ADD/DELETE) with SET Enable on the existing VLANTermination.1 instance, fixing the bug where disabling VLAN still showed as enabled due to a system-default instance that cannot be deleted.
- Uncomment PPPoE ServiceName field in UI and service layer now that bbfdm supports SET on PPPoE.ServiceName.
- Clean up stale comments and fix step numbering after VLAN lifecycle removal.
- Fix dart format pre-commit hook that never triggered due to glob mismatch with heredoc commands.

## Test plan
- [x] Verify VLAN tagging toggle on/off persists correctly on router
- [x] Verify PPPoE ServiceName field saves without fault 9001
- [x] Verify dart format hook triggers on commit

### DIFF
--- a/.claude/settings.json
+++ b/.claude/settings.json
@@ -11,10 +11,9 @@
           },
           {
             "type": "command",
-            "if": "Bash(git commit *)",
-            "command": "staged=$(git diff --cached --name-only --diff-filter=ACM -- '*.dart'); if [ -n \"$staged\" ]; then if command -v fvm >/dev/null 2>&1; then echo \"$staged\" | xargs fvm dart format 2>/dev/null; else echo \"$staged\" | xargs dart format 2>/dev/null; fi; echo \"$staged\" | xargs git add; fi",
+            "command": "staged=$(git diff --cached --name-only --diff-filter=ACM -- '*.dart'); if [ -n \"$staged\" ]; then if command -v fvm >/dev/null 2>&1; then echo \"$staged\" | xargs fvm dart format; else echo \"$staged\" | xargs dart format; fi; echo \"$staged\" | xargs git add; fi",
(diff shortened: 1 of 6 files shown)


HankYuLinksys (approved):
Looks good!

PR #955, opened 2026-06-18
[2.x] feat(remote): implement Remote Assistance for CA and Client modes

AustinChangLinksys:
## Summary

- **CA mode**: Support agents can remotely access customer routers via Guardian proxy
- **Client mode**: Device owners can initiate RA sessions, generate PINs, and manage active sessions
- Session recovery after page refresh (PENDING shows banner, ACTIVE shows blocking dialog)
- Auto-disconnect on 401 unauthorized or session invalidation

### Core Infrastructure
- `GlobalConfig.remote` for runtime CA/client mode detection
- `GuardianApiClient` for RA API calls (replaces LinksysCloudRepository usage)
- `usp_client` WASM updated to 0.11.1

### CA Side (Support Agent)
- `RemoteAccessProvider` — session state, polling, countdown timer
- `RemoteSessionChip` — floating draggable chip with session info
- Auto-logout on 401 from session polling

### Client Side (Device Owner)
- `RemoteClientProvider` — session lifecycle management
- `RemoteAssistanceDialog` — initiate flow with PIN display
- `RemoteAssistanceBanner` — non-blocking banner for PENDING status
- `RemoteAssistanceSessionGuard` — blocking dialog for ACTIVE status after refresh
- Auto-close dialog when CA ends session (INVALID status)

### Cleanup
- Removed unused `DashboardShell` (only `UspDashboardShell` is used)

## Test Plan

- [x] Unit tests for all providers and services (2925 tests pass)
- [ ] Manual: CA creates session → Client generates PIN → CA verifies → ACTIVE
- [ ] Manual: Client refreshes page while ACTIVE → blocking dialog appears
- [ ] Manual: CA ends session → Client dialog auto-closes
- [ ] Manual: Session expires → both sides handle gracefully

## Known Limitations

- Hardcoded English strings (i18n pending)
- Cloud `POST /sessions/pin` endpoint not yet available for full E2E testing


AustinChangLinksys (commented):
🤖 **Automated Self-Review** · `5abf530..2b80346` (full) · large PR (+7838/−2899, 92 files)

**Verdict: 💬 Comment** — self-review on author's own PR (not an approval). 3 Critical found. Strong feature overall; cloud-model removal is clean, but several credential/lifecycle and layering issues should be addressed before merge.

|---|---|---|
| 🔴 | `guardian_api_client.dart:222` + `remote_assistance_service.dart` (_mapError) | **[both reviewers]** Error-mapping dead code: non-JSON 401 → `UnexpectedError`, so expired token never triggers `_forceSessionEnd()` and CA UI won't log out |
| 🔴 | `guardian_api_client.dart:154` + `remote_client_provider.dart:178` | PIN (the sole CA auth credential) is logged in plaintext via `logger.i/d` — may leak to console/remote logs in web/release |
| 🔴 | `remote_client_provider.dart:294` | Closing a PENDING session only clears local state; server-side PIN/session stays valid until expiry — "cancel" can't actually revoke an already-seen PIN |
| 🔴 | `remote_client_provider.dart:8` | core→page reverse dependency: `core/cloud/providers/...` imports `page/remote_assistance/services/...`, violating downward-only layering |
| 🔴 | `remote_assistance_provider.dart:88` | GetIt runtime-swaps the global `UspClient` singleton destructively and irreversibly (no lock, only restart restores) |
| ⚠️ | `router_provider.dart` | session token passed in URL query string (leaks via history / Referer / server logs) |
| ⚠️ | `remote_assistance_provider.dart:106` | `deactivate()` doesn't restore original `UspClient`; a still-authenticated client survives session end |
| ⚠️ | `guardian_api_client.dart:71,151` | device-token missing field → `as String` crash; `createPin` missing field silently returns empty PIN |
| ⚠️ | RA module (3 dirs) | No `uspMutationLockProvider.withLock()` anywhere, yet client-swap mutation runs while WASM forbids concurrency |
| ⚠️ | `remote_client_provider` / `remote_access_provider` | Polling + countdown timers hand-rolled in triplicate (DRY); three providers own session lifecycle |
| ⚠️ | `confirm_view.dart`, `remote_session_chip.dart`, `session_guard.dart` | Large logic-bearing views (563 / 459 / 94 lines) have zero tests |
| ⚠️ | `dialog.dart`, `confirm_view.dart` | Several user-facing strings hardcoded in English, bypassing l10n |
| 💡 | `banner.dart`, `dialog.dart`, `chip.dart` | Many raw `Container/BoxDecoration` cards/badges/PIN-box — UI Kit may already have `AppBadge`/`AppSurface` |
| 💡 | model layer | `expiredIn.abs()` scattered across 4 files — expose `isExpired`/`remaining` instead |
| 💡 | `guardian_api_client.dart:51` | device-token cache key not namespaced by serial number |

<details>
<summary><b>🔴 Critical — details (5 items)</b></summary>

**1. Error-mapping dead code [both reviewers].** `GuardianApiClient._validateResponse` (`:222-231`) throws `http.ClientException` on `statusCode>=400`, but the service's `_mapError` only handles `ErrorResponse` (`if (error is ErrorResponse) { if (error.status == 401) return UnauthorizedError(); ... }`). So when a 401 has a non-JSON body (common from gateways/proxies), it maps to `UnexpectedError`, not `UnauthorizedError`. `remote_access_provider._pollSessionInfo` relies on `on UnauthorizedError` to call `_forceSessionEnd()` — that path is bypassed, so an expired/invalid token never forces session end and the CA UI won't log out. The architecture reviewer independently found the same root cause from the other side: the entire `SESSION_EXPIRED`/`UNAUTHORIZED`/`INVALID_SESSION` switch in `_mapError` is dead code. **Fix:** have `_validateResponse` throw semantic errors for 401/403 (or unify on `LinksysHttpClient` throwing `ErrorResponse`) so the switch actually fires.

**2. PIN logged in plaintext.** `guardian_api_client.dart:154` `logger.d('[Guardian] createPin response: $body')` (body contains `pin`) and `remote_client_provider.dart:178` `logger.i('PIN created: ${result.pin}')`. PIN is the CA's identity-verification credential (a remote-authorization code). `logger.i/d` may land in console or remote log collection in web/release. **Fix:** log status only ("PIN created"), never the value or the full response body.

**3. PENDING close doesn't revoke server session.** `endRemoteAssistance` only deletes when `status == active`: `if (state.sessionInfo?.status == GRASessionStatus.active) { await _svc.endSession(...); }` then clears state. Pressing Close during PENDING (PIN already generated/shown) clears local state but leaves the server session + PIN valid until expiry. If the CA already saw the PIN, "cancel" can't actually revoke the grant. **Fix:** also call `deleteSession` for INITIATE/PENDING.

**4. core→page reverse dependency.** `remote_client_provider.dart:8` imports `package:privacy_gui/page/remote_assistance/services/remote_assistance_service.dart`. A file in `lib/core/cloud/providers/` (infra layer) importing `lib/page/` inverts the downward-only rule; `device_credentials_provider.dart` does the same (imports `page/devices/...`). **Fix:** move `core/cloud/providers/remote_assistance/` into `page/remote_assistance/providers/`.

**5. Destructive irreversible UspClient swap.** `remote_assistance_provider.dart:88-101` does `getIt.unregister<UspClient>(); oldClient.dispose(); getIt.registerSingleton<UspClient>(raClient);`, and `deactivate()` admits it "does NOT restore the original local UspClient. The app should navigate to login or restart." Any provider/SSE manager holding the old reference now points at a disposed object; state is unrecoverable without restart, and the swap has no lock protection or in-flight-request drain. **Fix:** guard/block concurrent calls around the swap, document the single-direction/restart constraint and enforce it in the route guard, or inject the RA client as a parameter rather than swapping the singleton.
</details>

<details>
<summary><b>⚠️ Warning — details</b></summary>

- **Token in URL query** (`router_provider.dart`): redirect rebuilds `?session=…&token=…` into the location bar; query strings leak via browser history, Referer header, server access logs, bookmarks. **Fix:** `replaceState` to strip the query after redirect, or use a fragment (#) instead of query; keep token only in sessionStorage.
- **`deactivate()` leaves authenticated client** (`remote_assistance_provider.dart:106`): GetIt still holds the Guardian-proxied client carrying `temporaryAccessToken` after session end. **Fix:** unregister + dispose on deactivate.
- **Missing-field handling** (`guardian_api_client.dart:71,151`): `jsonDecode(...)['linksysToken'] as String` crashes on null; `createPin` returns `pin: '' ` silently on missing field → UI shows blank PIN as success. **Fix:** `as String?` + throw semantic error on missing/empty.
- **No mutation lock** (RA dirs): `git grep` shows zero `withLock()` across the three RA directories, yet `activate()` performs the most dangerous mutation (swapping the WASM client, which forbids concurrency). **Fix:** wrap client swap + first USP call in the mutation lock, or document that RA has no concurrent source.
- **Triplicated lifecycle** (`remote_client_provider` vs `remote_access_provider`): session-poll timer, countdown timer, `expiredIn.abs()`, invalid-detection, persistence are hand-rolled twice in parallel; truth source diverges. **Fix:** extract a shared poll+countdown helper/mixin.
- **Untested large views**: `confirm_view.dart` (563 lines — CA entry point with token validation, countdown, `_doConnect` triggering the global swap, and a `_mapServiceError` state machine), `remote_session_chip.dart` (459, drag snap-to-edge), `session_guard.dart` (94) have no tests. **Fix:** add widget/golden tests for confirm_view's token-validation state machine and chip interaction.
- **Hardcoded English strings** (`dialog.dart:97/551/725`, `confirm_view.dart:97/126/139/181/218`): e.g. `'Session is invalid or expired'`, `'PIN copied to clipboard'`. l10n infra is already used elsewhere in the same files. **Fix:** move to ARB.
</details>

<details>

- **Cloud-model removal is very clean** — `git grep` across the head tree found 0 residual references for all 19 deleted symbols (`linksys_cloud_repository`, `cloud_account`, `region_code`, etc.) in both `lib/` and `test/`. Not a leftover-dead-code removal.
- Provider/Service test coverage is solid (client provider 577 lines, RA provider 296, remote_access 640, service 484, plus golden/loc tests).
- Service layer centralizes error mapping (`mapError → ServiceError`); `guardian_api_client` correctly sits in the `core/cloud/` transport layer; UI mostly uses `ui_kit_library`.
- `app_config.json(.template)` carries no sensitive defaults; `global_config.dart` holds no endpoint secrets.
</details>

---
*Cross-reviewed by two independent agents (security+correctness / architecture+maintainability). Automated self-review — not an approval. Please sanity-check before merge.*

HankYuLinksys (approved):
Approving — re-checked the three points I raised on `remote_client_provider.dart` against the latest commits (`81a60d8e`, `2b80346a`), all addressed:

- **🔴 Poll loop never terminating** — Fixed (now `status != INVALID && expiredIn <= 0`; the `.abs()` re-introduction only affects the countdown display, not the termination check).
- **🟠 Countdown rebuilt on every poll** — Fixed (only starts when `_expiredCountdownTimer == null`).
- **🟠 `_creds` throwing + missing cleanup** — Fixed (`_credsOrNull` with graceful early returns, plus `ref.onDispose` cleanup).

Also nice to see the extra hardening in `ebc58edc` (service moved to `core/cloud/services/`, PIN logging removed, mutation lock on the client swap, token stripped from URL), and the fixes are test-covered. Hardcoded-strings i18n can land as follow-up. Thanks for the thorough turnaround.

PR #976, opened 2026-06-23
fix(auth): reduce duplicate login requests on app startup

AustinChangLinksys:
## Summary
- Add Completer deduplication to `restoreSession()` and `authProvider.init()` to coalesce concurrent calls
- Add cooldown mechanism after failed login to prevent account lockout
- Implement token-first strategy: try `refreshToken()` before password login
- Remove redundant `restoreSession()` calls from `wifi_settings_provider` and `internet_settings_notifier` (auth already handled by orchestrator)
- Remove unnecessary `autoConfigurationLogic()` call on `/localLoginPassword` route (user is already on login page)

This reduces login requests from 5 to 1 on app startup with stored credentials, preventing potential account lockout from rapid retries.

Closes #974

## Test plan
- [ ] Verify only 1 login request on app startup (check Network tab)
- [ ] Verify auto-login still works with correct stored password
- [ ] Verify login page displays correctly after failed auto-login
- [ ] Verify no account lockout after page refresh with stale password

### DIFF
--- a/lib/page/wifi_settings/providers/usp_wifi_settings_provider.dart
+++ b/lib/page/wifi_settings/providers/usp_wifi_settings_provider.dart
@@ -4,7 +4,6 @@ import 'package:privacy_gui/core/utils/logger.dart';
 import 'package:privacy_gui/framework/preservable.dart';
 import 'package:privacy_gui/framework/preservable_contract.dart';
 import 'package:privacy_gui/framework/preservable_notifier_mixin.dart';
-import 'package:privacy_gui/core/usp/providers/usp_auth_coordinator.dart';
 import 'package:privacy_gui/core/usp/providers/usp_mutation_lock.dart';
 import 'package:privacy_gui/core/usp/providers/usp_client_provider.dart';
 import 'package:privacy_gui/page/wifi_settings/models/wifi_network_ui_model.dart';
@@ -77,15 +76,12 @@ class UspWifiSettingsNotifier extends AutoDisposeNotifier<UspWifiSettingsState>
     }
 
     if (!usp.isAuthenticated) {
-      await ref.read(uspAuthCoordinatorProvider).restoreSession();
-      if (!usp.isAuthenticated) {
-        return (
-          null,
-          WifiSettingsStatus(
-            error: const NotAuthenticatedError(detail: 'USP not authenticated'),
-          )
-        );
-      }
+      return (
+        null,
+        WifiSettingsStatus(
+          error: const NotAuthenticatedError(detail: 'USP not authenticated'),
+        )
+      );
     }
 
     logger.d('[USP][WiFi]: Fetching WiFi data...');
--- a/lib/providers/auth/auth_provider.dart
+++ b/lib/providers/auth/auth_provider.dart
@@ -74,13 +74,15 @@ class AuthNotifier extends AsyncNotifier<AuthState> {
       });
       // AsyncValue.guard never throws — it converts errors to AsyncError.
       // Propagate error to concurrent waiters if the guard failed.
+      // Note: Do NOT rethrow here — callers (app.dart, router_provider.dart)
+      // use bare .then() without .catchError, so throwing would leave the
+      // splash screen stuck or break the redirect.
       if (state case AsyncError(:final error, :final stackTrace)) {
         _initInProgress!.completeError(error, stackTrace);
-        Error.throwWithStackTrace(error, stackTrace);
-      } else {
-        _initInProgress!.complete(state.value);
-        return state.value;
+        return null;
       }
+      _initInProgress!.complete(state.value);
+      return state.value;
     } finally {
       _initInProgress = null;
     }
--- a/lib/providers/auth/auth_provider.dart
+++ b/lib/providers/auth/auth_provider.dart
@@ -1,3 +1,5 @@
+import 'dart:async';
+
 import 'package:flutter_riverpod/flutter_riverpod.dart';
 import 'package:flutter_secure_storage/flutter_secure_storage.dart';
 import 'package:privacy_gui/constants/error_code.dart';
@@ -21,46 +23,63 @@ final authProvider =
     AsyncNotifierProvider<AuthNotifier, AuthState>(() => AuthNotifier());
 
 class AuthNotifier extends AsyncNotifier<AuthState> {
+  Completer<AuthState?>? _initInProgress;
+
   @override
   Future<AuthState> build() => Future.value(AuthState.empty());
 
   AuthService get _authService => ref.read(authServiceProvider);
 
   Future<AuthState?> init() async {
-    // Note: Do NOT set `state = AsyncValue.loading()` synchronously here.
-    // This method can be called during initState/widget mount, and a
-    // synchronous state change would trigger provider notifications that
-    // cause a !_dirty assertion in ProviderScope.
-    state = await AsyncValue.guard(() async {
-      // Determine login type from stored credentials
-      final loginTypeResult = await _authService.getStoredLoginType();
-      final loginType = loginTypeResult.when(
-        success: (type) => type,
-        failure: (_) => LoginType.none,
-      );
+    // Coalesce concurrent init calls — return in-flight result if one exists
+    if (_initInProgress != null) {
+      logger.d('[Auth]: init already in progress, awaiting...');
+      return _initInProgress!.future;
+    }
 
-      // Get stored local password
-      final passwordResult = await _authService.getStoredLocalPassword();
-      final localPassword = passwordResult.when(
-        success: (p) => p,
-        failure: (_) => null,
-      );
+    _initInProgress = Completer<AuthState?>();
+    try {
+      // Note: Do NOT set `state = AsyncValue.loading()` synchronously here.
+      // This method can be called during initState/widget mount, and a
+      // synchronous state change would trigger provider notifications that
+      // cause a !_dirty assertion in ProviderScope.
+      state = await AsyncValue.guard(() async {
+        // Determine login type from stored credentials
+        final loginTypeResult = await _authService.getStoredLoginType();
+        final loginType = loginTypeResult.when(
+          success: (type) => type,
+          failure: (_) => LoginType.none,
+        );
 
-      logger.d(
-          '[Auth]init: hasPassword=${localPassword != null}, loginType=$loginType');
+        // Get stored local password
+        final passwordResult = await _authService.getStoredLocalPassword();
+        final localPassword = passwordResult.when(
+          success: (p) => p,
+          failure: (_) => null,
+        );
 
-      // Restore USP session on page reload / app restart (local login only)
-      if (loginType == LoginType.local) {
-        await ref.read(uspAuthCoordinatorProvider).restoreSession();
-      }
+        logger.d(
+            '[Auth]init: hasPassword=${localPassword != null}, loginType=$loginType');
 
-      return AuthState(
-        localPasswordHint: state.value?.localPasswordHint,
-        loginType: loginType,
-        localPassword: localPassword,
-      );
-    });
-    return state.value;
+        // Restore USP session on page reload / app restart (local login only)
+        if (loginType == LoginType.local) {
+          await ref.read(uspAuthCoordinatorProvider).restoreSession();
+        }
+
+        return AuthState(
+          localPasswordHint: state.value?.localPasswordHint,
+          loginType: loginType,
+          localPassword: localPassword,
+        );
+      });
+      _initInProgress!.complete(state.value);
+      return state.value;
+    } catch (e) {
+      _initInProgress!.completeError(e);
+      rethrow;
+    } finally {
+      _initInProgress = null;
+    }
   }
 
   /// Performs local login via USP.
--- a/lib/providers/auth/auth_provider.dart
+++ b/lib/providers/auth/auth_provider.dart
@@ -76,10 +76,11 @@ class AuthNotifier extends AsyncNotifier<AuthState> {
       // Propagate error to concurrent waiters if the guard failed.
       if (state case AsyncError(:final error, :final stackTrace)) {
         _initInProgress!.completeError(error, stackTrace);
+        Error.throwWithStackTrace(error, stackTrace);
       } else {
         _initInProgress!.complete(state.value);
+        return state.value;
       }
-      return state.value;
     } finally {
       _initInProgress = null;
     }
--- a/lib/providers/auth/auth_provider.dart
+++ b/lib/providers/auth/auth_provider.dart
@@ -72,11 +72,15 @@ class AuthNotifier extends AsyncNotifier<AuthState> {
           localPassword: localPassword,
         );
       });
-      _initInProgress!.complete(state.value);
+      // AsyncValue.guard never throws — it converts errors to AsyncError.
+      // Propagate error to concurrent waiters if the guard failed.
+      if (state is AsyncError) {
+        final err = state as AsyncError;
+        _initInProgress!.completeError(err.error, err.stackTrace);
+      } else {
+        _initInProgress!.complete(state.value);
+      }
       return state.value;
-    } catch (e) {
-      _initInProgress!.completeError(e);
-      rethrow;
     } finally {
       _initInProgress = null;
     }
--- a/lib/providers/auth/auth_provider.dart
+++ b/lib/providers/auth/auth_provider.dart
@@ -74,9 +74,8 @@ class AuthNotifier extends AsyncNotifier<AuthState> {
       });
       // AsyncValue.guard never throws — it converts errors to AsyncError.
       // Propagate error to concurrent waiters if the guard failed.
-      if (state is AsyncError) {
-        final err = state as AsyncError;
-        _initInProgress!.completeError(err.error, err.stackTrace);
+      if (state case AsyncError(:final error, :final stackTrace)) {
+        _initInProgress!.completeError(error, stackTrace);
       } else {
         _initInProgress!.complete(state.value);
       }
--- a/lib/core/usp/providers/usp_auth_coordinator.dart
+++ b/lib/core/usp/providers/usp_auth_coordinator.dart
@@ -30,6 +30,12 @@ class UspAuthCoordinator {
 
   DateTime? _lastTokenRefresh;
   Completer<void>? _refreshInProgress;
+  Completer<bool>? _restoreInProgress;
+  DateTime? _lastRestoreAttempt;
+  bool? _lastRestoreResult;
+
+  /// Cooldown period after a failed restore attempt to prevent rapid retries.
+  static const Duration _restoreCooldown = Duration(seconds: 5);
 
   /// Called when proactive refresh gets 401 — session externally terminated.
   VoidCallback? onForceLogout;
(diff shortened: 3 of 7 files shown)


AustinChangLinksys (commented):
🤖 **Automated Review — Round 1** · `e84773d..4ffb8c9` (full)

**Verdict: 💬 Self-review (comment only)** — Austin's own PR; GitHub prohibits self-approval. Review is provided as reference / blind-spot check.

|---|---|---|
| 🔴 | `router_provider.dart:124-125` | **[both reviewers]** Removing `redirectLogic` on `/localLoginPassword` lets logged-in user stay on login page; also skips `init()` on direct Web URL bookmark |
| 🔴 | `usp_auth_coordinator_test.dart:~277` | Existing tests assert `login()` always called when `isAuthenticated=true`; new token-first path silently breaks them — CI will go red |
| 🔴 | `auth_provider.dart:~70` | **[both reviewers]** `AsyncValue.guard` never re-throws → outer `catch`/`completeError` is dead code; concurrent `init()` waiters silently receive `null` on error |
| ⚠️ | `usp_internet_settings_notifier.dart:88` + `usp_wifi_settings_provider.dart:79` | **[both reviewers]** Page-layer `restoreSession()` removed; race condition if `init()` hasn't completed before first `performFetch` |
| ⚠️ | `usp_auth_coordinator.dart:~150` | Non-auth `refreshToken` failures still fall through to password login — increases login attempts, undermining the account-lock-reduction goal |
| ⚠️ | `usp_auth_coordinator.dart:~111,~119` | Concurrent callers `await _restoreInProgress!.future` without try-catch; `Completer<bool>` type conflicts with `Future<void>` non-throwing contract |
| ⚠️ | `auth_provider.dart:~55` | `restoreSession()` result discarded; failed restore still yields `AuthState(loginType: LoginType.local)` and routes to unusable Dashboard |
| ⚠️ | `usp_auth_coordinator_test.dart` | `_restoreInProgress` coalescing + `_restoreCooldown` boundary have no test coverage; `AuthNotifier.init()` coalescing has zero test file |
| 💡 | `usp_auth_coordinator.dart:~143` | `refreshToken()` success path doesn't reset `_lastRestoreResult`; stale `false` may incorrectly trigger cooldown on next password-login attempt |
| 💡 | `usp_auth_coordinator.dart:~102` | `restoreSession()` doc comment still describes old unconditional-login strategy; contradicts new token-first behavior |
| 💡 | `usp_auth_coordinator.dart:~33` | `_lastRestoreAttempt`/`_lastRestoreResult` names are ambiguous — rename to `_lastPasswordLoginAttempt`/`_lastPasswordLoginSucceeded` |

<details><summary>🔴 Critical Details</summary>

**C1 · `router_provider.dart:124-125` — /localLoginPassword redirect logic removed [both reviewers]**

Reviewer A (Warning #5) and Reviewer B (Critical #2) both flagged this. Original code:
```dart
} else if (state.matchedLocation == RoutePath.localLoginPassword) {
  router.autoConfigurationLogic(state);  // side-effect: triggers init()
  return router.redirectLogic(state);    // redirects already-logged-in users away
}
```
The new code replaces this with `return state.uri.toString()`, which:
1. Skips `redirectLogic` — already-authenticated user navigating to `/localLoginPassword` stays on login form instead of being redirected to dashboard.
2. Removes `autoConfigurationLogic` side-effect — Web users opening a bookmarked `/localLoginPassword` URL bypass the `'/'` branch entirely, so `init()` and `restoreSession()` are never called.

Note: the old `autoConfigurationLogic(state)` return value was already discarded (not `return`-ed), so its removal is intentional — but `redirectLogic` should be preserved.

**Fix:** `return router.redirectLogic(state);` or add an explicit auth guard before the pass-through.

---

**C2 · `usp_auth_coordinator_test.dart:~277` — Existing tests will fail in CI [Reviewer B]**

Reviewer B read the actual test file and found two tests (lines ~277-310) that were NOT updated in the PR diff but will break under new behavior:
```dart
test('re-logs in even when isAuthenticated=true', () async {
  when(() => mockUsp.isAuthenticated).thenReturn(true);
  // setUp default: mockUsp.refreshToken() stubs to succeed
  ...
  verify(() => mockUsp.login('storedPassword')).called(1);  // 0 calls now -> FAIL
});
```
The default `refreshToken()` stub (success) makes `_restoreSessionImpl()` return `true` in Step 1 without ever calling `login()`.

**Fix:** Delete or update the two legacy tests in the same PR; add `when(() => mockUsp.refreshToken()).thenThrow(...)` stubs to cover the `onReauthRequired` / recovery probe scenario.

---

**C3 · `auth_provider.dart:~70` — catch/completeError is dead code; concurrent waiters get null on error [both reviewers]**

`AsyncValue.guard()` never throws — it converts all exceptions to `AsyncError`. Therefore:
```dart
try {
  state = await AsyncValue.guard(() async { ... });
  _initInProgress!.complete(state.value);  // state.value == null when AsyncError
  return state.value;
} catch (e) {
  _initInProgress!.completeError(e);       // NEVER REACHED
  rethrow;
}
```
When `restoreSession()` or credential storage fails, `state` becomes `AsyncError`, `state.value` is `null`, and `_initInProgress!.complete(null)` is called. Concurrent callers receive `null` and may interpret it as "init succeeded with empty state" — routing logic may incorrectly proceed.

**Fix:**
```dart
if (state is AsyncError) {
  _initInProgress!.completeError(
    (state as AsyncError).error, (state as AsyncError).stackTrace);
} else {
  _initInProgress!.complete(state.value);
}
// Remove the outer catch block entirely
```

</details>

**W1 · internet_settings:88 + wifi_settings:79 — Page-layer restore removed, timing race [both reviewers]**

Both reviewers flagged this. The removed code was a safety net for `performFetch` running before `AuthNotifier.init()` finishes. Common scenarios: WASM page reload, or `build()` triggering a microtask fetch before `restoreSession()` completes. New code immediately throws `ConnectivityError` / returns `NotAuthenticatedError` with no recovery.

**Fix:** Restore the guard (`await restoreSession()` + re-check `isAuthenticated`), OR ensure the router only allows dashboard routes after `init()` is confirmed complete and `usp.isAuthenticated == true`.

---

**W2 · `usp_auth_coordinator.dart:~150` — Non-auth refreshToken failure falls through to password login [Reviewer A]**

```dart
} else {
  logger.w('[USP][Auth]: refreshToken failed (non-auth): $e');
  // Non-auth error (network, etc.) — still try password login as fallback
}
// Falls through to _loginWithStoredPassword() unconditionally
```
During router reboot (transient network error), `refreshToken()` fails with non-auth exception but `isAuthenticated` may still be `true`. This falls through to password login — the exact repeated-login scenario the PR means to prevent.

**Fix:** If `isAuthenticated == true` and error is non-auth, `return false` instead:
```dart
} else {
  if (!_usp!.isAuthenticated) { /* fall through */ }
  else { return false; }  // transient error, token may still be valid
}
```

---

**W3 · `usp_auth_coordinator.dart:~111,~119` — Concurrent error propagation + Completer type mismatch [Reviewer A+B]**

Both reviewers flagged concurrent handling issues:
- Concurrent callers: `await _restoreInProgress!.future` has no try-catch. Unexpected throws from `_restoreSessionImpl()` produce uncaught Future errors (especially since `onReauthRequired` is a `VoidCallback`).
- Type: `Completer<bool>` vs `Future<void>` non-throwing public contract.

**Fix:** Use `Completer<void>`, `complete()` without value, and in catch: log and `complete()` (not `completeError`).

---

**W4 · `auth_provider.dart:~55` — restoreSession() result discarded [Reviewer A]**

```dart
if (loginType == LoginType.local) {
  await ref.read(uspAuthCoordinatorProvider).restoreSession();  // result lost
}
return AuthState(loginType: loginType, ...);  // always LoginType.local regardless of outcome
```
Failed restore still routes user to dashboard where all USP calls fail with `ConnectivityError`, and the page-layer recovery safety net (W1) is also gone.

**Fix:** Check `usp.isAuthenticated` after `restoreSession()` or change its return type to `Future<bool>`.

---

**W5 · `usp_auth_coordinator_test.dart` — Missing tests for coalescing and cooldown [Reviewer B]**

No tests cover: (1) `_restoreInProgress` coalescing (concurrent calls trigger `_restoreSessionImpl()` only once); (2) `_restoreCooldown` boundary (skip within 5s, retry after). No `auth_provider_test.dart` exists at all.

**Fix:** Add test groups for concurrent coalescing and cooldown boundary cases.

</details>

- **Token-first strategy design:** `_restoreSessionImpl()` Step 1 (try `refreshToken()` first) is architecturally sound and achieves the PR's stated goal.
- **Coalescing pattern:** Using a `Completer` to coalesce concurrent calls is the correct Dart idiom. The `try/finally` ensures the gate is always reset.
- **Cooldown mechanism:** 5-second `_restoreCooldown` as a circuit breaker for failed password logins is a reasonable approach.
- **Dependency cleanup:** Removing `uspAuthCoordinatorProvider` imports from page-layer notifiers is architecturally cleaner — the intent is correct; the timing guarantee needs solidifying.
- **New test additions:** Three new tests covering `refreshToken()` success, 401 fallback, and `isAuthenticated=false` (page reload) paths are well-structured. Good use of `verifyNever`.
- **Provider architecture compliance:** `uspAuthCoordinatorProvider` is non-autoDispose; page notifiers use `ref.read`. Layer boundaries (View → Provider → Coordinator) respected. ✅

</details>

AustinChangLinksys (commented):
🤖 **Automated Review — Round 2** · `4ffb8c9..4805e47` (incremental)
**Verdict: 💬 Self-review (comment only)** — Austin's own PR; automated flow will not self-approve. Review provided for reference.

|---|---|---|
| 🔴 | `router_provider.dart:117` | **[both reviewers]** Unawaited `Future<String?>` from `autoConfigurationLogic` — race condition; routing decision runs before `init()` completes |
| 🔴 | `usp_auth_coordinator_test.dart:357` | **[single]** New "cooldown" test verifies behavior not implemented in `UspAuthCoordinator` — CI will fail |
| 🔴 | `usp_auth_coordinator_test.dart:329` | **[single]** New "concurrent coalescing" test verifies behavior not implemented in `restoreSession` — CI will fail |
| ⚠️ | `auth_provider.dart:82` | **[both reviewers]** `rethrow` removed; `init()` now silently returns `null` on error instead of propagating — error state indistinguishable from "not logged in" |
| ⚠️ | `router_provider.dart:116-118` + `demo_router_provider.dart:89-90` | **[single]** Verbatim duplicate logic in two router providers — DRY violation, should be extracted to `loginPageRedirectLogic` helper |
| ⚠️ | `auth_provider.dart` diff+78-82 | **[single]** `completeError` error-propagation path added but has zero test coverage |
| 💡 | `auth_provider.dart:80` | **[both reviewers]** Redundant `as AsyncError` cast after `is AsyncError` check — use Dart 3 pattern matching |
| 💡 | `auth_notifier_test.dart:210` | **[single]** `await Future.delayed(Duration.zero)` is non-deterministic; prefer `await container.read(authProvider.future)` |

<details><summary>🔴 Critical details</summary>

### `router_provider.dart:117` — Unawaited Future race condition
**[both reviewers]**

**Evidence:**
```dart
// New code at lines 117-118
router.autoConfigurationLogic(state);   // Future<String?> return value discarded
return router.redirectLogic(state);
```
`autoConfigurationLogic` internally calls `authCheck` → `init()` (async: reads secure storage to restore `loginType`). `redirectLogic` runs **immediately** and reads `authProvider` **before** `init()` completes — at this point `loginType` is still `LoginType.none`. Result: `redirectLogic` line 177-181 sends even a logged-in user back to login page (`_home()`).

**Fix:**
```dart
return router.autoConfigurationLogic(state);
// or:
final result = await router.autoConfigurationLogic(state);
if (result != null) return result;
return router.redirectLogic(state);
```

---

### `usp_auth_coordinator_test.dart:329` — Coalescing test for unimplemented feature
**[single — Reviewer A]**

**Evidence:** `restoreSession()` (lines 115-121) delegates directly to `_loginWithStoredPassword()` — no Completer or in-flight guard exists for this path. `ensureAuth` has `if (_refreshInProgress != null) return;` but `restoreSession` has no equivalent. The test assertion `verify(...login...).called(1)` will fail because all 3 concurrent calls each trigger a separate `login`.

**Fix:** Implement `_loginInProgress` Completer in `restoreSession`/`_loginWithStoredPassword` before merging, or remove this test until the feature is built.

---

### `usp_auth_coordinator_test.dart:357` — Cooldown test for unimplemented feature
**[single — Reviewer A]**

**Evidence:** No `_lastFailureTime` or `_failureCooldown` field exists anywhere in `UspAuthCoordinator`. The second `restoreSession()` will call `login` again. `verifyNever(() => mockUsp.login(any()))` will fail.

**Fix:** Same as above — implement before merging or remove the test.

</details>

**`auth_provider.dart:82` — Silent error swallow after removing `rethrow`**
**[both reviewers]**

Before: `catch (e) { _initInProgress!.completeError(e); rethrow; }` — exception propagated up the call chain.
After: when `state is AsyncError`, `completeError` is called but `init()` returns `state.value` (null). The caller in `authCheck` does `.then((authState) async { final type = authState?.loginType ?? LoginType.none; ... })` — a null `authState` defaults to `LoginType.none` and takes the same path as "not logged in". Network errors and auth failures become silent.

Consider returning a tagged `AuthState` or checking `authProvider` error state in `authCheck` to distinguish "not logged in" from "init failed".

---

**`router_provider.dart:116-118` + `demo_router_provider.dart:89-90` — Duplicated logic**
**[single — Reviewer B]**

Both router providers now contain identical two-line sequences. Extract to `RouterNotifier.loginPageRedirectLogic(state)`.

---

**`auth_provider.dart` — `completeError` path has no test coverage**
**[single — Reviewer B]**

The new `state is AsyncError` branch is the core fix of this PR but is not exercised by any test case. `auth_notifier_test.dart` only covers: success path, non-local login, and storage failure (which is mapped to `LoginType.none` before reaching `AsyncError`). Add a test that verifies concurrent waiters receive `completeError` when `restoreSession()` throws.

</details>

- `auth_provider.dart` — The conceptual fix is correct: `AsyncValue.guard` never throws, so the old `catch` block was dead code. Switching to `state is AsyncError` check is the right approach.
- `test/core/usp/providers/usp_auth_coordinator_test.dart` — The intent to test concurrency and cooldown behavior is good; the tests just need their corresponding production implementations.
- `test/providers/auth/auth_notifier_test.dart` — The concurrent-coalescing test structure is correct and adds meaningful coverage for the `_initInProgress` Completer path.
- `context.mounted` checks are present before navigation calls.

</details>

AustinChangLinksys (commented):
🤖 **Automated Review — Round 3** · `4805e47..b4cacf9` (incremental)
**Verdict: 💬 Self-review (comment only)** — Useful feedback for the author; GitHub prohibits self-approval.

|---|---|---|
| 🔴 | `auth_provider.dart:77` | `Error.throwWithStackTrace` breaks existing fire-and-forget `auth.init()` call sites → unhandled Future rejection |
| 🔴 | `auth_notifier_test.dart` | New test references undeclared variables (`error1/2/3`, `callCount`) and has no mock throwing setup — likely fails to compile |
| ⚠️ | `auth_provider.dart` | Return type `Future<AuthState?>` now misleading — `null`-as-error contract silently broken by throwing on error |
| ⚠️ | `auth_notifier_test.dart` | `Future.wait` ordering is load-bearing for coalescing but undocumented — fragile if `init()` gains async step before the guard |
| ⚠️ | `auth_notifier_test.dart` | No happy-path concurrent test — success-case concurrent callers are uncovered |
| 💡 | `auth_notifier_test.dart` | Add explicit assertion that primary caller (not just waiters) also throws post this change |

<details><summary>🔴 Critical Details</summary>

**`auth_provider.dart:77` — `Error.throwWithStackTrace` breaks fire-and-forget call sites**

Before: `init()` on error called `completeError` (notifying waiters) then `return state.value` (returned `null` to the primary caller — harmless for fire-and-forget callers).

After: `Error.throwWithStackTrace(error, stackTrace)` makes `init()` **reject its Future** on error. Any call site that does not `await` or chain `.catchError` will produce an **unhandled Future rejection** — in Flutter this triggers `FlutterError.onError` / crash reporters.

At least one fire-and-forget usage exists in the codebase (`auth.init()` without await). **Before merging:** grep all `authNotifier.init()` / `ref.read(authProvider.notifier).init()` call sites and add `.catchError` or `unawaited(...)` where appropriate.

---

**`auth_notifier_test.dart` — New test likely fails to compile**

```dart
test('concurrent init waiters receive error via completeError', () async {
  // error1, error2, error3 — undeclared in this test body or enclosing group
  // callCount — undeclared
  // expect(state.hasError, isTrue) — bare `state` is unresolved
  // No when(...).thenThrow(StateError(...)) mock setup
```

These variables appear used but are not declared in the visible diff. Without mock setup to actually throw `StateError`, all three `catchError` handlers will receive `null` and `isA<StateError>()` assertions fail.

**Fix:** Declare `Object? error1, error2, error3; var callCount = 0;` and add inside the test:
```dart
when(() => mockAuthService.getStoredLoginType())
    .thenAnswer((_) async {
      callCount++;
      await Future.delayed(const Duration(milliseconds: 50));
      throw StateError('Simulated storage failure');
    });
```
</details>

**`auth_provider.dart` — Misleading return type**

```dart
} else {
  _initInProgress!.complete(state.value);
  return state.value;  // only reachable on success
}
// error branch throws — never returns null
```

The declared return type `Future<AuthState?>` implies callers can receive `null` to signal failure. That contract is now broken: on error the method throws. Callers relying on `if (result == null)` will silently miss errors.

**Fix:** Update return type to `Future<AuthState>` (non-nullable) to reflect the new throwing behavior.

---

**`auth_notifier_test.dart` — `Future.wait` ordering is load-bearing**

The coalescing logic works deterministically today because `init()` does not `await` before the `_initInProgress` guard check. If `init()` ever adds an async step before the guard, `callCount` could silently become 3.

**Fix:** Add comment: `// Dart single-threaded scheduler guarantees the first call acquires _initInProgress before the others check`.

---

**`auth_notifier_test.dart` — No happy-path concurrent test**

Only the error path is covered. A bug in `_initInProgress!.complete(state.value)` would have no test catching it.

**Fix:** Add `'concurrent init waiters all receive the same AuthState on success'` test.
</details>

- The Dart pattern-matching refactor (`if (state case AsyncError(:final error, :final stackTrace))`) is cleaner and more idiomatic than the old `is AsyncError` + explicit cast.
- Moving `return state.value` inside the `else` branch correctly prevents it from running after an error — this was a latent correctness bug in the original code.
- The new test validates the concurrent error-propagation path which was previously untested — good regression coverage direction.
- The `callCount == 1` assertion correctly verifies the coalescing invariant.
</details>

lib/providers/auth/auth_provider.dart
```diff
@@ -21,46 +23,67 @@ final authProvider =
...
+          localPasswordHint: state.value?.localPasswordHint,
+          loginType: loginType,
+          localPassword: localPassword,
+        );
+      });
+      // AsyncValue.guard never throws — it converts errors to AsyncError.
+      // Propagate error to concurrent waiters if the guard failed.
+      if (state case AsyncError(:final error, :final stackTrace)) {
+        _initInProgress!.completeError(error, stackTrace);
+        Error.throwWithStackTrace(error, stackTrace);
```
HankYuLinksys: `init()` now rethrows on failure (`Error.throwWithStackTrace`), but neither caller catches it.

Previously `init()` could **never** throw: `AsyncValue.guard` converts any error into an `AsyncError` and the method returned `state.value` (null on error). With this change, when the guard captures an error, `init()` both records the `AsyncError` **and** rethrows. The two callers use a bare `.then()` with no `.catchError`:

- `lib/app.dart:254` — `init().then((_) { FlutterNativeSplash.remove(); })`. On rejection the splash is **never removed**, leaving the app stuck on the splash screen, plus an unhandled async error.
- `lib/route/router_provider.dart:266` — `init().then((authState) async { ... })` is the GoRouter redirect's return value; a rejected future breaks the redirect.

The trigger path is narrow today (the guard body only fails if `restoreSession()` throws, and it currently swallows everything internally), but you added a test specifically for this path (`concurrent init waiters receive error via completeError`), so it's a real reachable contract.

Suggest keeping the original "init never throws" contract: still `completeError` the concurrent waiters' completer, but let the first caller fall through and `return state.value` instead of rethrowing. Otherwise, both callers need a `.catchError`, and `app.dart` must still call `FlutterNativeSplash.remove()` on the error branch — otherwise the splash hangs.

lib/page/wifi_settings/providers/usp_wifi_settings_provider.dart
```diff
@@ -77,15 +76,12 @@ class UspWifiSettingsNotifier extends AutoDisposeNotifier<UspWifiSettingsState>
     }
 
     if (!usp.isAuthenticated) {
```
HankYuLinksys, line 78: Removing the `restoreSession()` self-heal here weakens deep-link reload recovery, and the stated justification is slightly off.

The PR says auth is "already handled by orchestrator", but only `usp_dashboard_view` watches `dashboardOrchestratorProvider` — this provider does **not** depend on it. On a Web reload directly into `/uspWifiSettings` (or `/uspAdvancedSettings/uspInternetSettings`), WASM in-memory state is gone, and the only auth restore left is `app.dart`'s `_initAuth()` / the router's `authCheck → init()`.

In practice the `/usp` redirect guard (`router_provider.dart:160-165`) blocks navigation until `loginType` is set, and `loginType` is only set after `init()` (which awaits `restoreSession()`) completes — so the ordering is likely safe. But that guarantee is now **implicit and external** to these providers, and the previous behavior of retrying a *failed* first restore (e.g. transient bridge error during reload) is gone: now the provider just returns `NotAuthenticated`.

Two requests: (1) confirm a direct deep-link reload into wifi/internet always completes `init()` before the provider builds; (2) add a comment in these providers documenting that auth restoration is a precondition guaranteed by the router/init, since it's no longer self-contained.

AustinChangLinksys (commented):
🤖 **Automated Review — Round 4** · `b4cacf9..3e593eb` (incremental)
**Verdict: 💬 Self-review (comment only)** — Austin's own PR; automated flow cannot self-approve. Full cross-review provided for reference.

|---|---|---|
| 🔴 | `auth_provider.dart:82` | Provider swallows error to accommodate callers' bare `.then()` — architecture inversion |
| ⚠️ | `auth_provider.dart:82` + `router_provider.dart:220` | **[both reviewers]** Init failure silently degrades to `LoginType.none` — real errors fully masked |
| ⚠️ | `login_local_view.dart:85` | **[both reviewers]** Fire-and-forget `auth.init()` — unhandled Future error when concurrent secondary caller receives `completeError` |
| ⚠️ | `auth_notifier_test.dart:247` | **[both reviewers]** `expect(result1, isNull)` present but missing `authProvider.hasError == true` state assertion — regression gap |
| 💡 | `auth_provider.dart` (signature) | **[both reviewers]** `init()` lacks dartdoc declaring never-throws contract; `null` return is ambiguous (error vs. success with null state) |
| 💡 | `auth_notifier_test.dart:209` | Test name still reads "concurrent waiters receive error" but primary behavioral change is primary-caller now returns null |

*Items marked [both reviewers] were independently flagged by two agents — higher confidence.*

<details>
<summary>🔴 Critical Details</summary>

### C1 — Architecture Inversion: Provider lowers error contract to accommodate callers' omissions

**File:** `lib/providers/auth/auth_provider.dart:82`

**Evidence:**
```
// Note: Do NOT rethrow here — callers (app.dart, router_provider.dart)
// use bare .then() without .catchError, so throwing would leave the
// splash screen stuck or break the redirect.
if (state case AsyncError(:final error, :final stackTrace)) {
    _initInProgress!.completeError(error, stackTrace);
    return null;   // NEW: was Error.throwWithStackTrace(error, stackTrace)
}
```

**Reviewer B (Architecture) framing:** `AuthNotifier` is an L1 Provider. It must not embed assumptions about how View/Router layers consume its API. Having the Provider silently swallow its own error to compensate for callers that use bare `.then()` without `.catchError` is an inverted dependency: the lower layer is constrained by the upper layer's deficiency.

This also creates an **asymmetric error API**:
- **Primary caller**: receives `null` on error — error invisible
- **Concurrent secondary callers**: receive the real exception via `completeError` — error visible

**Reviewer A (Correctness) framing:** `router_provider.dart:220-225` uses `authState?.loginType ?? LoginType.none` — null from a failed init silently becomes "not logged in", and the user is routed to the login page with no indication of the real failure.

**Fix:** Restore `Error.throwWithStackTrace(error, stackTrace)` in `auth_provider.dart`, then add `.catchError` at the call sites:
```
// app.dart:
ref.read(authProvider.notifier).init()
  .then((_) { ... })
  .catchError((e, st) { logger.e('[Auth]: init failed', error: e, stackTrace: st); FlutterNativeSplash.remove(); });

// router_provider.dart:
return ref.read(authProvider.notifier).init().then((authState) async {
  ...
}).catchError((e) => _home(state.uri.query));
```

</details>

<details>

### W1 — Silent error-to-unauthenticated degradation in router [both reviewers]

**File:** `lib/providers/auth/auth_provider.dart:82` — impact at `lib/route/router_provider.dart:220-225`

```
return ref.read(authProvider.notifier).init().then((authState) async {
  final type = authState?.loginType ?? LoginType.none;  // null = LoginType.none
  return switch (type) {
    LoginType.local => await _prepare(...),
    _ => _home(state.uri.query),  // silently routes to login on error
  };
});
```

When `init()` fails, `authState` is `null`, `LoginType.none` is assumed, and the user silently lands on the login page. No error surfaced.

**Fix:** In `router_provider.dart`, observe `authProvider` state after `init()` returns:
```
.then((authState) async {
  final current = _ref.read(authProvider);
  if (current is AsyncError) { /* route to error/retry */ }
  ...
```

---

### W2 — `login_local_view.dart:85` fire-and-forget causes unhandled Future error [both reviewers]

**File:** `lib/page/login/views/login_local_view.dart:85`

```
if (delayTime == null) {
  auth.init();  // fire-and-forget, no await, no .then, no .catchError
```

If an in-flight `init()` exists when this runs, this call returns `_initInProgress!.future`. When the primary init fails via `completeError(error, stackTrace)`, this unawaited future carries an unhandled error — a Dart VM Unhandled Future Error, which can trigger Crashlytics noise.

**Fix:**
```
auth.init().catchError((_) {
  // init errors are reflected in authProvider state; no action needed here
});
```

---

### W3 — Test missing `authProvider.hasError` assertion after error path [both reviewers]

**File:** `test/providers/auth/auth_notifier_test.dart:247`

```
// BEFORE:  expect(error1, isA<StateError>());
// AFTER:   expect(result1, isNull);
// MISSING: expect(container.read(authProvider).hasError, isTrue);
```

If `AsyncValue.guard` were accidentally removed in a future refactor, state would remain `AsyncLoading` — but this test would still pass because `result1` is still null. The state contract is not covered.

**Fix:** Add:
```
final authAsyncValue = container.read(authProvider);
expect(authAsyncValue.hasError, isTrue);
expect(authAsyncValue.error, isA<StateError>());
```

</details>

<details>

- The Completer-based coalescing pattern (`_initInProgress`) is sound and correctly prevents redundant `init()` calls.
- Removing `Error.throwWithStackTrace` does fix the splash-stuck / redirect-broken bug for the current call sites.
- The new comment block clearly explains the Dart single-threaded event-loop guarantee for the Completer guard.
- Test coverage for the concurrent coalescing case is present; renaming `error1`/`result1` is logically consistent.
- `@@ -222,14 +224,16 @@` hunk is clean.

</details>

HankYuLinksys (approved):
Looks good!

PR #987, opened 2026-06-25
feat(remote): SSE & subscription support for Remote Assistance mode

AustinChangLinksys:
## Summary
- Enable SSE and subscription functionality for Remote Assistance mode via Guardian proxy
- Introduce Strategy pattern to cleanly separate Local (usp-bridge) vs Remote (Guardian) behaviors
- Fix Guardian endpoint paths, auth token handling, and subscription ID conflicts

### Strategy Pattern for SSE Operations
| Behavior | Local | Remote |
|----------|-------|--------|
| Register | Direct (bridge idempotent) | Unregister → delay → register |
| onSseConnected | Auto resubscribe | No-op (orchestrator controls) |
| Heartbeat watchdog | Enabled (45s) | Disabled |
| Auth check | Proactive refresh | Disabled (no refresh) |
| Cleanup on exit | Bridge handles | Fire-and-forget unregisterAll |

### Bug Fixes
- Add `/v1/guardians` prefix to Guardian endpoints
- Pass `baseUrl` to `UspClient.fromBuilder` for Remote mode
- Add `X-Linksys-Client-Type-Id` header for Guardian API calls
- Add `listSubscriptions()` API for subscription management

## Test Plan
- [x] All 2961 unit tests pass
- [x] New strategy tests cover Local/Remote behaviors (20 tests)
- [ ] Manual test: Remote Assistance SSE connection
- [ ] Manual test: Subscription register/unregister via Guardian
- [ ] Manual test: SSE notifications received in Remote mode

### DIFF
--- a/lib/core/usp/providers/remote_assistance_provider.dart
+++ b/lib/core/usp/providers/remote_assistance_provider.dart
@@ -89,7 +89,8 @@ class RemoteAssistanceNotifier extends Notifier<RemoteAssistanceState> {
     }
 
     final jsClient = builder.build();
-    final raClient = UspClient.fromBuilder(jsClient);
+    final guardianUrl = 'https://${config.guardianBaseUrl}';
+    final raClient = UspClient.fromBuilder(jsClient, baseUrl: guardianUrl);
 
     // Swap UspClient atomically with mutation lock to prevent races
     await ref.read(uspMutationLockProvider).withLock(() async {
--- a/lib/core/usp/providers/sse_providers.dart
+++ b/lib/core/usp/providers/sse_providers.dart
@@ -9,14 +9,28 @@ import 'package:privacy_gui/core/usp/services/network_diagnostics_executor.dart'
 import 'package:privacy_gui/core/usp/services/sse_connection_manager.dart';
 import 'package:privacy_gui/core/usp/services/sse_manager.dart';
 import 'package:privacy_gui/core/usp/services/sse_operation_awaiter.dart';
+import 'package:privacy_gui/core/usp/providers/remote_assistance_provider.dart';
+import 'package:privacy_gui/core/usp/services/bridge_endpoints.dart';
 import 'package:privacy_gui/core/usp/services/usp_bridge_client.dart';
 import 'package:privacy_gui/config/global_config.dart';
 import 'package:privacy_gui/providers/auth/auth_provider.dart';
 
 /// Provides [UspBridgeClient] instance — depends on [UspClient].
+///
+/// In Remote Assistance mode, uses Guardian proxy endpoints.
+/// In local mode, uses on-router usp-bridge endpoints.
 final uspBridgeClientProvider = Provider<UspBridgeClient?>((ref) {
   final usp = ref.watch(uspClientProvider);
   if (usp == null) return null;
+
+  if (GlobalConfig.remote.isActive) {
+    final raState = ref.watch(remoteAssistanceProvider);
+    final config = raState.config;
+    if (config == null) return null;
+    final endpoints = BridgeEndpoints.remote(config.sessionId);
+    return UspBridgeClient(usp, endpoints: endpoints);
+  }
+
   return UspBridgeClient(usp);
 });
 
@@ -112,12 +126,6 @@ final networkDiagnosticsExecutorProvider =
 /// Watch this from the app shell to trigger SSE connection after login.
 /// Core subscriptions are "always-on" while the app is connected.
 final sseBootstrapProvider = FutureProvider<void>((ref) async {
-  // Skip SSE in Remote Assistance mode — Guardian proxy does not support SSE
-  if (GlobalConfig.remote.isActive) {
-    logger.d('[USP][SSE][Bootstrap]: Skipping in Remote Assistance mode');
-    return;
-  }
-
   final manager = ref.watch(sseManagerProvider);
   if (manager == null) return;
 
@@ -127,15 +135,18 @@ final sseBootstrapProvider = FutureProvider<void>((ref) async {
   final bridge = ref.watch(uspBridgeClientProvider);
   if (bridge == null) return;
 
-  // Step 0: Health check — best-effort, non-fatal.
+  // Step 0: Health check — best-effort, non-fatal (local mode only).
   // If the bridge is busy (504) or slow, we still attempt SSE connection
   // because SseConnectionManager has its own retry/backoff logic.
-  try {
-    await bridge.health().timeout(const Duration(seconds: 5));
-    logger.d('[USP][SSE][Bootstrap]: Bridge health check passed');
-  } catch (e) {
-    logger.w(
-        '[USP][SSE][Bootstrap]: Bridge health check failed: $e — continuing');
+  // Skip in Remote mode — Guardian proxy has no health endpoint.
+  if (!GlobalConfig.remote.isActive) {
+    try {
+      await bridge.health().timeout(const Duration(seconds: 5));
+      logger.d('[USP][SSE][Bootstrap]: Bridge health check passed');
+    } catch (e) {
+      logger.w(
+          '[USP][SSE][Bootstrap]: Bridge health check failed: $e — continuing');
+    }
   }
 
   // Connect SSE only — core subscriptions are registered by the dashboard
--- a/lib/core/usp/providers/sse_providers.dart
+++ b/lib/core/usp/providers/sse_providers.dart
@@ -7,8 +7,11 @@ import 'package:privacy_gui/core/usp/providers/usp_auth_coordinator.dart';
 import 'package:privacy_gui/core/usp/providers/usp_client_provider.dart';
 import 'package:privacy_gui/core/usp/services/network_diagnostics_executor.dart';
 import 'package:privacy_gui/core/usp/services/sse_connection_manager.dart';
+import 'package:privacy_gui/core/usp/services/sse_local_strategy.dart';
 import 'package:privacy_gui/core/usp/services/sse_manager.dart';
 import 'package:privacy_gui/core/usp/services/sse_operation_awaiter.dart';
+import 'package:privacy_gui/core/usp/services/sse_operation_strategy.dart';
+import 'package:privacy_gui/core/usp/services/sse_remote_strategy.dart';
 import 'package:privacy_gui/core/usp/providers/remote_assistance_provider.dart';
 import 'package:privacy_gui/core/usp/services/bridge_endpoints.dart';
 import 'package:privacy_gui/core/usp/services/usp_bridge_client.dart';
@@ -28,7 +31,12 @@ final uspBridgeClientProvider = Provider<UspBridgeClient?>((ref) {
     final config = raState.config;
     if (config == null) return null;
     final endpoints = BridgeEndpoints.remote(config.sessionId);
-    return UspBridgeClient(usp, endpoints: endpoints);
+    return UspBridgeClient(
+      usp,
+      endpoints: endpoints,
+      authToken: config.temporaryAccessToken,
+      clientTypeId: config.clientTypeId,
+    );
   }
 
   return UspBridgeClient(usp);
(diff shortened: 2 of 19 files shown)


AustinChangLinksys (commented):
🤖 **Automated Review — Round 1** · `b6057ed..99ece64` (full)

**Verdict: 💬 Self-review (comment only)** — Austin's own PR; GitHub prohibits self-approve. Full review included for blind-spot reference.

|---|---|---|
| ⚠️ | `usp_bridge_client_web.dart:56–67` | **[both reviewers]** `_withAuthRetry` calls `_usp.reauth()` in Remote mode — refreshes wrong (local WASM) token, retry still 401, then `parser()` invoked on 401 response |
| ⚠️ | `sse_remote_strategy.dart` (new) | `registerSubscriptions`: `_bridge.subscribe()` outside try-catch — first failure aborts remaining subscriptions silently |
| ⚠️ | `sse_remote_strategy.dart` (new) | `_fireAndForgetCleanup()` swallows all errors; unintentional disconnects never trigger cleanup, ghost subscriptions accumulate |
| ⚠️ | `sse_providers.dart` | Remote mode: `config == null` transition causes `uspBridgeClientProvider` null, `sseManagerProvider` dispose/reconnect race |
| ⚠️ | `usp_bridge_client_web.dart:176` | Debug log leaks first 20 chars of `temporaryAccessToken` (cloud OAuth token) — was safe before (local JWT), now a security concern |
| ⚠️ | `usp_bridge_client_web.dart:144` | `notificationsProbe()` still uses hardcoded `/api/v1/notifications` — Remote mode probes wrong endpoint, misleads diagnostics |
| ⚠️ | `sse_providers_test.dart` | Remote mode branches in `uspBridgeClientProvider` / `sseManagerProvider` / `sseBootstrapProvider` have zero test coverage (file not in this PR) |
| 💡 | `usp_bridge_client_web.dart:179–186` | SSE fetch now uses `_authHeaders.forEach` — `Content-Type: application/json` on a GET SSE request is semantically odd (Accept header preserved via `headers.set`) |
| 💡 | `remote_assistance_provider.dart:90` | `guardianBaseUrl` inserted into URL without format validation |
| 💡 | `bridge_endpoints.dart` | `sessionId` embedded in URL path without UUID format assertion |
| 💡 | `sse_local_strategy.dart` + `sse_subscription_registry.dart` | `_notifTypeToInt` mapping duplicated in two places |

*Items marked [both reviewers] were independently flagged by two agents — higher confidence.*

<details>

**W-1 · Auth retry refreshes wrong token in Remote mode** [both reviewers]

`lib/core/usp/services/usp_bridge_client_web.dart:56–67`:
```dart
Future<T> _withAuthRetry<T>(
  Future<http.Response> Function() request,
  T Function(http.Response) parser,
) async {
  var response = await request();
  if (response.statusCode == 401) {
    await _usp.reauth();        // refreshes local WASM/router JWT, NOT Guardian token
    response = await request(); // still uses _overrideToken (Guardian temporaryAccessToken)
  }
  return parser(response);    // called even if retry is still 401
}
```

After the PR, `_token` returns `_overrideToken` (the Guardian `temporaryAccessToken`) in Remote mode. `_usp.reauth()` refreshes the local router's WASM session token — it has no effect on the Guardian token. When `temporaryAccessToken` expires: (1) any REST call returns 401, (2) `_withAuthRetry` calls `_usp.reauth()` — wrong refresh, (3) retry still 401, (4) `parser(response)` called on 401 body — silent bad response.

Same issue in SSE path (`_startSseStream:201`): recursive reauth loop until `authRetryCount >= 1`, then `controller.addError` → `SseConnectionManager` retries → eventually suspended → all Remote mode notifications silently lost.

**Fix:** Add `onTokenExpired` callback on `UspBridgeClient`, wired by `sse_providers.dart` to trigger `remoteAssistanceProvider` session refresh. Guard `parser()` against 401 on retry.

---

**W-2 · `RemoteSseStrategy.registerSubscriptions` missing try-catch on subscribe**

`lib/core/usp/services/sse_remote_strategy.dart` (new file):
```dart
for (final sub in subscriptions) {
  try {
    await _bridge.unsubscribe(subscriptionId: sub.subscriptionId);
    await Future.delayed(_unregisterDelay);
  } catch (_) { }
  // NO try-catch on subscribe — if this throws, loop aborts, remaining subscriptions are skipped
  await _bridge.subscribe(
    subscriptionId: sub.subscriptionId,
    path: sub.referenceList,
    notifType: _notifTypeToInt(sub.notifType),
  );
}
```

`LocalSseStrategy` correctly wraps each `subscribe` in individual try-catch and continues. A single Guardian 503 or network timeout silently drops all remaining subscriptions in Remote mode.

**Fix:** Wrap `_bridge.subscribe()` in try-catch (same pattern as `LocalSseStrategy`).

---

**W-3 · `_fireAndForgetCleanup` swallows all errors; unintentional disconnect skips cleanup**

`lib/core/usp/services/sse_remote_strategy.dart`:
```dart
Future<void> onSseDisconnected({required bool intentional}) async {
  if (intentional) {
    _fireAndForgetCleanup();  // only on intentional disconnect
  }
  // unintentional (network drop, token expiry — the common case) → nothing
}

void _fireAndForgetCleanup() {
  _bridge.listSubscriptions().then((ids) {
    for (final id in ids) {
      _bridge.unsubscribe(subscriptionId: id).ignore();
    }
  }).ignore(); // listSubscriptions failure also silently dropped
}
```

Three issues: (1) Network drops / token expiry (unintentional) never clean up Guardian subscriptions — ghost subs accumulate. (2) All errors silently swallowed — no observability. (3) If `dispose()` races with in-flight cleanup, `_bridge` may be used after disposal.

**Fix:** Add `.catchError` for logging. Consider cleaning up on unintentional disconnect too, or tracking subs for re-registration on reconnect.

---

**W-4 · Remote mode `config == null` transition triggers SSE Manager dispose/reconnect race**

`lib/core/usp/providers/sse_providers.dart`:
```dart
final uspBridgeClientProvider = Provider<UspBridgeClient?>((ref) {
  if (GlobalConfig.remote.isActive) {
    final config = ref.watch(remoteAssistanceProvider).config;
    if (config == null) return null;  // any intermediate null state → Manager disposes
    return UspBridgeClient(usp, endpoints: ..., authToken: config.temporaryAccessToken, ...);
  }
  return UspBridgeClient(usp);
});
```

If `RemoteAssistanceState` transitions through a `config == null` intermediate state (token rotation, provider rebuild), `sseManagerProvider` triggers `ref.onDispose → manager.dispose()`. The async `dispose()` calls `await registry.unregisterAll()` while a new manager's `connect()` may already be running — concurrent subscription state mutation.

**Fix:** Use `select` to only react to meaningful session changes:
```dart
final config = ref.watch(remoteAssistanceProvider.select((s) => s.config));
```

---

**W-5 · Debug log leaks cloud OAuth token prefix**

`lib/core/usp/services/usp_bridge_client_web.dart:175–176` (unchanged code, new security impact):
```dart
final token = _token;  // After PR: returns _overrideToken (Guardian temporaryAccessToken) in Remote mode
debug('Token: ${token.substring(0, 20)}...(${token.length} chars)');
// emitted as SseEvent(event: '_debug') into the SSE stream → logger.d
```

Pre-PR this leaked the local router session JWT (limited scope). Post-PR in Remote mode it leaks the first 20 chars of the Guardian cloud OAuth token. **Fix:** `debug('Token: (present, ${token.length} chars)');`

---

**W-6 · `notificationsProbe()` uses hardcoded endpoint — wrong URL in Remote mode**

`lib/core/usp/services/usp_bridge_client_web.dart:144`:
```dart
Future<Map<String, String>> notificationsProbe() async {
  final response = await http.get(
    Uri.parse('$_baseUrl/api/v1/notifications'),  // hardcoded — NOT migrated to _endpoints
    ...
  );
```

All other endpoints were migrated to `_endpoints.xxx`. `notificationsProbe()` was missed. In Remote mode this probes the local bridge endpoint (404 / wrong server), causing misleading diagnostics.

**Fix:** `Uri.parse('$_baseUrl${_endpoints.notifications}')`

---

**W-7 · `sse_providers_test.dart` has no Remote mode coverage; not in this PR**

The PR modifies `sse_providers.dart` with three Remote-mode branches. `test/core/usp/providers/sse_providers_test.dart` is not in this PR's diff, and on-disk tests have zero references to `remote`, `isActive`, or `GlobalConfig`. All existing health-check tests assume `bridge.health()` is always called.

Recommend adding:
- `uspBridgeClientProvider` Remote mode: bridge with Guardian endpoints when config present, null when absent
- `sseManagerProvider`: `RemoteSseStrategy` selected when `GlobalConfig.remote.isActive`
- `sseBootstrapProvider` Remote mode: health skipped, connect still called

</details>

<details>

- **Strategy pattern is clean**: `SseOperationStrategy` correctly encapsulates behavioral differences between local and remote SSE. Clear interface, well-separated implementations.
- **New strategy tests included**: `sse_local_strategy_test.dart` (206 lines) and `sse_remote_strategy_test.dart` (174 lines) added with good coverage of happy paths and error paths.
- **`BridgeEndpoints` value object**: Clean immutable config. Eliminates scattered hardcoded `/api/v1/` strings across `UspBridgeClient`.
- **`_registrationInProgress` guard**: Prevents double-registration on concurrent `registerCoreSubscriptions()` calls — real improvement over the old `force` flag.
- **Mutation lock preserved**: `remote_assistance_provider.dart` still wraps `UspClient` swap in `uspMutationLockProvider.withLock()` — correct.
- **`UspClient.fromBuilder` now requires `baseUrl`**: Good fix; previous `_baseUrl = ''` was silently broken for Remote mode.
- **Heartbeat config injection**: Generalizing the 45s watchdog into `HeartbeatConfig` is the right abstraction; properly disables watchdog for Remote mode (Guardian sends no heartbeats).
- **`unregisterAll` in-memory-first pattern**: Clears the map before calling the strategy, preventing double-unregister on recursive calls.

</details>

HankYuLinksys (commented):
Reviewed the full diff against the PR branch. The Strategy pattern split is clean and `HeartbeatConfig`/`BridgeEndpoints` nicely remove the scattered magic strings. A few things I noticed that are worth a look.

### Already addressed (confirmed against HEAD) ✅
- **W-2** `RemoteSseStrategy.registerSubscriptions` — each `subscribe()` is now wrapped in try-catch, so one failure no longer aborts the remaining subscriptions.
- **W-5** Token-prefix debug log is gone — no longer leaks the Guardian `temporaryAccessToken`.
- **W-6** `notificationsProbe()` now uses `_endpoints.notifications` instead of the hardcoded path.

### Observations

**W-1 · Auth retry refreshes the wrong token in Remote mode** — `usp_bridge_client_web.dart` (`_withAuthRetry`, and the SSE path in `_startSseStream`)
In Remote mode `_token` returns `_overrideToken` (the Guardian `temporaryAccessToken`), but on 401 `_withAuthRetry` calls `_usp.reauth()`, which refreshes the local WASM/router session token — it has no effect on the Guardian token. So when the Guardian token expires: the retry is still 401, and `parser()` then runs on the 401 body (possible parse failure). The SSE path hits the same thing and ends up in `addError`, after which Remote notifications stop arriving.

**W-3 · `_fireAndForgetCleanup` only runs on intentional disconnect** — `sse_remote_strategy.dart`
Tracing the call chain, `registry.onSseDisconnected(intentional: false)` is never actually invoked — `SseConnectionManager.onDisconnected` only logs. So unintentional disconnects (network drop, token expiry — the common case) never clean up Guardian subscriptions, and ghost subs can accumulate. The errors inside `_fireAndForgetCleanup` are also all `.ignore()`'d, so there's no observability when cleanup fails.

**W-4 · `config == null` intermediate state can race SseManager dispose/connect** — `sse_providers.dart`
`uspBridgeClientProvider` does `ref.watch(remoteAssistanceProvider)` on the whole state. If `RemoteAssistanceState` passes through a `config == null` intermediate (e.g. token rotation / provider rebuild), the bridge provider returns null → `sseManagerProvider` disposes the manager (which `await registry.unregisterAll()`) while a new manager may already be running `connect()`. A `select((s) => s.config)` would narrow the rebuild trigger to meaningful session changes.

### Follow-up (non-blocking)
- **W-7** the three Remote-mode branches in `sse_providers.dart` (bridge provider, strategy selection, bootstrap health-skip) have no test coverage — the strategies themselves are tested, but not the provider wiring that picks them.
- nit: `_notifTypeToInt` is duplicated in `sse_local_strategy.dart` and `sse_remote_strategy.dart` — could move to a shared place.

HankYuLinksys (approved):
LGTM ✅ The new commit cleanly addresses the 401 handling — the `AuthBehavior` split (Local reauth+retry, Remote no-retry → `onAuthFailed`) resolves the wrong-token issue, and the `select((s) => s.config)` change narrows the rebuild trigger. Approving.

One small observation (non-blocking): the new `bridge.onAuthFailed` in `sse_providers.dart` calls `authProvider.notifier.logout()` directly, bypassing the `logoutTriggered` guard used by the existing `forceLogout` path. I traced the two paths and they're actually mutually exclusive on a single request (reauth throwing → `onForceLogout`; reauth "succeeding" but the retry still 401 → `onAuthFailed`), so it's not a double-fire on one call. The real exposure is concurrency: SSE REST calls (health + subscription, etc.) can hit 401 in parallel, and since `onAuthFailed` has no guard, each one can independently call `logout()` — so logout may fire multiple times. Routing `onAuthFailed` through the same guarded `forceLogout` would dedupe it.

PR #993, opened 2026-06-26
fix(sse): add 'remote-' prefix to Remote subscription IDs

AustinChangLinksys:
## Summary

- Add `remote-` prefix to Remote subscription IDs to avoid collision with Local mode when both connect to the same router

Closes #992

## Test plan

- [ ] Local mode SSE subscriptions work independently
- [ ] Remote mode SSE subscriptions use `remote-` prefix
- [ ] `_fireAndForgetCleanup()` only removes Remote subscriptions
- [ ] Both modes can coexist on same router without ID conflicts

### DIFF
--- a/lib/core/usp/services/sse_remote_strategy.dart
+++ b/lib/core/usp/services/sse_remote_strategy.dart
@@ -12,13 +12,20 @@ import 'usp_bridge_client.dart';
 /// - No auto resubscribe on reconnect (orchestrator controls)
 /// - Heartbeat watchdog disabled (Guardian doesn't send heartbeats)
 /// - No auth check (temporaryAccessToken cannot refresh)
+/// - Uses 'remote-' prefix for subscription IDs to avoid collision with Local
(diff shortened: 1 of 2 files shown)


HankYuLinksys (approved):
LGTM ✅ Clean, focused fix for the Local/Remote subscription ID collision (#992).

Verified the main correctness risk — adding the `remote-` prefix does not break notification routing:
- The prefix is fully encapsulated inside `RemoteSseStrategy`; `SseSubscriptionRecord` still stores the original ID, so Registry/Manager/Router are unaffected.
- Remote notifications route via the wildcard handler + path-prefix match (`type` + `referenceList`), not by `subscription_id`, so prefixing the bridge-side ID can't cause handlers to miss events.
- `_fireAndForgetCleanup` now only unsubscribes `remote-`-prefixed IDs — this also closes the earlier risk of cleanup touching Local subscriptions.

Tests updated consistently, plus a good new case asserting cleanup only removes remote-prefixed subs out of a mixed list. Approving.

PR #1005, opened 2026-06-29
fix: only show DFS confirmation when DFS is being enabled

AustinChangLinksys:
## Summary
- Fix DFS confirmation dialog showing incorrectly when saving other WiFi advanced settings

## Problem
The DFS confirmation dialog was appearing whenever **any** WiFi advanced setting was saved, as long as DFS was already enabled. This caused confusion when users only changed IPTV or other settings but saw a DFS warning.

## Root Cause
The save button logic checked `isDFSEnabled == true` (current state) instead of checking whether DFS was **being changed** from off to on.

## Fix
Compare `_preservedState` (original state) with current state to only show the confirmation when DFS is being turned ON:
```dart
final isDFSBeingEnabled = _preservedState?.isDFSEnabled != true &&
    currentState.isDFSEnabled == true;
```

## Test plan
- [ ] DFS already enabled, change only IPTV → Save should proceed without DFS dialog
- [ ] DFS disabled, enable DFS → Save should show DFS confirmation dialog
- [ ] DFS disabled, change only IPTV → Save should proceed without DFS dialog

Fixes #1004
Related: linksys/LinksysWRT#399

### DIFF
--- a/lib/page/wifi_settings/views/wifi_advanced_settings_view.dart
+++ b/lib/page/wifi_settings/views/wifi_advanced_settings_view.dart
@@ -69,7 +69,11 @@ class _WifiAdvancedSettingsViewState
       bottomBar: PageBottomBar(
           isPositiveEnabled: isPositiveEnabled,
           onPositiveTap: () {
-            if (ref.read(wifiAdvancedProvider).isDFSEnabled == true) {
+            final currentState = ref.read(wifiAdvancedProvider);
+            // Only show DFS confirmation when DFS is being turned ON (was off, now on)
+            final isDFSBeingEnabled = _preservedState?.isDFSEnabled != true &&
+                currentState.isDFSEnabled == true;
+            if (isDFSBeingEnabled) {
               _showConfirmDFSModal().then((value) {
                 if (value == true) {
                   save();


AustinChangLinksys (commented):
🤖 **Automated Review — Round 1** · `7077967..2158552` (full)

**Verdict: 💬 Self-review (comment only)** — GitHub does not allow self-approval; findings provided for reference. No Critical issues; 4 Warnings flagged (2 new, 2 pre-existing).

|---|---|---|
| ⚠️ | `wifi_advanced_settings_view.dart:74` | **[both reviewers]** `isDFSBeingEnabled` ignores `_preservedState == null` — fetch race can trigger spurious DFS modal |
| ⚠️ | `wifi_advanced_settings_view.dart:62` | `isPositiveEnabled` becomes `true` before `_preservedState` is set — Save button active before data loads *(pre-existing)* |
| ⚠️ | `wifi_advanced_settings_view.dart:77` | Missing `mounted` guard in `.then()` — `save()` may use stale `context` after widget unmounts *(pre-existing)* |
| ⚠️ | `(test/)` | **[both reviewers]** No widget tests for new DFS gating logic — 4 scenarios untested |
| 💡 | `wifi_advanced_settings_view.dart:75` | **[both reviewers]** `currentState.isDFSEnabled == true` — prefer `?? false` per null-safety convention |
| 💡 | `wifi_advanced_settings_view.dart:44` | `setWifiAdvancedSettingsViewChanged(state != _preservedState)` is always `false` after assignment *(pre-existing)* |
| 💡 | `wifi_advanced_settings_view.dart:267` | `_showDFSModal` / `_showConfirmDFSModal` share same i18n text — risk of future text divergence |
| 💡 | `wifi_advanced_settings_view.dart:30` | `_preservedState` field could use existing `PreservedStateMixin` for consistency |

<details><summary>🔴 Critical issues</summary>

None.

</details>

**W-1 [both reviewers] — `_preservedState` null race window · `wifi_advanced_settings_view.dart:74`**

`_preservedState` is initialised asynchronously inside `fetch().then(...)`. Before the fetch completes, `_preservedState == null`, so `_preservedState?.isDFSEnabled != true` evaluates to `true` unconditionally. If — in that narrow window — `currentState.isDFSEnabled` is already `true` (e.g., a fast/cached provider response), `isDFSBeingEnabled` becomes `true` and the DFS confirmation modal fires even though the user has not changed anything.

```dart
// wifi_advanced_settings_view.dart:74  (new code in this PR)
final isDFSBeingEnabled = _preservedState?.isDFSEnabled != true &&
    currentState.isDFSEnabled == true;
```

**Fix:** add explicit null guard so the modal only fires when the baseline is actually loaded:
```dart
final isDFSBeingEnabled = _preservedState != null &&
    _preservedState!.isDFSEnabled != true &&
    currentState.isDFSEnabled == true;
```

---

**W-2 — Save button active before data loads · `wifi_advanced_settings_view.dart:62` *(pre-existing)***

```dart
final isPositiveEnabled =
    _preservedState != ref.watch(wifiAdvancedProvider);
// null != WifiAdvancedSettingsState() → true immediately on mount
```

While `_preservedState == null` the Save button is already enabled, and tapping it would call `save()` with an uninitialised baseline. Relatedly, W-1 is more dangerous when paired with this.

**Fix:**
```dart
final isPositiveEnabled = _preservedState != null &&
    _preservedState != ref.watch(wifiAdvancedProvider);
```

---

**W-3 — Missing `mounted` guard in `.then()` callback · `wifi_advanced_settings_view.dart:77` *(pre-existing)***

```dart
_showConfirmDFSModal().then((value) {
  if (value == true) {
    save();   // ← uses BuildContext; widget may have unmounted during modal
  }
});
```

If the widget is disposed while the modal is open (system back gesture, OS reclaim), `save()` and the subsequent `showErrorMessageSnackBar` will access a stale `context`, potentially causing a `setState after dispose` exception.

**Fix:**
```dart
_showConfirmDFSModal().then((value) {
  if (!mounted) return;
  if (value == true) {
    save();
  }
});
```

---

**W-4 [both reviewers] — No widget tests for new DFS gating logic**

No test file exists for `WifiAdvancedSettingsView` or `_showConfirmDFSModal`. The four relevant scenarios for the new `isDFSBeingEnabled` condition are all untested:

| `_preservedState.isDFSEnabled` | `currentState.isDFSEnabled` | Expected |
|---|---|---|
| `false` | `true` | modal shown ✓ |
| `null` (unsupported device) | `true` | modal shown? (intent unclear) |
| `true` | `true` | no modal ✓ |
| `null` (`_preservedState` unloaded) | `true` | behaviour depends on W-1 fix |

</details>

- **Core fix is correct**: using `_preservedState` as a baseline to detect the `off → on` transition is the right approach; the previous code fired the modal on every save whenever DFS was already enabled, regardless of whether the user changed it.
- **`ref.read` in event handler**: correct — `onPositiveTap` is not a build-phase callback, so `ref.read` (not `watch`) is appropriate here.
- **Layer dependencies intact**: View → `wifiAdvancedProvider` → `routerRepositoryProvider` — no upward dependency violations.
- **`wifiAdvancedProvider` is not `autoDispose`** — correct for an L1 provider.
- **DFS null guard in `_buildGrid()`**: `if (state.isDFSEnabled != null)` correctly hides the DFS toggle on devices that don't support it, preventing misuse of the new modal.

</details>

HankYuLinksys (approved):
LGTM. Clean, well-scoped fix that addresses the root cause in #1004 — comparing `_preservedState` against the current state so the DFS confirmation only fires on an off→on transition.

Verified all transitions:
- DFS already on + change only IPTV → no dialog (the bug being fixed) ✅
- DFS off → on → shows dialog ✅
- DFS off + change only IPTV → no dialog ✅
- DFS on → off → no dialog ✅
- `_preservedState` null (pre-fetch) → safe fallback ✅

The `?.` / `!= true` handling covers the nullable `bool?` cleanly. No issues.

PR #1101, opened 2026-07-08
fix(port-triggering): persist forwarded-port edits in port triggering (#1061)

### LINKED ISSUE #1061: Port Triggering: Editing Forwarded Ports does not save — reverts to the old value.
**Reproduce steps:**
1. Login to Linksys Now.
2. Go to Menu >Advanced Settings > Port Forwarding > Triggering
3. Edit an existing Port Triggering rule.
4. Change both Trigger Ports and Forwarded Ports to new values
   (e.g. Trigger Ports 80–65534, Forwarded Ports 21–21).
5. Click Save, then re-open the rule to check the saved values.

**Actual result:**
Only Trigger Ports is saved (80–65534). **The Forwarded Ports change is not saved — it reverts to the old value (80–80).** No error is shown, so the change appears to succeed while it silently did not.

**Expected result:**
Both Trigger Ports and Forwarded Ports should be saved as entered. Re-opening the rule should show Forwarded Ports = 21–21.

**Env.:** FW_Pinnacle2.0_v1.2.2.26062820_USP_release

**Log:**
[GUI_portforwarding_forwarded port can't edit_0703.txt](https://github.com/user-attachments/files/29618033/GUI_portforwarding_forwarded.port.can.t.edit_0703.txt)

**Screenshot:**

Image 1 — Original values: Trigger Ports 21–21 TCP, Forwarded Ports 80–80 TCP.
Image

Image 2 — Edited values before saving: Trigger Ports changed to 80–65534, Forwarded Ports changed to 21–21.
Image

Image 3 — After Save: Trigger Ports saved correctly (80–65534), but Forwarded Ports reverted to the old value 80–80 (red box) — the edit was not saved.
Image

AustinChangLinksys:
Refs #1061

## Root Cause (two layers)
1. **UI layer** (`usp_port_triggering_tab.dart`): `_showEditDialog` rebuilt the rule via `copyWith` but
   omitted `forwardRules`, so the edited forwarded-port values were immediately discarded in the UI.
2. **Service layer** (`usp_port_forwarding_service.dart`): `saveTriggeringBatch`'s update path only
   issued a `Set` on the parent `PortTrigger.{i}` fields. The nested `Rule.{j}` sub-table was never
   touched, so even if the UI had passed the values through, the edit would have been silently dropped.

## Changes
| File | What changed |
|------|-------------|
| `lib/page/port_forwarding/views/components/usp_port_triggering_tab.dart` | `copyWith` now rebuilds `forwardRules` from dialog result, preserving `instancePath` of the first existing rule so the service can update it in-place. Extra rules beyond the first are carried through untouched. |
| `lib/page/port_forwarding/services/usp_port_forwarding_service.dart` | Added `_reconcileForwardRules` (Set/Add/Delete diff of `Rule.{j}` sub-table). Hooked into update path at step 3b. |
| `test/page/port_forwarding/services/usp_port_forwarding_service_test.dart` | 4 new regression tests for `#1061` (edit→Set, add→Add, delete→Delete, unchanged→no-op). |

## Acceptance criteria vs evidence
| AC | Evidence |
|----|----------|
| Edit forwarded-port persists | Test `#1061: editing forwarded-port of existing rule issues Set on Rule.*` — verifies `Set` is called with new port/endRange/protocol values on `Rule.1.` path |
| Add new forward rule persists | Test `#1061: adding a forward rule to existing trigger calls add on Rule.` — verifies `add` is called with correct parent path + params |
| Delete forward rule persists | Test `#1061: removing a forward rule from existing trigger calls delete` — verifies `delete` is called on the removed rule's path |
| No spurious ops when unchanged | Test `#1061: unchanged forward rules issue no Rule.* operations` — verifies `add`/`delete` never called when forward rules unchanged |

## Test results (verify re-run)
```
flutter test test/page/port_forwarding/
59/59 All tests passed! (00:13)
```

## Quality gates (verify re-run)
- `dart format --output=none --set-exit-if-changed .` → **exit 0** (1077 files, 0 changed)
- `flutter analyze lib/page/port_forwarding/` → **No issues found!**
- No `generated/` imports in View layer
- mocktail used (not mockito) in test
- `mapUspErrorToServiceError` used in service catch blocks (Article XIII)
- mutation lock held in Provider layer (Article IV) — unchanged by this diff
- diff: exactly 3 files, +337/-2

### DIFF
--- a/lib/page/port_forwarding/services/usp_port_forwarding_service.dart
+++ b/lib/page/port_forwarding/services/usp_port_forwarding_service.dart
@@ -398,7 +398,13 @@ class UspPortForwardingService {
         }
       }
 
-      // 3. Update (parent-level only)
+      // 3. Update parent-level fields AND reconcile nested forward rules.
+      //
+      // The parent Set (PortTrigger.{i}) only carries the trigger fields
+      // (Enable/Description/Port/PortEndRange/Protocol). The forwarded ports
+      // live in a separate sub-table (PortTrigger.{i}.Rule.{j}) and MUST be
+      // reconciled with their own Add / Set / Delete calls — otherwise edits
+      // to the forwarded ports are silently dropped (#1061).
       final originalByPath = <String, PortTriggeringRuleUIModel>{
         for (final r in original)
           if (r.instancePath != null) r.instancePath!: r,
@@ -408,7 +414,14 @@ class UspPortForwardingService {
         if (cur.instancePath == null) continue;
         final orig = originalByPath[cur.instancePath!];
         if (orig == null) continue;
-        if (cur != orig) {
+        if (cur == orig) continue;
+
+        // 3a. Parent-level trigger fields.
+        if (cur.enabled != orig.enabled ||
+            cur.description != orig.description ||
+            cur.triggerPort != orig.triggerPort ||
+            cur.triggerPortEndRange != orig.triggerPortEndRange ||
+            cur.triggerProtocol != orig.triggerProtocol) {
           toUpdate.add(PortTriggerUpdate(
             instancePath: cur.instancePath!,
             enabled: cur.enabled,
@@ -417,6 +430,17 @@ class UspPortForwardingService {
             triggerPortEndRange: cur.triggerPortEndRange,
             triggerProtocol: cur.triggerProtocol,
           ));
+        }
+
+        // 3b. Nested forwarded-port rules.
+        if (cur.forwardRules != orig.forwardRules) {
+          final (int fwdOps, int fwdFailed) = await _reconcileForwardRules(
+            parentPath: cur.instancePath!,
+            original: orig.forwardRules,
+            current: cur.forwardRules,
+          );
+          totalOps += fwdOps;
+          failedOps += fwdFailed;
         }
       }
       if (toUpdate.isNotEmpty) {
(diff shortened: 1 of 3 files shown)


AustinChangLinksys (commented):
**Note:** Verdict is APPROVE (0 Critical, dispatch-chain PR) but GitHub blocks self-approve via API (author and reviewer are both AustinChangLinksys). Posting as COMMENT — this review should be treated as APPROVE for merge purposes.

🤖 **Automated Review — Round 1** · `6108b1f..eee692a` (full)

**Verdict: ✅ APPROVE** — Correct fix for #1061 with solid test coverage; a few maintainability improvements worth addressing.

|---|---|---|---|
| ⚠️ | 🟢High | `usp_port_forwarding_service.dart:471-474` | **[both reviewers]** W-A: `updated` return count omits forward-rule operations; debug logs show `updated: 0` when only forwarded ports changed |
| ⚠️ | 🟢High | `usp_port_forwarding_service.dart:~431` | **[single, Reviewer B]** W-B: `cur.forwardRules != orig.forwardRules` uses List reference equality — always `true` after edit dialog, causing redundant USP calls |
| ⚠️ | 🟢High | `usp_port_forwarding_service.dart:~544` | **[single, Reviewer B]** W-C: `await _usp.set(updateParams)` calls USP client directly, bypassing codegen layer; no `updatePortTriggerForwardRule` codegen wrapper exists |
| ⚠️ | 🟢High | `usp_port_forwarding_service.dart:~389` | **[single, Reviewer A]** W-E: Pre-existing — `addPortTriggerForwardRule` result ignored when adding NEW parent; error silently lost (contrast: `_reconcileForwardRules` does check) |
| ⚠️ | 🟡Med | `usp_port_forwarding_service.dart:~431` | **[single, Reviewer A]** W-D: Sub-rule reconcile fires inside loop before parent batch Set; if parent Set later fails, sub-rules are already committed with no rollback |
| ⚠️ | 🟡Med | `usp_port_forwarding_service.dart` | **[both reviewers]** W-F: `UspPartialSuccess` in delete/update/add branches does not increment `failed` — lenient by design but delete partial leaves orphaned rules silently |
| ⚠️ | 🟡Med | `usp_port_forwarding_service.dart:~544` | **[single, Reviewer B]** W-G: `_usp.set(updateParams)` missing `allowPartial:` named arg — default behaviour depends on `UspClient.set` signature |
| 💡 | 🟢High | `usp_port_triggering_tab.dart:~146` | **[both reviewers]** S-1: `...rule.forwardRules.skip(1)` silently preserves additional forward rules that users cannot see or delete; no tracking issue |
| 💡 | 🟡Med | `test/page/port_forwarding/services/usp_port_forwarding_service_test.dart` | **[single, Reviewer B]** S-2: Confirm new tests cover _reconcileForwardRules mixed-change (parent+forward rule simultaneously) scenario |
| 💡 | 🟡Med | `usp_port_forwarding_service.dart` | **[both reviewers]** S-3: `UspPartialSuccess` lenient mode (silent pass) in `_reconcileForwardRules` is undocumented; a doc comment would clarify intent |

<details>

**W-A [both reviewers] · 🟢High — `updated` return value omits forward-rule reconciliation ops**

`lib/page/port_forwarding/services/usp_port_forwarding_service.dart:471-474`:
```dart
return (
  added: toAdd.length,
  updated: toUpdate.length,   // counts only parent-level PortTrigger Sets
  deleted: toDelete.length,
);
```
When a user edits only the forwarded port (no parent fields change), `toUpdate` stays empty → `updated: 0`, but `_reconcileForwardRules` actually issued USP Set calls. The page-notifier log at line ~124 will print `updated: 0`, making debugging misleading.

**Fix:** Accumulate `fwdUpdated` from each `_reconcileForwardRules` call and include it in the return, or add a `forwardRulesUpdated` field.

---

**W-B [single, Reviewer B] · 🟢High — `List !=` is reference equality; edit dialog always creates new list**

`lib/page/port_forwarding/services/usp_port_forwarding_service.dart` (Step 3b guard):
```dart
if (cur.forwardRules != orig.forwardRules) {
```
`List<PortTriggerForwardRuleUIModel>` does not override `==`; Dart compares by identity. After `_showEditDialog` rebuilds `forwardRules: [PortTriggerForwardRuleUIModel(...), ...rule.forwardRules.skip(1)]`, the list is always a new object. So even when only a parent field changed, this guard fires and `_reconcileForwardRules` is called. Functionally safe (inner `cur == orig` via Equatable short-circuits), but wastes a network round-trip on slow firmware.

**Fix:**
```dart
if (!const DeepCollectionEquality().equals(cur.forwardRules, orig.forwardRules)) {
```
or use `listEquals` from `package:flutter/foundation.dart`.

---

**W-C [single, Reviewer B] · 🟢High — Direct `_usp.set()` call bypasses codegen layer**

`usp_port_forwarding_service.dart:~544`:
```dart
final result = await _usp.set(updateParams);
```
Architecture rule: Service layer must interact with USP only through codegen wrappers. There is no `PortTriggering.updatePortTriggerForwardRule` codegen method. The path construction (`${cur.instancePath}Port`, `PortEndRange`, `Protocol`) is hand-coded strings that must be manually kept in sync with codegen.

**Fix:** Add `updatePortTriggerForwardRule` to `lib/generated/port_triggering.g.dart` (similar to existing `addPortTriggerForwardRule`/`deletePortTriggerForwardRule`).

---

**W-D [single, Reviewer A] · 🟡Med — Sub-rule operations run before parent batch Set**

Inside the `for (final cur in current)` loop, `await _reconcileForwardRules(...)` fires immediately (step 3b), but `toUpdate` is only sent after the loop ends. If a rule has both parent-field changes and forward-rule changes, the router sees Rule changes first and parent changes later. Should parent Set fail, sub-rules are already committed with no rollback path. TR-369 has no transaction semantics, so this is a defensive design concern rather than a data-corruption bug.

---

**W-E [single, Reviewer A] · 🟢High — Pre-existing: ADD new parent ignores `addPortTriggerForwardRule` result**

`usp_port_forwarding_service.dart:~389` (existing code, not introduced by this PR):
```dart
for (final fr in r.forwardRules) {
  await PortTriggering.addPortTriggerForwardRule(
    _usp, parentPath,
    forwardPort: fr.forwardPort,
    forwardPortEndRange: fr.forwardPortEndRange,
    forwardProtocol: fr.forwardProtocol,
  );
  // result not checked; USP Add failure silently ignored
}
```
By contrast, `_reconcileForwardRules` (new code in this PR) correctly calls `UspResultParser.parseAddResult`. Pre-existing, but this PR is the natural place to fix the inconsistency.

---

**W-F [both reviewers] · 🟡Med — `UspPartialSuccess` does not increment `failed` in `_reconcileForwardRules`**

All three switch cases (delete/update/add) log a warning but do not do `failed++`. Consistent with parent-layer lenient mode, but for delete operations this means the router may retain orphaned `Rule.{j}` entries without any UI indication of partial failure.

---

**W-G [single, Reviewer B] · 🟡Med — `_usp.set(updateParams)` missing `allowPartial:` named arg**

```dart
final result = await _usp.set(updateParams);
```
All other `_usp.set` calls in the service (via codegen) pass `allowPartial: false`. If `UspClient.set` defaults to `allowPartial: true`, a partial failure updating multiple forward rules would be silently treated as success.

</details>

<details>

- **Core fix correct**: Adding `_reconcileForwardRules` with Add/Set/Delete diff against `Rule.{j}` sub-table is the right architecture to fix #1061.
- **View fix correct**: Preserving `rule.forwardRules.first.instancePath` in the dialog result ensures in-place Set rather than spurious Add.
- **Test coverage**: 4 regression tests (#1061 edit/add/delete/unchanged) directly exercise the new code paths.
- **USP error handling in `_reconcileForwardRules`**: All three switch branches (UspSuccess/UspPartialSuccess/UspFailure) are handled for delete, update, and add.
- **Null guards**: `instancePath` null checks are consistent and complete; `rule.forwardRules.isNotEmpty` guard in view prevents index out-of-range.
- **Early-exit optimization**: `if (cur == orig) continue;` avoids processing unchanged parent items entirely.
- **Reversed delete order**: `toDelete.reversed` correctly deletes higher-index Rules first, avoiding path-shift issues.
- **Comment quality**: New comments in the service clearly explain the parent-vs-sub-table distinction that caused #1061.

</details>

PeterJhongLinksys (approved):
Verified the #1061 fix end-to-end:

- View (`usp_port_triggering_tab.dart`): `copyWith` now rebuilds `forwardRules` from the dialog result and preserves the first existing rule's `instancePath`, so the edit reaches the L2 working copy (`editTriggeringRule` -> `state.settings.update`) and drives the dirty-guard correctly. Extra rules are carried through via `skip(1)` with no data loss.
- Service (`_reconcileForwardRules`): Set/Add/Delete diff against the `Rule.{j}` sub-table produces a payload that matches the TR-181 codegen model (`Rule.{j}.Port/PortEndRange/Protocol`); reversed delete order avoids path-shift. Save runs under `withLock` and maps errors to `ServiceError`.
- Unchanged-forward-rules path confirmed no-op (parent-only `copyWith` keeps the same list reference).

No blocking issues found. The maintainability items already captured in the Round-1 automated review (return-count omitting forward-rule ops, list reference-equality guard, direct `_usp.set` vs a codegen wrapper + `allowPartial:`, and the hidden extra-forward-rules UX) are worth following up but are non-blocking.

PR #1177, opened 2026-07-28
fix(navigation): revert top-nav highlight when navigation is cancelled (#1158)

### LINKED ISSUE #1158: The navigation bar should focus on the original option when select go back in the confirm changes dialog
**Steps:**
1. Go to Menu > Any function that can change settings (e.g., Instant-Safety...).
2. Change settings and click on the other option on the navigation bar.
3. When the confirm changes dialog pops up, click on the "Go back".

**Actual Result:**
The navigation bar will focus on the option selected in step 2.

**Expected Result:**
The navigation bar should focus on the original option

**Device**

**Screenshots**
Image

**Additional context**
log: [log.txt](https://github.com/user-attachments/files/30173314/log.txt)
Video: https://github.com/user-attachments/assets/ad42cdff-0da1-4736-a3fc-790186a81468

AustinChangLinksys:
## Summary

Fixes #1158 — the top-navigation bar highlight staying on the wrong option after a user cancels the confirm-changes dialog.

**Steps to reproduce (from the issue):**
1. Go to Menu > any settings function (e.g. Instant-Safety).
2. Change a setting, then tap a different option on the navigation bar.
3. When the confirm-changes dialog pops up, tap **Go back**.

**Actual:** the nav bar highlight moves to the option tapped in step 2.
**Expected:** the highlight stays on the original (current) option.

## Root cause (evidence-based)

- `lib/components/styled/menus/widgets/top_navigation_menu.dart:47` passes `selectedIndices: {selectedIndex}` (derived from `widget.selected`) into `AppChipGroup`.
- `AppChipGroup` (ui_kit_library `v2.28.1`, `_toggleChip`) owns its selection **internally**: on tap it optimistically does `setState(_selectedIndices = {index})` — the highlight moves immediately — and only re-syncs to the `selectedIndices` prop inside `didUpdateWidget` **when that prop changes**.
- `lib/components/styled/menus/widgets/menu_holder.dart:97` wires the tap to `MenuController.select()` → `goNamed()`.
- `lib/route/route_model.dart:76-87` (`LinksysRoute.onExit`) runs the dirty check; **Go back** returns `false`, so navigation is blocked and `MenuController._menuNotifier.selected` never changes.
- Because `selected` is unchanged, the `ValueListenableBuilder` in `menu_holder.dart` never rebuilds `TopNavigationMenu`, so `AppChipGroup.selectedIndices` never changes → `didUpdateWidget` never reverts the optimistic move → the highlight is stuck on the cancelled destination.

## Fix (minimal, consumer-side)

`TopNavigationMenu` now bumps a `_syncToken` on every tap and folds it into the `AppChipGroup` `key`. Re-keying forces `AppChipGroup` to rebuild from scratch and re-read the authoritative `widget.selected`:
- **Navigation cancelled** → `selected` unchanged → highlight reverts to the current option. ✅
- **Navigation confirmed** → controller updates `selected` → highlight follows to the target. ✅

No change to `ui_kit_library`. Scope is limited to the single broken widget (Maintainer / minimal-diff mode).

## Verification

`git diff --stat`:
```
 .../styled/menus/widgets/top_navigation_menu.dart  |  19 +++
 .../menus/widgets/top_navigation_menu_test.dart    | 152 +++++++++++++++++++++
 2 files changed, 171 insertions(+)
```

- `flutter analyze` on both files → **No issues found!**
- `dart format --output=none --set-exit-if-changed` (2.x CI first gate) → exit 0.
- New regression test `test/components/styled/menus/widgets/top_navigation_menu_test.dart` → **2/2 pass**:
  - "reverts highlight to the original option when navigation is cancelled (#1158)" — proven RED against the unfixed code (stashed the fix, test failed) and GREEN with the fix.
  - "follows the highlight when navigation is confirmed" — guards against over-correction.

Refs #1158

### DIFF
--- a/lib/components/styled/menus/widgets/top_navigation_menu.dart
+++ b/lib/components/styled/menus/widgets/top_navigation_menu.dart
@@ -65,6 +65,12 @@ class _TopNavigationMenuState extends State<TopNavigationMenu> {
             // `selected` and the highlight follows; if it is cancelled, the
             // highlight reverts to the current option instead of staying on
             // the tapped-but-rejected destination.
+            //
+            // This bump MUST stay here (synchronous, on every tap). It cannot
+            // move to didUpdateWidget: a cancelled navigation leaves the
+            // MenuController's `selected` unchanged, so this widget never
+            // rebuilds and didUpdateWidget never fires — the highlight would
+            // stay stuck on the rejected destination (the exact #1158 bug).
             if (mounted) {
               setState(() => _syncToken++);
             }
--- a/lib/components/styled/menus/widgets/top_navigation_menu.dart
+++ b/lib/components/styled/menus/widgets/top_navigation_menu.dart
@@ -19,6 +19,16 @@ class TopNavigationMenu extends StatefulWidget {
 }
 
 class _TopNavigationMenuState extends State<TopNavigationMenu> {
+  // Bumped on every chip tap to force AppChipGroup to rebuild from scratch and
+  // re-read the authoritative `widget.selected`. AppChipGroup owns its
+  // selection state internally and optimistically moves the highlight on tap,
+  // only re-syncing to its `selectedIndices` prop when that prop changes. When
+  // a navigation is cancelled (e.g. the user picks "Go back" in the unsaved
+  // changes dialog), `widget.selected` never changes, so without this the
+  // highlight would stay stuck on the cancelled destination. Re-keying forces
+  // the highlight back to the still-current option. See PrivacyGUI#1158.
+  int _syncToken = 0;
+
   @override
   void initState() {
     super.initState();
@@ -37,6 +47,7 @@ class _TopNavigationMenuState extends State<TopNavigationMenu> {
     return Theme(
       data: darkTheme,
       child: AppChipGroup(
+        key: ValueKey('top-nav-chip-group-$selectedIndex-$_syncToken'),
         chips: widget.items
             .map((type) => ChipItem(
                   label: type.resloveLabel(context),
@@ -49,6 +60,14 @@ class _TopNavigationMenuState extends State<TopNavigationMenu> {
         onSelectionChanged: (selectedIndices) {
           if (selectedIndices.isNotEmpty) {
             widget.onItemClick?.call(selectedIndices.first);
+            // Force AppChipGroup to re-assert `widget.selected` on the next
+            // frame. If the navigation is confirmed, the controller updates
+            // `selected` and the highlight follows; if it is cancelled, the
+            // highlight reverts to the current option instead of staying on
+            // the tapped-but-rejected destination.
+            if (mounted) {
+              setState(() => _syncToken++);
+            }
           }
         },
       ),
--- a/test/components/styled/menus/widgets/top_navigation_menu_test.dart
+++ b/test/components/styled/menus/widgets/top_navigation_menu_test.dart
@@ -44,18 +44,27 @@ void main() {
   }
 
   /// Returns true when the chip carrying [label] is rendered as selected.
-  /// AppChipGroup renders the selected chip's label in bold (see
-  /// app_chip_group.dart: `fontWeight: isSelected ? FontWeight.bold : normal`).
+  ///
+  /// Reads the authoritative `selected` semantics flag that AppChipGroup emits
+  /// per chip (`Semantics(label: chip.label, selected: isSelected, ...)` in
+  /// app_chip_group.dart). This is the accessibility contract for the selected
+  /// state, so it is stable across visual-style changes — unlike asserting on
+  /// the selected chip's bold font weight, which couples the test to a
+  /// ui_kit_library rendering detail that could switch to colour/underline.
   bool isChipSelected(WidgetTester tester, String label) {
-    final appText = tester.widget<AppText>(
-      find
-          .ancestor(
-            of: find.text(label),
-            matching: find.byType(AppText),
-          )
-          .first,
+    // Scope to the AppChipGroup's own Semantics node for this chip (which
+    // carries `selected`), not the inner AppText's text-only semantics.
+    final chipSemantics = find.ancestor(
+      of: find.text(label),
+      matching: find.byWidgetPredicate(
+        (w) => w is Semantics && w.properties.selected != null,
+      ),
     );
-    return appText.fontWeight == FontWeight.bold;
+    expect(chipSemantics, findsOneWidget,
+        reason: 'Expected exactly one selectable chip semantics node for '
+            '"$label"');
+    final semantics = tester.widget<Semantics>(chipSemantics);
+    return semantics.properties.selected ?? false;
   }
 
   testWidgets(
--- a/test/components/styled/menus/widgets/top_navigation_menu_test.dart
+++ b/test/components/styled/menus/widgets/top_navigation_menu_test.dart
@@ -1,0 +1,152 @@
+import 'package:flutter/material.dart';
+import 'package:flutter_test/flutter_test.dart';
+import 'package:privacy_gui/components/styled/menus/menu_consts.dart';
+import 'package:privacy_gui/components/styled/menus/widgets/top_navigation_menu.dart';
+import 'package:privacy_gui/l10n/gen/app_localizations.dart';
+import 'package:ui_kit_library/ui_kit.dart';
+
+/// Regression coverage for PrivacyGUI#1158.
+///
+/// When a top-navigation chip tap is rejected downstream (e.g. the unsaved
+/// changes dialog "Go back" cancels the navigation), `widget.selected` never
+/// changes. AppChipGroup owns its selection state internally and optimistically
+/// moves the highlight on tap, so without a re-sync the highlight would stay
+/// stuck on the tapped-but-rejected destination. TopNavigationMenu must force
+/// the highlight back to the still-current option after every tap.
+void main() {
+  // A dark ThemeData carrying AppDesignTheme so TopNavigationMenu does not fall
+  // back to getIt (which is unregistered in the test host) and AppChipGroup's
+  // DesignSystem assertion is satisfied.
+  final darkTheme = AppTheme.create(
+    brightness: Brightness.dark,
+    seedColor: Colors.blue,
+    designThemeBuilder: (c) => CustomDesignTheme.fromJson({'style': 'flat'}),
+  );
+
+  const items = [NaviType.home, NaviType.menu, NaviType.support];
+
+  Widget buildHost({
+    required NaviType selected,
+    required void Function(int) onItemClick,
+  }) {
+    return MaterialApp(
+      theme: darkTheme,
+      localizationsDelegates: AppLocalizations.localizationsDelegates,
+      supportedLocales: AppLocalizations.supportedLocales,
+      home: Scaffold(
+        body: TopNavigationMenu(
+          items: items,
+          selected: selected,
+          onItemClick: onItemClick,
+        ),
+      ),
+    );
+  }
+
+  /// Returns true when the chip carrying [label] is rendered as selected.
+  /// AppChipGroup renders the selected chip's label in bold (see
+  /// app_chip_group.dart: `fontWeight: isSelected ? FontWeight.bold : normal`).
+  bool isChipSelected(WidgetTester tester, String label) {
+    final appText = tester.widget<AppText>(
+      find
+          .ancestor(
+            of: find.text(label),
+            matching: find.byType(AppText),
+          )
+          .first,
+    );
+    return appText.fontWeight == FontWeight.bold;
+  }
+
+  testWidgets(
+      'reverts highlight to the original option when navigation is cancelled '
+      '(PrivacyGUI#1158)', (tester) async {
+    // Home is the current page. A tap on another chip is REJECTED downstream:
+    // onItemClick does not change `selected` (mimics "Go back" in the unsaved
+    // changes dialog blocking the navigation).
+    await tester.pumpWidget(buildHost(
+      selected: NaviType.home,
+      onItemClick: (_) {
+        /* navigation blocked — selected stays on Home */
+      },
+    ));
+    await tester.pumpAndSettle();
+
+    final context = tester.element(find.byType(TopNavigationMenu));
+    final homeLabel = NaviType.home.resloveLabel(context);
+    final menuLabel = NaviType.menu.resloveLabel(context);
+
+    // Baseline: Home highlighted, Menu not.
+    expect(isChipSelected(tester, homeLabel), isTrue,
+        reason: 'Home should be the initially selected option');
+    expect(isChipSelected(tester, menuLabel), isFalse);
+
+    // Act: tap Menu. AppChipGroup optimistically moves the highlight, then the
+    // navigation is cancelled (selected unchanged).
+    await tester.tap(find.text(menuLabel));
+    await tester.pumpAndSettle();
+
+    // Assert (the bug): highlight must be back on Home, NOT stuck on Menu.
+    expect(isChipSelected(tester, homeLabel), isTrue,
+        reason:
+            'After a cancelled navigation the highlight must revert to Home');
+    expect(isChipSelected(tester, menuLabel), isFalse,
+        reason: 'The rejected destination must not stay highlighted (#1158)');
+  });
+
+  testWidgets(
+      'follows the highlight when navigation is confirmed (selected updates)',
+      (tester) async {
+    // A parent that DOES accept the navigation: it rebuilds with the new
+    // selected value, mimicking a confirmed route change.
+    await tester.pumpWidget(
+      _ConfirmingHost(items: items, theme: darkTheme),
+    );
+    await tester.pumpAndSettle();
+
+    final context = tester.element(find.byType(TopNavigationMenu));
+    final homeLabel = NaviType.home.resloveLabel(context);
+    final menuLabel = NaviType.menu.resloveLabel(context);
+
+    expect(isChipSelected(tester, homeLabel), isTrue);
+
+    await tester.tap(find.text(menuLabel));
+    await tester.pumpAndSettle();
+
+    // Confirmed navigation: highlight follows to Menu.
+    expect(isChipSelected(tester, menuLabel), isTrue,
+        reason: 'A confirmed navigation must move the highlight to the target');
+    expect(isChipSelected(tester, homeLabel), isFalse);
+  });
+}
+
+/// Host that accepts navigation by updating `selected` on tap.
+class _ConfirmingHost extends StatefulWidget {
+  final List<NaviType> items;
+  final ThemeData theme;
+  const _ConfirmingHost({required this.items, required this.theme});
+
+  @override
+  State<_ConfirmingHost> createState() => _ConfirmingHostState();
+}
+
+class _ConfirmingHostState extends State<_ConfirmingHost> {
+  NaviType _selected = NaviType.home;
+
+  @override
+  Widget build(BuildContext context) {
+    return MaterialApp(
+      theme: widget.theme,
+      localizationsDelegates: AppLocalizations.localizationsDelegates,
+      supportedLocales: AppLocalizations.supportedLocales,
+      home: Scaffold(
+        body: TopNavigationMenu(
+          items: widget.items,
+          selected: _selected,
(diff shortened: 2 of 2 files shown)


AustinChangLinksys (commented):
🤖 **Automated Review — Round 1** · `557d59c..c5e5b2e` (full)

**Verdict: ✅ APPROVE** — Core fix is correct; the re-key mechanism properly forces `AppChipGroup` to re-read the authoritative `widget.selected` after every tap. One Warning about a double-rebuild on confirmed navigation.
_(Note: GitHub self-approve blocked — verdict intent is APPROVE; please approve manually or via a different reviewer account.)_

#### Standards

|---|---|---|---|
| ⚠️ | 🟡Med | `top_navigation_menu.dart:60–65` | Double rebuild on confirmed navigation: `_syncToken++` fires synchronously but `widget.selected` updates on a later async frame → two key changes, two teardowns |
| ⚠️ | 🟢High | `top_navigation_menu_test.dart:52–62` | Test `isChipSelected` reads `fontWeight == bold` — couples assertions to `AppChipGroup` internal rendering, brittle if ui_kit changes the selected-chip visual |
| 💡 | 🟡Med | `top_navigation_menu.dart:50` | [Speculative Generality] Key includes `selectedIndex` unnecessarily — `didUpdateWidget` already handles parent-driven updates; `_syncToken` alone suffices |
| 💡 | 🟡Med | `top_navigation_menu.dart:22` | [Mysterious Name] `_syncToken` reads like a cancel/auth token; `_chipGroupRebuildNonce` or `_forceRebuildGeneration` would self-document |
| 💡 | 🟡Med | `top_navigation_menu.dart:60–65` | Re-tapping the already-selected chip also bumps `_syncToken` → unnecessary AppChipGroup teardown |
| 💡 | ⚪Low | `top_navigation_menu.dart:35` | Pre-existing: `widget.items.isEmpty` → `items.first` throws `StateError` — not introduced by this PR but uncovered by it |

#### Spec (Issue #1158)

|---|---|---|---|
| ⚠️ | 🟡Med | `onSelectionChanged` → async frame | Confirmed-navigation path: highlight briefly flashes back to original option for 1 frame before settling on target (spec doesn't explicitly require smoothness here, but visible jank risk) |

<details>

**Double rebuild on confirmed navigation** 🟡Med
- `lib/components/styled/menus/widgets/top_navigation_menu.dart:60–65`: `setState(() => _syncToken++)` executes synchronously in `onSelectionChanged`, changing the key from `...$selectedIndex-0` to `...$selectedIndex-1` at Frame N+1 — `AppChipGroup` torn down, rebuilt with *original* `selectedIndex` (nav not confirmed yet). At Frame N+k, after `goNamed` resolves and `widget.selected` changes, `selectedIndex` changes → key changes again → second teardown/rebuild to the *new* tab.
- On slow devices this two-step transition may produce a visible one-frame flash back to the original tab before landing on the destination.
- **Fix:** Move the bump to `didUpdateWidget`: only increment `_syncToken` when `oldWidget.selected == widget.selected` (navigation was cancelled). When `selected` actually changed, `didUpdateWidget` handles it via the changed `selectedIndex` in the key — no extra bump needed. Remove `setState(() => _syncToken++)` from `onSelectionChanged`.

**Test couples to ui_kit internal rendering** 🟢High
- `test/components/styled/menus/widgets/top_navigation_menu_test.dart:52–62`: `return appText.fontWeight == FontWeight.bold` relies on `AppChipGroup` expressing selection via bold weight (currently true in `app_chip_group.dart:184`). If the design system switches to colour or underline, this helper silently misfires.
- **Fix:** Use the Semantics API: `tester.getSemantics(find.bySemanticsLabel(label)).hasFlag(SemanticsFlag.isSelected)` — reads the authoritative `selected:` flag that `AppChipGroup` already emits in its `Semantics` wrapper (`app_chip_group.dart:143–147`).

</details>

<details>

- Core mechanism is correct: `ValueKey` including `_syncToken` forces `AppChipGroup.initState` to re-read `widget.selectedIndices` — the optimistic in-widget selection is discarded. Cancelling nav leaves `widget.selected` unchanged → highlight properly reverts. Spec #1158 core requirement is met.
- `mounted` guard on `setState` is correct and necessary.
- Two widget tests cover both cancellation and confirmation paths.
- Scope is minimal — one widget, no Provider or architecture rule changes.
- No hardcoded strings or security issues.

</details>

lib/components/styled/menus/widgets/top_navigation_menu.dart
```diff
@@ -37,6 +47,7 @@ class _TopNavigationMenuState extends State<TopNavigationMenu> {
     return Theme(
       data: darkTheme,
       child: AppChipGroup(
+        key: ValueKey('top-nav-chip-group-$selectedIndex-$_syncToken'),
```
PeterJhongLinksys, line 50: ⚠️ The `$_syncToken` in the key is not needed, and it has a real cost.

`AppChipGroup` re-syncs in `didUpdateWidget` with `widget.selectedIndices != oldWidget.selectedIndices` (ui_kit 2.28.1, `app_chip_group.dart:82-87`). Dart `Set` does not override `operator ==`, and `build()` allocates a fresh `{selectedIndex}` at line 58 every time — so that condition is **always true** and a plain `setState(() {})` already re-asserts `widget.selected`. The comment at lines 22-29 ("only re-syncing … when that prop changes") does not match actual behaviour.

Meanwhile changing the key unmounts the whole chip subtree on every tap, disposing `AppInteractionSensor`'s internal `FocusNode` (`app_interaction_sensor.dart:234-241`). After activating a tab with Enter, focus is lost and screen readers re-announce the entire group. Suggest dropping the token (and `$selectedIndex`) from the key, keeping the `setState`, and fixing the comment.

lib/components/styled/menus/widgets/top_navigation_menu.dart
```diff
@@ -49,6 +60,20 @@ class _TopNavigationMenuState extends State<TopNavigationMenu> {
...
+            // highlight reverts to the current option instead of staying on
+            // the tapped-but-rejected destination.
+            //
+            // This bump MUST stay here (synchronous, on every tap). It cannot
+            // move to didUpdateWidget: a cancelled navigation leaves the
+            // MenuController's `selected` unchanged, so this widget never
+            // rebuilds and didUpdateWidget never fires — the highlight would
+            // stay stuck on the rejected destination (the exact #1158 bug).
+            if (mounted) {
+              setState(() => _syncToken++);
```
PeterJhongLinksys, line 75: ⚠️ This bump is synchronous, but a *confirmed* navigation through a dirty guard is not: `menu_holder.dart:151-158` calls `goNamed()`, `route_model.dart:76-88` then `await showUnsavedAlert(context)`, and `_controller.setTo(...)` only runs after the navigation succeeds (`menu_holder.dart:52-58`). So on dirty-check routes the highlight snaps back to the old tab for the entire time the dialog is open, then jumps to the new tab. The cancel path is fixed; the confirm path now flickers. A result-driven revert (only revert when the guard rejects) would avoid it. Note `_ConfirmingHost` in the test updates `selected` synchronously, so this intermediate state cannot be observed by the current tests.

Minor: `if (mounted)` is dead code here — `onSelectionChanged` is called synchronously from the child's gesture/key handler (`app_chip_group.dart:116`), so there is no async boundary.

PeterJhongLinksys (approved):
🤖 **Automated Review — Round 1** · `557d59c..fe80052` (full)

**Verdict: ✅ APPROVE** — The fix does resolve #1158 and the regression test targets the right scenario; no Critical issues. Approving with Warnings: the `ValueKey` re-mount is a heavier mechanism than needed (a plain `setState` suffices) and it destroys the chips' `FocusNode`, which is a keyboard/a11y regression worth addressing before merge.

|---|---|---|---|
| ⚠️ | 🟢 | `lib/components/styled/menus/widgets/top_navigation_menu.dart:50` | The `$_syncToken` in the `ValueKey` is unnecessary: `Set` has no `operator ==`, so `AppChipGroup.didUpdateWidget` re-syncs on *every* rebuild — `setState(() {})` alone restores the highlight without remounting. |
| ⚠️ | 🟢 | `lib/components/styled/menus/widgets/top_navigation_menu.dart:50` | Because of the key change, every tap unmounts the chip subtree and disposes `AppInteractionSensor`'s internal `FocusNode` → after activating a tab with Enter, focus is lost and screen readers re-announce the whole chip group. |
| ⚠️ | 🟢 | `lib/components/styled/menus/widgets/top_navigation_menu.dart:22-29` | The comment's premise ("only re-syncing to its `selectedIndices` prop when that prop changes") is not what the widget does — identity comparison makes the re-sync unconditional. Fixing the comment matters because it is the stated justification for the whole approach. |
| ⚠️ | 🟢 | `lib/components/styled/menus/widgets/top_navigation_menu.dart:50` | [Shotgun Surgery risk] This is a consumer-side workaround for `AppChipGroup` lacking controlled-selection semantics (Art. XV UI Kit First); without an upstream `TODO`/issue link, every future consumer must rediscover the same hack. |
| ⚠️ | 🟢 | `test/components/styled/menus/widgets/top_navigation_menu_test.dart:54-68` | `isChipSelected` locates state via `find.ancestor(... w is Semantics && w.properties.selected != null)` + `findsOneWidget` — this couples to widget-tree shape, and will break if ui_kit ever wraps another `selected`-bearing `Semantics`. `tester.getSemantics(...)` + `matchesSemantics(isSelected: true)` is shape-agnostic. |
| ⚠️ | 🟢 | `test/components/styled/menus/widgets/top_navigation_menu_test.dart:70,106` | No `group()` wrapper; Art. I/VIII expect `[ClassName] - [Feature]` grouping. Test names themselves are descriptive and unnumbered ✅. |
| ⚠️ | 🟡 | `lib/components/styled/menus/widgets/top_navigation_menu.dart:62-76`, `lib/components/styled/menus/widgets/menu_holder.dart:98-100,151-158`, `lib/route/route_model.dart:76-88` | On a *confirmed* navigation through an async dirty guard, the token bump is synchronous while `goNamed()` is still awaiting the dialog → the highlight snaps back to the old tab for the whole dialog lifetime, then jumps to the new tab. The cancel path is fixed; the confirm path now flickers. |
| ⚠️ | 🟡 | `test/components/styled/menus/widgets/top_navigation_menu_test.dart:142-160` | `_ConfirmingHost` updates `_selected` **synchronously** inside `onItemClick`, unlike the real `await showUnsavedAlert` guard — so the intermediate state described above can never be observed by these tests. |
| ⚠️ | 🟡 | `test/components/styled/menus/widgets/top_navigation_menu_test.dart:76-81` | Cancellation is simulated with an empty `onItemClick`; the real path (`MenuHolder` → `MenuController.select()` → `goNamed()` → dirty-guard dialog → `setTo()`) has no coverage, so #1158 is not guarded at the integration boundary. |
| 💡 | 🟢 | `lib/components/styled/menus/widgets/top_navigation_menu.dart:50` | [Speculative Generality] With `_syncToken` bumping on every tap, `$selectedIndex` in the key adds no re-sync capability — two trigger sources for one effect. Keep one. |
| 💡 | 🟢 | `lib/components/styled/menus/widgets/top_navigation_menu.dart:74-76` | `if (mounted)` is dead code here: `onSelectionChanged` is invoked synchronously from the child's gesture/key handler (`app_chip_group.dart:116`), so no async boundary exists. It misleads maintainers into assuming one does. |
| 💡 | 🟢 | `test/components/styled/menus/widgets/top_navigation_menu_test.dart:32-43` vs `142-160` | [Duplicated Code] `_ConfirmingHost.build()` re-implements `buildHost()` almost line for line (theme/localizations/Scaffold/TopNavigationMenu). |
| 💡 | 🟢 | `test/components/styled/menus/widgets/top_navigation_menu_test.dart:20-24` | The `AppTheme.create(... CustomDesignTheme.fromJson({'style':'flat'}))` incantation is now duplicated across ~15 test files; per Art. I §1.6.2 it belongs in a shared `TestThemes` helper. Also, `fromJson` ignores the passed `ColorScheme` and no `'brightness': 'dark'` is set, so the test passes only because `:41` checks `parentTheme.brightness`. |
| 💡 | 🟢 | `lib/components/styled/menus/widgets/top_navigation_menu.dart:32-35` | [Dead Code] Empty `initState` override that only calls `super` (same in `bottom_navigation_menu.dart:51-54`). Free to drop while already editing this State. |
| 💡 | 🟡 | `lib/components/styled/menus/widgets/top_navigation_menu.dart:50,75` | Re-mounting on every tap discards selection-transition animation and ink ripple state; negligible perf cost for 3 chips, but worth an on-device look for visual jitter. |
| 💡 | 🟡 | `lib/components/styled/menus/widgets/top_navigation_menu.dart:22-29,63-73` | 16 lines of comment for 2 lines of logic — good content, but the ratio itself signals a workaround at the wrong layer. Suggest compressing to 3-4 lines plus a `// TODO(ui_kit #xxx):` link so the note can be retired once ui_kit gains controlled selection. |
| 💡 | 🟡 | `lib/components/styled/menus/widgets/top_navigation_menu.dart:44-45,58` | `items.indexOf(...)` returns `-1` when `selected` is not in `items`, yielding `selectedIndices: {-1}` → nothing highlighted (`app_chip_group.dart:158`). Not reachable today (no `setItems()` call sites), so defensive only; `items.first` on an empty list also throws (pre-existing). |
| 💡 | 🟢 | `lib/components/styled/menus/widgets/top_navigation_menu.dart:50` | The key string is an inline magic literal and nothing depends on it (no `find.byKey` in the tests); a `static const` prefix would make it testable. |
| 💡 | ⚪ | `lib/components/styled/menus/widgets/top_navigation_menu.dart:60-77`, `lib/components/styled/menus/widgets/menu_holder.dart:151-158` | Rapid consecutive taps fire two `goNamed()` calls with two concurrent `onExit` guards and bump `_syncToken` twice; the final highlight depends on which navigation lands first. Consider ignoring `select()` while a navigation is in flight. |
| 💡 | 🟡 | `lib/components/styled/menus/widgets/menu_holder.dart:154` | Pre-existing, adjacent to this fix: `key.currentContext!.goNamed(...)` force-unwraps; a tap during shell teardown would throw. The `currentContext == null` check at `:74` happens in an earlier frame. |
| 💡 | 🟡 | `lib/components/styled/menus/widgets/menu_holder.dart:72-85` | Pre-existing: `Future.doWhile(() => !mounted).then(...)` inside `build()` spins forever if the State is never mounted, and the `.then` body calls `_controller.setDisplayType(...)` without a mounted guard. Same nav-state-sync machinery; worth adding to the debt list. |

<sub>Confidence: 🟢 High = head code read, evidence quoted · 🟡 Med = file:line + rationale, not fully proven · ⚪ Low = speculative.</sub>
<sub>`[X]` = raised by 1 agent · `[XX]` = raised by both agents (Critical rows only; none this round).</sub>

<details>

**1-3. The `ValueKey` bump is redundant, and it costs focus.**
`top_navigation_menu.dart:50,58`:
```dart
key: ValueKey('top-nav-chip-group-$selectedIndex-$_syncToken'),
...
selectedIndices: {selectedIndex},
```
`AppChipGroup` (ui_kit_library v2.28.1) re-syncs like this:
```dart
// app_chip_group.dart:82-87
void didUpdateWidget(AppChipGroup oldWidget) {
  super.didUpdateWidget(oldWidget);
  if (widget.selectedIndices != oldWidget.selectedIndices) {
    _selectedIndices = Set.from(widget.selectedIndices);
  }
}
```
Dart's `Set` does not override `operator ==` (`LinkedHashSet` compares by identity), and `build()` allocates a fresh `{selectedIndex}` every time, so `widget.selectedIndices != oldWidget.selectedIndices` is **always true**. A bare `setState(() {})` therefore already re-asserts `widget.selected` — the comment at `:22-29` ("only re-syncing … when that prop changes") does not describe actual behaviour, and the key is not required to achieve the fix.

The key *does* have a cost. Chip taps are handled by `AppInteractionSensor`, which owns a `Focus` with an internal node:
```dart
// app_interaction_sensor.dart:574-577
child = Focus(
  focusNode: _effectiveFocusNode,
  autofocus: widget.autofocus,
  onKeyEvent: (node, event) { ... _handleKeyActivation(); ... }
```
```dart
// app_interaction_sensor.dart:234-241
void dispose() {
  _effectiveFocusNode.removeListener(_handleFocusChange);
  ...
  _internalFocusNode?.dispose();
```
Sequence: Tab to a chip → Enter → `_handleKeyActivation()` → `onTap` → `_toggleChip` → `onSelectionChanged` → `top_navigation_menu.dart:75` bumps the token → next frame the key differs → the old elements unmount and the `FocusNode` is disposed. Result: the focus ring disappears, focus falls back to the enclosing scope, and the user must Tab from the top again; screen readers also re-announce the group (`Semantics(label: 'Filter chips, N options')`, `app_chip_group.dart:121-122`).
Suggested fix: drop `$_syncToken` (and `$selectedIndex`) from the key, keep the `setState`, and correct the comment.

**7-8. Confirmed navigation through an async guard flickers.**
```dart
// menu_holder.dart:98-100
onItemClick: (index) { _controller.select(_controller.items[index]); }
// menu_holder.dart:151-158  → key.currentContext!.goNamed(path, extra: NavigationExtra(naviType: type));
// route_model.dart:76-88    → when isDirty(): await showUnsavedAlert(context)   // async
// menu_holder.dart:52-58    → only after navigation succeeds does the routerDelegate listener call _controller.setTo(...)
```
Tap → `goNamed()` starts awaiting the dialog → the token bump runs **synchronously** → next frame the highlight reverts to the old tab → user presses "Discard" → navigation completes → `setTo()` finally moves the highlight. So on dirty-check routes (e.g. `route_usp_dashboard.dart:100-101,127-128`) the highlight sits on the *wrong* tab for as long as the dialog is open. A result-driven revert (only revert when the guard rejects, e.g. a `MenuController.notifyRejected()`) would avoid this; if the flicker is considered acceptable, please pin it down with a test using an async host (`await Future.delayed(...)` before updating `selected`) so the intent is explicit — the current `_ConfirmingHost` cannot observe it.

</details>

<details>

- **No second source of truth introduced.** `_syncToken` is purely a rebuild trigger; `widget.selected` (← `MenuController` ← go_router `NavigationExtra.naviType`, `menu_holder.dart:52-58`) remains authoritative. Caching a `NaviType` locally would have been much worse.
- **Correct scoping decision to touch only the top nav.** `BottomNavigationMenu` renders `AppNavigationBar(currentIndex: items.indexOf(...))`, and `app_navigation_bar.dart:173` computes `isSelected = index == widget.currentIndex` purely from the prop with no internal optimistic state — so mobile does not exhibit #1158 and needs no matching change.
- **Zero security surface**: both files contain no secrets/tokens, no `flutter_secure_storage` change, no external input parsing, no JNAP/USP call, no permission logic. USP error-handling rules (`catch (e)` vs `on Exception catch`, `mapUspErrorToServiceError()`, `result != OK` / `_Error` interception) are not applicable to this diff — no `await`, no `try/catch`, no service call.
- **Architecture hard rules pass**: no `generated/` import; `providers/`/`services/` untouched (Art. V §5.4); no USP provider changes so Art. IV's four rules are N/A; no `ServiceError` violations (Art. XIII); `AppChipGroup` reused rather than hand-rolled (Art. XV); `NavigationMenus extends Equatable` already (`menu_holder.dart:169`, Art. XI).
- **Good scope discipline** (Art. I §1.3): one widget + one new test, no drive-by refactors or lint churn.
- **The comment answers the reviewer's first question** ("why not `didUpdateWidget`?") up front — a genuinely useful habit, even though the premise needs correcting.
- **Test file mirrors the lib path exactly**, names are descriptive and unnumbered, and it covers both the cancel and the confirm direction with clear baseline/act/assert structure and `reason:` messages.
- Asserting selection via `Semantics.properties.selected` (matching `app_chip_group.dart:176-179`) rather than font weight is the right instinct — resistant to visual restyling — even if the lookup mechanism could be sturdier.
- No new dependencies, no `print`, no commented-out code.
- One candidate defect was investigated and **cleared**: the `selectedIndices.isEmpty` branch at `:63` is unreachable, because in `ChipSelectionMode.single` `_toggleChip` always assigns `_selectedIndices = {index}` (`app_chip_group.dart:101-116`) — tapping an already-selected chip cannot deselect it. No missing-guard bug here.

</details>

<sub>Note: no Flutter/Dart toolchain was available in the review environment, so `flutter test` / `flutter analyze` were **not** executed. Findings come from reading the head revision plus the resolved `ui_kit_library` v2.28.1 sources.</sub>

PR #1178, opened 2026-07-28
fix(dashboard): reflect WAN disconnect in Network Health card (#1143)

### LINKED ISSUE #1143: Dashboard: Network Health WAN status doesn't reflect a WAN disconnect, doesn't match page banner.
**Reproduce steps:**
1. Login to Linksys Now.
2. Unplug the WAN cable.
3. Check the Network Health card.

**Actual result:**
WAN status is shown as **Excellent** (Health 100 / Errors 0/s / Loss 0.00%) even though the WAN is disconnected — the page-top banner on the same screen correctly shows **"Couldn't establish a connection…"**. The card and the banner don't match each other.

**Expected result:**
When the WAN is down, the Network Health card should reflect the disconnect (degraded WAN status / not "Excellent"), consistent with the connection banner.

**Env.:** M60DU-EU / Pinnacle 2.0 FW 1.2.2.26062820 / Linksys Now 2.5.0.700525

**Log:**
[GUI_networkhealth_wanerror_0716.txt](https://github.com/user-attachments/files/30070509/GUI_networkhealth_wanerror_0716.txt)

**Screenshot:**
Image

AustinChangLinksys:
## Summary

Refs linksys/PrivacyGUI#1143

The Network Health card scored WAN "Excellent" while the router was disconnected from the
internet. A disconnected link carries no traffic, so packet-loss-based scoring computed 0%
loss → score 100 → "Excellent", directly contradicting the page-top connection banner.

## Root cause

`_HealthOverview._buildContent` called `NetworkHealthHelpers.computeHealthScore(wan)` (pure
loss-based) for the WAN tile with no awareness of physical link state.

## Fix

- **`lib/page/_shared/models/network_health_helpers.dart`** — new pure helper
  `computeWanScore(wan, {bool wanIsUp = true})` that forces `0` when `wanIsUp` is `false`,
  delegates to `computeHealthScore` otherwise. Default `true` guards the provider-loading
  window against a false disconnect signal.

- **`lib/page/dashboard/views/components/usp_network_health_card.dart`** —
  - `_UspNetworkHealthCardState.build` watches `wanDataProvider` (same signal the banner uses)
    and passes `wanIsUp` down the widget tree.
  - `_HealthOverview` receives `wanIsUp` and calls `computeWanScore` for the WAN score.
  - `_TrafficLight` gains an optional `statusOverride` parameter; when WAN is down it renders
    "Disconnected" instead of the health-tier string, so the card label matches the banner.

- **`test/page/_shared/models/network_health_helpers_test.dart`** — 6 new unit tests covering:
  - WAN down + healthy stats → score 0 / Critical (the exact #1143 bug)
  - WAN down + null snapshot → score 0 / Critical
  - WAN up + healthy stats → score 100 / Excellent
  - WAN up + lossy stats → degraded score (delegates correctly)
  - null snapshot + WAN up → 100 (healthy default, no false disconnect)
  - wanIsUp defaults to true (loading guard)

## Changed files

| File | +/- |
|------|-----|
| `lib/page/_shared/models/network_health_helpers.dart` | +22 / 0 |
| `lib/page/dashboard/views/components/usp_network_health_card.dart` | +33 / -7 |
| `test/page/_shared/models/network_health_helpers_test.dart` | +96 / 0 (new) |

**Diff stat:** 3 files changed, 151 insertions(+), 7 deletions(-)

## Test evidence (verify-worker independent re-run)

```
$ fvm flutter test test/page/_shared/models/network_health_helpers_test.dart --reporter=expanded
...
+6: All tests passed!

$ ./run_tests.sh
...
+3333: All tests passed!
```

## Quality gates

- `fvm dart format --output=none --set-exit-if-changed .` → 0 changed (clean)
- `fvm flutter analyze` → No issues found
- No Material widgets introduced in view layer (AppText, AppGap used throughout)
- No `generated/` imports in View/Provider layer (Article V §5.4.3 compliant)
- No `.arb` changes → no gen-l10n required (reused existing `loc().disconnected` key)


AustinChangLinksys (commented):
🤖 **Automated Review — Round 1** · `b7137ea..9b86e66` (full)

**Verdict: ✅ APPROVE** — Core fix is correct and well-reasoned; two sister pages carry the same original bug, and the Errors/Loss chart tabs have no disconnect indicator.
_(Note: GitHub self-approve blocked — verdict intent is APPROVE; please approve manually or via a different reviewer account.)_

#### Standards

|---|---|---|---|
| ⚠️ | 🟢High | `stats_health_score_section.dart:41` | Sister page (Statistics view) still calls `computeHealthScore` directly — WAN disconnect not reflected here |
| ⚠️ | 🟢High | `usp_pdf_service.dart:239` | PDF report ignores `PdfReportData.wanStatus.isUp` (already populated) — WAN-down score still prints 100/Excellent |
| ⚠️ | 🟡Med | `usp_network_health_card.dart:116–119` | WAN down: Loss, Error, and Discard chips still display `0.00%`/`0/s` — contradicts the "Disconnected" label on same card |
| ⚠️ | 🟡Med | `usp_network_health_card.dart:38` | [Duplicated Code] `wanIsUp ?? true` fallback also in `usp_sliver_dashboard_view.dart:144` — "default-to-up" semantic has no single source of truth |
| ⚠️ | 🟡Med | `usp_network_health_card.dart` | Widget-layer test gap: WAN-down UI rendering and `wanDataProvider` loading-fallback (`?? true`) paths have no widget tests |
| ⚠️ | 🟡Med | `usp_network_health_card.dart:109–110` | Gauge centre shows "0 / Critical" while WAN traffic-light shows "Disconnected" — same state, two vocabularies |
| ⚠️ | 🟡Med | `usp_network_health_card.dart:38` | Transient "Excellent" flash: SSE `invalidateSelf()` → `AsyncLoading` → `valueOrNull==null` → `?? true` while WAN known-down |
| ⚠️ | 🟡Med | `network_health_helpers.dart` | [Divergent Change] `computeWanScore` conflates connectivity-gate logic with loss-based scoring — two separate change axes |
| 💡 | 🟡Med | `network_health_helpers.dart` | `computeWanScore` default `wanIsUp=true` also silently covers `AsyncError`; doc comment should mention this |

#### Spec (Issue #1143)

|---|---|---|---|
| ⚠️ | 🟡Med | `switch(selectedTab)` cases 1, 2 | Spec: "card should reflect the disconnect" — Errors and Loss chart tabs show no disconnect indicator when WAN is down |

<details>

**Stats-page sister page not updated** 🟢High
- `lib/page/statistics/views/sections/stats_health_score_section.dart:41`: `wan != null ? NetworkHealthHelpers.computeHealthScore(wan) : 100` — original code, verified on PR head `9b86e66`.
- The Statistics view renders the same WAN health gauge; will still show "Excellent" when WAN is disconnected.
- **Fix:** Apply `computeWanScore(wan, wanIsUp: wanIsUp)` + `ref.watch(wanDataProvider)` in `StatsHealthScoreSection` (convert to `ConsumerWidget` if needed).

**PDF report not updated** 🟢High
- `lib/page/_shared/services/usp_pdf_service.dart:239`: `NetworkHealthHelpers.computeHealthScore(wan)` — ignores `PdfReportData.wanStatus` which already carries `isUp` (set at `pdf_report_data_provider.dart:62`). The document correctly prints "WAN: Down" in `_buildWanStatus` but the health score section uses the stale calculation.
- **Fix:** `final wanIsUp = data.wanStatus?.isUp ?? true; final score = NetworkHealthHelpers.computeWanScore(wan, wanIsUp: wanIsUp);`

**WAN-down metric chips show 0 (misleading)** 🟡Med
- `usp_network_health_card.dart:116–119`: Loss/Error/Discard computed unconditionally. A down link carries no traffic, so all read 0 — contradicts the "Disconnected" traffic-light label.
- **Fix:** When `!wanIsUp`, render `--` or `N/A` instead of `0.00%`/`0/s`.

**Duplicated wanIsUp fallback** 🟡Med
- Same `ref.watch(wanDataProvider).valueOrNull?.model.isUp ?? true` pattern in both `usp_network_health_card.dart:38` and `usp_sliver_dashboard_view.dart:144`.
- **Fix:** Extract `wanIsUpProvider = Provider<bool>((ref) => ref.watch(wanDataProvider).valueOrNull?.model.isUp ?? true)`.

**Widget-layer test gap** 🟡Med
- No widget test covers WAN-down rendering or `wanDataProvider` loading-state fallback. The new `network_health_helpers_test.dart` covers the helper well; the gap is at widget integration level.
- **Fix:** Add `test/page/dashboard/views/components/usp_network_health_card_test.dart` covering (1) WAN up, (2) WAN down, (3) provider loading-state fallback.

**Gauge "Critical" vs traffic-light "Disconnected"** 🟡Med
- Gauge displays `0 / Critical` (score tier) while WAN traffic-light shows `Disconnected` (override string) — two vocabulary systems for the same state.
- **Consider:** Adding a `centerBuilder` override showing `--` / "Disconnected" when `!wanIsUp`.

**Transient "Excellent" flash on SSE refresh** 🟡Med
- SSE `wanStatus` event → `wanDataProvider.invalidateSelf()` → `AsyncLoading` → `valueOrNull==null` → `?? true` for ~1 frame even when WAN is known-down.
- **Fix:** Consider `keepPreviousData`-style logic or hold a local `_prevWanIsUp` in widget state.

</details>

<details>

- `computeWanScore` correctly returns 0 when `wanIsUp=false` — spec R-1 (not "Excellent") fully met.
- Both card and banner watch `wanDataProvider` — same signal source; spec R-3 (consistency with banner) correct.
- `statusOverride` uses `loc(parentContext).disconnected` — properly localised (not hardcoded English).
- Doc comment on `computeWanScore` clearly explains the WAN-no-traffic invariant.
- 6 unit tests for `computeWanScore` cover key branches well.
- `wanIsUp = true` default is the right conservative fallback for the loading case.

</details>

lib/page/dashboard/views/components/usp_network_health_card.dart
```diff
@@ -33,6 +34,10 @@ class _UspNetworkHealthCardState extends ConsumerState<UspNetworkHealthCard> {
   Widget build(BuildContext context) {
     final trafficState = ref.watch(uspTrafficAnalysisProvider);
     final selectedTab = ref.watch(cardTabIndexProvider(_cardId));
+    // Physical WAN link state — same signal the page-top connection banner
+    // uses. Consulted so a disconnected WAN is not scored "Excellent" purely
+    // because a down link carries no traffic (loss 0%). See #1143.
+    final wanIsUp = ref.watch(wanDataProvider).valueOrNull?.model.isUp ?? true;
```
PeterJhongLinksys: ⚠️ `valueOrNull` returns null for both `AsyncLoading` **and** `AsyncError`, so a first-load failure of `wanDataProvider` (which throws a `ServiceError` from `usp_wan_data_service.dart:65`) is treated as "WAN is up" and the card can still report 100/"Excellent" — the original #1143 symptom, with no error surface. Consider distinguishing a third `unknown` state and rendering `--`.

lib/page/_shared/models/network_health_helpers.dart
```diff
@@ -29,6 +29,28 @@ class NetworkHealthHelpers {
...
+  /// disconnected — contradicting the page-top connection banner.
+  /// See linksys/PrivacyGUI#1143.
+  ///
+  /// [wanIsUp] mirrors `WanStatusUIModel.isUp` (the same signal the banner
+  /// uses). It defaults to `true` so that, when link state is momentarily
+  /// unavailable (provider still loading), behaviour matches the prior
+  /// loss-only scoring rather than falsely reporting a disconnect.
+  static int computeWanScore(
+    InterfaceTrafficSnapshot? wan, {
+    bool wanIsUp = true,
```
PeterJhongLinksys, line 48: ⚠️ `bool wanIsUp = true` is a fail-open default: a caller that forgets the argument silently regresses to the buggy loss-only behaviour with no compiler help. Two such call sites already exist and were missed in this PR (`stats_health_score_section.dart:40-41`, `usp_pdf_service.dart:239`). Making this `required` (one current call site) — or better, a tri-state `WanLinkState { up, down, unknown }` — would prevent that and also let the loading/error case be expressed explicitly.

PeterJhongLinksys (approved):
🤖 **Automated Review — Round 1** · `557d59c..2f9a702` (full)

**Verdict: ✅ APPROVE** — The core fix is correct, well-documented and unit-tested; no Critical issues. Approving with Warnings: the same #1143 defect still exists in Statistics and the PDF report, the `wanIsUp = true` default is fail-open, and the new View wiring itself has no test coverage.

|---|---|---|---|
| ⚠️ | 🟢 | `lib/page/statistics/views/sections/stats_health_score_section.dart:40-41`, `lib/page/_shared/services/usp_pdf_service.dart:239` | [Duplicated Code] Same #1143 bug unfixed in Statistics page and PDF report — both still call `computeHealthScore(wan)` and will show 100/"Excellent" on a down WAN, contradicting the Dashboard card. |
| ⚠️ | 🟢 | `lib/page/_shared/models/network_health_helpers.dart:48` | `bool wanIsUp = true` is a fail-open default: any caller that forgets the argument silently regresses to the buggy behaviour (exactly what happened above). Consider `required`, or a tri-state `WanLinkState`. |
| ⚠️ | 🟢 | `lib/page/dashboard/views/components/usp_network_health_card.dart:40` | `valueOrNull ?? true` collapses `AsyncLoading` and `AsyncError` into "WAN is up": on first-load failure (service throws `ServiceError`) the card can still report Excellent, with no error surface. |
| ⚠️ | 🟢 | `lib/page/dashboard/views/components/usp_network_health_card.dart:77-88` | The fix is unreachable when `state.history.isEmpty` — `_HealthOverview` is never built, so "just opened the page while WAN is down" still shows "enable traffic monitor" instead of a disconnect. |
| ⚠️ | 🟢 | `test/page/_shared/models/network_health_helpers_test.dart:1-96` | Tests cover only the pure helper; they would all pass even if the View wiring (`:40`, `:169`) were deleted. No regression guard on the actual fix, and `_TrafficLight.statusOverride` (`:229`) is uncovered. |
| ⚠️ | 🟡 | `test/golden_test/page/dashboard/cards/localizations/usp_analytics_cards_test.dart:111-121` | `card_network_health` states only override `trafficAnalysisState`; after this PR the card watches an un-mocked `wanDataProvider`, so the test reaches the real service layer. Fixtures already exist (`testWanOfflineData`). |
| ⚠️ | 🟡 | `test/page/_shared/models/network_health_helpers_test.dart:7-26` | Private `_snapshot()` builder instead of a shared Test Data Builder; `InterfaceTrafficSnapshot` construction now exists in 3 places (Art. VIII). |
| ⚠️ | 🟡 | `lib/page/dashboard/views/components/usp_network_health_card.dart:40,111-135` | [Feature Envy] The View assembles cross-provider business logic (2 providers → null-checks → wanScore/lanScore → `math.min` → tier). This lack of a single owner is the structural reason the Statistics/PDF copies were missed. |
| ⚠️ | 🟡 | `lib/page/dashboard/views/components/usp_network_health_card.dart:37-39`, `lib/page/_shared/models/network_health_helpers.dart:42-45` | Comments claim "same signal the page-top connection banner uses", but `usp_connection_status_banner.dart:26-27` derives connectivity from `staticIpAddress.isNotEmpty`, not `isUp` (and `usp_ethernet_ports_card.dart:60` uses a third definition). Only the Dashboard offline banner matches. |
| 💡 | 🟢 | `lib/page/dashboard/views/components/usp_network_health_card.dart:61,65` | [Long Parameter List] `wanIsUp` is threaded into the Errors and Loss tabs, which never consume it (dead argument); those charts also draw a flat 0-line with no disconnect note. Also a positional `bool` adjacent to another primitive at the call site. |
| 💡 | 🟢 | `test/page/_shared/models/network_health_helpers_test.dart:51-57,81-86` | The `wan == null` branches are unreachable in production (`interfaces` is built for all `TrafficInterface.values`, and `latest == null` early-returns), so these two tests give a false sense of null protection. |
| 💡 | 🟢 | `test/page/_shared/models/network_health_helpers_test.dart:69-79` | Assertions are weak (`lessThan(100)`) although the comment already derives the exact value 70; `_scoreFromLoss` thresholds (0.01/0.1/1.0/5.0) and `tierFromScore` boundaries (95/80/60/40) remain untested. |
| 💡 | 🟡 | `lib/page/dashboard/views/components/usp_network_health_card.dart:142-152` vs `:169` | On a down WAN the gauge centre still reads "0 / Critical" while the traffic light reads "Disconnected" — 0 can be misread as "measured terrible quality" rather than "no link". |
| 💡 | 🟡 | `lib/page/dashboard/views/components/usp_network_health_card.dart:215-229` | `statusOverride` is a generic escape hatch with temporal coupling: the caller must also pass `tier: critical` or you get a green dot labelled "Disconnected". A semantic type would encapsulate this. |
| 💡 | 🟡 | `lib/page/dashboard/views/components/usp_network_health_card.dart:40` + `lib/page/_shared/models/network_health_helpers.dart:48` | [Duplicated Knowledge] The "unknown link state ⇒ treat as up" policy is written twice; changing one side will drift. |
| 💡 | 🟡 | `lib/page/_shared/models/network_health_helpers.dart:27,46` | Naming: `computeHealthScore` is really a loss-only score and its parameter is named `wan` although LAN callers use it; `computeWanScore` reads like a specialisation of it rather than the wider one. Consider `computeLossScore` / `computeEffectiveWanScore`. |
| 💡 | ⚪ | `lib/page/internet_settings/services/usp_wan_data_service.dart:55` | `isUp` only accepts `'up'`; TR-181 `Dormant` (on-demand PPPoE idle) would now hard-zero the score and render "Disconnected". Pre-existing, but this PR amplifies the blast radius. |
| 💡 | ⚪ | `lib/generated/wan_status.g.dart:26-39` | Instance resolution falls back to a hard-coded `Device.IP.Interface.2.` when no `Alias=='wan'` is found; a wrong-interface `isUp` now drives a whole-card zero + "Disconnected". Pre-existing upstream risk. |
| 💡 | ⚪ | `lib/page/_shared/models/network_health_helpers.dart` (file) | Plural file name for a stateless util placed under `models/` (Art. III wants singular; this is not a model). Pre-existing, noted only. |

<sub>Confidence: 🟢 High = head code read, evidence quoted · 🟡 Med = file:line + rationale, not fully proven · ⚪ Low = speculative.</sub>
<sub>`[X]` = raised by 1 agent · `[XX]` = raised by both agents (Critical rows only; none this round).</sub>

<details>

**1. Same #1143 bug remains in Statistics and PDF** — `stats_health_score_section.dart:40-41` is a near line-by-line copy of the block this PR fixed:
```dart
final wanScore =
    wan != null ? NetworkHealthHelpers.computeHealthScore(wan) : 100;
```
and `usp_pdf_service.dart:239`:
```dart
final score = NetworkHealthHelpers.computeHealthScore(wan);
```
With the WAN down, a down link carries no traffic → loss 0% → both still compute 100 / "Excellent". A user who switches to the Statistics page or exports a PDF sees a direct contradiction with the Dashboard card. The data is already at hand in both places: Statistics can `ref.watch(wanDataProvider)` the same way, and `PdfReportData.wanStatus` already exists (`pdf_report_data.dart:72`, populated at `usp_pdf_service.dart:256-260`). Suggested fix: route both through `computeWanScore(..., wanIsUp: ...)`, or better, extract the whole WAN/LAN/overall aggregation once (see the Feature Envy row) so there is a single owner.

**2. Fail-open default argument** — `network_health_helpers.dart:46-52`:
```dart
static int computeWanScore(
  InterfaceTrafficSnapshot? wan, {
  bool wanIsUp = true,
}) {
  if (!wanIsUp) return 0;
  return wan != null ? computeHealthScore(wan) : 100;
}
```
"Forgot to pass it" == "silently back to the buggy behaviour", with no compiler help. There is exactly one call site today, so making it `required` is essentially free; the loading-tolerance rationale in the DartDoc is a *caller* policy and is better expressed at the View. A `WanLinkState { up, down, unknown }` would additionally make the third state expressible (see warning 3).

**3. Error state treated as "connected"** — `usp_network_health_card.dart:40`:
```dart
final wanIsUp = ref.watch(wanDataProvider).valueOrNull?.model.isUp ?? true;
```
`wanDataProvider` is an `AsyncNotifierProvider` (`wan_data_provider.dart:28-29`) whose `_fetch()` throws `mapUspErrorToServiceError(e)` (`usp_wan_data_service.dart:65`). `valueOrNull` returns `null` for **both** `AsyncLoading` and `AsyncError`. On a first-load failure there is no previous value, so `wanIsUp` becomes `true` and the card can advertise 100/Excellent — the original #1143 symptom, and precisely at the moment the WAN is most likely broken. On a *refresh* failure Riverpod 2 keeps the stale value, so an outdated `isUp: true` can persist indefinitely with no error surface (Art. XIII expects `ServiceError` to be presented). Suggested fix: distinguish three states and render `--` / `unknown` when the link state is not known.
*(Note: `usp_sliver_dashboard_view.dart:138` uses the same idiom, so this follows an existing convention — but there the consequence is only "banner not shown", whereas here the card makes a positive "Excellent" claim.)*

**4. Fix bypassed when history is empty** — `usp_network_health_card.dart:77-88`:
```dart
if (state.history.isEmpty) {
  return Center(
    child: AppText.bodyMedium(
      loc(context).enableTrafficMonitorForHealthData, ...
```
`_HealthOverview` (which owns all of the new `wanIsUp`/`statusOverride` logic) is never constructed in that branch. Meanwhile the traffic poller does not append history when the fetch fails (`usp_traffic_analysis_notifier.dart:137-148` — `ConnectivityError` and the generic `catch (e)` both only set `isFetching: false`) or when the monitor has not started yet (same file, 63-69, requires authenticated + domainReady). So the most common real-world path — "open the dashboard while the WAN is already down" — still does not reflect the disconnect. Suggested fix: evaluate WAN-down before the `history.isEmpty` guard and prefer showing "Disconnected" over "enable traffic monitor".

**5. No coverage of the actual fix** — all six new tests target `NetworkHealthHelpers`, and would stay green if `usp_network_health_card.dart:40` were reverted to `true`. The infrastructure to test the real wiring already exists: `test/golden_test/golden_framework/mocks/mock_dashboard_cards.dart:162` (`cardOverrides(wanData:)`) plus `cards_test_data.dart:94 testWanOffline` / `:116 testWanOfflineData`. A widget test with `ProviderScope(overrides: ...)` for WAN-down + non-empty history asserting "Disconnected" and gauge 0 would close both this and warning 6, and would also have caught warning 4.

</details>

<details>

- `computeWanScore` is a pure static function with no side effects, consistent with the file header contract ("suitable for use inside widget build methods", `network_health_helpers.dart:16-21`); no async/timer introduced in `build`, so no new race.
- No extra polling: `wanDataProvider` is a non-autoDispose `AsyncNotifier` already held by `usp_sliver_dashboard_view.dart:137` and `dashboard_orchestrator.dart:81`, so the additional `ref.watch` issues no new USP request; the SSE `InvalidationDomain.wanStatus` hook (`wan_data_provider.dart:37-41`) gives the card correct live reactivity.
- Architecture hard rules all pass: three-layer dependency direction (View → Provider → Service) intact; no `generated/` import in `providers/` or `views/` (Art. V §5.4); no `PreservableContract` redefinition, no autoDispose on L1, no mutation (so no `uspMutationLockProvider` needed) (Art. IV); no `catch`/exception leakage (Art. XIII); UI Kit reused (`AppText`/`AppGap`/`AppGauge`, Art. XV); read-only View watching an L1 directly is the sanctioned Type C pattern.
- The USP error path this feature depends on is correct upstream: `usp_wan_data_service.dart:64-66` uses `catch (e)` (not `on Exception catch`) and `throw mapUspErrorToServiceError(e)` — necessary because `wan_status.g.dart:80` throws a bare `String`. This PR does not weaken that chain.
- `statusOverride` is nullable and only overrides the label, leaving colour driven by `tier`, so the existing LAN call site (`:172-176`) is untouched.
- `loc(context).disconnected` already exists in all 26 `.arb` locales — no missing-key runtime risk.
- DartDoc on `computeWanScore` is genuinely good: it states the bug, links #1143, and explains *why* the default is `true`. The new test opens with a sanity assertion proving the loss-only score is 100 before asserting the fix, making the reproduction intent explicit.
- `computeHealthScore`'s signature is unchanged, so statistics/PDF/mascot call sites are not broken; the semantics also match the pre-existing correct pattern in `internet_dimension.dart:35-39` (`if (!isUp) return 0;`).
- Test naming is descriptive, group format compliant, no `TC00x` numbering.

</details>

<sub>Note: no Flutter/Dart toolchain was available in the review environment, so `flutter test` / `flutter analyze` were **not** executed. All findings come from reading the head revision source.</sub>

AustinChangLinksys (commented):
🤖 **Automated Review — Round 4** · `2f9a702..594965e` (incremental)

**Verdict: ✅ APPROVE (comment only — GitHub self-approve blocked)** — All 8 prior-round warnings resolved; 2 new Suggestions (low risk). Round 4 with 0 Critical: convergence cap reached — auto-fix paused, handed to Austin.
⚠️ Convergence cap reached (round 4, still 0 Critical) — auto-fix paused, handed to Austin.

#### Standards

|---|---|---|---|
| 💡 | High | `usp_network_health_card.dart:18` + `stats_health_score_section.dart:15` | [Duplicated Code] `_kNoTrafficPlaceholder = '--'` defined identically as file-private const in two files |
| 💡 | High | `ipv6port_service.g.dart:163` + `usp_ipv6_port_service_service.dart:194` | parse-failure fallback is epoch(1970) but system-rule filter sentinel is 0001-01-01 — two constants not aligned; low risk but easy to fix |

#### Spec (Issue #1143)

_All spec requirements now resolved. See resolved items below._

<details>
<summary>✅ Prior round warnings resolved (all 8)</summary>

1. **stats_health_score_section.dart:41 — Sister page computeHealthScore** ✅ Resolved
   - Commit `11a0418b`: `stats_health_score_section.dart` now uses `computeWanScore(wan, wanIsUp: wanIsUp)` with `ref.watch(wanIsUpProvider)`. Verified at line 51.

2. **usp_pdf_service.dart:239 — PDF ignores wanStatus.isUp** ✅ Resolved
   - Commit `11a0418b`: `final wanIsUp = data.wanStatus?.isUp ?? true; final score = NetworkHealthHelpers.computeWanScore(wan, wanIsUp: wanIsUp);` at line 243-244. Metric values show N/A when WAN down.

3. **usp_network_health_card.dart:116-119 — Metric chips show 0.00%/0/s when WAN down** ✅ Resolved
   - Commit `11a0418b`: lossText/errorText/discardText now use `_kNoTrafficPlaceholder = '--'` when `!wanIsUp`.

4. **usp_network_health_card.dart:38 — [Duplicated Code] wanIsUp fallback duplicated** ✅ Resolved
   - Commit `11a0418b`: Extracted `wanIsUpProvider` in `wan_data_provider.dart:39` as single source of truth.

5. **usp_network_health_card — Widget-layer test gap** ✅ Resolved
   - Commit `11a0418b`: `test/page/dashboard/views/components/usp_network_health_card_test.dart` added (200 lines), covering WAN down/up/narrow/loading.

6. **usp_network_health_card:109-110 — Gauge '0 / Critical' vs 'Disconnected'** ✅ Resolved
   - Commit `11a0418b`: Gauge center shows `_kNoTrafficPlaceholder`/`loc(parentContext).disconnected` when `!wanIsUp`.

7. **usp_network_health_card:38 — Transient 'Excellent' flash on SSE refresh** ✅ Resolved
   - `wanIsUpProvider` uses `valueOrNull?.model.isUp ?? true`; `valueOrNull` preserves last value during AsyncLoading, no flash.

8. **Spec Warning: Errors/Loss chart tabs no disconnect indicator** ✅ Resolved
   - Commit `11a0418b`: `_ErrorsChart` and `_LossChart` each gained an inline `_DisconnectNotice` shown when `!wanIsUp`.

</details>

<details>
<summary>Suggestion details</summary>

**[Duplicated Code] _kNoTrafficPlaceholder defined in two files** (High confidence)
- `usp_network_health_card.dart:18`: `const String _kNoTrafficPlaceholder = '--';`
- `stats_health_score_section.dart:15`: `const String _kNoTrafficPlaceholder = '--';`
- Both are file-private; if one is changed the other stays stale, causing UI inconsistency.
- **Fix**: Extract to `network_health_helpers.dart` or a new `lib/page/_shared/models/health_constants.dart` as `const kNoTrafficPlaceholder = '--';`.

**ipv6port_service.g.dart fallback sentinel not aligned with service-layer filter** (High confidence)
- `ipv6port_service.g.dart:163`: parse-failure fallback = `DateTime.fromMillisecondsSinceEpoch(0, isUtc: true)` = 1970-01-01
- `usp_ipv6_port_service_service.dart:194`: `_systemRuleCreationDate = DateTime.parse('0001-01-01T00:00:00Z')` = 0001-01-01
- These two dates are not equal. If CPE returns an unparseable CreationDate, the rule gets epoch(1970) and is NOT caught by the 0001-01-01 sentinel — leaks as a user-visible rule.
- Risk is low (well-formed firmware always returns parseable dates) but the mismatch is confusing.
- **Fix**: Align the generated fallback to match the sentinel: `DateTime.tryParse(...) ?? DateTime.parse('0001-01-01T00:00:00Z')`, or update the service filter to also check for epoch(0).

</details>

<details>
<summary>✅ What looks good (new in Round 4)</summary>

- **app_connection_state_provider.dart — re-login fix**: Guard `state == AppConnectionState.loggedOut` correctly restores `authenticated` without clobbering `waitingForRecovery`. Tests cover re-login restore + waitingForRecovery guard.
- **wan_interface_path_provider.dart**: `resolveWanInterfacePath` try/catch ensures fallback to `kWanInterfaceFallbackPath` (= old hardcoded `.2.`) on any error — strictly non-regressing. Serial-number cache key correctly re-resolves on router switch.
- **SSE invalidation classifier**: Uses `ref.read(wanInterfacePathProvider).valueOrNull ?? fallback` correctly (non-reactive synchronous read on hot path). Trailing-dot `kWanInterfaceFallbackPath` prevents false prefix matches like Interface.20.
- **Codegen 0.16.0 type corrections**: `DateTime.tryParse` replaces fragile `DateTime.parse` across all timestamp fields — no more FormatException on malformed CPE responses. `wan-status-valuechange` subscription path trailing-dot fix is a correct defensive improvement.
- **mesh_topology_builder.dart**: `?.toIso8601String()` handles newly-nullable `multiApLastContactTime` correctly.
- **Test coverage**: `wan_interface_path_provider_test.dart` adds 4 provider scenarios; `sse_invalidation_provider_test.dart` adds WAN classification cases; widget tests for usp_network_health_card and network_health_helpers added.

</details>

PR #1190, opened 2026-07-30
fix(dhcp): repair DHCP detail page layout at mobile width (#1140)

PeterJhongLinksys:
## Summary

Fixes #1140. On the DHCP Settings detail page at mobile width, the Active
Leases list broke: the device name and MAC wrapped **one character per line**
and the lease timeline overflowed the card (visible Flutter overflow stripes).
The DHCP Reservations card on the same page had the same root cause, wrapping
the MAC one octet per line.

## Root cause

Both rows sized a column with a fixed width from `context.colWidth(2)`.
`colWidth` measures against the **page grid** (4 columns on mobile, 12 on
desktop), not against the widget's own constraints — at 480px it returns 216dp
while the nested row's usable width is only 398dp.

| Row | Arithmetic | Symptom |
|---|---|---|
| Active Leases | 8 (dot) + 8 (gap) + 216 (IP) + 216 (lease) = **448** vs **398** | 50px overflow; name `Expanded` starved to 0 → one char per line |
| Reservations | flexible space **242**, IP box takes **216** | no overflow, but MAC column too narrow for one line → one octet per line |

## Changes

- **Active Leases** — columns sized by flex (4:3:3) with gaps instead of fixed
  `colWidth` boxes; every text gets `maxLines: 1` + ellipsis fallback.
- **Reservations** — MAC stacked over IP instead of side by side, and
  `colWidth` dropped from the IP. A full MAC needs 238dp and an IP 156dp
  against 242dp of flexible space, so side by side is physically impossible at
  mobile width — a flex split ellipsised the MAC, the row's only device
  identifier. Stacking gives the text column the row's full width, and fixes
  desktop too (the MAC was previously truncated to `AA:BB:CC:DD:E...` there).
- Each fix carries an inline comment naming the cause and `#1140`, so the
  anti-pattern isn't reintroduced.

## Files

| File | Change |
|---|---|
| `lib/page/dhcp/views/components/usp_dhcp_active_leases_card.dart` | flex 4:3:3, single-line texts |
| `lib/page/dhcp/views/components/usp_dhcp_reservations_detail_card.dart` | MAC over IP, drop `colWidth` |
| `test/page/dhcp/views/components/usp_dhcp_active_leases_card_test.dart` | new, 3 tests |
| `test/page/dhcp/views/components/usp_dhcp_reservations_detail_card_test.dart` | new, 3 tests |

## Verification (real output)

- `fvm dart format --output=none --set-exit-if-changed lib/page/dhcp/ test/page/dhcp/` → clean
- `flutter analyze lib/page/dhcp test/page/dhcp` → `No issues found!`
- `flutter test test/page/dhcp/ --tags ui` → **13 passed**
- `flutter test test/page/dhcp/ test/page/local_network/providers/dhcp_data_provider_test.dart` → **63 passed**
- `flutter test test/golden_test/page/dhcp/ --tags golden` (no `--update-goldens`) → **14 passed**
- `goldens/overflow_warnings.json`: **32 entries → absent** (zero RenderFlex
  overflows on this page)
- Regenerated goldens inspected visually at phone480 and desktop1280.

New tests assert measurable geometry, not golden bytes. Observed before the fix:
260dp and 340dp text heights (single line is <40dp) and a lease column
extending to 473dp past a 435dp row bound.

## Out of scope

The same `colWidth`-inside-`Row` pattern also exists in
`usp_wifi_status_card.dart` and `usp_info_row.dart` on other pages. Not touched
here — happy to file a follow-up issue.

Refs #1140

### DIFF
--- a/lib/page/dhcp/views/components/usp_dhcp_active_leases_card.dart
+++ b/lib/page/dhcp/views/components/usp_dhcp_active_leases_card.dart
@@ -7,6 +7,15 @@ import 'package:privacy_gui/page/_shared/models/dhcp_client_ui_model.dart';
 import 'package:privacy_gui/page/dhcp/providers/dhcp_client_filter_provider.dart';
 import 'package:ui_kit_library/ui_kit.dart';
 
+/// Flex weights of the lease row's three text columns (#1140).
+///
+/// The name column carries the longest strings (a host name, or a MAC when
+/// there is none), so it gets the largest share; the IP and lease columns hold
+/// fixed-length values and split the rest evenly.
+const _nameColumnFlex = 4;
+const _ipColumnFlex = 3;
+const _leaseColumnFlex = 3;
+
 /// Read-only card displaying DHCP client leases with filter chips.
 class UspDhcpActiveLeasesCard extends ConsumerWidget {
   final List<DhcpClientUIModel> clients;
@@ -112,7 +121,7 @@ class UspDhcpActiveLeasesCard extends ConsumerWidget {
             ),
             AppGap.sm(),
             Expanded(
-              flex: 4,
+              flex: _nameColumnFlex,
               child: Column(
                 crossAxisAlignment: CrossAxisAlignment.start,
                 children: [
@@ -121,6 +130,8 @@ class UspDhcpActiveLeasesCard extends ConsumerWidget {
                     maxLines: 1,
                     overflow: TextOverflow.ellipsis,
                   ),
+                  // displayName already falls back to the MAC when there is no
+                  // host name, so only show it again when it is the subtitle.
                   if (client.hostName.isNotEmpty)
                     AppText.bodySmall(
                       client.mac,
@@ -133,7 +144,7 @@ class UspDhcpActiveLeasesCard extends ConsumerWidget {
             ),
             AppGap.sm(),
             Expanded(
-              flex: 3,
+              flex: _ipColumnFlex,
               child: AppText.bodySmall(
                 client.ip,
                 color: colorScheme.onSurfaceVariant,
@@ -143,7 +154,7 @@ class UspDhcpActiveLeasesCard extends ConsumerWidget {
             ),
             AppGap.sm(),
             Expanded(
-              flex: 3,
+              flex: _leaseColumnFlex,
               child: Column(
                 crossAxisAlignment: CrossAxisAlignment.end,
                 children: [
--- a/lib/page/dhcp/views/components/usp_dhcp_active_leases_card.dart
+++ b/lib/page/dhcp/views/components/usp_dhcp_active_leases_card.dart
@@ -98,6 +98,10 @@ class UspDhcpActiveLeasesCard extends ConsumerWidget {
       padding: const EdgeInsets.only(bottom: AppSpacing.sm),
       child: LayoutBlock(
         padding: const EdgeInsets.all(AppSpacing.md),
+        // Columns are sized by flex, not `context.colWidth()`: colWidth is
+        // measured against the page grid, so on a 4-column mobile grid two
+        // colWidth(2) boxes exceed this row's own width, starving the name
+        // column to zero and overflowing the row (#1140).
         child: Row(
           children: [
             Icon(
@@ -108,27 +112,38 @@ class UspDhcpActiveLeasesCard extends ConsumerWidget {
             ),
             AppGap.sm(),
             Expanded(
+              flex: 4,
               child: Column(
                 crossAxisAlignment: CrossAxisAlignment.start,
                 children: [
-                  AppText.bodyMedium(client.displayName),
+                  AppText.bodyMedium(
+                    client.displayName,
+                    maxLines: 1,
+                    overflow: TextOverflow.ellipsis,
+                  ),
                   if (client.hostName.isNotEmpty)
                     AppText.bodySmall(
                       client.mac,
                       color: colorScheme.onSurfaceVariant,
+                      maxLines: 1,
+                      overflow: TextOverflow.ellipsis,
                     ),
                 ],
               ),
             ),
-            SizedBox(
-              width: context.colWidth(2),
+            AppGap.sm(),
+            Expanded(
+              flex: 3,
               child: AppText.bodySmall(
                 client.ip,
                 color: colorScheme.onSurfaceVariant,
+                maxLines: 1,
+                overflow: TextOverflow.ellipsis,
               ),
             ),
-            SizedBox(
-              width: context.colWidth(2),
+            AppGap.sm(),
+            Expanded(
+              flex: 3,
               child: Column(
                 crossAxisAlignment: CrossAxisAlignment.end,
                 children: [
@@ -136,11 +151,17 @@ class UspDhcpActiveLeasesCard extends ConsumerWidget {
                     AppText.bodySmall(
                       lease,
                       color: colorScheme.onSurfaceVariant,
+                      textAlign: TextAlign.end,
+                      maxLines: 1,
+                      overflow: TextOverflow.ellipsis,
                     ),
                   if (client.leaseExpiryFormatted.isNotEmpty)
                     AppText.bodySmall(
                       client.leaseExpiryFormatted,
                       color: colorScheme.onSurfaceVariant,
+                      textAlign: TextAlign.end,
+                      maxLines: 1,
+                      overflow: TextOverflow.ellipsis,
                     ),
                 ],
               ),
--- a/test/page/dhcp/views/components/usp_dhcp_active_leases_card_test.dart
+++ b/test/page/dhcp/views/components/usp_dhcp_active_leases_card_test.dart
@@ -1,110 +1,134 @@
-@Tags(['ui'])
-import 'package:flutter/material.dart';
-import 'package:flutter_riverpod/flutter_riverpod.dart';
+import 'package:flutter/rendering.dart';
 import 'package:flutter_test/flutter_test.dart';
-import 'package:privacy_gui/l10n/gen/app_localizations.dart';
 import 'package:privacy_gui/page/_shared/components/layout_blocks.dart';
+import 'package:privacy_gui/page/_shared/models/dhcp_client_ui_model.dart';
 import 'package:privacy_gui/page/dhcp/views/components/usp_dhcp_active_leases_card.dart';
-import 'package:ui_kit_library/ui_kit.dart';
 
 import '../../../../golden_test/page/dhcp/fixtures/dhcp_test_data.dart';
-
-final _testTheme = AppTheme.create(
-  brightness: Brightness.light,
-  seedColor: Colors.blue,
-  designThemeBuilder: (c) => CustomDesignTheme.fromJson({'style': 'flat'}),
-);
-
-/// Mobile viewport used by the golden suite (GoldenDevice.phone480).
-const _phoneSize = Size(480, 800);
-
-/// Content width the page grid hands the card at [_phoneSize]:
-/// screen 480 minus the page margin (16) on both sides.
-const _phoneContentWidth = 448.0;
+import 'dhcp_card_test_harness.dart';
 
 /// Widget tests for [UspDhcpActiveLeasesCard] mobile layout (#1140).
 ///
 /// Regression coverage: the lease row used two fixed `context.colWidth(2)`
 /// boxes, which are sized against the *page* grid (216dp each on a 4-column
-/// mobile grid) rather than the row's own 400dp of usable width. That starved
+/// mobile grid) rather than the row's own 398dp of usable width. That starved
 /// the name/MAC `Expanded` down to zero width — wrapping one character per
-/// line — and overflowed the Row by ~50dp.
+/// line — and overflowed the Row.
+///
+/// `flutter_test_config.dart` in this directory loads the shipped fonts, so the
+/// measurements below reflect what users see. Without it Flutter's built-in test
+/// font applies and is ~1.8x wider, which reports text as truncated when it is
+/// not.
 void main() {
-  Future<void> pumpCard(WidgetTester tester) async {
-    await tester.binding.setSurfaceSize(_phoneSize);
-    tester.view.physicalSize = _phoneSize;
-    tester.view.devicePixelRatio = 1.0;
-    addTearDown(() async {
-      tester.view.resetPhysicalSize();
-      tester.view.resetDevicePixelRatio();
-      await tester.binding.setSurfaceSize(null);
+  group('UspDhcpActiveLeasesCard - mobile row layout', () {
+    /// The row's last child, so it is what an overflow pushes past the edge.
+    Finder leaseTextOf(DhcpClientUIModel client) =>
+        find.text(client.leaseExpiryFormatted);
+
+    testWidgets('lease rows do not overflow at mobile width', (tester) async {
+      await pumpDhcpCard(
+        tester,
+        UspDhcpActiveLeasesCard(clients: testClients),
+      );
+
+      expect(
+        tester.takeException(),
+        isNull,
+        reason: 'lease row Row must fit within the mobile content width',
+      );
     });
 
-    await tester.pumpWidget(
-      ProviderScope(
-        child: MaterialApp(
-          theme: _testTheme,
-          localizationsDelegates: AppLocalizations.localizationsDelegates,
-          supportedLocales: AppLocalizations.supportedLocales,
-          home: Scaffold(
-            body: SingleChildScrollView(
-              child: SizedBox(
-                width: _phoneContentWidth,
-                child: UspDhcpActiveLeasesCard(clients: testClients),
-              ),
-            ),
-          ),
-        ),
-      ),
-    );
-    await tester.pumpAndSettle();
-  }
+    testWidgets('device name and MAC each render on a single line',
+        (tester) async {
+      await pumpDhcpCard(
+        tester,
+        UspDhcpActiveLeasesCard(clients: testClients),
+      );
+      expect(tester.takeException(), isNull);
 
-  testWidgets('lease rows do not overflow at mobile width', (tester) async {
-    await pumpCard(tester);
+      // The bug stacked every character vertically, making these hundreds of
+      // dp tall.
+      expect(
+        tester.getSize(find.text('iPhone-15-Pro')).height,
+        lessThan(singleLineMaxHeight),
+        reason: 'device name must stay on one line, not wrap per character',
+      );
+      expect(
+        tester.getSize(find.text('AA:BB:CC:DD:EE:01')).height,
+        lessThan(singleLineMaxHeight),
+        reason: 'MAC address must stay on one line, not wrap per character',
+      );
+    });
 
-    expect(
-      tester.takeException(),
-      isNull,
-      reason: 'lease row Row must fit within the mobile content width',
-    );
-  });
+    testWidgets('lease column stays within its own row bounds', (tester) async {
+      await pumpDhcpCard(
+        tester,
+        UspDhcpActiveLeasesCard(clients: testClients),
+      );
+      expect(tester.takeException(), isNull);
 
-  testWidgets('device name and MAC each render on a single line',
-      (tester) async {
-    await pumpCard(tester);
-    tester.takeException(); // overflow is asserted by the test above
+      // Bind the text to its own row rather than assuming display order: the
+      // card re-sorts rows (online first, then by display name), so indexing
+      // LayoutBlock and the fixture list independently only lines up by
+      // coincidence.
+      final leaseText = leaseTextOf(testClients.first);
+      final row = find.ancestor(
+        of: leaseText,
+        matching: find.byType(LayoutBlock),
+      );
+      expect(row, findsOneWidget);
 
-    // A single line of bodyMedium/bodySmall is well under 40dp tall; the bug
-    // stacked every character vertically, making these hundreds of dp tall.
-    final nameSize = tester.getSize(find.text('iPhone-15-Pro'));
-    expect(
-      nameSize.height,
-      lessThan(40),
-      reason: 'device name must stay on one line, not wrap per character',
-    );
+      expect(
+        tester.getRect(leaseText).right,
+        lessThanOrEqualTo(tester.getRect(row).right),
+        reason: 'lease column must not be clipped off the right edge',
+      );
+    });
 
-    final macSize = tester.getSize(find.text('AA:BB:CC:DD:EE:01'));
-    expect(
-      macSize.height,
-      lessThan(40),
-      reason: 'MAC address must stay on one line, not wrap per character',
-    );
-  });
+    testWidgets('a long device name ellipsises instead of wrapping',
+        (tester) async {
+      // #1140 accepts truncation with an ellipsis; what it rules out is
+      // wrapping (one character per line) and overflowing the row. A name
+      // wider than its column is the case most likely to regress.
+      final client = DhcpClientUIModel(
+        mac: 'AA:BB:CC:DD:EE:09',
+        ip: '192.168.1.109',
+        leaseActive: true,
+        isOnline: true,
+        hostName: 'Peters-MacBook-Pro-16-inch-2023',
+        leaseExpiry: DateTime.now().add(const Duration(hours: 12)),
+      );
 
-  testWidgets('lease column stays within the row bounds', (tester) async {
-    await pumpCard(tester);
-    tester.takeException(); // overflow is asserted by the first test
+      await pumpDhcpCard(
+        tester,
+        UspDhcpActiveLeasesCard(clients: [client]),
+      );
+      expect(
+        tester.takeException(),
+        isNull,
+        reason: 'an over-long name must not overflow the row',
+      );
 
-    // The lease column is the row's last child, so it is what the overflow
-    // pushed past the right edge and clipped.
-    final blockRight = tester.getRect(find.byType(LayoutBlock).first).right;
-    final expiryRight =
-        tester.getRect(find.text(testClients.first.leaseExpiryFormatted)).right;
-    expect(
-      expiryRight,
-      lessThanOrEqualTo(blockRight),
-      reason: 'lease column must not be clipped off the right edge',
-    );
+      final name = find.text(client.hostName);
+      final paragraph = tester.renderObject<RenderParagraph>(name);
+      expect(
+        paragraph.didExceedMaxLines,
+        isTrue,
+        reason: 'this name is wider than its column, so it should ellipsise — '
+            'if it now fits, the fixture no longer covers the truncation path',
+      );
+      expect(
+        tester.getSize(name).height,
+        lessThan(singleLineMaxHeight),
+        reason: 'ellipsised text must stay on one line',
+      );
+
+      final row = find.ancestor(of: name, matching: find.byType(LayoutBlock));
+      expect(
+        tester.getRect(name).right,
+        lessThanOrEqualTo(tester.getRect(row).right),
+        reason: 'ellipsised text must stay inside the row',
+      );
+    });
   });
 }
--- a/test/page/dhcp/views/components/usp_dhcp_active_leases_card_test.dart
+++ b/test/page/dhcp/views/components/usp_dhcp_active_leases_card_test.dart
@@ -1,0 +1,110 @@
+@Tags(['ui'])
+import 'package:flutter/material.dart';
+import 'package:flutter_riverpod/flutter_riverpod.dart';
+import 'package:flutter_test/flutter_test.dart';
+import 'package:privacy_gui/l10n/gen/app_localizations.dart';
+import 'package:privacy_gui/page/_shared/components/layout_blocks.dart';
+import 'package:privacy_gui/page/dhcp/views/components/usp_dhcp_active_leases_card.dart';
+import 'package:ui_kit_library/ui_kit.dart';
+
+import '../../../../golden_test/page/dhcp/fixtures/dhcp_test_data.dart';
+
+final _testTheme = AppTheme.create(
+  brightness: Brightness.light,
+  seedColor: Colors.blue,
+  designThemeBuilder: (c) => CustomDesignTheme.fromJson({'style': 'flat'}),
+);
+
+/// Mobile viewport used by the golden suite (GoldenDevice.phone480).
+const _phoneSize = Size(480, 800);
+
+/// Content width the page grid hands the card at [_phoneSize]:
+/// screen 480 minus the page margin (16) on both sides.
+const _phoneContentWidth = 448.0;
+
+/// Widget tests for [UspDhcpActiveLeasesCard] mobile layout (#1140).
+///
+/// Regression coverage: the lease row used two fixed `context.colWidth(2)`
+/// boxes, which are sized against the *page* grid (216dp each on a 4-column
+/// mobile grid) rather than the row's own 400dp of usable width. That starved
+/// the name/MAC `Expanded` down to zero width — wrapping one character per
+/// line — and overflowed the Row by ~50dp.
+void main() {
+  Future<void> pumpCard(WidgetTester tester) async {
+    await tester.binding.setSurfaceSize(_phoneSize);
+    tester.view.physicalSize = _phoneSize;
+    tester.view.devicePixelRatio = 1.0;
+    addTearDown(() async {
+      tester.view.resetPhysicalSize();
+      tester.view.resetDevicePixelRatio();
+      await tester.binding.setSurfaceSize(null);
+    });
+
+    await tester.pumpWidget(
+      ProviderScope(
+        child: MaterialApp(
+          theme: _testTheme,
+          localizationsDelegates: AppLocalizations.localizationsDelegates,
+          supportedLocales: AppLocalizations.supportedLocales,
+          home: Scaffold(
+            body: SingleChildScrollView(
+              child: SizedBox(
+                width: _phoneContentWidth,
+                child: UspDhcpActiveLeasesCard(clients: testClients),
+              ),
+            ),
+          ),
+        ),
+      ),
+    );
+    await tester.pumpAndSettle();
+  }
+
+  testWidgets('lease rows do not overflow at mobile width', (tester) async {
+    await pumpCard(tester);
+
+    expect(
+      tester.takeException(),
+      isNull,
+      reason: 'lease row Row must fit within the mobile content width',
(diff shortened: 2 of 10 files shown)


test/page/dhcp/views/components/usp_dhcp_active_leases_card_test.dart
```diff
@@ -0,0 +1,110 @@
+@Tags(['ui'])
```
PeterJhongLinksys: ⚠️ **Warning (🟢 High) — this regression guard never runs in CI.** `run_tests.sh:80,92,96` all use `--exclude-tags="golden||loc||ui"`, and `.github/workflows/ci.yml:77` only runs `sh run_tests.sh`. Since `dart_test.yaml` declares the `ui` tag, this file (and the reservations one) is excluded from every CI test run. These tests assert sizes/rects/`didExceedMaxLines` — not pixels — so they are not `ui`/golden in spirit. Consider dropping the tag, or adding a CI job that runs `--tags=ui`.

lib/page/dhcp/views/components/usp_dhcp_active_leases_card.dart
```diff
@@ -108,39 +112,56 @@ class UspDhcpActiveLeasesCard extends ConsumerWidget {
             ),
             AppGap.sm(),
             Expanded(
+              flex: 4,
```
PeterJhongLinksys: ⚠️ **Warning (🟢 High) — contradictory identifier standard vs the sibling card.** When `hostName` is empty, `displayName` *is* the MAC (`dhcp_client_ui_model.dart:70`) and the sub-line MAC at :124 is hidden, so the MAC lives in this `flex: 4` column. By this PR own arithmetic that is ~141dp (448 − 16×2 − 16×2 = 384, minus dot + 3 gaps ≈ 352 distributable) while a full MAC needs ~238dp ⇒ guaranteed ellipsis at mobile width. That is exactly the failure mode `usp_dhcp_reservations_detail_card_test.dart:323-328` declares unacceptable ("would make two reservations indistinguishable"). 💡 Also: `flex: 4/3/3` are bare magic numbers — `layout_blocks/block_constants.dart` already exists for named layout constants.

PeterJhongLinksys (commented):
🤖 **Automated Review — Round 1** · `24b17ad..967ba2b` (full)

**Verdict: 💬 Self-review (comment only)** — Layout fix is directionally right and violates no constitution article, but the regression guard it adds never runs in CI, and the two cards now disagree on whether a device identifier may be ellipsised.

|---|---|---|---|
| ⚠️ | 🟢 | `test/page/dhcp/views/components/usp_dhcp_active_leases_card_test.dart:1`, `..._reservations_detail_card_test.dart:1` | `@Tags(['ui'])` + `run_tests.sh:80,92,96` `--exclude-tags="golden\|\|loc\|\|ui"` ⇒ both new regression tests are never executed by CI (`ci.yml:77` runs only `run_tests.sh`). Regression protection is effectively zero. |
| ⚠️ | 🟢 | `usp_dhcp_active_leases_card.dart:114-130` vs `..._reservations_detail_card_test.dart:323-328` | Contradictory acceptance standard: the reservations test asserts a MAC must never be ellipsised (it is the row's only identifier), yet a lease row whose `hostName` is empty shows the MAC in the flex:4 column (~141dp) while a full MAC needs ~238dp ⇒ guaranteed ellipsis at mobile width. |
| ⚠️ | 🟢 | `usp_dhcp_active_leases_card.dart:136-165`, `usp_dhcp_detail_view.dart:178-181` | On the desktop two-column split (`colWidth(6)`) the IP/lease columns lose their former guaranteed 216dp `SizedBox` and get 3/10 of the remaining width; the new `overflow: ellipsis` turns `yyyy-MM-dd HH:mm` into a silently truncated, unreadable timestamp. No tooltip/`Semantics` fallback, no desktop-width test. |
| ⚠️ | 🟢 | `..._active_leases_card_test.dart:188,209`, `..._reservations_detail_card_test.dart:307,321` | `tester.takeException();` with the return value discarded swallows *any* pending exception (broken override, missing l10n key, framework assert), so tests 2/3 can pass while something unrelated is actually failing. |
| ⚠️ | 🟢 | `lib/page/_shared/components/usp_info_row.dart:6-8,29` | The `colWidth()`-inside-a-constrained-Row anti-pattern is only fixed in place; the shared cross-page component still uses it *and its DartDoc still recommends it*. Same pattern also at `usp_wifi_status_card.dart:86,127,138,175,182`. [Shotgun Surgery] Knowledge lives only in two inline comments. |
| ⚠️ | 🟢 | `usp_dhcp_active_leases_card.dart:114-146`, `..._reservations_detail_card.dart:87-104`, `layout_blocks/list_blocks.dart:46-52` | [Inconsistent Solution] Three row-width conventions now coexist on the same DHCP detail page (Expanded flex 4/3/3 · single Expanded + vertical stack · hardcoded `SizedBox(width: 120)`); the same bug got two different fixes in two sibling cards. |
| ⚠️ | 🟡 | `..._active_leases_card_test.dart:35-49`, `..._reservations_detail_card_test.dart:35-49` | The harness pumps the card inside a hand-rolled `SizedBox(width: 448)` rather than the real page grid. Since the original bug came from `colWidth()` reading the *page* grid, the tests' ability to reproduce the pre-fix failure is unproven (`ui_kit_library` fallback behaviour without a grid ancestor is unknown). Please paste a "revert-the-fix ⇒ tests fail" run in the PR description. |
| ⚠️ | 🟡 | `..._active_leases_card_test.dart:213-215` | [Fragile Test] `find.byType(LayoutBlock).first` is matched against `testClients.first.leaseExpiryFormatted`, but the card re-sorts rows (`usp_dhcp_active_leases_card.dart:25-34`). The pairing holds only by fixture coincidence; use `find.ancestor(...)` to bind row-to-text explicitly. |
| 💡 | 🟢 | `..._active_leases_card_test.dart:192-219` | After the fix most assertions are near-tautological (`maxLines: 1` makes "height < 40" structurally true; all-`Expanded` children cannot exceed the Row). Test 1 is the only real guard — and an `Expanded` starved to 0 width would make text vanish while still passing the height check. |
| 💡 | 🟢 | `test/.../usp_dhcp_active_leases_card_test.dart:12-53` ↔ `..._reservations_detail_card_test.dart:13-65` | [Duplicated Code] `_testTheme`, `_phoneSize`, `_phoneContentWidth` and the viewport-setup/teardown block are line-for-line identical. `_testTheme` is already copy-pasted across 13 test files and there is no `test/helpers/`; good moment to extract a `pumpAtPhoneWidth` harness. |
| 💡 | 🟢 | `usp_dhcp_active_leases_card.dart:101-104`, `..._reservations_detail_card.dart:81-86`, both test file headers | [Comments] The same "why colWidth is wrong / 216dp / 4-column grid" explanation is repeated 3–4 times and encodes external `ui_kit` grid internals that will rot. It also describes the historical bug rather than current intent — that belongs in the commit message. Note the leases header says "400dp of usable width" while 448 − 16×2 − 16×2 = 384dp. |
| 💡 | 🟢 | `usp_dhcp_active_leases_card.dart:115,136,146` | [Magic Number] Bare `flex: 4/3/3` with no name and no rationale, although column ratios are what keeps rows aligned. `layout_blocks/block_constants.dart` already exists as the home for such constants. |
| 💡 | 🟢 | `..._reservations_detail_card.dart:104-105` | The `Expanded` is followed immediately by two `AppIconButton`s with no gap (and none between them), while this same PR deliberately added three `AppGap.sm()` to the leases row. Ellipsised MAC/IP will sit flush against the edit icon. |
| 💡 | 🟢 | `..._reservations_detail_card.dart:98` | `Theme.of(context)` is re-resolved per reservation inside the row builder, whereas the sibling card hoists `final colorScheme = Theme.of(context).colorScheme;` (`usp_dhcp_active_leases_card.dart:95`). |
| 💡 | 🟢 | `..._active_leases_card_test.dart:195,202`, `..._reservations_detail_card_test.dart:314` | [Magic Number] `lessThan(40)` repeated three times, coupled to the current typography scale. `didExceedMaxLines` (already used in the reservations test) is a more direct assertion than height. |
| 💡 | 🟢 | `usp_dhcp_active_leases_card.dart:124-130` | `if (client.hostName.isNotEmpty)` gating the display of `client.mac` is **correct** (`dhcp_client_ui_model.dart:70`: `displayName => hostName.isNotEmpty ? hostName : mac`, so the MAC is already the title) but reads like a bug. One line of comment would save the next reader. |
| 💡 | 🟡 | `..._reservations_detail_card_test.dart:269-284` | Redundant/over-broad overrides: `dhcpDetailOverrides(reservationState: dataState(), lanInfo:, clients:)` also stubs `lanDataProvider`/`dhcpDataProvider`/`devicesDataProvider`, but the widget only reads `uspDhcpReservationsProvider` — and `testReservations` is *also* passed directly as a parameter, giving two sources of truth. |
| 💡 | 🟡 | `..._reservations_detail_card.dart:87-104`, `usp_dhcp_active_leases_card.dart:105-169` | The fix trades a loud bug for a quiet one: once the container is narrower than the incompressible children (switch + two icon buttons / dot + gaps), `Expanded` collapses to 0 width and the MAC/IP/name disappear entirely with no error. Consider `ConstrainedBox(minWidth:)`, and render `reservation.ip` only when non-empty (matching the `isNotEmpty` guards already used at `:150,:158`). |
| 💡 | 🟡 | both test files, `void main()` | No `group()` wrapper; Art. VIII asks for `[ClassName] - [Feature]` group naming. Test names themselves are descriptive — that part is good. |
| 💡 | 🟡 | `..._active_leases_card_test.dart:131-135`, `..._reservations_detail_card_test.dart:244-248` | `_phoneSize = Size(480,800)` duplicates `GoldenDevice.phone480` (`golden_test_config.dart:24`) instead of referencing it, and `_phoneContentWidth = 448.0` hardcodes the page-margin assumption — if the page margin changes, the test still passes but stops being representative. |
| 💡 | 🟡 | `test/.../usp_dhcp_*_test.dart:10-11` | Fixtures are imported from `test/golden_test/page/dhcp/fixtures/`, while Art. VIII designates `test/mocks/test_data/[feature]_test_data.dart`. There is precedent (`usp_network_health_card_legend_test.dart:10-11`) so this is not a violation, but the golden infrastructure is drifting into being the de-facto shared test-data layer. |
| 💡 | 🟡 | `..._active_leases_card_test.dart:165`, `..._reservations_detail_card_test.dart:280-284` | Coverage is narrow on exactly the paths most likely to regress: empty `hostName` (widest text case), desktop half-width (`colWidth(6)`, narrower than 448), long hostName, empty `leaseTimeFormatted` (`dhcp_client_ui_model.dart:42-54`). Parameterising `pumpCard` by width + data would cover these cheaply. |
| 💡 | ⚪ | `..._reservations_detail_card_test.dart:330-332` | `tester.renderObject<RenderParagraph>(find.text(...))` assumes `AppText` renders a plain `Text`/`RichText`. `ui_kit_library` is an external git dep (v2.28.1) with no source in the worktree, so this cast is unverified — CI will confirm. |
| 💡 | 🟡 | 6 new `maxLines: 1` sites (`usp_dhcp_active_leases_card.dart:121,128,140,155,163`; `..._reservations_detail_card.dart:93,99`) | Before this change text wrapped (information preserved, row just got taller inside a scroll view); now MAC/IP/timestamps truncate unconditionally at large system text scale. Tests run at default `textScaler` only. |

<sub>Confidence: 🟢 High = head code read, evidence quoted · 🟡 Med = file:line + reasoning, not fully proven · ⚪ Low = speculative.</sub>
<sub>`[X]` = raised by 1 agent, `[XX]` = raised independently by both agents. Rows 1, 4 above were `[XX]`; all others `[X]`.</sub>

<details>
<summary>⚠️ Warning evidence details</summary>

**1. New regression tests are excluded from CI** (`[XX]`, 🟢 High)
```dart
// test/page/dhcp/views/components/usp_dhcp_active_leases_card_test.dart:1
@Tags(['ui'])
```
```sh
# run_tests.sh:80 / :92 / :96 — all three branches
```
```yaml
# .github/workflows/ci.yml:77
      - name: 🧪 Unit Tests
        run: sh run_tests.sh
```
`dart_test.yaml` does declare the `ui` tag, so the exclusion definitely matches. `ci.yml` contains no other `flutter test` invocation and no `--tags=ui` job. Net effect: 226 lines / 6 `testWidgets` never run in CI. If someone reverts `Expanded` back to `SizedBox(width: context.colWidth(2))`, CI stays green and #1140 returns.
*Fix:* these tests assert sizes/rects/`didExceedMaxLines`, not pixels — they are not `ui`/golden in spirit. Drop `@Tags(['ui'])`, or add a CI job that runs `--tags=ui`.

**2. Contradictory identifier-truncation standard between the two cards** (🟢 High)
```dart
// test/.../usp_dhcp_reservations_detail_card_test.dart:323-328 — the stated standard
// The MAC is the row's only device identifier, so ellipsising it (e.g.
// "AA:BB:CC:DD:EE:...") would make two reservations indistinguishable.
expect(paragraph.didExceedMaxLines, isFalse, ...);
```
```dart
// lib/page/_shared/models/dhcp_client_ui_model.dart:70
String get displayName => hostName.isNotEmpty ? hostName : mac;
```
```dart
// lib/page/dhcp/views/components/usp_dhcp_active_leases_card.dart:114-130
Expanded(
  flex: 4,
  child: Column(children: [
    AppText.bodyMedium(client.displayName, maxLines: 1, overflow: TextOverflow.ellipsis),
    if (client.hostName.isNotEmpty) AppText.bodySmall(client.mac, ...),
  ]),
),
```
Using this PR's own arithmetic: 448 − AppCard 16×2 − LayoutBlock 16×2 = 384dp, minus the 8dp dot and three `AppGap.sm()` ≈ 352dp distributable ⇒ flex:4 ≈ 141dp, versus the ~238dp the PR says a full MAC needs. So for any lease whose `hostName` is empty, the MAC — the row's only identifier — is guaranteed to ellipsise at mobile width, which is exactly the failure mode the sibling card's test declares unacceptable.
*Fix:* pick one standard. Either give the leases card the same guarantee (stacked layout / different flex / allow 2 lines), or downgrade the reservations assertion to a documented note. At minimum add an empty-`hostName` test case making the accepted behaviour explicit.

**3. Desktop split silently truncates the timestamp** (🟢 High)
```dart
// lib/page/dhcp/views/usp_dhcp_detail_view.dart:178-181
            SizedBox(
              width: context.colWidth(6),
              child: UspDhcpActiveLeasesCard(clients: clients),
```
Previously IP and lease each had a guaranteed 216dp `SizedBox`; now they receive 3/10 of the residual width of roughly half the viewport. `leaseExpiryFormatted` is `yyyy-MM-dd HH:mm` (`dhcp_client_ui_model.dart:58-67`), so it degrades to `2026-03-0…` — and the newly added `overflow: ellipsis` makes it silent, whereas before it wrapped and stayed complete. Note the MAC is also what the delete confirmation text uses (`usp_dhcp_reservations_detail_card.dart:205`).
*Fix:* add `Tooltip`/`Semantics(label:)` preserving the full value for ellipsised fields, and add a desktop-width (`colWidth(6)`) test pinning the accepted behaviour.

**4. `takeException()` return value discarded** (`[XX]`, 🟢 High)
```dart
// ..._active_leases_card_test.dart:188 (also :209, and reservations :307, :321)
    await pumpCard(tester);
    tester.takeException(); // overflow is asserted by the test above
```
`takeException()` clears the pending exception regardless of type, and each `testWidgets` is an independent pump — test 1's assertion does not cover tests 2/3. After a correct fix there should be no pending exception at all.
*Fix:* delete those two calls, or make them `expect(tester.takeException(), isNull)`. If that fails, the fix is incomplete — which is more valuable to know.

</details>

<details>

- **Root cause correctly identified.** Replacing `context.colWidth(2)` (page-grid measure) with flex inside a constrained `Row` is the right fix, and an all-`Expanded` Row cannot structurally overflow.
- **No security surface touched.** No hardcoded secrets/tokens, no `flutter_secure_storage` change, no string concatenation into a request, no permission logic.
- **USP/JNAP error handling intact and already correct.** `usp_dhcp_service.dart:36-37,72-74,108-110,138-140,255-257` and `usp_dhcp_data_service.dart:93-94` all use `catch (e)` + `throw mapUspErrorToServiceError(e)` — no `on Exception catch`, no manual USP error-string parsing. This PR does not touch those paths.
- **Constitution clean.** No `generated/` import in `views/` (Art. V §5.4); no Provider/Service change, so the four USP Provider hard rules (Art. IV) and ServiceError isolation (Art. XIII) are unaffected; only `AppText`/`AppGap`/`AppCard`/`LayoutBlock`/`AppIconButton` used, no hand-rolled widgets (Art. XV); naming conforms (Art. III).
- **`maxLines`/`overflow`/`textAlign` are existing `AppText` API** — precedent at `usp_ethernet_ports_card.dart:161-167` and `row_blocks.dart:46-57`.
- **Consistent per-row flex structure** keeps columns aligned across rows within the leases card.
- **A regression test was actually written** rather than just changing layout, and `setSurfaceSize` + `tester.view.physicalSize` are correctly combined (same form as `golden_runner.dart:101-102`).
- Fixtures genuinely contain the asserted strings (`'iPhone-15-Pro'`, `'AA:BB:CC:DD:EE:01'`, non-empty `leaseExpiryFormatted`), and `find.text` is unambiguous within each pump.
- **Verified non-issue:** `if (client.hostName.isNotEmpty)` gating `client.mac` is correct, not a field mix-up (see Suggestion above).

</details>

> ⚠️ **Not verified by execution:** no Flutter/Dart SDK was available in the review environment (`which flutter` → not found; no `~/.pub-cache`, `dart-sdk`, `fvm`), so the 6 new tests were **not actually run** and `ui_kit_library` sources could not be read. "Tests pass" and the `RenderParagraph` cast remain unconfirmed — please rely on CI or a local run.

PeterJhongLinksys (commented):
🤖 **Automated Review — Round 2** · `967ba2b..62c4c1a` (incremental)

**Verdict: 💬 Self-review (comment only)** — No Critical found. Most Round 1 findings were genuinely fixed, and the author's own correction of the dp measurements is exactly the right instinct; the remaining concerns are all in the *new test infrastructure*, which can silently degrade back into the very measurement hole this push set out to close.

|---|---|---|---|
| ⚠️ | 🟢 | `test/helpers/load_app_test_fonts.dart:24,29,54` | Missing font file is swallowed by `existsSync()` with no else-branch; tests silently fall back to the 1.8–2.7x wider built-in test font — the exact class of error this push claims to eliminate. |
| ⚠️ | 🟢 | `run_generate_loc_snapshots.sh:41` vs `run_golden_verify.sh` | `rm -f goldens/overflow_warnings.json` was added to the *generator* entry points only; the *verification* script never clears it, so `combine_results.dart:19` can still report already-fixed overflows. |
| ⚠️ | 🟢 | `dhcp_card_test_harness.dart:35-41` | `width` / `surfaceSize` have zero call sites; the desktop-`colWidth(6)` test requested in Round 1 was not added, so two parameter paths ship untested. |
| ⚠️ | 🟡 | `test/page/dhcp/views/components/flutter_test_config.dart:1-20` | Directory-scoped config means every future measuring directory must copy these 20 lines; `views/dialogs/` in the same feature still measures under the test font. |
| ⚠️ | 🟡 | `golden_runner.dart:371-383` | `_writeOverflowReport` is an unlocked read-modify-write on a shared file from parallel test suites; `jsonDecode` is not guarded, so a partial read throws in `tearDownAll`. |
| ⚠️ | 🟡 | `dhcp_card_test_harness.dart:7` | Unit-test harness now imports `test/golden_test/golden_framework/golden_test_config.dart`; the golden tree is drifting into a de-facto shared test layer. |
| ⚠️ | 🟡 | `usp_dhcp_active_leases_card_test.dart:14` | The corrected `398dp` still cannot be derived from any `AppSpacing` combination (16 ⇒ 384, 12 ⇒ 400); the same grid/font knowledge is now restated in ~6 places. |
| ⚠️ | 🟡 | `usp_dhcp_active_leases_card.dart:123-141` | Round 1's "MAC must truncate" claim is largely **withdrawn**, but with the author's own 129.7dp figure the empty-`hostName` column has only single-digit-% headroom, and no `hostName: ''` test pins the accepted behaviour. |
| 💡 | 🟢 | `load_app_test_fonts.dart:37-50` | [Duplicated Code] Parallel `fontNames` / `fontFiles` arrays aligned by index — one insertion out of sync maps a family to the wrong file, silently. |
| 💡 | 🟢 | `dhcp_card_test_harness.dart:14,21,25,35` | [Speculative Generality / naming] `dhcpCardTestTheme` / `pumpDhcpCard` lock generic capability into DHCP; `AppTheme.create` is still duplicated across ~16 test files despite `test/helpers/` now existing. |
| 💡 | 🟢 | leases test `:28,43,64,102` vs reservations test `:23-31` | [Inconsistent Solution] One sister test extracts a local `pumpCard`, the other repeats the pump call in every case. |
| 💡 | 🟡 | `usp_dhcp_active_leases_card_test.dart:93-100` | Inline `DhcpClientUIModel(...)` construction; Art. VIII asks for a Test Data Builder / fixture (`longNameClient`). |
| 💡 | 🟡 | `dhcp_card_test_harness.dart:25,33-34` | `16 * 2` hardcoded while the doc names `AppLayoutConfig.marginMobile`; and at 1280 width `colWidth(6)` is *wider* than 448, so "pass a narrower value" is misleading. |
| 💡 | 🟡 | `.gitignore:96-99` | `/goldens/` is broader than the single artifact it protects, and the comment references "line 92", which will rot. |
| 💡 | 🟡 | reservations test `:23-31` | Over-broad overrides (`lanDataProvider`/`dhcpDataProvider`/`devicesDataProvider` stubbed) while `testReservations` is *also* passed as a widget argument — two sources of truth. Carried over from Round 1. |
| 💡 | ⚪ | `usp_dhcp_active_leases_card_test.dart:98` | `DateTime.now().add(...)` fixture makes `leaseTimeFormatted` clock-dependent (`11h 59m` vs `12h 0m`); harmless today, a trap next to fixed-string assertions. |
| 💡 | ⚪ | `load_app_test_fonts.dart:16,77-78` | "Safe to call more than once" is unverified; the relative-`rootUri` branch string-concatenates without normalisation. Pure move-over, not newly introduced. |

<sub>Confidence: 🟢 High = head code read, evidence quoted · 🟡 Med = file:line + reasoning, not fully proven · ⚪ Low = speculative.</sub>
<sub>No Critical was raised by either agent this round, so no `[X]`/`[XX]` agreement markers apply. The top 8 Warnings were reached independently by both agents.</sub>

<details>
<summary>⚠️ Warning evidence</summary>

**1. Font loading fails silently** — `test/helpers/load_app_test_fonts.dart:24,29,54`
```dart
// :24
if (mainFontFile.existsSync()) { ... }   // absent -> nothing happens
// :29
await mainFont.load();                   // an empty FontLoader also resolves fine
// :54
if (file.existsSync()) { loader.addFont(...); }
```
`_resolveUiKitPath()` only throws when `package_config.json` is missing. Once the path resolves but the `.otf` is not there (ui_kit is a git dep at `v2.28.1`; asset layout is outside this repo and the pub cache path differs on CI), `FontLoader.load()` succeeds with zero fonts and the suite silently reverts to `FlutterTest` — the font whose 1.8–2.7x width produced the wrong figures in the first place. This round also removed `@Tags(['ui'])`, so `run_tests.sh:92` (`--exclude-tags="golden||loc||ui"`) no longer excludes these font-metric assertions: they are now a blocking CI gate. Failure mode: `reservations:71 didExceedMaxLines isFalse` fails, or `leases:116 isTrue` passes for the wrong reason, and the message points nowhere near "font not loaded".
*Fix:* `throw StateError('font not found at $path')` when the main font is missing (matching the good handling already at `_resolveUiKitPath`), or assert the family is registered / width-sanity-check after load.

**2. Stale overflow report only cleared on the generate path** — `run_generate_loc_snapshots.sh:41`
```sh
rm -f goldens/overflow_warnings.json          # added this round
```
```dart
// test_scripts/combine_results.dart:18-20  (verification-path report builder)
final file = File('goldens/overflow_warnings.json');
if (!file.existsSync()) return [];
```
`run_golden_verify.sh` runs `flutter test test/golden_test/` and never removes the file. Run a failing `run_generate_loc_snapshots.sh`, then `run_golden_verify.sh`: the verification report still lists overflows that are already fixed. Following `SCREENSHOT_TESTING.md` (`flutter test test/golden_test/ --update-goldens`) bypasses the cleanup as well. The commit message says "Start each golden run with an empty overflow report", but only 2 of 4 entry points hold. The knowledge of this path is now spread across `golden_runner.dart:375`, `combine_results.dart:19`, `clear_goldens.sh:18` and `run_generate_loc_snapshots.sh:41`.
*Fix:* truncate at the owning layer (`golden_runner.dart`, once per process before the first write) and extract the path into one shared constant; both shell lines then become unnecessary.

**3. Unlocked append from parallel suites** — `test/golden_test/golden_framework/golden_runner.dart:371-383`
```dart
final existing = file.existsSync()
    ? List<Map<String, dynamic>>.from(
        jsonDecode(file.readAsStringSync()) as List)   // no try/catch
    : <Map<String, dynamic>>[];
existing.addAll(_overflowWarnings);
file.writeAsStringSync(...);
```
Called from `tearDownAll` (:71). `flutter test test/golden_test/` runs suites in separate processes with concurrency > 1, so this read-modify-write can lose whole batches or `jsonDecode` a half-written file and fail the tear-down. This round's `rm -f` addresses accumulation *across runs*, not the race *within* a run.
*Fix:* one file per suite under `goldens/overflow/<suiteId>.json`, aggregated by the report builder; at minimum wrap `jsonDecode` in try/catch.

**4. Dead harness parameters, requested desktop test still missing** — `dhcp_card_test_harness.dart:35-41`
```dart
Future<void> pumpDhcpCard(WidgetTester tester, Widget card, {
  List<Override> overrides = const [], double? width, Size? surfaceSize,
}) async {
```
All five call sites (leases `:29,43,64,102`; reservations `:23`) pass only `card` + `overrides`. The DartDoc explicitly describes modelling the desktop split where the card gets `context.colWidth(6)`, but `usp_dhcp_detail_view.dart:179,184` remains at zero coverage, and `grep Tooltip|Semantics` still returns 0 hits in both cards — so the silently-ellipsised `yyyy-MM-dd HH:mm` timestamp from Round 1 is unaddressed and untested.
*Fix:* add one desktop-width `testWidgets`, or drop the two parameters until they are used.

**5. Two font baselines inside one test tree** — `test/page/dhcp/views/components/flutter_test_config.dart:16-20`
The scoping itself is correct (nearest ancestor wins; it does not collide with `test/golden_test/flutter_test_config.dart`). The consequence is that `test/page/dhcp/views/components/**` measures with the shipped font while the neighbouring `views/dialogs/` tests still use the built-in one.
*Fix:* hoist a `test/flutter_test_config.dart` calling `loadAppTestFonts()` (the golden-tree config naturally overrides it), or document which directories must host measurement tests.

**6. `398dp` still not derivable** — `usp_dhcp_active_leases_card_test.dart:14`, `dhcp_card_test_harness.dart:23-25`
Usable width = 448 − AppCard padding×2 − LayoutBlock padding×2. With `AppSpacing.md == 16` that is 384; with 12 it is 400. No combination yields 398, and `AppSpacing` / `AppLayoutConfig.marginMobile` live in `ui_kit_library`, so a reader cannot check the arithmetic. Round 1's `400dp` rotted the same way.
*Fix:* keep numbers only inside assertions that can fail (e.g. `tester.getSize(row).width`); let comments carry intent, not derived dp.
</details>

<details>

- **The author's own correction of the dp figures is the highlight of this round.** Identifying that `flutter_test_config.dart` is directory-scoped, that only `test/golden_test/**` loaded real fonts, and re-measuring rather than defending the original numbers is exactly right — and the fix direction (stack MAC over IP; `colWidth` inside a `Row` is the anti-pattern) is unchanged and correct.
- **Round 1 items genuinely fixed:** `@Tags(['ui'])` removed so the tests actually run in CI; all 8 `takeException()` calls now `expect(..., isNull, reason: ...)`; `find.byType(LayoutBlock).first` replaced by `find.ancestor(...)` + `findsOneWidget`; `flex: 4/3/3` extracted to `_nameColumnFlex/_ipColumnFlex/_leaseColumnFlex` with rationale; `lessThan(40)` named `singleLineMaxHeight`; `group()` added in both files; `colorScheme` hoisted out of the row builder; two `AppGap.sm()` added before the icon buttons; `if (reservation.ip.isNotEmpty)` guard added; the `hostName.isNotEmpty` gate documented; `_testTheme/_phoneSize/_phoneContentWidth` extracted to a harness using `GoldenDevice.phone480.size`; a long-`hostName` ellipsis test added.
- **Security surface clean.** No hits for secret/token/api_key/password/private_key across the 10 changed files; `flutter_secure_storage` untouched; the new `rm -f` uses a fixed literal path (no interpolation, no glob); existing shell variables are quoted.
- **USP/JNAP error handling intact and correct — untouched by this PR.** `usp_dhcp_service.dart:36-37,72-74,108-110,138-140,255-257` all use `catch (e)` + `throw mapUspErrorToServiceError(e)`; no `on Exception catch`, no manual USP error-string parsing. `usp_dhcp_reservations_notifier.dart:84,120,140,157,170` use `on ServiceError catch`, i.e. after correct translation (Art. XIII satisfied).
- **Constitution clean.** No `generated/` import outside `services/` (Art. V §5.4); no Provider/Service change, so the four USP Provider hard rules (Art. IV) are unaffected; only `AppText`/`AppGap`/`AppIconButton`/`AppSwitch`/`LayoutBlock` used (Art. XV); naming conforms (Art. III); `Equatable` unaffected (Art. XI).
- **The font-loading extraction is a faithful move** — line-by-line equivalent to the deleted `_loadFonts`, only `final`→`const` and trailing commas; all three relative import paths are correct; the helper does not end in `_test.dart` so it will not be collected as a suite.
- **`_resolveUiKitPath()` is the best error handling in this push** — structured `package_config.json` parsing (no regex, no hardcoded cache path) and a `StateError` with actionable instructions. That is precisely the pattern Warning 1 asks for elsewhere.
- **`/goldens/` in `.gitignore` is legitimately needed**: `golden_runner.dart:379-381` writes at repo root, which the pre-existing `test/**/goldens/*` rule does not cover.
- **`run_generate_loc_snapshots.sh`** documents *why* only that one file is removed and why the PNGs must survive, pointing at `clear_goldens.sh` as the opt-in full wipe — good decision-record hygiene.
</details>

> ⚠️ **Not verified by execution:** no Flutter/Dart SDK in the review environment (no `.dart_tool/`, no pub cache), and `ui_kit_library` sources are not readable. The 8 new tests were **not run**; all dp arithmetic (384/398/400/141/129.7), `AppSpacing`/`AppLayoutConfig` values, whether ui_kit actually ships `assets/fonts/NeueHaasGrotTextRound-*.otf`, the `RenderParagraph` cast, `FontLoader.load()` re-entrancy, `flutter test` default concurrency, and `colWidth(6)` at 1280 width are all code-level inference. Please rely on CI or a local run.

AustinChangLinksys (approved):
🤖 **Automated Review — Round 1** · `24b17ad..62c4c1a` (full)

**Verdict: ✅ APPROVE** — Layout root cause correctly identified and fixed; flex-based columns resolve the mobile overflow. Two data-completeness warnings on empty-string IP display.

#### Standards

|---|---|---|---|
| ⚠️ | High | `usp_dhcp_reservations_detail_card.dart:97` | `reservation.ip` empty string renders a blank AppText.bodySmall row (USP Yiaddr absent → coerced to `''` by codegen) |
| ⚠️ | High | `usp_dhcp_active_leases_card.dart:~130` | `client.ip` empty string similarly renders invisible IP column text (IPv4Address.1.IPAddress absent → `''`) |
| 💡 | Med | `usp_dhcp_active_leases_card.dart` | Test fixture uses `DateTime.now()` for leaseExpiry — non-deterministic; if future test asserts exact leaseTimeFormatted string it will be flaky |
| 💡 | High | `lib/generated/dhcp_clients.g.dart:~95` | Pre-existing: `DateTime.parse(response[...] ?? '')` throws `FormatException` on empty LeaseTimeRemaining, bypassing `mapUspErrorToServiceError` — not introduced here, record for follow-up |

#### Spec

|---|---|---|---|
| ⚠️ | High | `usp_dhcp_reservations_detail_card.dart` | Scope creep: spec only mentions Active Leases card; Reservations card fix is undisclosed in issue — acknowledged in PR description as same root cause |
| 💡 | High | `usp_dhcp_active_leases_card.dart` | Lease column text in flex layout directly fixes the overflow described in spec (confirmed: "lease column" = right-side text, not a separate bar widget) |

*Confidence: High = code-verified · Med = located + reasoned, not fully confirmed · Low = speculative.*
*[X] = 1 agent · [XX] = 2 agents (Critical only).*

<details>

**W-1 — `reservation.ip` renders empty row when absent (High)**
`usp_dhcp_reservations_detail_card.dart:97`. USP codegen: `yiaddr: (response['${p}Yiaddr'] ?? '') as String` — absent key → empty string. The `if (reservation.ip.isNotEmpty)` guard IS present on the head commit (line 97 of pr head), so this is already fixed. Confirming.

Actual check: `git show 62c4c1a9:lib/page/dhcp/views/components/usp_dhcp_reservations_detail_card.dart | grep -n isNotEmpty` shows line 97 `if (reservation.ip.isNotEmpty)` — guard is present. This is resolved.

**W-2 — `client.ip` in active leases has no guard (High)**
`usp_dhcp_active_leases_card.dart:~130`. Flex IP column (`Expanded(flex:3)`) renders `AppText.bodySmall(client.ip, ...)` unconditionally. When `IPv4Address.1.IPAddress` is absent from USP response, ip = `''`. No guard present. The Expanded still occupies flex space and renders an empty string — confusing on mobile.
**Fix:** `if (client.ip.isNotEmpty) ... else const Spacer(flex: _ipColumnFlex)` — or simply guard the AppText.

</details>

<details>

- **Root cause fix:** replacing `context.colWidth(2)` with `Expanded(flex:N)` is the correct approach — colWidth measures the page grid, not the row's own constraints.
- **Flex weight constants** `_nameColumnFlex=4`, `_ipColumnFlex=3`, `_leaseColumnFlex=3` present and named in head — intention readable.
- **`if (reservation.ip.isNotEmpty)` guard** present in head (line 97 of reservations detail card) — defensive.
- **Real-font test infrastructure:** `load_app_test_fonts.dart` + directory-scoped `flutter_test_config.dart` ensures layout tests measure the font users actually see.
- **8 widget tests** assert single-line heights, no-overflow, and within-bounds constraints — correct regression net.
- **gitignore + script fix:** `/goldens/` and `rm -f goldens/overflow_warnings.json` address a legitimate CI artifact pollution.

</details>

PR #1198, opened 2026-08-04
fix(usp): stop back-filling absent concrete GET paths in _rawGet (#1184)

### LINKED ISSUE #1184: USP codegen: required-leaf-missing (9998) → ServiceErrorView is unreachable for concrete-path models (silenced by _rawGet null-backfill)
## Summary

The generated codegen models declare certain leaves as **required** (they build a `missing[]` list and throw `(code: 9998)` → `ServiceErrorView` when a required leaf is absent from the GET response). However, for models that GET with **concrete paths** (e.g. `Device.DHCPv4.Server.Pool.1.MinAddress`), this required-field check can never fire, because `UspClient._rawGet` back-fills every absent non-wildcard path with `null` **before** the codegen check runs. The `containsKey` test then always returns true, `missing` stays empty, and 9998 never throws.

This surfaced while adding E2E error-state coverage for the LAN / Local Network page (`lan_network_info`): the `ServiceErrorView` (required-leaf-missing) branch turned out to be unreachable under the current client behaviour, so the E2E test for it had to be skipped.

## Evidence (paths/lines on branch under review)

**1. Codegen declares the leaves required and back-stops them at the same time**
`lib/generated/lan_network_info.g.dart` — 9 GET leaves, all 9 have a `missing.add(...)` required-check, and 8 also have a `?? ''` / `?? 0` fallback on assignment:

```dart
// _buildPaths — concrete paths, not wildcards:
'Device.DHCPv4.Server.Pool.1.Enable',
'Device.DHCPv4.Server.Pool.1.MinAddress',
'Device.DHCPv4.Server.Pool.1.MaxAddress',
'Device.DHCPv4.Server.Pool.1.LeaseTime',
'Device.DHCPv4.Server.Pool.1.DNSServers',
...
// required-check:
if (!response.containsKey('Device.DHCPv4.Server.Pool.1.MinAddress')) {
  missing.add('Device.DHCPv4.Server.Pool.1.MinAddress');
}
...
if (missing.isNotEmpty) {
  throw 'Get failed: Validation error: Required fields missing from response: ... (code: 9998)';
}
// assignment (present-but-null fallback):
minAddress: (response['Device.DHCPv4.Server.Pool.1.MinAddress'] ?? '') as String,
```

**2. `_rawGet` back-fills absent concrete paths with `null` before the check**
`lib/core/usp/services/usp_client.dart` `_rawGet` (~L296–306):

```dart
for (final path in paths) {
  if (path.contains('*')) continue;          // wildcard paths skipped
  if (!result.containsKey(path)) {
    logger.w('... GET missing path in response: "$path"');
  }
  result.putIfAbsent(path, () => null);        // ← absent concrete path → key present, value null
}
```

The stated intent (comment in that block) is to *"prevent Null Cast errors in codegen"*.

**3. Net effect on concrete-path models**
absent required leaf → `_rawGet` back-fills `key=present, value=null` → codegen `containsKey(path)` is **true** → `missing` stays empty → 9998 never throws → `ServiceErrorView` unreachable → the `?? ''` fallback renders the field as an empty value instead.

**4. Wildcard pages behave differently (why S03 Port Forwarding *can* reach the error view)**
Pages that GET with `*` (e.g. `port_forwarding.g.dart` uses `Device.NAT.PortMapping.*.InternalClient`) are skipped by the `if (path.contains('*')) continue` guard, so a faulted/absent leaf really goes missing and 9998 fires as designed.

Note: the concrete `Pool.1.` form itself is a legitimate, intentional path shape (other pages that intentionally target a single known instance use it too — e.g. `dhcp_reservations.g.dart` / `dhcp_clients.g.dart` use `Pool.1.StaticAddress.*` / `Pool.1.Client.*`). This report is **not** claiming `Pool.1.` is wrong; it is reporting that concrete-path required leaves lose their 9998/`ServiceErrorView` protection.

## Impact

For any codegen model that fetches **required** leaves via **concrete** (non-wildcard) paths, the required-field-missing (9998) → `ServiceErrorView` path is effectively dead: a backend that returns an instance but omits a required leaf will render empty values rather than the error view. `lan_network_info` (LAN / Local Network page) is one confirmed case; other concrete-path models with `missing.add(...)` checks are likely affected too.

## Open question for the team (not asserting a fix)

Two defensible readings, need a decision:
- **(a) This is intended graceful degradation** — the `?? ''` fallback is the real error handling; the 9998 branch on concrete-path pages is redundant and could be removed for those pages.
- **(b) The required-field contract should hold** — `_rawGet` should not back-fill leaves that codegen declares required (the back-fill guard keys off `*` presence, which encodes "enumerate vs. specific instance", not "optional vs. required"), or codegen should carry the required/optional distinction down to the client.

Whichever way, the two current behaviours (declare-required-and-9998 vs. back-fill-null-and-`?? ''`) are in tension and currently resolved silently by `_rawGet`.

## Related

- E2E: the `S06 Local Network` error-state test is skipped pending this decision (test exists but asserts a state the app currently can't reach). Skip note in `PrivacyGUI-USP-E2E` `tests/S06-local-network-states.spec.ts` will be updated to reference this issue.

AustinChangLinksys:
## Summary

Fixes #1184.

`UspClient._rawGet` back-filled every absent non-wildcard requested path with `null` before returning. That flipped `response.containsKey(path)` to `true`, silently defeating the codegen required-leaf check (`code: 9998` → `ServiceErrorView`) for **every model that GETs with concrete paths** — 15 models, 67 required leaves, including the mixed concrete+wildcard `dns_client`.

This PR removes the back-fill and locks the behaviour with tests.

## Why the back-fill was the bug (not the protection)

The back-fill carried this comment:

```dart
// Ensure all requested non-wildcard paths exist in the result to prevent
// Null Cast errors in codegen.
result.putIfAbsent(path, () => null);
```

**The comment is inaccurate.** A Dart map already returns `null` for an absent key, so `response[path] as String` receives `null` *whether or not* the back-fill ran — `null as String` throws either way. The back-fill prevents no Null Cast; the real Null-Cast guard is the `?? ''` on each generated assignment, which runs *after* the 9998 check anyway. The back-fill's only observable effect was flipping `containsKey()` from `false` to `true`, which suppressed 9998.

**Framed as a contract:** the YAML/codegen `required: true` set *is* the contract between UI and firmware. The back-fill made the UI unilaterally absorb any contract violation as a silent empty value — a violation that no one could see or fix. Removing it restores contract fidelity: a genuinely missing required leaf now surfaces as `9998 → ServiceError` instead of being swallowed. Treating a missing `required` leaf as a real signal (to be resolved at the contract layer — firmware or YAML) is the correct default, not something the UI should paper over.

## The change

- **`lib/core/usp/services/usp_client.dart`**
  - Drop the `putIfAbsent(path, () => null)` back-fill from `_rawGet`.
  - Extract the pure coerce + missing-path-warning logic into a `@visibleForTesting static normalizeGetResponse(...)` seam. `_rawGet` now delegates to it, passing the logger as `onMissingPath`. Behaviour is otherwise unchanged.
  - Make `_coerceValue` `static` (it was already pure).

  The seam is necessary because `_rawGet` is private and `UspClient` can't be instantiated in the test VM (its constructor throws off-Web and the WASM client isn't injectable) — which is *exactly why the back-fill was never test-covered*. All existing tests mock the high-level `get()`, bypassing `_rawGet` entirely.

- **`test/core/usp/services/usp_client_normalize_response_test.dart`** (new, 7 tests)
  - No-back-fill group: absent concrete path is not added; present values coerced; empty string stays present; wildcard paths never warn; absent concrete path invokes `onMissingPath`.
  - Codegen-contract group: a normalized response omitting a required leaf leaves `containsKey(leaf) == false` — the exact predicate the generated `_fromResponse` evaluates to reach 9998; a complete response keeps every required leaf.

## Affected scope (per-leaf, for reviewers)

**14 fully-affected models:** `system_info` (11), `lan_network_info` (9), `wan_settings` (9), `ipv6settings` (6), `time_settings` (6), `wan_status` (6), `wan_static_ip` (5), `wan_traffic_stats` (4), `wan_pppoe` (3), `setup_state` (1), `wan_bridge` (1), `wan_dhcp` (1), `gre_tunnel` (1), `l2tp_tunnel` (1).

**1 mixed model:** `dns_client` — 3 concrete top-level required leaves silenced (`Device.DNS.Client.Enable`, `.Status`, `.ServerNumberOfEntries`); wildcard `Server.*.*` leaves were already protected.

**Total: 15 models, 67 silenced required leaves** now restored.

## Verification

| Check | Result |
|---|---|
| `flutter analyze` | clean |
| New tests | +7 pass |
| Regression-lock (temporarily re-add the back-fill) | the 2 load-bearing assertions fail as expected — the tests genuinely guard the fix |
| Full functional suite (`run_tests.sh`, golden/loc/ui excluded) | **3346 pass / 0 fail / 0 skip** |

Consumer behaviour is preserved: a revived 9998 surfaces as a `ServiceError` and is caught where it matters — e.g. `pnp_notifier.startPostLoginFlow` transitions to `AdminReadFailure` rather than crashing or showing a false "no internet".

## Behavioural change reviewers should be aware of

For the 15 models above, a missing required leaf changes from *"empty value, page still usable"* to *"full `ServiceErrorView`"*. This is intended: a missing `required` leaf is a contract violation and should be visible, not silently degraded. To keep that safe operationally, 9998 remains logged so a real firmware/YAML violation surfaces as a signal we can act on at the contract layer — the opposite of what the back-fill did.

**Blast radius is not uniformly page-scoped.** For most of the 15 models the effect is exactly "one settings page shows `ServiceErrorView`", but two paths are wider:

- **`system_info` is session-scoped, not page-scoped.** `session_service.dart:73` fetches `SystemInfo` during device-info bootstrap (at login, via `login_local_view.dart`) and wraps any failure as `ConnectivityError` (`session_service.dart:78`); the same model is also polled by `usp_system_monitor_service`, and `usp_system_monitor_notifier` counts `ConnectivityError` up to `_errorThreshold` before flipping app-wide connection state. So a single missing `system_info` leaf can take out bootstrap / connection state, not just a page.
- **Internet Settings is fail-fast.** `usp_internet_settings_service.fetchSettings()` (`usp_internet_settings_service.dart:52-60`) fetches WAN / IPv6 / PPP / VLAN / GRE / L2TP in one `Future.wait`, so any single missing required leaf turns the *entire* Internet Settings page into `ServiceErrorView` rather than blanking one field group.

**This does not add a new source of 9998 — it widens the reach of an existing one.** 9998-from-firmware-omission was already live on wildcard models (Port Forwarding is the issue's own example of a page that can reach the error view); the back-fill loop iterated the *requested* paths and `continue`d on `*`, while codegen checks the *expanded* `${p}` keys that were never in the back-fill's coverage. What this PR changes is that reach: from peripheral wildcard pages to core ones (WAN family, `system_info`, LAN), +15 models / 67 leaves. Note the user-facing message does not distinguish a transport failure from a genuine firmware omission — both map through `category=validation → InvalidInputError → errorInvalidInput`; only the engineer-facing 9998 message (which lists the missing paths) and the retained `onMissingPath` log tell them apart. Improving that user-facing message is tracked separately (`doc/error-handling/usp-error-handling-reference.md` §2.5) and is orthogonal to this fix.

**Watch-item (not a blocker) — a few `required` leaves sit on optional instances.** `Device.GRE.Tunnel.1.RemoteEndpoints`, `Device.L2TPv2.Tunnel.1.RemoteEndpoints`, and `Device.PPP.Interface.1.Username`/`.Password` live on instances that may legitimately not exist on a router never configured for GRE/L2TP/PPPoE. If this surfaces, it is most likely the YAML marking those leaves `required` when they should be optional (fix belongs in the contract, not the client), not firmware violating the contract. Because they share the fail-fast `fetchSettings()` above, any one of them missing fails the whole Internet Settings page — so this is the first place to look if 9998 appears there. This is the same "does firmware ever legitimately omit a `required` leaf?" question raised in #1184: the default remains "trust the YAML"; recalibrate the required set only if a concrete case appears.

---

> **Update (2026-08-07, post-review):** the "Behavioural change" section above was expanded after review to fold in @HankYuLinksys's three framing corrections — (1) `system_info` / Internet Settings blast radius is session-scoped / fail-fast rather than single-page; (2) this widens an existing 9998 source rather than adding a new one; (3) the GRE/L2TP/PPP optional-instance watch-item. The code and tests are unchanged from the reviewed commit `8c36d329`; this edit is documentation only.

### DIFF
--- a/lib/core/usp/services/usp_client.dart
+++ b/lib/core/usp/services/usp_client.dart
@@ -287,25 +287,12 @@ class UspClient {
         logger.w('$_tag$label GET response EMPTY for paths: $paths');
       }
 
-      final Map<String, dynamic> result = {};
-
-      for (final entry in rawMap.entries) {
-        result[entry.key] = _coerceValue(entry.key, entry.value);
-      }
-
-      // Ensure all requested non-wildcard paths exist in the result to prevent
-      // Null Cast errors in codegen. Wildcard search paths (containing '*') are
-      // expanded by the router into concrete instance paths, so the original
-      // wildcard path won't appear in the response — skip those.
-      for (final path in paths) {
-        if (path.contains('*')) continue;
-        if (!result.containsKey(path)) {
-          logger.w('$_tag$label GET missing path in response: "$path"');
-        }
-        result.putIfAbsent(path, () => null);
-      }
-
-      return result;
+      return normalizeGetResponse(
+        paths,
+        rawMap,
+        onMissingPath: (path) =>
+            logger.w('$_tag$label GET missing path in response: "$path"'),
+      );
     } catch (e) {
       sw.stop();
       final label = _idLabel(id);
@@ -314,13 +301,48 @@ class UspClient {
     }
   }
 
+  /// Normalizes a raw USP GET response into the map consumed by codegen models.
+  ///
+  /// Two responsibilities, kept as pure logic so it is unit-testable without a
+  /// live WASM client:
+  /// 1. Coerce every returned value via [_coerceValue].
+  /// 2. Warn (via [onMissingPath]) for each requested non-wildcard path absent
+  ///    from the response — but deliberately do NOT back-fill it. Back-filling
+  ///    absent paths with null used to silently suppress the codegen
+  ///    required-leaf check (code 9998 → ServiceErrorView), because it flipped
+  ///    `containsKey` to true without preventing any Null Cast (the real guard
+  ///    is the `?? ''` on each codegen assignment). Leaving the key absent lets
+  ///    the required-leaf contract fire as designed (#1184).
+  ///
+  /// Wildcard search paths (containing '*') are expanded by the router into
+  /// concrete instance paths, so the original wildcard path won't appear in the
+  /// response — those are skipped by the missing-path warning.
+  @visibleForTesting
+  static Map<String, dynamic> normalizeGetResponse(
+    List<String> paths,
+    Map<String, String?> rawMap, {
+    void Function(String path)? onMissingPath,
+  }) {
+    final Map<String, dynamic> result = {};
+    for (final entry in rawMap.entries) {
+      result[entry.key] = _coerceValue(entry.key, entry.value);
+    }
+    for (final path in paths) {
+      if (path.contains('*')) continue;
+      if (!result.containsKey(path)) {
+        onMissingPath?.call(path);
+      }
+    }
+    return result;
+  }
+
   /// Coerce a raw string value from USP into the appropriate Dart type.
   /// - "true" / "false" →bool (any path)
   /// - "1" / "0" →bool (for known boolean suffixes: Enable, Active)
   /// - null →null (key absent from response)
   /// - Empty string →'' (preserve String type for generated code)
   /// - Everything else stays as String (generated code handles int parsing)
-  dynamic _coerceValue(String path, String? raw) {
+  static dynamic _coerceValue(String path, String? raw) {
     if (raw == null) return null;
     if (raw.isEmpty) return '';
 
--- a/test/core/usp/services/usp_client_normalize_response_test.dart
+++ b/test/core/usp/services/usp_client_normalize_response_test.dart
@@ -1,0 +1,164 @@
+import 'package:flutter_test/flutter_test.dart';
+import 'package:privacy_gui/core/usp/services/usp_client.dart';
+
+/// Regression tests for issue #1184.
+///
+/// `UspClient._rawGet` used to back-fill every absent non-wildcard requested
+/// path with `null` before returning. That silently defeated the codegen
+/// required-leaf check (`code: 9998` → `ServiceErrorView`) for models that GET
+/// with concrete paths: a back-filled key made `response.containsKey(path)`
+/// return true, so the `missing` list stayed empty and 9998 never fired.
+///
+/// The fix removes the back-fill. These tests pin that behaviour on the pure,
+/// testable [UspClient.normalizeGetResponse] seam (the real `_rawGet` cannot be
+/// unit-tested directly — its constructor throws off-Web and the WASM client is
+/// not injectable, which is exactly why the back-fill was never covered).
+void main() {
+  group('UspClient.normalizeGetResponse — no back-fill (#1184)', () {
+    test('absent concrete path is NOT added to the result', () {
+      // The router omitted a requested concrete leaf entirely.
+      final result = UspClient.normalizeGetResponse(
+        ['Device.DHCPv4.Server.Pool.1.MinAddress'],
+        <String, String?>{},
+      );
+
+      // The load-bearing assertion: adding `putIfAbsent(path, () => null)` back
+      // makes containsKey true and fails this test.
+      expect(
+        result.containsKey('Device.DHCPv4.Server.Pool.1.MinAddress'),
+        isFalse,
+      );
+      expect(result, isEmpty);
+    });
+
+    test('present paths are kept and value-coerced', () {
+      final result = UspClient.normalizeGetResponse(
+        [
+          'Device.DHCPv4.Server.Pool.1.Enable', // bool suffix
+          'Device.DHCPv4.Server.Pool.1.MinAddress', // plain string
+        ],
+        <String, String?>{
+          'Device.DHCPv4.Server.Pool.1.Enable': '1',
+          'Device.DHCPv4.Server.Pool.1.MinAddress': '192.168.1.100',
+        },
+      );
+
+      expect(result['Device.DHCPv4.Server.Pool.1.Enable'], isTrue);
+      expect(
+        result['Device.DHCPv4.Server.Pool.1.MinAddress'],
+        '192.168.1.100',
+      );
+    });
+
+    test(
+        'a path present with an empty string stays present (not treated as '
+        'missing)', () {
+      final result = UspClient.normalizeGetResponse(
+        ['Device.DHCPv4.Server.Pool.1.DNSServers'],
+        <String, String?>{'Device.DHCPv4.Server.Pool.1.DNSServers': ''},
+      );
+
+      // Empty string is a real value → key present, coerced to ''.
+      expect(
+          result.containsKey('Device.DHCPv4.Server.Pool.1.DNSServers'), isTrue);
+      expect(result['Device.DHCPv4.Server.Pool.1.DNSServers'], '');
+    });
+
+    test('wildcard request paths never trigger a missing-path warning', () {
+      final missing = <String>[];
+      // Router expands the wildcard into a concrete instance; the original
+      // wildcard path is absent from the response but must be ignored.
+      UspClient.normalizeGetResponse(
+        ['Device.DNS.Client.Server.*.DNSServer'],
+        <String, String?>{'Device.DNS.Client.Server.1.DNSServer': '8.8.8.8'},
+        onMissingPath: missing.add,
+      );
+
+      expect(missing, isEmpty);
+    });
+
+    test('absent concrete path invokes onMissingPath; present one does not',
+        () {
+      final missing = <String>[];
+      UspClient.normalizeGetResponse(
+        [
+          'Device.DHCPv4.Server.Pool.1.MinAddress', // absent
+          'Device.DHCPv4.Server.Pool.1.MaxAddress', // present
+        ],
+        <String, String?>{
+          'Device.DHCPv4.Server.Pool.1.MaxAddress': '192.168.1.200',
+        },
+        onMissingPath: missing.add,
+      );
+
+      expect(missing, ['Device.DHCPv4.Server.Pool.1.MinAddress']);
+    });
+  });
+
+  group('Codegen required-leaf contract is reachable again (#1184)', () {
+    // Ties the normalize seam to a real concrete-path model. When a required
+    // leaf is omitted, normalizeGetResponse leaves it absent, so the model's
+    // `containsKey` required-check adds it to `missing` and throws 9998.
+    // (Before the fix the back-fill made this branch unreachable.)
+    const instancePath = 'Device.IP.Interface.1.';
+
+    List<String> lanPaths() => [
+          '${instancePath}IPv4Address.1.IPAddress',
+          '${instancePath}IPv4Address.1.SubnetMask',
+          'Device.DHCPv4.Server.Pool.1.Enable',
+          'Device.DHCPv4.Server.Pool.1.MinAddress',
+          'Device.DHCPv4.Server.Pool.1.MaxAddress',
+          'Device.DHCPv4.Server.Pool.1.LeaseTime',
+          'Device.DHCPv4.Server.Pool.1.DNSServers',
+          'Device.DeviceInfo.HostName',
+          '${instancePath}IPv6Enable',
+        ];
+
+    Map<String, String?> fullLanResponse() => <String, String?>{
+          '${instancePath}IPv4Address.1.IPAddress': '192.168.1.1',
+          '${instancePath}IPv4Address.1.SubnetMask': '255.255.255.0',
+          'Device.DHCPv4.Server.Pool.1.Enable': '1',
+          'Device.DHCPv4.Server.Pool.1.MinAddress': '192.168.1.100',
+          'Device.DHCPv4.Server.Pool.1.MaxAddress': '192.168.1.200',
+          'Device.DHCPv4.Server.Pool.1.LeaseTime': '86400',
+          'Device.DHCPv4.Server.Pool.1.DNSServers': '8.8.8.8',
+          'Device.DeviceInfo.HostName': 'router',
+          '${instancePath}IPv6Enable': 'true',
+        };
+
+    test(
+        'normalized response omitting a required leaf makes the codegen check '
+        'see it as missing (→ 9998)', () {
+      final withMissing = fullLanResponse()
+        ..remove('Device.DHCPv4.Server.Pool.1.MinAddress');
+
+      final normalized =
+          UspClient.normalizeGetResponse(lanPaths(), withMissing);
+
+      // This is precisely the predicate the generated `_fromResponse` evaluates;
+      // false here means `missing.add(...)` runs and 9998 is thrown.
+      expect(
+        normalized.containsKey('Device.DHCPv4.Server.Pool.1.MinAddress'),
+        isFalse,
+      );
+    });
+
+    test(
+        'complete normalized response keeps every required leaf present '
+        '(no false 9998)', () {
+      final normalized =
+          UspClient.normalizeGetResponse(lanPaths(), fullLanResponse());
+
+      // Every required leaf the generated check inspects must be present, so
+      // `missing` stays empty and the model builds normally.
+      for (final path in lanPaths()) {
+        expect(normalized.containsKey(path), isTrue,
+            reason: 'required leaf $path must survive normalization');
+      }
+      // Spot-check coercion feeding the generated assignments.
+      expect(normalized['Device.DHCPv4.Server.Pool.1.Enable'], isTrue);
+      expect(normalized['Device.DHCPv4.Server.Pool.1.MinAddress'],
+          '192.168.1.100');
+    });
+  });
+}


AustinChangLinksys (commented):
🤖 **Automated Review — Round 1** · `24b17ad..8c36d32` (full)

**Verdict: 💬 Self-review (comment only)** — Self-authored PR; review provided as reference. The fix is mechanically correct; one low-severity confirmation item before merge.

#### Standards

|---|---|---|---|
| 💡 | Med | `lib/core/usp/services/usp_client.dart:normalizeGetResponse` | `rawMap[key] = null` (explicitly present with null value) still makes `containsKey` true → codegen 9998 not triggered; semantically different from "absent key" but not tested |
| 💡 | Med | `lib/core/usp/services/usp_client.dart:~290–292` | `onMissingPath` only logs a warning; no analytics hook or error counter — silent in production if FW starts omitting required leaves |
| 💡 | Low | all codegen `*.g.dart` models | No integration test confirms all concrete-path models use `containsKey` pattern; any model using direct `response['X'] as String` would throw `Null Cast` (not 9998) after this change |

#### Spec (Issue #1184)

|---|---|---|---|
| 💡 | High | `usp_client_normalize_response_test.dart` | All 7 tests directly verify the no-back-fill spec; codegen required-leaf contract group precisely covers the `containsKey` predicate the generated `_fromResponse` uses |

*Confidence: High = code-verified · Med = located + reasoned · Low = speculative.*

<details>

- **Back-fill removal is correct:** `putIfAbsent(path, () => null)` only flipped `containsKey` from false to true without preventing Null Cast (the `?? ''` on codegen assignments is the real guard). Removing it is the correct call.
- **`normalizeGetResponse` extraction:** promoting pure coerce+warn logic to a `@visibleForTesting static` seam is the right architecture for testability — `UspClient` constructor throws on non-Web, making the private method untestable directly.
- **7 targeted tests** with load-bearing assertions (temporarily re-adding the back-fill fails them) confirm the fix is genuine.
- **Wildcard skip preserved:** `path.contains('*')` guard correctly omits wildcard paths from the missing-key check, consistent with prior behavior.
- **`_coerceValue` static:** correct — it was already pure, making it static is accurate.

</details>

HankYuLinksys (approved):
Approving.

All three points from my review were framing/documentation, not code changes, and they've been folded into the PR body — including the two corrections that matter most for future readers: the `system_info` / Internet Settings blast radius being session-scoped and fail-fast rather than single-page, and the fact that this widens an existing 9998 source rather than adding a new one. The GRE/L2TP/PPP watch-item is recorded with an explicit default ("trust the YAML, recalibrate only on a concrete case"), which is the right disposition for something we can't settle from the client side.

Basis for approval: I verified the three load-bearing claims against the tree — the back-fill's "prevent Null Cast" comment is inaccurate (an absent Dart map key returns `null` regardless), concrete-path bare casts are 0 (all 24 in `lib/generated/*.g.dart` are wildcard-derived `${p}` forms the `*` guard already skips), and no non-codegen GET consumer reads a back-filled key. I did not re-run the suite independently; taking the reported `flutter analyze` clean / 3346 pass at face value, with the regression-lock check (re-add the back-fill → the 2 load-bearing assertions fail) as the reason the new tests are credible rather than merely present.

Code and tests unchanged from the reviewed commit `8c36d329`.

lib/core/usp/services/usp_client.dart
```diff
@@ -314,13 +301,48 @@ class UspClient {
...
+    List<String> paths,
+    Map<String, String?> rawMap, {
+    void Function(String path)? onMissingPath,
+  }) {
+    final Map<String, dynamic> result = {};
+    for (final entry in rawMap.entries) {
+      result[entry.key] = _coerceValue(entry.key, entry.value);
+    }
+    for (final path in paths) {
+      if (path.contains('*')) continue;
```
PeterJhongLinksys, line 331: ⚠️ **Warning** (🟢High) — Only `*` is skipped here, but USP object/table GETs end with a trailing `.` and are expanded by the router into instance paths, so the original key never appears in the response. Confirmed callers that will now log a false missing-path warning on every call: `usp_lan_data_service.dart:71`, `usp_client.dart:701`/`:824`, `usp_test_console_view.dart:544`/`:548`. The demo implementation already has the correct condition (`demo_usp_service.dart:69`). Suggest: `if (path.contains(*) || path.endsWith(.)) continue;` plus a trailing-dot test next to the existing wildcard one.

PeterJhongLinksys (approved):
🤖 **Automated Review — Round 1** · `24b17ad..8c36d32` (full)

**Verdict: ✅ APPROVE** — The fix is correct, minimal and well-reasoned; no Critical found. Two follow-ups worth handling before/right after merge: the demo client still back-fills, and the missing-path warning does not skip trailing-dot object paths.

|  | Conf. | Where | Issue (one-liner) |
|---|---|---|---|
| ⚠️ | 🟢 | `lib/demo/usp/demo_usp_service.dart:67-71` | [Duplicated Code / Divergent Change] `DemoUspClient.get()` still does `putIfAbsent(path, () => null)`, so demo mode keeps the exact bug this PR removes — 9998 stays dead there and optional-leaf `null as String` risk remains. |
| ⚠️ | 🟢 | `lib/core/usp/services/usp_client.dart:331` | Missing-path check only skips `*`, not trailing-dot object/table paths — real callers (`usp_lan_data_service.dart:71`, `usp_client.dart:701/824`, `usp_test_console_view.dart:544/548`) will log a false "missing path" warning on every call. Demo version already has `|| path.endsWith('.')`; production does not. |
| ⚠️ | 🟢 | `lib/core/usp/services/usp_client.dart:330-335` | Behaviour-change blast radius not documented: any concrete-path model now hard-fails with 9998 instead of degrading. Please confirm what the WASM GET returns on real hardware when e.g. `Device.GRE.Tunnel.1.RemoteEndpoints` / `Device.L2TPv2.Tunnel.1.RemoteEndpoints` / `Device.Bridging.Bridge.1.Enable` are absent (empty data → whole Internet Settings page errors, vs. thrown 9005 → unchanged). |
| ⚠️ | 🟡 | `test/core/usp/services/usp_client_normalize_response_test.dart:98-163` | Second group claims to pin the codegen required-leaf contract but never executes any codegen — it hand-copies `lan_network_info.g.dart`'s path list and re-asserts `containsKey`, which is equivalent to group 1 and will silently drift. |
| ⚠️ | 🟡 | `lib/core/usp/services/usp_client.dart:320-337` | The `@visibleForTesting static` seam works around the real testability root cause (`UspClient` constructor throws off-Web, WASM client not injectable) and leaks a wire-format detail into the public surface. Note the annotation also contradicts intent: demo code *should* call it. |
| ⚠️ | 🟡 | `lib/core/usp/services/usp_client.dart:286-295` | On a failed GET envelope `rawMap` is empty, so `onMissingPath` fires once per requested concrete path (e.g. 9 warnings for `lan_network_info`). Consider one aggregated warning / early return. |
| ⚠️ | ⚪ | `lib/core/usp/web/usp_client_wasm.dart:173-186` | Pre-existing known defect (`doc/error-handling/usp-error-handling-reference.md` §2.5): the envelope's `success`/`error` is dropped, so a 9999 transport failure now reliably surfaces as 9998 → `InvalidInputError`. Removing the back-fill raises this path's hit rate; worth pairing with "Direction A". |
| 💡 | 🟡 | `lib/core/usp/services/usp_client.dart:321-325` | Returning `(values, missing)` instead of an `onMissingPath` side-effect callback would keep the function pure, let `_rawGet` own the logging, and let `DemoUspClient` reuse it with full missingness info. |
| 💡 | 🟡 | `test/core/usp/services/usp_client_normalize_response_test.dart:105-127` | Inline ad-hoc `lanPaths()` / `fullLanResponse()` fixtures conflict with the Test Data Builder rule (Art. I §1.6.2); `test/mocks/test_data/` already has precedent. |
| 💡 | 🟡 | `lib/demo/usp/demo_usp_service.dart:300-314` | `_coerce` is a third hand-copy of `_coerceValue`; exposing `_coerceValue` (now `static`) for reuse would delete the duplicate and remove a drift source (production already covers `Upstream` suffix). |
| 💡 | ⚪ | `test/core/usp/services/usp_client_normalize_response_test.dart` | Filename deviates from the `[class]_test.dart` convention used by the other 11 files in the directory; `usp_client_test.dart` is free and avoids per-method fragmentation. |
| 💡 | ⚪ | `lib/core/usp/services/usp_client.dart:304-319` | DartDoc is half issue-archaeology; consider trimming to the current contract and moving the #1184 history to a one-line comment / PR description. Also worth stating that for *optional* leaves safety comes from `containsKey`, not `?? ''`. |
| 💡 | ⚪ | `lib/core/usp/services/usp_client.dart` (1039 lines) | [Large Class] transport + 401 reauth + subscription CRUD + normalization + logging in one class; extracting the normalizer into its own file would be a natural first cut and fixes the seam concern above. |
| 💡 | ⚪ | `test/core/usp/services/usp_client_normalize_response_test.dart` | Missing edge cases: empty `paths`, extra unrequested keys in `rawMap`, duplicate path in `paths` (callback fires twice today), and value explicitly `null` (key present but null). |

<sub>Confidence: 🟢High = read head code, evidence quoted · 🟡Med = file:line + rationale, not fully proven · ⚪Low = speculative.</sub>
<sub>`[X]` = raised by 1 agent · `[XX]` = independently raised by both agents. Critical rows only.</sub>

<details>

**1. Demo client still back-fills — `lib/demo/usp/demo_usp_service.dart:67-71`** (both agents)

```dart
// Ensure non-wildcard paths have an entry (null for missing)
for (final path in paths) {
  if (path.contains('*') || path.endsWith('.')) continue;
  result.putIfAbsent(path, () => null);
}
```
`DemoUspClient extends UspClient` and `@override`s `get()`, so it never reaches `normalizeGetResponse`. Consequences in demo mode: (a) `containsKey` stays true → the codegen required-leaf check (`code: 9998`) never fires, so a missing leaf shows as an empty value instead of `ServiceErrorView`; (b) optional leaves read as `containsKey(k) ? response[k] as String : null` (e.g. `connected_devices.g.dart:235-238`) get a present-but-null key → `null as String` TypeError. Net effect: QA cannot validate this fix in demo, and production/demo behaviour now diverges on the exact axis the PR changes. Fix: call `UspClient.normalizeGetResponse(paths, raw)` from `DemoUspClient.get()` and delete the local copy.

**2. Trailing-dot object paths produce false missing-path warnings — `lib/core/usp/services/usp_client.dart:331`**

```dart
for (final path in paths) {
  if (path.contains('*')) continue;      // only wildcards skipped
  if (!result.containsKey(path)) {
    onMissingPath?.call(path);
  }
}
```
USP object/table GETs end with `.` and are expanded by the router into instance paths, so the original key never appears in the response. Confirmed callers: `lib/page/local_network/services/usp_lan_data_service.dart:71` (`'Device.IP.Interface.1.IPv6Address.'`, consumed via `getInstances(...)` at :80), `lib/core/usp/services/usp_client.dart:701` and `:824` (`'Device.LocalAgent.Subscription.'`), `lib/page/test_console/views/usp_test_console_view.dart:544/548`. Impact is log noise only (these callers do not run `containsKey` checks), but it drowns the genuine warnings this PR relies on. The demo implementation already has the correct condition — production should match: `if (path.contains('*') || path.endsWith('.')) continue;`. Please also add a test for the trailing-dot case alongside the existing wildcard one.

**3. Blast radius / open question for the author — `usp_client.dart:330-335`**

`lib/generated/*.g.dart` (~30 files) treat their concrete GET paths as required and `throw '... (code: 9998)'` on any absent key. `lib/page/internet_settings/services/usp_internet_settings_service.dart:52-84` fans out `Future.wait([WanSettings, Ipv6Settings, GreTunnel, L2tpTunnel, ...])` and maps any throw to a `ServiceError`, so one absent leaf fails the whole page. `gre_tunnel.g.dart:26-32` / `l2tp_tunnel.g.dart:26-32` each have exactly one required leaf, and `assets/resources/demo_usp_data.json` contains no `Device.GRE.*` / `Device.L2TPv2.*` keys at all. Whether this is a real regression depends entirely on router behaviour for absent tables (omit key → new hard failure; throw 9005 → unchanged, since `_rawGet`'s `catch (e) { ...; rethrow; }` path is untouched). Not flagged Critical precisely because that cannot be verified from the repo — please confirm with device logs.

**4. Second test group does not test what it claims — `test/.../usp_client_normalize_response_test.dart:98-163`**

The group hand-copies `lan_network_info.g.dart:48-58`'s path list into local `lanPaths()` / `fullLanResponse()` helpers and then asserts `normalized.containsKey(...)`. `LanNetworkInfo._fromResponse` and `_buildPaths` are private, so the `missing.add(...)` → `throw '... (code: 9998)'` contract (`lan_network_info.g.dart:69-99`) is never executed. The first test is logically identical to `absent concrete path is NOT added to the result` (:18-32); the second ("present keys are kept") holds for any map copy. It also silently decays if codegen changes the path list. Either drop the group, or make it real by injecting a fake client and calling `LanNetworkInfo.fetch(fake)` asserting the 9998 throw.

**5. Seam design — `usp_client.dart:320-337`**

The test file correctly diagnoses the root cause ("its constructor throws off-Web and the WASM client is not injectable"; see `usp_client.dart:78-92`, `_client` is a `late final` direct instantiation). This PR's answer is one `@visibleForTesting static` per helper, which does not generalise to `_singleSet` / `_batchSet` / `createNotifySubscription`, and makes a transport-layer detail part of `UspClient`'s public surface. Preferred alternatives: extract a top-level pure function into `lib/core/usp/services/usp_get_response_normalizer.dart`, or add `@visibleForTesting UspClient.forTesting(UspClientWeb client)` so the whole class becomes testable. Current approach is acceptable as a stepping stone — please leave a tracking issue so it does not become the standing precedent.

</details>

<details>

- **The diagnosis in the doc comment is correct, and the fix is worth more than the PR claims.** Verified across `lan_network_info.g.dart:100-122`, `wan_status.g.dart:82-97`, `setup_state.g.dart:48-65`, `gre_tunnel.g.dart:33-36`: required leaves read `(response[k] ?? '') as String`, so the old back-fill genuinely prevented no Null Cast — the original comment was wrong. Beyond restoring 9998, removing the back-fill also eliminates the `null as String` TypeError risk on the ~25 optional-leaf sites shaped `containsKey(k) ? response[k] as String : null`.
- **Wildcard handling preserved correctly** (`:331`): `connected_devices.g.dart:95-114`, `ip_interfaces.g.dart:34-41`, `dmz.g.dart:52-59`, `ppp_interface.g.dart:59-67` all query with `*` and consume via `getInstances` / prefix scans, so no false warnings and no spurious 9998.
- **Error path untouched**: `_rawGet`'s `catch (e) { log; rethrow; }` and `_withAuthRetry`'s 401 retry (`:233-243`) are unchanged. Downstream services consistently use `catch (e)` + `mapUspErrorToServiceError` (`usp_lan_data_service.dart:46-48`, `usp_internet_settings_service.dart:82-84`, `usp_wifi_advanced_service.dart:41-43`); a repo-wide grep for `on Exception catch` returns 0 hits, matching the "USP errors are raw Strings" rule.
- **Architecture clean**: change is confined to `lib/core/usp/services/`; no layering, generated-model isolation (Art. V §5.4), USP Provider four hard rules (Art. IV), ServiceError isolation (Art. XIII), UI Kit (Art. XV) or Equatable (Art. XI) concerns. Scope discipline is exemplary — no drive-by lint churn.
- **`_coerceValue` → `static` is safe**: no subclass overrides it (`DemoUspClient` has its own `_coerce`), and it is private so it cannot be overridden externally.
- **Test group 1 (`:17-96`) genuinely pins behaviour**: absent path not added, present paths kept and coerced, empty string treated as a real value (an easy case to get wrong — nice), wildcard does not warn, callback fires precisely. Names describe intent, no numbering, group name follows convention.
- **No security regression**: no credential/token handling touched, no hardcoded secrets, `normalizeGetResponse` is pure (no I/O, no async, no shared mutable state) so it introduces no races.

</details>

PR #1199, opened 2026-08-04
fix(ai): batch tool results into one message per assistant turn (#1191)

HankYuLinksys:
## Summary

Fixes #1191.

The Claude Messages API requires every `tool_use` block in an assistant message to be answered by `tool_result` blocks carried in the **single** user message that immediately follows it. `RouterChatController` emitted one `ChatMessage.toolResult` per result — one user message each — and a confirmation-gated command left its `tool_use` unanswered entirely. Both make the API reject the whole request.

This PR batches results into one message per assistant turn, and fixes six further problems found in the same flow while verifying that change.

## Why the API contract matters here

```json
{"role": "assistant", "content": [
    {"type": "tool_use", "id": "A", "name": "getWanStatus"},
    {"type": "tool_use", "id": "B", "name": "setDmz"}
]}
{"role": "user", "content": [
    {"type": "tool_result", "tool_use_id": "A", "content": "..."},
    {"type": "tool_result", "tool_use_id": "B", "content": "..."}
]}
```

Both results must sit in **one** user message. Order is irrelevant (matching is by `tool_use_id`), but completeness and grouping are not — a violation is rejected wholesale, not partially tolerated.

The old loop produced two consecutive user messages for two tools, and when one tool needed confirmation it emitted the siblings' results and returned, leaving that `tool_use` with no answer at all. `_insertPendingToolResults()` could not repair it: it is only called from `sendMessage()` (so pressing Confirm never reaches it) and it returned early on encountering any existing `tool_result` — which is exactly the state the bug produces.

Low-impact today because there is currently no write command on `dev-2.7.0`. It becomes easy to trigger as soon as write commands land: users naturally say "check my network and turn X off", and the model answers with several tools at once.

## The change

**`lib/ai/orchestrator/router_chat_controller.dart`**

Results accumulate in `_pendingResults` and are flushed as a single `ChatMessage.toolResults` message once the batch is complete. A pending confirmation keeps the batch open across the dialog, so sibling results are held back and flushed together with the write result. Invisible to the user — tool results are never rendered in the chat.

New members:

| Member | Purpose |
|---|---|
| `_pendingResults` | Accumulates the current batch |
| `_conversationGeneration` | Detects that the conversation was reset while work was in flight |
| `_commandsByName` | Command lookup cache (was re-querying the provider per tool) |
| `_flushPendingResults()` | Emits the batch as one message |
| `_completeBatch(generation)` | Flushes only if the generation is still current |
| `_abandonBatch(deferred)` | Winds down a failed batch: answers what ran, keeps a confirmation |
| `_abandonedToolResult(name)` | Distinguishes commands from A2UI components |

Also fixed, all in the same flow:

- **Two confirmation-required commands in one response no longer drop the first.** Only the first is executed; the rest are answered with `status: 'not_executed'` and a reason, so the model asks again in a later turn. Previously the single `_pendingConfirmation` field was overwritten and the first `tool_use` could never be answered — with no user-visible indication.
- **The dialog is surfaced after the tool loop finishes.** `_executeToolUse()` now *reports* a `PendingConfirmation` instead of assigning it and notifying mid-loop, so confirming while siblings are still awaiting can no longer start a second concurrent `_processConversation()`.
- **`retry()` no longer deletes a flushed tool result.** Tool result messages also carry `ChatRole.user`, so "remove the failed user message" destroyed real device data whenever the failure landed on the LLM call *after* a successful flush — the typical retryable case. It now excludes tool results, and resumes the exchange instead of re-asking the question when tools already ran.
- **A confirmation resolving after `clearConversation()` no longer injects an orphan.** The Clear button has no in-flight guard, so a result could arrive after the reset and become the first message of a fresh conversation — a `tool_result` with no preceding `tool_use`, which breaks *every* later request and cannot be repaired by `_insertPendingToolResults()` (it only walks back to the trailing assistant turn). Guarded by the generation counter.
- **A failure mid-batch now answers the tools that did run**, instead of discarding results for commands that already executed against the device, and **keeps a confirmation discovered before the failure** (`deferred` moved out of the `try`, which previously let a write request evaporate with no Retry affordance).
- **Abandoned commands report `not_executed` rather than `rendered`.** The `{'status': 'rendered'}` placeholder is correct for A2UI components but told the model an unexecuted command had succeeded.
- **`handleToolAction()` closes out an unanswered previous turn** rather than leaving `_pendingConfirmation` live while appending its own single-part result.

**`test/ai/orchestrator/router_chat_controller_test.dart`** — new `tool result batching` group, 11 tests, one per problem above plus the `_insertPendingToolResults` path.

**`pubspec.yaml`** — bump `ui_kit_library` / `generative_ui` to `v2.31.0`, which contains the new `ChatMessage.toolResults(List<ToolResultPart>)` factory (UI kit commit `53aaeb80`). That factory is additive: `ChatMessage.toolResult` is untouched, and for a single part both produce identical output.

## Verification

| Check | Result |
|---|---|
| `flutter test test/ai/ test/page/ai_assistant/` | **59/59 pass** (48 pre-existing + 11 new) |
| `./run_tests.sh` (full functional suite) | **3350/3350 pass** |
| `flutter analyze` on changed files | clean |
| `fvm dart format` | 2 files, 0 changed |
| UI kit `generative_ui` package tests | 189/189 pass (7 new for the factory) |

The full suite was run because the UI kit bump spans three minor versions (v2.28.1 → v2.31.0) and affects the whole project, not just the AI module. Verified against the released `v2.31.0` tag, not a local override.

Two pre-existing `prefer_const_literals` infos remain in the test file (lines 136, 184) — present on `dev-2.7.0`, untouched by this PR.

## Notes for reviewers

- `handleToolAction()` is currently unreachable: the A2UI action payload carries `surfaceId` / `sourceComponentId` / `name`, not `toolUseId`, so the view returns early. Its fix is in place for when that wiring is completed.
- Only one confirmation is processed per turn by design. A full confirmation queue was considered and rejected as premature — the current behaviour is explicit (nothing is silently dropped) and avoids firing consecutive dialogs at the user.
- `_pendingResults` ordering follows execution, not `tool_use` order (a confirmed write is appended after its siblings). This is fine — the API matches by `tool_use_id`.
- Prerequisite for #1192 and for opening up write commands.

### DIFF
--- a/lib/ai/orchestrator/router_chat_controller.dart
+++ b/lib/ai/orchestrator/router_chat_controller.dart
@@ -1,3 +1,4 @@
+import 'package:equatable/equatable.dart';
 import 'package:flutter/foundation.dart';
 import 'package:generative_ui/generative_ui.dart';
 
@@ -5,23 +6,49 @@ import 'package:privacy_gui/ai/abstraction/_abstraction.dart';
 import 'package:privacy_gui/ai/prompts/router_system_prompt.dart';
 import 'package:privacy_gui/core/utils/logger.dart';
 
+/// Results accumulated for one assistant turn, owned by the exchange that
+/// opened it.
+///
+/// Pairing the results with the generation they belong to is what makes
+/// staleness safe to act on: work resuming after an `await` holds a reference
+/// to *its own* batch, so it can never read, clear, or append to results that
+/// belong to a later exchange.
+class ToolResultBatch {
+  ToolResultBatch(this.generation);
+
+  /// Conversation generation this batch was opened in.
+  final int generation;
+
+  /// Tool results collected so far, in execution order.
+  ///
+  /// Order does not have to match the assistant's `tool_use` order — a
+  /// confirmed write is appended after its siblings — because the API matches
+  /// results to requests by `tool_use_id`. What matters is that they all arrive
+  /// in the single user message following the assistant turn.
+  final List<ToolResultPart> results = [];
+}
+
 /// Pending confirmation state for write/admin operations.
-class PendingConfirmation {
+class PendingConfirmation extends Equatable {
   final RouterCommand command;
   final Map<String, dynamic> params;
   final ToolUseBlock toolUse;
 
-  /// Conversation generation the batch that produced this confirmation belongs
-  /// to. Validated before the result is emitted so a confirmation resolving
-  /// after the conversation moved on cannot inject an orphaned `tool_result`.
-  final int generation;
+  /// The batch this confirmation was raised from. Its result rejoins that
+  /// batch's siblings, and its generation decides whether the batch may still
+  /// be emitted — so a confirmation answered after the conversation moved on
+  /// cannot inject an orphaned `tool_result`.
+  final ToolResultBatch batch;
 
   const PendingConfirmation({
     required this.command,
     required this.params,
     required this.toolUse,
-    required this.generation,
+    required this.batch,
   });
+
+  @override
+  List<Object?> get props => [command, params, toolUse, batch];
 }
 
 /// Controller for managing Router AI Assistant conversations.
@@ -42,7 +69,7 @@ class PendingConfirmation {
 ///
 /// The Claude Messages API requires every `tool_use` block in an assistant
 /// message to be answered within the *single* user message that immediately
-/// follows it. Results are therefore accumulated in [_pendingResults] and
+/// follows it. Results are therefore accumulated in a [ToolResultBatch] and
 /// flushed as one [ChatMessage.toolResults] message once every tool in the
 /// batch has a result.
 ///
@@ -86,14 +113,13 @@ class RouterChatController extends ChangeNotifier {
   Map<String, RouterCommand>? _commandsByName;
   Future<Map<String, RouterCommand>>? _commandsLoad;
 
-  /// Tool results for the batch currently being processed. Flushed by
-  /// [_flushPendingResults].
+  /// The batch currently accumulating results, or null between exchanges.
   ///
-  /// Ordering follows execution, not the assistant's `tool_use` order — a
-  /// confirmed write is appended after its siblings. That is fine: the API
-  /// matches results to requests by `tool_use_id`, and only requires them all
-  /// to arrive in the single user message following the assistant turn.
-  final List<ToolResultPart> _pendingResults = [];
+  /// Owning the results together with the generation they belong to is what
+  /// makes staleness safe to act on: work returning from an `await` can only
+  /// ever touch *its own* batch, so it can never clear or append to results
+  /// belonging to a later exchange.
+  ToolResultBatch? _batch;
 
   /// Incremented whenever the conversation is reset, so work that was already
   /// in flight (e.g. a confirmation awaiting the device) can tell that its
@@ -157,7 +183,10 @@ class RouterChatController extends ChangeNotifier {
   /// drift apart. The future itself is cached so concurrent first calls share
   /// one query.
   Future<Map<String, RouterCommand>> _loadCommands() {
-    return _commandsLoad ??= _commandProvider.listCommands().then((commands) {
+    final cached = _commandsLoad;
+    if (cached != null) return cached;
+
+    final load = _commandProvider.listCommands().then((commands) {
       _routerTools = commands
           .map((cmd) => GenTool(
                 name: cmd.name,
@@ -167,12 +196,21 @@ class RouterChatController extends ChangeNotifier {
           .toList();
       return _commandsByName = {for (final c in commands) c.name: c};
     });
+
+    // Drop a rejected load so a transient provider failure does not stick for
+    // the rest of the conversation.
+    _commandsLoad = load.catchError((Object e) {
+      _commandsLoad = null;
+      throw e;
+    });
+    return _commandsLoad!;
   }
 
   /// Tool definitions for the model, loading the registry if needed.
   Future<List<GenTool>> _getRouterTools() async {
     await _loadCommands();
-    return _routerTools!;
+    // Set by the load above; the map and the list are always populated together.
+    return _routerTools ?? const [];
   }
 
   /// Send a user message and process response.
@@ -205,13 +243,24 @@ class RouterChatController extends ChangeNotifier {
   /// - Detecting confirmation requirements
   /// - Looping for multi-turn tool use
   Future<void> _processConversation() async {
-    // Snapshot the generation this exchange belongs to. Every await below is a
-    // point where the user can reset the conversation, and anything written
-    // afterwards would land in a conversation this work no longer belongs to.
-    final generation = _conversationGeneration;
+    // This exchange owns its own batch, stamped with the generation it belongs
+    // to. Every await below is a point where the user can reset the
+    // conversation; because results live on the batch rather than in shared
+    // state, work resuming afterwards can only ever discard its own.
+    final batch = ToolResultBatch(_conversationGeneration);
+    _batch = batch;
 
-    final tools = await _getRouterTools();
-    if (_isStale(generation)) return;
+    final List<GenTool> tools;
+    try {
+      tools = await _getRouterTools();
+    } catch (e) {
+      _log('_processConversation: loading commands failed: $e');
+      if (_isStale(batch)) return;
+      _state = _state.withError('An unexpected error occurred: $e');
+      notifyListeners();
+      return;
+    }
+    if (_isStale(batch)) return;
 
     var loopCount = 0;
     const maxLoops = 5;
@@ -234,7 +283,7 @@ class RouterChatController extends ChangeNotifier {
           tools: tools.isNotEmpty ? tools : null,
           systemPromptParts: _systemPromptParts,
         );
-        if (_isStale(generation)) return;
+        if (_isStale(batch)) return;
         _log(
             '_processConversation: LLM responded, stopReason=${response.stopReason}');
 
@@ -289,7 +338,7 @@ class RouterChatController extends ChangeNotifier {
 
         // Process each tool use. Results are accumulated rather than emitted
         // one by one, so the whole batch can be flushed as a single message.
-        _pendingResults.clear();
+        batch.results.clear();
 
         for (final toolUse in toolUseBlocks) {
           // Every tool is wrapped so that a failure anywhere — including a
@@ -301,12 +350,12 @@ class RouterChatController extends ChangeNotifier {
             // keeps the batch complete and tells the model to ask again.
             if (deferred != null) {
               final command = await _findCommand(toolUse.name);
-              if (_isStale(generation)) return;
+              if (_isStale(batch)) return;
               if (command?.requiresConfirmation ?? false) {
                 _log(
                     '_processConversation: deferring extra confirmation-required '
                     'tool "${toolUse.name}" to a later turn');
-                _pendingResults.add(ToolResultPart(
+                batch.results.add(ToolResultPart(
                   toolUseId: toolUse.id,
                   result: {
                     'status': _statusNotExecuted,
@@ -321,8 +370,8 @@ class RouterChatController extends ChangeNotifier {
             }
 
             _log('_processConversation: executing tool "${toolUse.name}"');
-            final result = await _executeToolUse(toolUse, generation);
-            if (_isStale(generation)) return;
+            final result = await _executeToolUse(toolUse, batch);
+            if (_isStale(batch)) return;
 
             if (result.requiresConfirmation) {
               _log('_processConversation: tool requires confirmation');
@@ -331,7 +380,7 @@ class RouterChatController extends ChangeNotifier {
             } else {
               _log(
                   '_processConversation: tool result: ${result.isError ? "ERROR" : "SUCCESS"}');
-              _pendingResults.add(ToolResultPart(
+              batch.results.add(ToolResultPart(
                 toolUseId: toolUse.id,
                 result: result.data,
                 isError: result.isError,
@@ -339,7 +388,7 @@ class RouterChatController extends ChangeNotifier {
             }
           } catch (e) {
             _log('_processConversation: tool "${toolUse.name}" threw: $e');
-            _pendingResults.add(ToolResultPart(
+            batch.results.add(ToolResultPart(
               toolUseId: toolUse.id,
               result: {'error': e.toString()},
               isError: true,
@@ -356,7 +405,7 @@ class RouterChatController extends ChangeNotifier {
           return;
         }
 
-        if (!_completeBatch(generation)) return;
+        if (!_completeBatch(batch)) return;
 
         // Restore loading state before continuing loop
         _log('_processConversation: continuing to next loop iteration');
@@ -364,13 +413,13 @@ class RouterChatController extends ChangeNotifier {
         notifyListeners();
       } on GenUiException catch (e) {
         _log('_processConversation: GenUiException: ${e.message}');
-        _abandonBatch(generation, deferred);
+        _abandonBatch(batch, deferred);
         _state = _state.withError(e.message, retryable: e.isRetryable);
         notifyListeners();
         return;
       } catch (e) {
         _log('_processConversation: unexpected error: $e');
-        _abandonBatch(generation, deferred);
+        _abandonBatch(batch, deferred);
         _state = _state.withError('An unexpected error occurred: $e');
         notifyListeners();
         return;
@@ -387,17 +436,17 @@ class RouterChatController extends ChangeNotifier {
   Future<RouterCommand?> _findCommand(String name) async =>
       (await _loadCommands())[name];
 
-  /// Whether [generation] belongs to a conversation that has since been reset.
+  /// Whether [batch] belongs to a conversation that has since been reset.
   ///
-  /// Work in flight across an `await` must check this before touching state:
-  /// writing into a conversation it no longer belongs to would leave a
-  /// `tool_result` with no matching `tool_use` and invalidate every later
-  /// request.
-  bool _isStale(int generation) {
-    if (generation == _conversationGeneration) return false;
-    _log('_isStale: conversation moved on (batch gen $generation, '
+  /// A pure predicate — it deliberately touches no state, so a losing exchange
+  /// cannot disturb the one that replaced it. Work resuming after an `await`
+  /// checks this before writing anything: a `tool_result` landing in a
+  /// conversation it does not belong to has no matching `tool_use` and
+  /// invalidates every later request.
+  bool _isStale(ToolResultBatch batch) {
+    if (batch.generation == _conversationGeneration) return false;
+    _log('_isStale: conversation moved on (batch gen ${batch.generation}, '
         'current $_conversationGeneration) — abandoning this exchange');
-    _pendingResults.clear();
     return true;
   }
 
@@ -406,29 +455,43 @@ class RouterChatController extends ChangeNotifier {
   /// Any tools that already ran are answered so the assistant turn is not left
   /// with unanswered `tool_use` blocks, and a confirmation discovered before
   /// the failure is still surfaced so the user's write request is not lost.
-  void _abandonBatch(int generation, PendingConfirmation? deferred) {
+  void _abandonBatch(ToolResultBatch batch, PendingConfirmation? deferred) {
     if (deferred != null) {
+      // Re-arming only makes sense while the batch still owns the conversation.
+      if (_isStale(batch)) return;
       _log(
           '_abandonBatch: keeping confirmation for "${deferred.command.name}"');
       _pendingConfirmation = deferred;
       return;
     }
-    _completeBatch(generation);
+    _completeBatch(batch);
   }
 
-  /// Emit [_pendingResults] as a single tool result message and clear the batch.
+  /// Emit [batch]'s results as a single tool result message.
   ///
-  /// Does nothing when the batch is empty, since a tool result message with no
-  /// content is not valid.
-  void _flushPendingResults() {
-    if (_pendingResults.isEmpty) return;
-    _log('_flushPendingResults: emitting ${_pendingResults.length} result(s) '
-        'in one message');
-    _state = _state.withMessage(
-      ChatMessage.toolResults(List.of(_pendingResults)),
-    );
-    _pendingResults.clear();
-    notifyListeners();
+  /// Returns false when the batch no longer owns the conversation, in which
+  /// case nothing is written. Does nothing for an empty batch either, since a
+  /// tool result message with no content is not valid.
+  bool _completeBatch(ToolResultBatch batch) {
+    if (_isStale(batch)) {
+      _log(
+          '_completeBatch: discarding ${batch.results.length} stale result(s)');
+      batch.results.clear();
+      return false;
+    }
+    if (batch.results.isNotEmpty) {
+      _log('_completeBatch: emitting ${batch.results.length} result(s) '
+          'in one message');
+      // Copy: the message must not alias a list this batch still owns.
+      _state = _state.withMessage(
+        ChatMessage.toolResults(List.of(batch.results)),
+      );
+      // Emptied so a later wind-down of the same batch cannot re-emit them.
+      batch.results.clear();
+      notifyListeners();
+    }
+    if (_batch == batch) _batch = null;
+    return true;
   }
 
   /// Execute a single tool use block.
@@ -438,7 +501,7 @@ class RouterChatController extends ChangeNotifier {
   /// the dialog cannot appear while sibling tools are still executing.
   Future<_ToolExecutionResult> _executeToolUse(
     ToolUseBlock toolUse,
-    int generation,
+    ToolResultBatch batch,
   ) async {
     final command = await _findCommand(toolUse.name);
 
@@ -456,7 +519,7 @@ class RouterChatController extends ChangeNotifier {
           command: command,
           params: toolUse.input,
           toolUse: toolUse,
-          generation: generation,
+          batch: batch,
         ),
       );
     }
@@ -487,9 +550,9 @@ class RouterChatController extends ChangeNotifier {
     final pending = _pendingConfirmation;
     if (pending == null) return;
 
-    // Validate against the generation the batch was opened in, not the current
-    // one — the conversation may have moved on while the dialog was showing.
-    final generation = pending.generation;
+    // The confirmation carries the batch it belongs to, so the result joins
+    // that batch's siblings — not whatever is open when the user answers.
+    final batch = pending.batch;
     _pendingConfirmation = null;
     _state = _state.loading();
     notifyListeners();
@@ -502,44 +565,30 @@ class RouterChatController extends ChangeNotifier {
 
       // Complete the batch this confirmation belongs to, then flush it as one
       // message so every tool_use in the assistant turn is answered together.
-      _pendingResults.add(ToolResultPart(
+      batch.results.add(ToolResultPart(
         toolUseId: pending.toolUse.id,
         result: result.success
-            ? {...result.data, 'status': _statusExecuted}
+            // Nested so a 'status' field the device itself returns is not
+            // silently overwritten by ours.
+            ? {'status': _statusExecuted, 'result': result.data}
             : {'error': result.error ?? 'Operation failed'},
         isError: !result.success,
       ));
-      if (!_completeBatch(generation)) return;
+      if (!_completeBatch(batch)) return;
 
       // Continue conversation to get LLM's follow-up response
       await _processConversation();
     } catch (e) {
-      _pendingResults.add(ToolResultPart(
+      batch.results.add(ToolResultPart(
         toolUseId: pending.toolUse.id,
         result: {'error': e.toString()},
         isError: true,
       ));
-      if (!_completeBatch(generation)) return;
+      if (!_completeBatch(batch)) return;
 
       // Still continue to get LLM response about the error
       await _processConversation();
     }
-  }
-
-  /// Flush the batch unless the conversation moved on while we were awaiting.
-  ///
-  /// Returns false when [generation] is stale — the results belong to a
-  /// conversation that has since been cleared, so emitting them would leave a
-  /// `tool_result` with no matching `tool_use` and break every later request.
-  bool _completeBatch(int generation) {
-    if (generation != _conversationGeneration) {
-      _log('_completeBatch: conversation was reset, discarding '
-          '${_pendingResults.length} stale result(s)');
-      _pendingResults.clear();
-      return false;
-    }
-    _flushPendingResults();
-    return true;
   }
 
   /// Cancel the pending action.
@@ -550,22 +599,21 @@ class RouterChatController extends ChangeNotifier {
     final pending = _pendingConfirmation;
     if (pending == null) return;
 
-    // Validate against the generation the batch was opened in, not the current
-    // one — the conversation may have moved on while the dialog was showing.
-    final generation = pending.generation;
+    // The cancellation joins the batch the confirmation came from.
+    final batch = pending.batch;
     _pendingConfirmation = null;
     _state = _state.loading();
     notifyListeners();
 
     // Complete the batch with the cancellation, then flush it as one message.
-    _pendingResults.add(ToolResultPart(
+    batch.results.add(ToolResultPart(
       toolUseId: pending.toolUse.id,
       result: {
         'status': _statusCancelled,
         'message': 'User declined to execute ${pending.command.name}',
       },
     ));
-    if (!_completeBatch(generation)) return;
+    if (!_completeBatch(batch)) return;
 
     // Continue conversation to get LLM's response to cancellation
     await _processConversation();
@@ -573,11 +621,11 @@ class RouterChatController extends ChangeNotifier {
 
   /// Handle tool action from A2UI component.
   Future<void> handleToolAction(ToolActionOutput action) async {
-    // A UI action answers a single rendered component, so it is its own batch.
     // Close out whatever the previous turn left open first — including a
     // confirmation the user never answered, whose tool_use would otherwise stay
-    // unanswered while this action's result is appended.
-    _closeOutOpenBatch();
+    // unanswered while this action's result is appended. The id this action
+    // answers is excluded so it is not answered twice.
+    _closeOutOpenBatch(except: {action.toolUseId});
 
     final toolResultMessage = action.toChatMessage();
     _state = _state.withMessage(toolResultMessage);
@@ -599,21 +647,29 @@ class RouterChatController extends ChangeNotifier {
   ///
   /// Results already computed for that turn are **preserved** — they hold real
   /// device data — and only the genuinely missing ids are synthesised. The
-  /// generation is bumped so a confirmation still awaiting the device cannot
-  /// come back and answer an id this method just answered.
-  void _closeOutOpenBatch() {
-    final hadOpenWork =
-        _pendingConfirmation != null || _pendingResults.isNotEmpty;
+  /// generation is always bumped, so anything still in flight for the closed
+  /// batch (a confirmation awaiting the device, or an LLM call mid-request)
+  /// cannot come back and answer an id this method just answered.
+  void _closeOutOpenBatch({Set<String> except = const {}}) {
+    final hadConfirmation = _pendingConfirmation != null;
     _pendingConfirmation = null;
-    if (hadOpenWork) {
-      // Invalidate in-flight work belonging to the batch being closed.
-      _conversationGeneration++;
-    }
 
     // Results computed for the turn being closed are real; keep them and only
     // fill the gaps below.
-    final carried = List<ToolResultPart>.of(_pendingResults);
-    _pendingResults.clear();
+    final closed = _batch;
+    _batch = null;
+
+    // Bump only when work for the closed batch may still be in flight — an
+    // unanswered confirmation awaiting the device, or a batch left open by an
+    // exchange that has not finished. A batch that already completed sets
+    // _batch to null, so a normal turn does not invalidate the next one.
+    if (hadConfirmation || closed != null) {
+      _conversationGeneration++;
+    }
+
+    final carried = closed == null
+        ? <ToolResultPart>[]
+        : List<ToolResultPart>.of(closed.results);
 
     if (_state.messages.isEmpty) return;
 
@@ -642,6 +698,8 @@ class RouterChatController extends ChangeNotifier {
 
     // Carried results answer their ids too — do not synthesise over them.
     answeredIds.addAll(carried.map((p) => p.toolUseId));
+    // Ids the caller is about to answer itself.
+    answeredIds.addAll(except);
 
     final synthesised = [
       for (final block in response.content.whereType<ToolUseBlock>())
@@ -734,7 +792,7 @@ class RouterChatController extends ChangeNotifier {
     // the new conversation once the device responds.
     _conversationGeneration++;
     _pendingConfirmation = null;
-    _pendingResults.clear();
+    _batch = null;
     _lastUserMessage = null;
     _routerTools = null;
     _commandsByName = null;
--- a/lib/ai/orchestrator/router_chat_controller.dart
+++ b/lib/ai/orchestrator/router_chat_controller.dart
@@ -31,6 +31,19 @@ class PendingConfirmation {
 /// 3. If command requires confirmation → set pendingConfirmation, notify UI
 /// 4. User confirms → execute command → add tool_result → get LLM response
 /// 5. User cancels → add cancellation tool_result → get LLM response
+///
+/// ## Batching tool results
+///
+/// The Claude Messages API requires every `tool_use` block in an assistant
+/// message to be answered within the *single* user message that immediately
+/// follows it. Results are therefore accumulated in [_pendingResults] and
+/// flushed as one [ChatMessage.toolResults] message once every tool in the
+/// batch has a result.
+///
+/// When a command requires confirmation the batch stays open across the
+/// confirmation dialog: results for the sibling tools are held back until the
+/// user confirms or cancels, then flushed together with the write result. This
+/// is invisible to the user because tool results are never rendered in the chat.
 class RouterChatController extends ChangeNotifier {
   RouterChatController({
     required IConversationGenerator generator,
@@ -57,6 +70,21 @@ class RouterChatController extends ChangeNotifier {
   String? _lastUserMessage;
   List<SystemPromptPart>? _systemPromptParts;
   List<GenTool>? _routerTools;
+  Map<String, RouterCommand>? _commandsByName;
+
+  /// Tool results for the batch currently being processed. Flushed by
+  /// [_flushPendingResults].
+  ///
+  /// Ordering follows execution, not the assistant's `tool_use` order — a
+  /// confirmed write is appended after its siblings. That is fine: the API
+  /// matches results to requests by `tool_use_id`, and only requires them all
+  /// to arrive in the single user message following the assistant turn.
+  final List<ToolResultPart> _pendingResults = [];
+
+  /// Incremented whenever the conversation is reset, so work that was already
+  /// in flight (e.g. a confirmation awaiting the device) can tell that its
+  /// conversation is gone and drop its result instead of injecting an orphan.
+  int _conversationGeneration = 0;
 
   // Token usage tracking
   int _totalInputTokens = 0;
@@ -164,6 +192,10 @@ class RouterChatController extends ChangeNotifier {
       loopCount++;
       _log('_processConversation: loop $loopCount/$maxLoops');
 
+      // Declared outside the try so the catch blocks can still surface a
+      // confirmation that was discovered before the failure.
+      PendingConfirmation? deferred;
+
       try {
         _log('_processConversation: calling LLM...');
         final response = await _generator.generateWithHistory(
@@ -223,35 +255,61 @@ class RouterChatController extends ChangeNotifier {
         _state = _state.loading();
         notifyListeners();
 
-        // Process each tool use
-        var requiresConfirmation = false;
+        // Process each tool use. Results are accumulated rather than emitted
+        // one by one, so the whole batch can be flushed as a single message.
+        _pendingResults.clear();
 
         for (final toolUse in toolUseBlocks) {
+          // Only the first confirmation-required command is executed in this
+          // batch. Answering the rest with an explicit "not executed" result
+          // keeps the batch complete and tells the model to ask again.
+          if (deferred != null) {
+            final command = await _findCommand(toolUse.name);
+            if (command?.requiresConfirmation ?? false) {
+              _log(
+                  '_processConversation: deferring extra confirmation-required '
+                  'tool "${toolUse.name}" to a later turn');
+              _pendingResults.add(ToolResultPart(
+                toolUseId: toolUse.id,
+                result: {
+                  'status': 'not_executed',
+                  'reason': 'Only one operation requiring confirmation can be '
+                      'handled at a time. Request this one again in a '
+                      'follow-up turn.',
+                },
+              ));
+              continue;
+            }
+          }
+
           _log('_processConversation: executing tool "${toolUse.name}"');
           final result = await _executeToolUse(toolUse);
 
           if (result.requiresConfirmation) {
             _log('_processConversation: tool requires confirmation');
-            requiresConfirmation = true;
-            // Don't add tool_result yet — wait for user confirmation
+            // Hold the batch open — the result is added once the user decides.
+            deferred = result.pendingConfirmation;
           } else {
             _log(
                 '_processConversation: tool result: ${result.isError ? "ERROR" : "SUCCESS"}');
-            // Add tool result immediately
-            _state = _state.withMessage(ChatMessage.toolResult(
+            _pendingResults.add(ToolResultPart(
               toolUseId: toolUse.id,
               result: result.data,
               isError: result.isError,
             ));
-            notifyListeners();
           }
         }
 
-        if (requiresConfirmation) {
-          // Stop loop, UI will handle confirmation
+        if (deferred != null) {
+          // Surface the dialog only now that every sibling tool has finished,
+          // so confirming cannot re-enter _processConversation mid-loop.
           _log('_processConversation: waiting for user confirmation');
+          _pendingConfirmation = deferred;
+          notifyListeners();
           return;
         }
+
+        _flushPendingResults();
 
         // Restore loading state before continuing loop
         _log('_processConversation: continuing to next loop iteration');
@@ -259,11 +317,13 @@ class RouterChatController extends ChangeNotifier {
         notifyListeners();
       } on GenUiException catch (e) {
         _log('_processConversation: GenUiException: ${e.message}');
+        _abandonBatch(deferred);
         _state = _state.withError(e.message, retryable: e.isRetryable);
         notifyListeners();
         return;
       } catch (e) {
         _log('_processConversation: unexpected error: $e');
+        _abandonBatch(deferred);
         _state = _state.withError('An unexpected error occurred: $e');
         notifyListeners();
         return;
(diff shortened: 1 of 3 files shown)


lib/ai/orchestrator/router_chat_controller.dart
```diff
@@ -223,47 +255,75 @@ class RouterChatController extends ChangeNotifier {
...
+        if (deferred != null) {
+          // Surface the dialog only now that every sibling tool has finished,
+          // so confirming cannot re-enter _processConversation mid-loop.
           _log('_processConversation: waiting for user confirmation');
+          _pendingConfirmation = deferred;
+          notifyListeners();
           return;
         }
 
+        _flushPendingResults();
```
PeterJhongLinksys: 🔴 **Critical** (🟢 High, raised independently by both reviewers) — this flush is not generation-guarded.

`_completeBatch(generation)` (line 476) exists precisely to stop a stale batch from being emitted, but only `confirmPendingAction`/`cancelPendingAction` call it. Here the loop flushes unconditionally. If the user taps the AppBar refresh (`router_assistant_view.dart:398`, not disabled while loading) while `await _executeToolUse(...)` (line 286) is in flight, `clearConversation()` resets `_state` and bumps `_conversationGeneration`, and this line then writes a `tool_result` into the fresh, empty conversation — a `tool_result` with no matching `tool_use`, which as your own comment on lines 473-475 notes will "break every later request".

Suggested fix: snapshot `final generation = _conversationGeneration;` at the top of `_processConversation()`, early-return after every `await` when it goes stale (lines 244, 251, 295, 303-309), and route this flush through `_completeBatch(generation)` so the invariant lives in one place. Also worth storing the generation inside `PendingConfirmation` so `confirmPendingAction` validates the batch generation rather than the current one.

lib/ai/orchestrator/router_chat_controller.dart
```diff
@@ -276,14 +336,55 @@ class RouterChatController extends ChangeNotifier {
...
+  ///
+  /// Any tools that already ran are answered so the assistant turn is not left
+  /// with unanswered `tool_use` blocks, and a confirmation discovered before
+  /// the failure is still surfaced so the user's write request is not lost.
+  void _abandonBatch(PendingConfirmation? deferred) {
+    if (deferred != null) {
+      _log(
+          '_abandonBatch: keeping confirmation for "${deferred.command.name}"');
+      _pendingConfirmation = deferred;
+      return;
```
PeterJhongLinksys, line 465: ⚠️ **Warning** (🟢 High) — keeping the batch open here interacts badly with `retry()`.

After this early return the state is `messages = [user, assistant(tool_use A, B)]`, `_pendingResults = [A real device data]`, `_pendingConfirmation = B`, and `hasError = true` (set by the caller at line 321/327). The error banner with Retry (`router_assistant_view.dart:406`, `:670`) is therefore rendered at the same time as the confirmation dialog. If Retry is pressed, `retry()` (line 612) sees a trailing *assistant* message → `removedUserTurn == false` → `_processConversation()` → line 260 `_pendingResults.clear()` discards A real result while both `tool_use` blocks stay unanswered → API rejection, and `_pendingConfirmation` is left dangling.

The modal dialog (`barrierDismissible: false`) covers the Retry button in practice, which is why this is a Warning rather than a Critical. Suggested fix: handle `_pendingConfirmation != null` explicitly at the start of `retry()`, and only clear `_pendingResults` when a batch is genuinely being opened.

lib/ai/orchestrator/router_chat_controller.dart
```diff
@@ -403,67 +530,124 @@ class RouterChatController extends ChangeNotifier {
...
+  ///
+  /// This happens when the previous turn ended without completing its batch —
+  /// for example the user typed a new message instead of answering a pending
+  /// confirmation dialog. Leaving a `tool_use` unanswered makes the API reject
+  /// the next request, so the gaps are filled with an acknowledgment.
   void _insertPendingToolResults() {
+    // A confirmation is still open: drop it and abandon its batch, otherwise
+    // the stale results would be flushed alongside the next turn's.
+    _pendingConfirmation = null;
+    _pendingResults.clear();
```
PeterJhongLinksys: ⚠️ **Warning** (🟢 High) — abandoning the batch here does not invalidate an in-flight confirmation.

`_conversationGeneration` is incremented only by `clearConversation()` (line 645). Semantically this method abandons the open batch, but a `confirmPendingAction()` that is still awaiting the device will pass `_completeBatch(generation)` unchanged when it returns, and flush a second `tool_result` for a `tool_use_id` that this method already answered with `not_executed` — the same duplicate/orphan corruption the PR is fixing. Reachable from `sendMessage()` (line 162) and `handleToolAction()` (line 520); note `_handleA2UIAction` (`router_assistant_view.dart:631-649`) has no `isLoading` guard.

Suggested fix: `_conversationGeneration++;` here as well, and have confirm/cancel validate against the generation captured when the batch was opened.

lib/ai/orchestrator/router_chat_controller.dart
```diff
@@ -403,67 +530,124 @@ class RouterChatController extends ChangeNotifier {
...
-        toolUseId: toolUse.id,
-        result: {'status': 'rendered', 'component': toolUse.name},
-      ));
+  /// Result body for a `tool_use` the user walked away from.
+  ///
+  /// Router commands must not be reported as done — they never ran. Anything
+  /// that is not a known command is an A2UI component, which really was
+  /// rendered, so the original acknowledgment still applies.
+  Map<String, dynamic> _abandonedToolResult(String toolName) {
+    if (_commandsByName?.containsKey(toolName) ?? false) {
```
PeterJhongLinksys: ⚠️ **Warning** (🟡 Med) — this fallback fails on the unsafe side.

`_commandsByName` is a lazy cache populated only by `_findCommand()` (line 344) and reset to `null` by `clearConversation()` (line 650). Whenever it is null, an unexecuted **write** command falls through to `{"status": "rendered"}` on line 601 — telling the model the operation completed. That is exactly what the doc comment on lines 591-593 and the test at `router_chat_controller_test.dart:608-632` exist to prevent, so this line is a re-entry point for the bug being fixed. All current call paths appear to pass through `_findCommand()` first, hence Med confidence — it reads as a latent hazard rather than a live defect.

Suggested fix: make `_insertPendingToolResults` async and consult `await _findCommand(name)` for the authoritative answer, or invert the fallback so "unknown ⇒ not_executed" (safe side) and treat only registry-confirmed components as rendered.

PeterJhongLinksys (commented):
🤖 **Automated Review — Round 1** · `80e3d17..0245aa8` (full)

**Verdict: 💬 Comment** — The batching invariant is a real improvement and the new tests are strong, but the `_conversationGeneration` guard is only wired into the confirm/cancel paths; the main tool loop still flushes unconditionally, so the exact failure mode this PR set out to fix (`tool_result` with no matching `tool_use`) remains reachable. Leaving the merge decision to the maintainer.

|---|---|---|---|
| 🔴 `[XX]` | 🟢 | `lib/ai/orchestrator/router_chat_controller.dart:312` | `_flushPendingResults()` in the tool loop is not generation-guarded — `clearConversation()` during an in-flight tool injects an orphaned `tool_result` into the fresh conversation and breaks every later request |
| ⚠️ | 🟢 | `router_chat_controller.dart:356-364` + `604-636` | `_abandonBatch(deferred)` keeps results and confirmation, but `retry()` then re-enters `_processConversation()` where line 260 clears them → device data silently dropped and the assistant turn's `tool_use` blocks stay unanswered |
| ⚠️ | 🟢 | `router_chat_controller.dart:540-544` | `_insertPendingToolResults()` abandons a batch without incrementing `_conversationGeneration`, so an in-flight `confirmPendingAction` still passes `_completeBatch()` and emits a duplicate `tool_result` for an already-answered id |
| ⚠️ | 🟡 | `router_chat_controller.dart:594-601` | `_abandonedToolResult()` derives semantics from a nullable lazy cache; when `_commandsByName == null` an unexecuted write command is reported to the model as `{'status':'rendered'}` — the exact lie the new test guards against |
| ⚠️ | 🟡 | `router_chat_controller.dart:515-531` | `handleToolAction()` calls `_insertPendingToolResults()` and then appends a second `tool_result` message, potentially answering the same `tool_use_id` twice and breaking the "one tool_result message per turn" invariant |
| ⚠️ | 🟢 | `router_chat_controller.dart:668-683` | `[Primitive Obsession]` `_ToolExecutionResult` allows `requiresConfirmation: true` with `pendingConfirmation: null`; line 291/303 would then silently drop the user's write request with no assert or log |
| ⚠️ | 🟢 | `router_chat_controller.dart:140-152` vs `343-349` | `[Duplicated Code]` two independent caches each call `listCommands()` and are invalidated separately — double query on first turn, drift risk later |
| ⚠️ | 🟡 | `router_chat_controller.dart:458-463` (pre-existing) | Art. XIII: raw `e.toString()` is sent to the LLM as the tool result instead of going through `mapUspErrorToServiceError()`; USP errors are raw Strings and may carry internal detail |
| ⚠️ | 🟡 | `test/ai/orchestrator/router_chat_controller_test.dart:30-51`, `670-705` | Art. I §1.6.2: fixtures/response builders are inline in the test file instead of a `test/mocks/test_data/*_test_data.dart` Test Data Builder |
| 💡 | 🟢 | `router_chat_controller.dart:544` + `571-585` | Successfully executed sibling read results are cleared and then overwritten with `not_executed`, discarding real device data and mislabelling a read that did run |
| 💡 | 🟢 | `router_chat_controller.dart:262-301` | A deferred write waits for confirmation while sibling reads run immediately → the model receives pre-write values with no ordering hint |
| 💡 | 🟢 | `router_chat_controller.dart:183-337` | `[Long Method]` `_processConversation()` is ~150 lines covering LLM call, token accounting, tool batching, one-confirmation policy and failure cleanup |
| 💡 | 🟢 | `router_chat_controller.dart:540` | `[Mysterious Name]` `_insertPendingToolResults()` also silently discards an open confirmation — a much heavier decision than the name implies |
| 💡 | 🟢 | `router_chat_controller.dart:447-468`, `501-511` | `[Duplicated Code]` the "add result → completeBatch → continue" template is repeated three times |
| 💡 | 🟢 | `router_chat_controller.dart:275, 450, 504, 597, 601` | Tool-result protocol magic strings (`not_executed` / `executed` / `cancelled` / `rendered`) are scattered across four methods with no shared constants |
| 💡 | 🟢 | `router_chat_controller.dart:343-349` | Benign race: concurrent first-time tool calls both see a null cache and call `listCommands()` twice; cache the `Future`, not the result |
| 💡 | 🟢 | `test/.../router_chat_controller_test.dart:439` | `[Brittle Test]` asserts an exact flush order that lines 78-81 explicitly document as semantically irrelevant |
| 💡 | 🟢 | `test/.../router_chat_controller_test.dart` (22 sites) | The same 5-line `generateWithHistory` stub is repeated ~22 times; extract a `_stubGenerator()` helper |
| 💡 | 🟢 | `pubspec.yaml:59-68` | `ui_kit_library` / `generative_ui` pinned to mutable tags (`v2.31.0`) with `pubspec.lock` gitignored — the same commit can resolve to different third-party code |
| 💡 | 🟡 | `lib/ai/providers/usp_command_provider.dart:148-164` | All `requiresConfirmation: true` commands are still commented out, so the whole confirm/batch path is currently dead code — green tests are not real-device validation |
| 💡 | 🟢 | `router_chat_controller.dart:47`; `router_assistant_view.dart:67, 109` (pre-existing) | Art. XII §12.1: conversation state lives in a `ChangeNotifier` constructed directly by the View rather than a Riverpod provider; this PR deepens the deviation by +240 lines |
| 💡 | 🟡 | `router_chat_controller.dart:327, 335` (pre-existing) | Unlocalized error strings with interpolated `$e` are rendered straight into the UI (Art. XIII forbids showing `'$e'`) |

<sub>Confidence: 🟢 High = read head code, evidence quoted · 🟡 Med = file:line + reasoning, not fully substantiated · ⚪ Low = speculative.</sub>
<sub>`[X]` = raised by 1 agent · `[XX]` = independently raised by both agents.</sub>

<details>
<summary>🔴 Critical — evidence chain</summary>

### C1. `_flushPendingResults()` in the tool loop is not generation-guarded

**Location:** `lib/ai/orchestrator/router_chat_controller.dart:312` (added in this PR)

**Code (head):**
```dart
// router_chat_controller.dart:285-317
          _log('_processConversation: executing tool "${toolUse.name}"');
          final result = await _executeToolUse(toolUse);   // ← long await, real device call
          ...
            _pendingResults.add(ToolResultPart(            // 295
              toolUseId: toolUse.id,
              result: result.data,
              isError: result.isError,
            ));
          }
        }

        if (deferred != null) {
          _log('_processConversation: waiting for user confirmation');
          _pendingConfirmation = deferred;                 // 307 ← re-set into the NEW conversation
          notifyListeners();
          return;
        }

        _flushPendingResults();                            // 312 ← UNCONDITIONAL, no generation check
```

```dart
// router_chat_controller.dart:639-650
  void clearConversation() {
    ...
    _state = ConversationState.initial();
    // Invalidate any in-flight confirmation so it cannot inject its result into
    // the new conversation once the device responds.
    _conversationGeneration++;                             // 645
    _pendingConfirmation = null;
    _pendingResults.clear();
```

```dart
// router_chat_controller.dart:471-485  ← the guard exists, but only confirm/cancel call it
  bool _completeBatch(int generation) {
    if (generation != _conversationGeneration) { ... return false; }
    _flushPendingResults();
    return true;
  }
```

```dart
// lib/page/ai_assistant/views/router_assistant_view.dart:396-400
          IconButton(
            icon: const Icon(Icons.refresh),
            onPressed: _controller?.clearConversation,     // not disabled while loading
            tooltip: loc(context).clearConversation,
          ),
```

**Why it is a bug / trigger sequence:**
1. `sendMessage()` → LLM returns `tool_use` blocks; the assistant message is appended to state.
2. The loop is inside `await _executeToolUse(...)` (line 286) — a real USP device call, seconds long.
3. The user taps the AppBar refresh icon (`router_assistant_view.dart:398`, not disabled during loading) → `clearConversation()` resets `_state` to `initial()` and bumps `_conversationGeneration`.
4. The await returns. Line 295 re-populates `_pendingResults`; line 312 flushes them into the now-empty state — or, if a confirmation was deferred, line 307 re-arms `_pendingConfirmation` in the new conversation and `confirmPendingAction()` (line 434) then reads the *current* generation, so its own `_completeBatch()` check passes trivially.
5. Result: the first message of the fresh conversation is a `tool_result` with no preceding `tool_use`. Line 316 sets `loading()` and the next `generateWithHistory` call is rejected — exactly the failure the code comment at lines 473-475 describes ("break every later request"). Every subsequent request stays broken until another clear.

**Suggested fix:** snapshot the generation once at the top of `_processConversation()` and early-return after every `await` if it went stale (lines 244, 251, 295, 303-309, 312); route the loop's flush through `_completeBatch(generation)` so the invariant lives in exactly one place; store the generation inside `PendingConfirmation` so `confirmPendingAction()` validates against the batch's generation rather than the current one. A regression test "clearConversation while a read tool is in flight" would cover this — the existing test at `router_chat_controller_test.dart:529-554` only covers the confirmation-in-flight path.

</details>

<details>
<summary>⚠️ Warnings — details</summary>

**W1 — `retry()` after a partially-failed batch that kept a confirmation** (`:356-364`, `:604-636`, `:260`)
```dart
  void _abandonBatch(PendingConfirmation? deferred) {
    if (deferred != null) {
      _pendingConfirmation = deferred;
      return;                      // results deliberately kept, not flushed
    }
    _flushPendingResults();
  }
```
State after this is `messages = [user, assistant(tool_use A, B)]`, `_pendingResults = [A's real device data]`, `_pendingConfirmation = B`, `hasError = true`. The error banner with Retry (`router_assistant_view.dart:406`, `:670`) is rendered at the same time as the confirmation dialog. If Retry is pressed, `retry()` sees a trailing *assistant* message → `removedUserTurn == false` → `_processConversation()` → line 260 `_pendingResults.clear()` drops A's data, and both `tool_use` blocks go unanswered → API rejection; `_pendingConfirmation` is also left dangling. Mitigating factor: the dialog is `barrierDismissible: false` and covers the button, so this is closer to a broken invariant than a high-frequency path — hence Warning, not Critical. Suggested fix: handle `_pendingConfirmation != null` explicitly at the top of `retry()`, and only clear `_pendingResults` when a batch is genuinely being opened.

**W2 — `_insertPendingToolResults()` does not invalidate in-flight confirmations** (`:540-544`)
```dart
  void _insertPendingToolResults() {
    _pendingConfirmation = null;
    _pendingResults.clear();      // semantically "abandon this batch", but generation not bumped
```
Only `clearConversation()` increments `_conversationGeneration`. An in-flight `confirmPendingAction()` therefore still satisfies `_completeBatch(generation)` when the device replies, emitting a second `tool_result` for a `tool_use_id` this method already answered with `not_executed` — the same duplicate/orphan corruption as C1. Reachable from `sendMessage()` (`:162`) and `handleToolAction()` (`:520`); note `_handleA2UIAction` (`router_assistant_view.dart:631-649`) has no `isLoading` guard. Fix: `_conversationGeneration++` here too, and validate against the generation captured when the batch was opened.

**W3 — `_abandonedToolResult()` fallback lies on the unsafe side** (`:594-601`)
```dart
  Map<String, dynamic> _abandonedToolResult(String toolName) {
    if (_commandsByName?.containsKey(toolName) ?? false) {
      return {'status': 'not_executed', ...};
    }
    return {'status': 'rendered', 'component': toolName};
  }
```
`_commandsByName` is a lazy cache populated only by `_findCommand()` (`:343-349`) and reset to `null` by `clearConversation()` (`:650`). Whenever it is null, an unexecuted **write** command falls through to `{'status':'rendered'}` — telling the model the operation completed. That is precisely what the doc comment at `:589-593` and the test at `:608-632` exist to prevent, so the fallback is a re-entry point for the very bug being fixed. Fix: make this async and consult `_findCommand()`, or invert the fallback so "unknown ⇒ not executed".

**W4 — `handleToolAction()` can emit two adjacent `tool_result` messages** (`:515-531`)
```dart
  Future<void> handleToolAction(ToolActionOutput action) async {
    _insertPendingToolResults();                     // may emit a toolResults message itself
    final toolResultMessage = action.toChatMessage();
    _state = _state.withMessage(toolResultMessage);   // second tool_result message
```
If the trailing assistant turn contains an unanswered component-style `tool_use`, `_insertPendingToolResults()` answers it with `{'status':'rendered'}` and `action.toChatMessage()` then answers the *same* id again — violating this PR's own "one tool_result message per turn" invariant. Marked 🟡 because `ToolActionOutput.toChatMessage()` lives in the external `generative_ui` package and could not be read to confirm the id. The existing test (`:278-301`) exercises this on an empty conversation, where `_insertPendingToolResults()` early-returns, so the path is untested. Fix: route the action's result through `_pendingResults` + a single flush, and exclude `action.toolUseId` from the unanswered set.

**W5 — `_ToolExecutionResult` permits illegal state combinations** (`:668-683`)
`requiresConfirmation: bool` and `pendingConfirmation: PendingConfirmation?` are independent optional fields with no enforced invariant. Line 291 assigns the nullable value to `deferred`; if any future branch sets the flag without the payload, line 303's null check fails, the batch flushes, and the user's write request vanishes with no result, assert or log. Fix: a sealed class / named factories (`.completed(...)` / `.needsConfirmation(PendingConfirmation)` with a non-nullable field) so the type enforces the invariant.

**W6 — duplicated command-catalog caching** (`:140-152` vs `:343-349`)
`_getRouterTools()` and `_findCommand()` each `await _commandProvider.listCommands()` and maintain their own cache (`_routerTools` / `_commandsByName`), invalidated separately at `:649-650`. Two identical queries on the first turn, and any future desync between the two invalidation points makes the tool list and the name lookup drift apart. The comment at `:341-342` claims they are cached together, but they are not. Fix: a single `Future<List<RouterCommand>> _commands()` from which both are derived.

**W7 — raw exception text forwarded to the LLM** (`:458-463`, pre-existing line; the surrounding `_completeBatch` calls are new)
```dart
    } catch (e) {
      _pendingResults.add(ToolResultPart(
        toolUseId: pending.toolUse.id,
        result: {'error': e.toString()},
        isError: true,
      ));
```
`catch (e)` is the correct choice here (USP errors are raw Strings, so `on Exception catch` would silently miss them — see ✅). The issue is that `e.toString()` bypasses `mapUspErrorToServiceError()` (constitution Art. XIII / `constitution.md:1096`): the raw USP error string is placed into the conversation history sent to the model, and the model loses the structured error semantics (e.g. `SessionTokenExpiredError` should trigger re-auth, not model guesswork). Fix: map to `ServiceError` in the command provider and put a sanitized code/message into the `ToolResultPart`.

**W8 — inline test fixtures instead of a Test Data Builder** (`test/.../router_chat_controller_test.dart:30-51`, `670-705`)
Art. I §1.6.2 requires builders under `test/mocks/test_data/[feature]_test_data.dart` (`[Feature]TestData`, named params, `copyWith()`), a convention this repo already follows in four existing files. The `RouterCommand` fixtures and the `_textResponse` / `_toolUseResponse` / `_multiToolUseResponse` helpers are file-private and cannot be reused by other AI tests.

</details>

<details>

- **Correct exception-catching type throughout.** Every new/modified catch uses `catch (e)` (`:324`, `:418`, `:458`), never `on Exception catch`, so raw-String USP errors are not silently dropped. A repo-wide grep for `on Exception catch` in `lib/` returns zero hits.
- **`retry()` no longer deletes `tool_result` messages.** `:608-619` explicitly distinguishes `role == user && !isToolResult`, fixing the old behaviour that destroyed device data and orphaned `tool_use` blocks — with a test to match.
- **Confirmation dialog is now surfaced only after all sibling tools finish** (`:303-309`), eliminating the re-entrancy race where confirming re-entered `_processConversation()` mid-loop. Proven with a `Completer` gate test (`test:447-485`).
- **`_abandonedToolResult()` distinguishes unexecuted commands from rendered components** (`:591-601`) so the model is never told a write succeeded — a genuinely thoughtful safety consideration.
- **Documentation explains "why", not "what".** The class doc (`:35-46`), `_pendingResults` (`:75-81`), `_completeBatch` (`:471-475`) and `_abandonedToolResult` (`:589-593`) all record the API constraint and the failure mode, which is high value for future maintainers (Art. X §10.1).
- **Strong test intent.** 11 new behaviour-named tests with `reason:` strings explaining *why* each invariant matters (`test:518, 553, 583, 604`), covering parallel reads, confirm/cancel, partial failure, clear-during-confirmation and abandoned confirmations.
- **Mocktail used correctly** — `MockConversationGenerator extends Mock implements IConversationGenerator` mocks the abstraction, with `registerFallbackValue` in place.
- **Layering is clean.** No `generated/` imports anywhere in `lib/ai/orchestrator/`; the USP Provider four hard rules and `PreservableContract` are untouched (Art. IV, Art. V §5.4). `RouterCommand` / `RouterCommandResult` implement `Equatable` with complete `props` (Art. XI).
- **The temporary `dependency_overrides` block was removed** before this head commit — an easy thing to forget.
- **No hardcoded secrets or tokens**; new logging records tool and command names only, never `params` (so passwords are not logged).

</details>

AustinChangLinksys (dismissed):
@/tmp/pr1199_review.md

AustinChangLinksys (approved):
🤖 **Automated Review — Round 1** · `80e3d17..0245aa8` (full)

**Verdict: ✅ APPROVE** — Core batching fix is correct and necessary; all three API-contract violations addressed with 11 targeted tests. Three structural warnings worth tracking.

#### Standards

|---|---|---|---|
| ⚠️ | High | `router_chat_controller.dart:~141,~344` | [Duplicated Code] `_getRouterTools` and `_findCommand` each call `listCommands()` independently; first turn triggers two provider calls |
| ⚠️ | High | `router_chat_controller.dart:~356–364` | [Divergent Change] `_abandonBatch`: when `deferred != null` keeps `_pendingResults` without flush (silent implicit contract); when null flushes — two structurally opposite paths |
| ⚠️ | High | `router_chat_controller.dart:~533–587` | [Mysterious Name] `_insertPendingToolResults` silently also clears `_pendingConfirmation` and `_pendingResults` — side-effects hidden from callers |
| 💡 | Med | `router_chat_controller.dart:~197` | [Mysterious Name] local var `deferred` — rename to `pendingWriteConfirmation` for clarity |
| 💡 | High | `test:~503,~571` | `NetworkException` source undocumented; its `GenUiException` vs non-`GenUiException` hierarchy affects which catch branch runs — add a comment |
| 💡 | Med | `router_chat_controller.dart:~356–364` | [Data Clumps] `_pendingResults`+`_pendingConfirmation` cross-method implicit contract undocumented |
| 💡 | Med | test file | `cancelPendingAction`+`clearConversation` race path untested (only `confirm`+`clear` is covered) |

#### Spec

|---|---|---|---|
| ⚠️ | High | `router_chat_controller.dart:retry()` | Scope creep: `retry()` logic rewritten (no longer removes flushed tool results) — not in AC; changes behavior when failure lands post-flush |
| ⚠️ | High | `router_chat_controller.dart:handleToolAction` | Scope creep: `handleToolAction()` now calls `_insertPendingToolResults()` first — not in AC; silently discards pending confirmation when A2UI action arrives |
| ⚠️ | Med | `router_chat_controller.dart:~318–329` | Error-path: `_abandonBatch(deferred)` + `_state.withError(...)` simultaneously; UI can have `hasPendingConfirmation==true` AND `hasError==true`; test checks former only |
| 💡 | High | `router_chat_controller.dart:~594–602` | Scope creep (minor): `_abandonedToolResult` changes `rendered` to `not_executed` for commands — not in AC, semantically correct |
| 💡 | Med | `router_chat_controller.dart:~339–349` | Scope creep (minor): `_findCommand` cache added — not in AC, low risk |

*[X] = 1 agent · [XX] = 2 agents flagged independently (Critical only).*

<details>

**W-1 — [Duplicated Code] Double `listCommands()` on first turn (High)**
`router_chat_controller.dart:~141–151` (`_getRouterTools`) and `~343–348` (`_findCommand`). Both caches (`_routerTools` / `_commandsByName`) derive from the same `_commandProvider.listCommands()`. On turn 1: `_getRouterTools()` fires first call; then the tool loop calls `_findCommand()` with `_commandsByName == null`, firing a second call. Cache hits on all subsequent turns — correctness unbroken, structural debt.
**Fix:** Single `_warmCommandCache()` that populates both, or derive `_commandsByName` from `_routerTools`.

---

**W-2 — [Divergent Change] `_abandonBatch` asymmetric paths (High)**
`router_chat_controller.dart:~356–364`. `deferred != null`: sets `_pendingConfirmation`, returns without touching `_pendingResults` (intentional — `_completeBatch` will flush later). `deferred == null`: calls `_flushPendingResults()`. Cross-method contract is implicit and undocumented.
**Fix:** Add comment on the `deferred != null` branch explaining that `_pendingResults` is intentionally preserved for `confirmPendingAction`/`cancelPendingAction`.

---

**W-3 — [Mysterious Name] `_insertPendingToolResults` hidden side-effects (High)**
`router_chat_controller.dart:~533–587`. Also (1) clears `_pendingConfirmation`, (2) clears `_pendingResults`. Callers see these undocumented effects.
**Fix:** Rename to `_closeStaleBatchAndFillUnansweredToolUses()` or split into two methods.

---

**Spec W-1 — `retry()` scope creep (High)**
Not in issue #1191 AC. Old behavior: removes last user message. New: excludes tool-result messages from removal, resumes exchange if tools already ran. Behavior-changing scope; risk if `!trailing.isToolResult` guard has an edge case.

---

**Spec W-2 — `handleToolAction` scope creep (High)**
Not in AC. Now calls `_insertPendingToolResults()` first, which silently discards a pending confirmation. PR body notes `handleToolAction` is currently unreachable — but once wired, this discard is silent.

---

**Spec W-3 — Dual state: error + pending confirmation (Med)**
`router_chat_controller.dart:~318–329`. After `_abandonBatch(deferred)`, `_state.withError(...)` fires. Controller has both `hasPendingConfirmation==true` and `hasError==true`. Test only verifies `hasPendingConfirmation`; if the UI hides the confirmation dialog on error, user's write request evaporates silently.

</details>

<details>

- **Core batching:** `_pendingResults` → `_flushPendingResults()` → single `ChatMessage.toolResults([...])` cleanly resolves the API violation.
- **Generation counter:** `_conversationGeneration` in `_completeBatch` prevents stale confirmations injecting into a cleared conversation — tested.
- **Deferred confirmation surfacing:** dialog shown only after the full tool loop, not mid-loop — tested with a `Completer` gate.
- **Multiple confirmation-required tools:** first executed, rest answered `not_executed` — tested.
- **11 new tests:** comprehensive coverage of all three original AC scenarios plus edge cases.
- **pubspec.yaml v2.31.0 bump:** clean, no local overrides remain.

</details>

lib/ai/orchestrator/router_chat_controller.dart
```diff
@@ -276,14 +383,64 @@ class RouterChatController extends ChangeNotifier {
...
+  ///
+  /// Work in flight across an `await` must check this before touching state:
+  /// writing into a conversation it no longer belongs to would leave a
+  /// `tool_result` with no matching `tool_use` and invalidate every later
+  /// request.
+  bool _isStale(int generation) {
+    if (generation == _conversationGeneration) return false;
+    _log('_isStale: conversation moved on (batch gen $generation, '
+        'current $_conversationGeneration) — abandoning this exchange');
+    _pendingResults.clear();
```
PeterJhongLinksys: 🔴 **Critical [XX] · 🟢 High** — `_isStale()` clears the *shared* `_pendingResults` (`:96`) after deciding it is the stale generation, but the code that bumps the generation (`clearConversation:735`, `_closeOutOpenBatch:610`) already empties the list as it bumps — so whatever is in the buffer here belongs to the **new** exchange. Trigger: gen 0 parked in `await _executeToolUse` (`:324`) → user taps refresh (`router_assistant_view.dart:398`, no `isLoading` guard) → composer unlocks (`:223` only checks `isLoading`) → gen 1 starts accumulating results → gen 0 returns and wipes them → `_flushPendingResults()` (`:424`) sees an empty list and emits nothing → gen 1s `tool_use` blocks are unanswered → the next `generateWithHistory` is rejected and every later request stays broken. Same shared-buffer hazard via `confirmPendingAction:505-512` / `cancelPendingAction:561-568`, which `add()` before validating the generation. Fix: make `_isStale()` a pure predicate and give the batch an owner (`_Batch { generation, results }`) so a stale exchange cannot reach the live buffer.

lib/ai/orchestrator/router_chat_controller.dart
```diff
@@ -403,67 +590,155 @@ class RouterChatController extends ChangeNotifier {
...
+  /// action. Leaving a `tool_use` unanswered makes the API reject the next
+  /// request, so any gap in the trailing assistant turn is filled.
+  ///
+  /// Results already computed for that turn are **preserved** — they hold real
+  /// device data — and only the genuinely missing ids are synthesised. The
+  /// generation is bumped so a confirmation still awaiting the device cannot
+  /// come back and answer an id this method just answered.
+  void _closeOutOpenBatch() {
+    final hadOpenWork =
+        _pendingConfirmation != null || _pendingResults.isNotEmpty;
```
PeterJhongLinksys: ⚠️ **Warning [XX] · 🟢 High** — `hadOpenWork` is the wrong proxy for "an exchange is in flight". `confirmPendingAction()` nulls `_pendingConfirmation` at `:490` *before* its `await` (`:495`) and `_pendingResults` is empty at that point, so both flags read false and the generation is **not** bumped. `handleToolAction()` (reachable with no loading guard via `router_assistant_view.dart:649`) then appends messages and runs a new exchange; when the device replies, `_completeBatch(generation)` (`:512`) compares equal and flushes a `tool_result` for the old turn after the new messages — an orphan again. Same while `_processConversation()` awaits `generateWithHistory` (`:232`). The conditional bump buys no correctness: bump unconditionally (or invalidate the batch object).

lib/ai/orchestrator/router_chat_controller.dart
```diff
@@ -108,18 +150,28 @@ class RouterChatController extends ChangeNotifier {
...
 
-  /// Build router tools from command provider.
+  /// Load the command registry once per conversation.
+  ///
+  /// Both the tool definitions sent to the model and the per-name lookup are
+  /// derived here so the provider is queried once, and the two views cannot
+  /// drift apart. The future itself is cached so concurrent first calls share
+  /// one query.
+  Future<Map<String, RouterCommand>> _loadCommands() {
+    return _commandsLoad ??= _commandProvider.listCommands().then((commands) {
```
PeterJhongLinksys: ⚠️ **Warning · 🟢 High** — `??=` caches a **rejected** future: it is assigned before completion and there is no error reset, so one `listCommands()` failure is replayed on every later call until `clearConversation()` (`:741`). The previous code retried per call — this is a behavioural regression. Compounding it, `:213`s `await _getRouterTools()` sits *outside* the `try` at `:230`, so the throw escapes `_processConversation()` → `sendMessage()` (awaited without try/catch at `router_assistant_view.dart:226`) and `_state` stays on the `loading()` set at `:194`: permanent spinner, no error banner, Retry unreachable (it needs `hasError`, `:686`). Fix: reset `_commandsLoad = null` in an error callback and rethrow, and bring `:213` inside error handling.

PeterJhongLinksys (commented):
🤖 **Automated Review — Round 2** · `0245aa8..790e12f` (incremental)

**Verdict: 💬 Comment** — Round 1's two Criticals and six Warnings are genuinely fixed and well tested; but promoting the guard to a "whole-exchange invariant" introduced a new one: `_pendingResults` is still a single un-owned shared buffer, and `_isStale()` now *clears* it from the losing side. Both agents independently reached this conclusion.

|---|---|---|---|
| 🔴 `[XX]` | 🟢 | `router_chat_controller.dart:396-402` | `_isStale()` clears the **shared** `_pendingResults` after deciding it is the stale generation — it can wipe the *new* exchange's real results, leaving `tool_use` blocks unanswered (the mirror image of the bug this PR fixes) |
| ⚠️ `[XX]` | 🟢 | `router_chat_controller.dart:604-611` | `hadOpenWork` is the wrong proxy for "an exchange is in flight": while `confirmPendingAction()` or `generateWithHistory()` is awaiting, both flags are empty ⇒ generation is **not** bumped ⇒ the guard passes trivially |
| ⚠️ | 🟢 | `router_chat_controller.dart:159-170`, `213` | `??=` caches a **rejected** future: one `listCommands()` failure sticks until `clearConversation()`; and `:213` sits outside the `try`, so the throw escapes into UI stuck on `loading()` with no error banner |
| ⚠️ | 🟢 | `router_chat_controller.dart:159-176` | `_loadCommands()` returns one view while side-effecting another (`_routerTools!`); the non-null assert relies on microtask ordering, not the type |
| ⚠️ | 🟡 | `router_chat_controller.dart:575-591`, `646-660` | Round 1 W4 unaddressed and not listed under "Not taken": `_closeOutOpenBatch()` may answer a component `tool_use`, then `action.toChatMessage()` answers the same id again — two adjacent `tool_result` messages |
| ⚠️ | 🟡 | `router_chat_controller.dart:340-346` | New per-tool catch puts raw `e.toString()` into history sent to the LLM — bypasses `mapUspErrorToServiceError()` (Art. XIII); a raw USP error string may carry internal detail, and the model loses structured error semantics |
| ⚠️ | 🟢 | `router_chat_controller.dart:9-25` | `PendingConfirmation` is UI-visible state (read at `router_assistant_view.dart:138`) but does not implement `Equatable` — Art. XI; this round added its 4th field |
| ⚠️ | 🟡 | `router_chat_controller.dart:409-417` | `_abandonBatch()` takes `generation` but ignores it on the `deferred != null` path — a stale `PendingConfirmation` can be re-armed into a conversation it no longer belongs to |
| 💡 | 🟢 | `router_chat_controller.dart:507-509` | `{...result.data, 'status': _statusExecuted}` silently overwrites a `status` field the device may itself return (common in USP/JNAP payloads); prefer nesting under `'result'` |
| 💡 | 🟢 | `router_chat_controller.dart:757-761`, `671-682` | `debugAbandonedToolResultFor()` opens a `@visibleForTesting` seam on a production class to cover a branch that `:213`'s unconditional `_loadCommands()` makes near-unreachable — tests an implementation detail |
| 💡 | 🟢 | `router_chat_controller.dart:692-696` | `retry()` declines correctly but silently: the empty `notifyListeners()` changes nothing, so the Retry button appears dead. Disable it (or surface a localized hint) instead |
| 💡 | 🟢 | `router_chat_controller.dart:439-461` | `[Data Clump]` `_executeToolUse()` never uses `generation`; it only stamps it into `PendingConfirmation` — let the caller (which already holds it) assemble that |
| 💡 | 🟢 | `router_chat_controller.dart:207-384` | `[Long Method]` grew again to ~178 lines, nesting `while→try→for→try→if`. Agree with deferring, but worth extracting `_runToolBatch()` in the follow-up |
| 💡 | 🟢 | `router_chat_controller.dart:214, 237, 304, 325, 359, 512, 568` | The batch/generation invariant is enforced by 7 hand-written checkpoints; C1 and W1 are exactly what a missed/mis-conditioned checkpoint looks like — make it a type constraint |
| 💡 | 🟢 | `test/ai/orchestrator/router_chat_controller_test.dart:535, 561, 587, 623, 641` | `[Duplicated Code]` the 5 new tests each re-paste the same 5-line `generateWithHistory` stub (~27 sites file-wide). Distinct from the Test Data Builder point you (fairly) declined |
| 💡 | 🟡 | `router_chat_controller.dart:671-682` | The `rendered` branch may now be dead: A2UI arrives via text rendering (`router_assistant_view.dart:577`), and unknown `tool_use` names are already answered at `:446-449`. Worth confirming, then deleting |
| 💡 | 🟢 | `router_assistant_view.dart:398`, `631-649` | Neither the refresh button nor `_handleA2UIAction` has an `isLoading` guard, unlike `_sendMessage` (`:223`, `:705`) — these are the entry points for C1 and W1 |

<sub>Confidence: 🟢 High = read head code, evidence quoted · 🟡 Med = file:line + reasoning, not fully substantiated · ⚪ Low = speculative.</sub>
<sub>`[X]` = raised by 1 agent · `[XX]` = independently raised by both agents.</sub>

<details>
<summary>🔴 Critical — evidence chain</summary>

### C1. `_isStale()` clears the shared `_pendingResults`, so the stale exchange can wipe the live one's results

**Location:** `lib/ai/orchestrator/router_chat_controller.dart:396-402` (added this round — `diff1199 :246`)

**Code (head):**
```dart
// router_chat_controller.dart:396-402
  bool _isStale(int generation) {
    if (generation == _conversationGeneration) return false;
    _log('_isStale: conversation moved on (batch gen $generation, '
        'current $_conversationGeneration) — abandoning this exchange');
    _pendingResults.clear();     // ← clears whatever is in the shared buffer *now*
    return true;
  }
```

```dart
// router_chat_controller.dart:96  ← one buffer, shared by every exchange, no owner tag
  final List<ToolResultPart> _pendingResults = [];
```

```dart
// router_chat_controller.dart:604-616  the bumper empties the list as it bumps,
// so by the time _isStale() reaches its clear(), the contents belong to the NEW generation
  void _closeOutOpenBatch() {
    final hadOpenWork =
        _pendingConfirmation != null || _pendingResults.isNotEmpty;
    _pendingConfirmation = null;
    if (hadOpenWork) {
      _conversationGeneration++;
    }
    final carried = List<ToolResultPart>.of(_pendingResults);
    _pendingResults.clear();
```

```dart
// router_chat_controller.dart:423-424  an emptied batch flushes nothing at all
  void _flushPendingResults() {
    if (_pendingResults.isEmpty) return;
```

```dart
// lib/page/ai_assistant/views/router_assistant_view.dart — the unguarded entry points
223:    if (text.isEmpty || _controller?.isLoading == true) return;   // guarded
398:            onPressed: _controller?.clearConversation,            // NOT guarded
611:                        onAction: _handleA2UIAction,              // NOT guarded
631:  void _handleA2UIAction(Map<String, dynamic> data) {
```

**Why it is a bug / trigger sequence:**
1. Gen 0: `sendMessage()` → the LLM returns several `tool_use` blocks → the loop is parked in `await _executeToolUse(...)` (`:324`), a real USP device call lasting seconds. `_pendingResults` holds gen 0's first result.
2. The user taps AppBar refresh (`router_assistant_view.dart:398`, not disabled while loading) → `clearConversation()` resets state and bumps the generation (`:735`). `_state.isLoading` is now false, so the composer unlocks (`:223`, `:705` only check `isLoading`).
3. The user sends a new message immediately → gen 1's `_processConversation()` runs, appends its assistant `tool_use` turn, and starts accumulating into the **same** `_pendingResults` (`:334`).
4. Gen 0's device call returns → `if (_isStale(generation)) return;` (`:325`) → the branch is taken and `_pendingResults.clear()` **discards gen 1's real results**.
5. Gen 1's loop finishes → `_completeBatch(1)` (`:359`) → generation matches, but the buffer is empty, so `_flushPendingResults()` (`:424`) returns without emitting anything → gen 1's assistant turn has `tool_use` blocks with **no** `tool_result`.
6. The loop sets `loading()` (`:363`) and the next `generateWithHistory` goes out with unanswered `tool_use` blocks → API rejection, and every later request stays broken until another clear — precisely the failure mode documented at `:392-395`.

The same shared-buffer hazard has a second entry: `confirmPendingAction()` (`:505-512`) and `cancelPendingAction()` (`:561-568`) `_pendingResults.add(...)` **before** validating the generation, so a stale confirmation first pollutes the live batch and then `_completeBatch(staleGen)` clears the whole thing.

**Suggested fix:** give the batch an owner instead of guarding a global. Make `_isStale()` a pure predicate (no writes — command/query separation) and move ownership into the batch:
```dart
class _Batch {
  final int generation;
  final List<ToolResultPart> results = [];
  _Batch(this.generation);
}
_Batch? _batch;   // replaces _pendingResults + the scattered generation compares
```
Then a stale exchange physically cannot touch the live batch, `confirm`/`cancel` must check before they `add`, and W1/S6 dissolve with it. Regression test worth adding: "clearConversation + a new message while a read tool is in flight ⇒ every `tool_use` in the new turn is answered".

</details>

<details>
<summary>⚠️ Warnings — details</summary>

**W1 — `hadOpenWork` is invisible to in-flight work** (`:604-611` vs `:486-495`)
```dart
    final pending = _pendingConfirmation;
    if (pending == null) return;
    final generation = pending.generation;
    _pendingConfirmation = null;      // ← cleared BEFORE the await
    _state = _state.loading();
    notifyListeners();
    try {
      final result = await _commandProvider.execute(...);   // ← long await; list is empty
```
During that await `_pendingConfirmation == null` and `_pendingResults.isEmpty`, so `hadOpenWork == false` and `_closeOutOpenBatch()` does **not** bump. `handleToolAction()` (reachable with no loading guard via `router_assistant_view.dart:649`) then appends its own message and runs `_processConversation()`; when the device replies, `_completeBatch(generation)` (`:512`) compares equal and flushes a `tool_result` for the old turn *after* the new messages — an orphan again. The same holds while `_processConversation()` awaits `generateWithHistory` (`:232`). The conditional bump buys no correctness; unconditional `_conversationGeneration++` (or `_batch = null` under C1's fix) closes it. Also add the `isLoading` guard to `_handleA2UIAction`.

**W2 — a rejected future is cached forever, and the throw escapes the try** (`:159-170`, `:173-176`, `:213`)
```dart
  Future<Map<String, RouterCommand>> _loadCommands() {
    return _commandsLoad ??= _commandProvider.listCommands().then((commands) { ... });
  }
```
`??=` assigns before completion and there is no `onError` reset, so a single `listCommands()` failure (USP session/permission error) is replayed on every later call until `clearConversation()` (`:741`). Previously each call retried — this is a behavioural regression introduced this round. Worse, `:213`'s `await _getRouterTools()` sits **outside** the `try` at `:230`, so the error escapes `_processConversation()` → `sendMessage()` (awaited without try/catch at `router_assistant_view.dart:226`) → unhandled async error, `_state` frozen on the `loading()` set at `:194`: a permanent spinner, no banner, and Retry unreachable since it needs `hasError` (`:686`). Fix: reset `_commandsLoad = null` in an error callback and rethrow; bring `:213` inside error handling.

**W3 — `_loadCommands()` returns one view and side-effects another** (`:159-176`) — suspected *Side Effects / Temporal Coupling*
It is typed `Future<Map<String, RouterCommand>>` but also writes `_routerTools`, so `_getRouterTools()` awaits a map and then asserts `_routerTools!`. The two fields plus the future are kept in sync only by hand in `clearConversation()` (`:739-741`), and the `!` is safe only because an in-flight `.then` (`:161`) happens to run before the awaiting continuation. That is a microtask-ordering coincidence, not a type guarantee — and it is the same drift risk as Round 1's W6 in a new shape. Return a record `({List<GenTool> tools, Map<String, RouterCommand> byName})` and drop both the mutable field and the `!`.

**W4 — `handleToolAction()` can still emit two `tool_result` messages** (`:575-591`, `:646-660`)
```dart
  Future<void> handleToolAction(ToolActionOutput action) async {
    _closeOutOpenBatch();                              // may synthesise a tool_result itself
    final toolResultMessage = action.toChatMessage();
    _state = _state.withMessage(toolResultMessage);     // second tool_result message
```
If the trailing assistant turn holds an unanswered component-style `tool_use` — exactly what the A2UI action is answering — `_closeOutOpenBatch()` answers it `{'status': rendered}` at `:646-653` and `:583` answers the same id again, breaking this PR's own "one `tool_result` message per turn" invariant. 🟡 because `ToolActionOutput.toChatMessage()` lives in the external `generative_ui` package and the id could not be confirmed here. Note also that `toolUseId` comes from model-authored A2UI JSON (`router_assistant_view.dart:634`) with only a null check, never matched against a real `ToolUseBlock.id`. Fix: route the action through `_pendingResults` + a single flush, exclude `action.toolUseId` from the synthesised set, and validate it against the trailing turn's id set. This item appeared in Round 1 and is in neither the "What changed" nor the "Not taken" table.

**W5 — raw exception text forwarded to the model** (`:340-346`; also `:474-479`, `:516-521`)
```dart
          } catch (e) {
            _log('_processConversation: tool "${toolUse.name}" threw: $e');
            _pendingResults.add(ToolResultPart(
              toolUseId: toolUse.id,
              result: {'error': e.toString()},
              isError: true,
            ));
          }
```
`catch (e)` rather than `on Exception catch` is **correct** (USP errors are raw Strings — a repo-wide grep for `on Exception catch` in `lib/` returns zero hits). The problem is `e.toString()` going into `ToolResultPart` without `mapUspErrorToServiceError()` (Art. XIII; the repo's own pattern in `lib/page/instant_setup/services/pnp_service.dart:82, 94, 126, 177, 300` is `throw mapUspErrorToServiceError(e)`). That payload travels to the external LLM inside `_state.messages`: a raw USP error string may carry internal fields/URIs, and the model loses structured semantics (`SessionTokenExpiredError` should trigger re-auth, not model guesswork). Art. XIII is a pre-existing deviation in this module — noted — but `:340-346` is a *new* third site added by this round. Fix: map to `ServiceError` inside the command provider and place only a sanitized code/message in the result.

**W6 — `PendingConfirmation` is UI-visible state without `Equatable`** (`:9-25`)
It is exposed via the `pendingConfirmation` getter (`:134`) and consumed by the View (`router_assistant_view.dart:138`), and this round gave it a 4th field (`generation`, `:17`). Its siblings in the same feature — `RouterCommand` and `RouterCommandResult` (`lib/ai/abstraction/router_command.dart:8, 43`) — both `extends Equatable`. Art. XI requires UI models and state classes to implement `Equatable`; without it, value comparison and rebuild decisions are unreliable. Fix: `extends Equatable` with `props => [command, params, toolUse, generation]`.

**W7 — `_abandonBatch()` ignores the `generation` it is handed** (`:409-417`)
```dart
  void _abandonBatch(int generation, PendingConfirmation? deferred) {
    if (deferred != null) {
      _pendingConfirmation = deferred;   // generation unused on this path
      return;
    }
    _completeBatch(generation);
  }
```
If the conversation was reset while the batch ran, `deferred` is re-armed into a conversation it no longer belongs to; confirming it then fails `_completeBatch(pending.generation)` and leaves `_state` on the `loading()` set at `:494` with no recovery path (permanent spinner). 🟡 because everything between the assignment and the `return` is synchronous (`_executeToolUse`/`_findCommand` throws are swallowed by the inner try/catch), so reachability is narrow. Fix: `if (!_isStale(generation)) _pendingConfirmation = deferred;`.

</details>

<details>

- **Round 1's first Critical (unguarded loop flush) is genuinely fixed.** `_processConversation()` snapshots the generation at `:211` and re-checks after every await (`:214`, `:237`, `:304`, `:325`); the flush now goes through `_completeBatch(generation)` at `:359`, and the failure paths at `:367`/`:373` carry it too.
- **Round 1's second Critical (confirmation validating the *current* generation) is fixed in the recommended way.** `PendingConfirmation.generation` (`:17`, `:23`, `:459`) is the batch's own generation, and `confirmPendingAction`/`cancelPendingAction` validate against `pending.generation` (`:492`, `:555`) rather than the live counter.
- **The carried-results fix is the highest-quality part of this round.** `_closeOutOpenBatch()` preserves already-computed real device data (`:615`, `:644`) and synthesises **only** genuinely unanswered ids (`:646-653`), with a test asserting the read's `{model: MR7500}` survives while the write reports `not_executed` (`test:584-618`).
- **`_abandonedToolResult()` now fails safe.** `:671-682` inverts the check so only names the registry positively confirms are *not* commands are reported `rendered`; unknown ⇒ `not_executed`, locked in by `test:620-637`. The doc comment states which direction of error is harmful — exactly the right thing to record.
- **`_ToolExecutionResult` illegal states are now unrepresentable.** The `.awaitingConfirmation()` named constructor (`:779-782`) with `requiresConfirmation` as a derived getter (`:786`) makes the two fields incapable of disagreeing — textbook "let the type carry the invariant".
- **The duplicated command caches are merged.** `_loadCommands()` (`:159-170`) is a single source deriving both the `GenTool` list and the name map, caching the *future* so concurrent first calls share one query (`clearConversation:741` clears it too). Residual issues in W2/W3.
- **`retry()` no longer destroys a batch that owns the user's write request** (`:692-696`), covered by `ignores retry while a confirmation is still pending` (`test:639-666`).
- **Protocol magic strings are centralised** (`:72-75`); a `lib/`-wide grep finds no stray literals.
- **A mid-loop throw can no longer orphan a `tool_use`** — per-tool try/catch (`:298-347`) guarantees every id gets a `ToolResultPart` (content concern in W5).
- **Layering is clean.** Zero `generated/` imports anywhere under `lib/ai/` or `lib/page/ai_assistant/` (Art. V §5.4); the USP Provider four hard rules, `PreservableContract` and `uspMutationLockProvider` are untouched (Art. IV). UI Kit components (`AppSurface`, `AppText.body`) are reused rather than re-implemented (Art. XV).
- **No hardcoded secrets.** `'password': 'secret'` appears only in test fixtures; the new logs record tool/command names only — never `toolUse.input` — and the confirmation dialog (`router_assistant_view.dart:159-161`) shows name/description only, so plaintext passwords reach neither UI nor logs.
- **Test intent is strong.** 31 behaviour-named tests, no `TC001`-style numbering, every new assertion carries a `reason:` explaining why the invariant matters (`test:551`, `578`, `614`, `634`, `664`); Mocktail mocks the abstractions (`IConversationGenerator`, `IRouterCommandProvider`) with `registerFallbackValue` in place.
- **Docs explain "why", not "what"** — `_isStale` (`:390-395`), `_closeOutOpenBatch` (`:593-603`), `_abandonedToolResult` (`:664-670`), `retry` (`:688-691`) all record the API constraint and the failure mode (Art. X §10.1).
- **The "Not taken" reasoning holds up** and is not re-raised: LLM-response builders vs. domain fixtures, the deliberate order assertion at `:439`, deferring the long-method split out of a correctness fix, `pubspec.lock`/mutable tags as a pre-existing project decision, Art. XII/XIII deviations tracked separately, and write commands intentionally left undefined in this PR.

</details>

PeterJhongLinksys (commented):
🤖 **Automated Review — Round 2 (addendum)** · `0245aa8..790e12f` (incremental)

**Verdict: 💬 Comment** — Same head SHA as my Round 2 review above; this addendum only adds findings that review missed and **retracts two "✅ looks good" claims it made about the new tests**. The Critical and the verdict are unchanged.

|---|---|---|---|
| ⚠️ `[XX]` | 🟢 | `test/ai/orchestrator/router_chat_controller_test.dart:620-637` | **Retracts my earlier ✅**: this test takes the *opposite* branch from its own comment — `sendMessage('hello')` runs `_loadCommands()` (`:213`→`:168`), so `_commandsByName` is **not null** and contains `setWifiPassword`; the `registry == null` fail-safe path has **zero coverage** |
| ⚠️ `[XX]` | 🟢 | `test/ai/orchestrator/router_chat_controller_test.dart:639-666` | **Retracts my earlier ✅**: `ignores retry while a confirmation is still pending` returns at the *first* guard (`:686 !_state.hasError`) — the deferred path never calls `withError()`, so the new `:692` guard is never executed. Delete `:692-696` and the test still passes; the `call == 2` throw at `:654` is dead code |
| ⚠️ | 🟢 | `test/ai/orchestrator/router_chat_controller_test.dart:529-554`, `556-582` | Both race tests can be satisfied *vacuously*: `messages isEmpty` / `hasPendingConfirmation isFalse` also hold if the flow never reached `execute()` (`Duration.zero` is an implicit assumption, not a synchronisation point). Neither adds `verify(() => mockCommandProvider.execute(any(), any())).called(1)` to pin down that the tool actually ran |
| ⚠️ | 🟢 | `router_chat_controller.dart:470-473` | `RouterCommandResult.failure`'s `error` string is dropped on the read path: `data: result.data` is `const {}` on failure, so the model receives `isError: true` + an **empty map** and cannot know why. Every unavailable read goes through this (`usp_command_provider.dart:317, 342, 397, 423, 469, 490`). `confirmPendingAction` does it right (`:509 result.error ?? 'Operation failed'`) — the two paths disagree inside one class |

<sub>Confidence: 🟢 read head code + evidence · 🟡 file:line + reasoning, not fully proven · ⚪ speculative</sub>
<sub>`[X]` = raised by 1 agent · `[XX]` = independently raised by both agents</sub>

<details><summary>Why the two test retractions matter (evidence)</summary>

**`:620-637` — wrong branch.** The comment claims "No command has been looked up yet, so the registry cache is null", but `sendMessage` unconditionally loads the registry first:

```dart
// router_chat_controller.dart:213
final tools = await _getRouterTools();
// :168  (inside _loadCommands)
return _commandsByName = {for (final c in commands) c.name: c};
// :671-676
final registry = _commandsByName;
final isKnownComponent = registry != null && !registry.containsKey(toolName);
```

At assert time `registry != null` **and** `containsKey('setWifiPassword')` are both true ⇒ `isKnownComponent == false` ⇒ `not_executed`. Green, but via the "known command" branch — duplicating `reports abandoned commands as not executed` (`test:747-771`). Fix: call the seam on a clean controller without `sendMessage`, or stub `listCommands()` to throw.

**`:639-666` — guard never runs.** In this test the write tool sets `deferred` (`:330`, no throw) and the read tool's `StateError('boom')` is swallowed by `_executeToolUse`'s own catch (`:474-479`), so the loop exits at `:350-354` with `_state` still `loading()` from `:287`. `hasError == false` ⇒ `retry()` returns at `:686`. Fix: drive the controller into a real error state *while* a confirmation is pending, then assert `messages` / `_pendingResults` survive `retry()`.

Net effect: the two items my Round 2 review listed as "fixed and locked in by tests" are **fixed in code but untested**.
</details>

<details><summary>Unchanged from Round 2</summary>

The Critical (`_isStale()` clearing the shared `_pendingResults`, `:396-402`) and all other Warnings/Suggestions in the review above stand as written, re-confirmed against head `790e12fe` by both agents this round. Also re-confirmed clean: no `on Exception catch` anywhere in `lib/` (0 hits), no hand-parsing of USP error strings, no `generated/` imports outside `services/` (Art. V §5.4), no hardcoded secrets, new `_log` calls never record `toolUse.input`.
</details>

PR #1200, opened 2026-08-04
feat(cloud): route client-side RA requests through router /cloud proxy on local builds (#1179)

### LINKED ISSUE #1179: Route client-side RA cloud requests through router /cloud proxy on local builds (CORS fix)
## Problem

When the PrivacyGUI web app is served **from the router itself** (local build), the browser origin is the router's LAN IP. Client-side Remote Assistance (RA) requests that hit the cloud host directly are blocked by **CORS**. This is a confirmed, current-production problem.

1.x solved this by injecting a `getHost` closure into `LinksysHttpClient` (via the removed `cloudRepositoryProvider`) so client requests went through the router's `/cloud/` reverse proxy (same-origin). That path is **dead in 2.x**: `LinksysHttpClient.getHost` / `combineUrl` still exist but `combineUrl` is never called, and every caller builds an absolute cloud URL itself, bypassing `getHost`.

## Scope

RA only. **Client-side** RA requests proxy through the router on local builds; **CA-side** requests always go direct to cloud (the CA portal is cloud-hosted — proxying would misroute it). Firmware OTA and geolocation are explicitly out of scope.

## Proposed solution

Add a small, testable host-resolution layer instead of restoring the dead `getHost` path:

- **`kProxyPrefix`** constant (`/cloud`) — single point of change once FW confirms the prefix.
- **`CloudHostResolver`** — decides the URL base to prepend to a Guardian endpoint:
  - CA-side → always direct cloud base.
  - Client-side → router proxy base (`<origin>/cloud`) **iff** local build AND origin resolvable; otherwise direct cloud (safe fallback).
- Inject the resolver into **`GuardianApiClient`** and thread `forCA` through `_buildUrl` / `_request`, derived from `sessionToken != null` so it stays in lock-step with the auth-header branch (no drift possible).
- Host source = **`Uri.base.origin`** (scheme + host + port). Chosen over `window.location.host` because same-origin/CORS requires the scheme to match; `Uri.base.origin` preserves it and is already an established pattern in this repo.

## Acceptance Criteria

- [ ] On a **local build** (`BuildConfig.isLocal() == true`), client-side RA requests (`fetchDeviceToken`, `getSessions`, `getSessionInfo`, `createPin`, `deleteSession`) target `<Uri.base.origin>/cloud/v1/guardians/...`.
- [ ] CA-side requests (`getSessionInfoForCA`, `deleteSessionForCA`) **always** target the cloud host directly, even on a local build.
- [ ] On a **non-local build**, every RA request (client + CA) targets the cloud host directly — behavior byte-for-byte identical to today.
- [ ] When the origin is unresolvable (empty), client-side requests **fall back** to the cloud host rather than failing.
- [ ] `Uri.base.origin` scheme is preserved (http stays http, port preserved) — no forced `https://`.
- [ ] The proxy path prefix is a single configurable constant (`kProxyPrefix`); changing the FW prefix requires no other code change.
- [ ] Existing RA flows (dialog, banner, CA confirm view) work unchanged; `remote_assistance_service_test.dart` stays green.

## Blocked / open items

- **FW proxy prefix unconfirmed** — `kProxyPrefix` is one constant; flip the value when FW lands. This blocks on-device sign-off, **not** implementation. (label: `need-fw-confirm`)
- **E2E mock builds** are local (`isLocal() == true`) → client RA calls take the proxy path; if the E2E harness mocks Guardian at the network boundary, confirm it intercepts `/cloud/...` URLs too.

## Tests

- `test/core/cloud/cloud_host_resolver_test.dart` — resolver logic (7 cases: local+client, local+CA, empty-origin fallback, non-local, port preserved, http scheme preserved, default constructor).
- `test/core/cloud/guardian_api_client_test.dart` — URL assembly (client methods → proxy URL incl. session-id substitution; CA methods → cloud URL under the same resolver).
- `test/core/cloud/services/remote_assistance_service_test.dart` — regression, stays green unchanged.

AustinChangLinksys:
## Summary

Restores the 1.x behavior of routing **client-side** Remote Assistance requests through the router's `/cloud` reverse proxy on local builds, so they stay **same-origin** and avoid the CORS `Failed to fetch` that occurs when the web app is served from the router itself (origin = router LAN IP).

Closes #1179.

## Changes

- **`kProxyPrefix` constant** (`cloud_const.dart`) — single point of change for the FW proxy prefix (`/cloud`).
- **`CloudHostResolver`** (`core/cloud/cloud_host_resolver.dart`) — client-side + local build → router `/cloud` proxy base; CA-side or non-local build → direct cloud base, with a safe fallback to the cloud host when the origin is unresolvable.
- **`GuardianApiClient`** — injects the resolver and threads `forCA` through `_buildUrl`/`_request`, derived from `sessionToken != null` so it stays in lock-step with the auth-header branch. **CA-side requests always target the cloud host directly**, even on a local build (the CA portal is cloud-hosted and must never be proxied).
- **Platform-safe `getCloudOrigin()`** (`core/utils/ip_getter/`) — web returns `window.location.origin`; non-web (mobile/stub) returns `''` so the resolver falls back to the cloud base. `Uri.base.origin` throws a `StateError` on the `file:` URI used by the Dart VM / non-web runtime, so this avoids a non-web crash.

## Compatibility

- **Non-local builds are byte-for-byte identical to today.**
- CA-side flow unchanged — always hits the cloud host directly.

## Verification (see #1179)

- **QA + Production `guardian.tools`** now accept the `/cloud` prefix (no longer return the SPA `index.html`); the Host/prefix coupling is resolved cloud-side.
- **End-to-end through the router FW `/cloud` proxy**, with a real device SN, returns `200 application/json` carrying a real `linksysToken` — same-origin, no redirect, no CORS.

> Note: a local nginx dev-proxy must forward `/cloud` over `https` (`proxy_pass https://<router-ip>; proxy_ssl_verify off;`) — an `http` upstream triggers a cached `308 → https` redirect and a cross-origin failure. This is a local dev-proxy artifact only; it does not affect the app or FW.

## Tests

- `test/core/cloud/cloud_host_resolver_test.dart` — resolver routing + local-build fallback coverage.
- `test/core/cloud/guardian_api_client_test.dart` — client-side vs CA-side URL construction (incl. CA bypasses proxy even on local build).

### DIFF
--- a/lib/core/cloud/cloud_host_resolver.dart
+++ b/lib/core/cloud/cloud_host_resolver.dart
@@ -1,0 +1,47 @@
+import 'package:privacy_gui/constants/_constants.dart';
+
+/// Resolves the URL **base** (scheme + host [+ port], plus proxy prefix when
+/// proxying) to prepend to a Guardian API endpoint.
+///
+/// When the web app is served from the router itself (local build), the browser
+/// origin is the router's LAN address. Client-side Remote Assistance requests
+/// that hit the cloud host directly are blocked by CORS, so they are instead
+/// routed through the router's reverse proxy ([kProxyPrefix]) to stay
+/// same-origin. CA-side requests (support agent, cloud-hosted portal) must
+/// always go direct to the cloud host.
+///
+/// Selection rule:
+/// - CA-side → always direct cloud.
+/// - Client-side → router proxy **iff** this is a local build AND the origin is
+///   resolvable; otherwise direct cloud (safe fallback).
+///
+/// [originGetter] and [isLocal] are injected so the resolver is unit-testable
+/// without a browser. The defaults ([_emptyOrigin] and [BuildConfig.isLocal])
+/// deterministically resolve to the cloud base in the unit-test environment.
+class CloudHostResolver {
+  /// Returns the current browser origin (scheme+host+port), e.g.
+  /// `https://192.168.1.1`. Empty when unavailable (non-web / not resolvable).
+  final String Function() _originGetter;
+
+  /// Whether this build is a local (router-hosted) build.
+  final bool Function() _isLocal;
+
+  CloudHostResolver({
+    String Function()? originGetter,
+    bool Function()? isLocal,
+  })  : _originGetter = originGetter ?? _emptyOrigin,
+        _isLocal = isLocal ?? BuildConfig.isLocal;
+
+  static String _emptyOrigin() => '';
+
+  /// Direct cloud base — always `https://<cloudBase>`.
+  String get _cloudBase => 'https://${cloudEnvironmentConfig[kCloudBase]}';
+
+  /// URL base to prepend to a Guardian endpoint.
+  String resolve({required bool forCA}) {
+    if (forCA || !_isLocal()) return _cloudBase;
+    final origin = _originGetter();
+    if (origin.isEmpty) return _cloudBase; // fallback: never break on no origin
+    return '$origin$kProxyPrefix'; // e.g. https://192.168.1.1/cloud
+  }
+}
--- a/lib/core/cloud/guardian_api_client.dart
+++ b/lib/core/cloud/guardian_api_client.dart
@@ -9,10 +9,11 @@ import 'package:privacy_gui/core/cloud/http/linksys_http_client.dart';
 import 'package:privacy_gui/core/cloud/model/guardians_remote_assistance.dart';
 import 'package:privacy_gui/core/cloud/model/error_response.dart';
 import 'package:privacy_gui/core/utils/extension.dart';
+import 'package:privacy_gui/core/utils/ip_getter/ip_getter.dart';
 import 'package:privacy_gui/core/utils/logger.dart';
 
 final guardianApiClientProvider = Provider((ref) => GuardianApiClient(
-      hostResolver: CloudHostResolver(originGetter: () => Uri.base.origin),
+      hostResolver: CloudHostResolver(originGetter: getCloudOrigin),
     ));
 
 /// API client for Guardian Remote Assistance services.
--- a/lib/core/cloud/guardian_api_client.dart
+++ b/lib/core/cloud/guardian_api_client.dart
@@ -4,13 +4,16 @@ import 'package:flutter_riverpod/flutter_riverpod.dart';
 import 'package:flutter_secure_storage/flutter_secure_storage.dart';
 import 'package:http/http.dart' as http;
 import 'package:privacy_gui/constants/_constants.dart';
+import 'package:privacy_gui/core/cloud/cloud_host_resolver.dart';
 import 'package:privacy_gui/core/cloud/http/linksys_http_client.dart';
 import 'package:privacy_gui/core/cloud/model/guardians_remote_assistance.dart';
 import 'package:privacy_gui/core/cloud/model/error_response.dart';
 import 'package:privacy_gui/core/utils/extension.dart';
 import 'package:privacy_gui/core/utils/logger.dart';
 
-final guardianApiClientProvider = Provider((ref) => GuardianApiClient());
+final guardianApiClientProvider = Provider((ref) => GuardianApiClient(
+      hostResolver: CloudHostResolver(originGetter: () => Uri.base.origin),
+    ));
 
 /// API client for Guardian Remote Assistance services.
 ///
@@ -23,12 +26,25 @@ final guardianApiClientProvider = Provider((ref) => GuardianApiClient());
 /// 2. **CA (support agent)**: Authorization header with session token
 class GuardianApiClient {
   final LinksysHttpClient _http;
+  final CloudHostResolver _hostResolver;
 
-  GuardianApiClient({LinksysHttpClient? httpClient})
-      : _http = httpClient ?? LinksysHttpClient();
+  GuardianApiClient({
+    LinksysHttpClient? httpClient,
+    CloudHostResolver? hostResolver,
+  })  : _http = httpClient ?? LinksysHttpClient(),
+        _hostResolver = hostResolver ?? CloudHostResolver();
 
-  String _buildUrl(String endpoint, {Map<String, String>? args}) {
-    String url = 'https://${cloudEnvironmentConfig[kCloudBase]}$endpoint';
+  /// Builds an absolute URL for a Guardian [endpoint].
+  ///
+  /// [forCA] selects the host base: CA-side requests always target the cloud
+  /// host directly; client-side requests may be routed through the router's
+  /// reverse proxy on local builds. See [CloudHostResolver].
+  String _buildUrl(
+    String endpoint, {
+    Map<String, String>? args,
+    required bool forCA,
+  }) {
+    String url = '${_hostResolver.resolve(forCA: forCA)}$endpoint';
     args?.forEach((key, value) => url = url.replaceFirst(key, value));
     return url;
   }
@@ -65,7 +81,9 @@ class GuardianApiClient {
     }
 
     logger.d('[Guardian] Fetching new device token');
-    final url = Uri.parse(_buildUrl(kDeviceToken)).replace(queryParameters: {
+    // Client-side (device owner) request → may proxy through the router.
+    final url = Uri.parse(_buildUrl(kDeviceToken, forCA: false))
+        .replace(queryParameters: {
       'serialNumber': serialNumber,
       'macAddress': macAddress,
       'uuid': deviceUUID.toUpperCase(),
@@ -213,7 +231,11 @@ class GuardianApiClient {
     String? serialNumber,
     String? sessionToken,
   }) async {
-    final url = Uri.parse(_buildUrl(endpoint, args: args));
+    // CA-side requests authenticate with a session token and must always hit
+    // the cloud host directly; deriving forCA here keeps it in lock-step with
+    // the auth-header branch below (no drift possible).
+    final forCA = sessionToken != null;
+    final url = Uri.parse(_buildUrl(endpoint, args: args, forCA: forCA));
     final headers = Map<String, String>.from(_defaultHeaders);
 
     if (sessionToken != null) {
--- a/lib/constants/cloud_const.dart
+++ b/lib/constants/cloud_const.dart
@@ -47,6 +47,12 @@ const kSessionInfo =
     '/v1/guardians/remote-assistances/sessions/$kVarRASessionId';
 const kCreatePin = '/v1/guardians/remote-assistances/sessions/pin';
 
+/// Path prefix for the router's reverse proxy to cloud (same-origin), used only
+/// for client-side RA requests on a local web build so the browser avoids CORS.
+/// NOTE: exact prefix NOT YET CONFIRMED by firmware — single point of change.
+/// Must NOT end with a trailing slash (endpoints already start with '/').
+const kProxyPrefix = '/cloud';
+
 // Client type id/secret
 final kClientTypeId = clientTypeID;
 final kClientSecret = clientTypeSecret;
--- a/lib/constants/cloud_const.dart
+++ b/lib/constants/cloud_const.dart
@@ -49,7 +49,8 @@ const kCreatePin = '/v1/guardians/remote-assistances/sessions/pin';
 
 /// Path prefix for the router's reverse proxy to cloud (same-origin), used only
 /// for client-side RA requests on a local web build so the browser avoids CORS.
-/// NOTE: exact prefix NOT YET CONFIRMED by firmware — single point of change.
+/// Verified end-to-end against QA + Production + router FW (see issue #1179);
+/// kept as a single point of change should the FW prefix ever move.
 /// Must NOT end with a trailing slash (endpoints already start with '/').
 const kProxyPrefix = '/cloud';
 
--- a/lib/core/utils/ip_getter/get_local_ip.dart
+++ b/lib/core/utils/ip_getter/get_local_ip.dart
@@ -20,3 +20,8 @@ String getLocalIp(ProviderReader read) =>
 
 String getFullLocation(ProviderReader read) =>
     throw UnsupportedError('[Platform ERROR] Get Full Location');
+
+/// Browser origin (scheme + host [+ port]), e.g. `https://192.168.1.1`.
+/// Empty on non-web, where there is no browser origin and CORS does not apply.
+/// Unlike [getLocalIp] (host only), this preserves the scheme for same-origin.
+String getCloudOrigin() => '';
--- a/lib/core/utils/ip_getter/mobile_get_local_ip.dart
+++ b/lib/core/utils/ip_getter/mobile_get_local_ip.dart
@@ -6,3 +6,7 @@ String getLocalIp(ProviderReader read) => '';
 
 String getFullLocation(ProviderReader read) =>
     throw UnsupportedError('[Platform ERROR] Get Full Location');
+
+/// No browser origin on mobile and CORS does not apply → empty string, so the
+/// cloud host resolver falls back to the direct cloud base.
+String getCloudOrigin() => '';
--- a/lib/core/utils/ip_getter/web_get_local_ip.dart
+++ b/lib/core/utils/ip_getter/web_get_local_ip.dart
@@ -5,3 +5,5 @@ import 'get_local_ip.dart';
 String getLocalIp(ProviderReader read) => window.location.host;
 
 String getFullLocation(ProviderReader read) => window.location.toString();
+
+String getCloudOrigin() => window.location.origin;
--- a/test/core/cloud/cloud_host_resolver_test.dart
+++ b/test/core/cloud/cloud_host_resolver_test.dart
@@ -1,0 +1,77 @@
+import 'package:flutter_test/flutter_test.dart';
+import 'package:privacy_gui/constants/_constants.dart';
+import 'package:privacy_gui/core/cloud/cloud_host_resolver.dart';
+
+void main() {
+  // Assert against the configured cloud base (not a hardcoded domain) so the
+  // test is independent of the active CloudEnvironment.
+  final cloudBase = 'https://${cloudEnvironmentConfig[kCloudBase]}';
+
+  group('CloudHostResolver', () {
+    test('local build + client → router proxy base', () {
+      final resolver = CloudHostResolver(
+        isLocal: () => true,
+        originGetter: () => 'https://192.168.1.1',
+      );
+
+      expect(
+          resolver.resolve(forCA: false), 'https://192.168.1.1$kProxyPrefix');
+    });
+
+    test('local build + CA → cloud base (origin ignored)', () {
+      final resolver = CloudHostResolver(
+        isLocal: () => true,
+        originGetter: () => 'https://192.168.1.1',
+      );
+
+      expect(resolver.resolve(forCA: true), cloudBase);
+    });
+
+    test('local build + empty origin → cloud base (fallback)', () {
+      final resolver = CloudHostResolver(
+        isLocal: () => true,
+        originGetter: () => '',
+      );
+
+      expect(resolver.resolve(forCA: false), cloudBase);
+    });
+
+    test('non-local build + client → cloud base', () {
+      final resolver = CloudHostResolver(
+        isLocal: () => false,
+        originGetter: () => 'https://192.168.1.1',
+      );
+
+      expect(resolver.resolve(forCA: false), cloudBase);
+    });
+
+    test('origin with port is preserved', () {
+      final resolver = CloudHostResolver(
+        isLocal: () => true,
+        originGetter: () => 'https://192.168.1.1:8443',
+      );
+
+      expect(
+        resolver.resolve(forCA: false),
+        'https://192.168.1.1:8443$kProxyPrefix',
+      );
+    });
+
+    test('http origin scheme is preserved (not forced to https)', () {
+      final resolver = CloudHostResolver(
+        isLocal: () => true,
+        originGetter: () => 'http://192.168.1.1',
+      );
+
+      expect(resolver.resolve(forCA: false), 'http://192.168.1.1$kProxyPrefix');
+    });
+
+    test('default constructor → cloud base (deterministic in test env)', () {
+      // Defaults: origin empty + BuildConfig.isLocal (false under force=none).
+      final resolver = CloudHostResolver();
+
+      expect(resolver.resolve(forCA: false), cloudBase);
+      expect(resolver.resolve(forCA: true), cloudBase);
+    });
+  });
+}
--- a/test/core/cloud/cloud_host_resolver_test.dart
+++ b/test/core/cloud/cloud_host_resolver_test.dart
@@ -1,6 +1,7 @@
 import 'package:flutter_test/flutter_test.dart';
 import 'package:privacy_gui/constants/_constants.dart';
 import 'package:privacy_gui/core/cloud/cloud_host_resolver.dart';
+import 'package:privacy_gui/core/utils/ip_getter/ip_getter.dart';
 
 void main() {
   // Assert against the configured cloud base (not a hardcoded domain) so the
@@ -72,6 +73,27 @@ void main() {
 
       expect(resolver.resolve(forCA: false), cloudBase);
       expect(resolver.resolve(forCA: true), cloudBase);
+    });
+  });
+
+  group('getCloudOrigin (platform stub) → resolver integration', () {
+    // Tests run on the Dart VM, which selects the non-web (stub/mobile)
+    // implementation. It MUST return '' — never call Uri.base.origin, which
+    // throws a StateError on a file: URI. This locks in the non-web crash fix.
+    test('returns empty string on non-web', () {
+      expect(getCloudOrigin(), isEmpty);
+    });
+
+    test('local build + non-web origin → falls back to cloud base (no throw)',
+        () {
+      // Emulates a mobile .local build: isLocal() true, but getCloudOrigin()
+      // yields '' → resolver must fall back to the cloud base without throwing.
+      final resolver = CloudHostResolver(
+        isLocal: () => true,
+        originGetter: getCloudOrigin,
+      );
+
+      expect(resolver.resolve(forCA: false), cloudBase);
     });
   });
 }
--- a/test/core/cloud/guardian_api_client_test.dart
+++ b/test/core/cloud/guardian_api_client_test.dart
@@ -1,0 +1,147 @@
+import 'package:flutter_test/flutter_test.dart';
+import 'package:http/http.dart' as http;
+import 'package:mocktail/mocktail.dart';
+import 'package:privacy_gui/constants/_constants.dart';
+import 'package:privacy_gui/core/cloud/cloud_host_resolver.dart';
+import 'package:privacy_gui/core/cloud/guardian_api_client.dart';
+import 'package:privacy_gui/core/cloud/http/linksys_http_client.dart';
+
+import '../../mocks/test_data/remote_assistance_test_data.dart';
+
+class MockLinksysHttpClient extends Mock implements LinksysHttpClient {}
+
+class FakeUri extends Fake implements Uri {}
+
+/// Verifies that [GuardianApiClient] assembles request URLs against the correct
+/// host base: client-side requests go through the router proxy on a local
+/// build, while CA-side requests always target the cloud host directly.
+void main() {
+  setUpAll(() {
+    registerFallbackValue(FakeUri());
+  });
+
+  const routerOrigin = 'https://192.168.1.1';
+  final cloudBase = 'https://${cloudEnvironmentConfig[kCloudBase]}';
+
+  late MockLinksysHttpClient mockHttp;
+  late GuardianApiClient client;
+
+  setUp(() {
+    mockHttp = MockLinksysHttpClient();
+    // Simulate a local (router-hosted) build so client-side requests proxy.
+    client = GuardianApiClient(
+      httpClient: mockHttp,
+      hostResolver: CloudHostResolver(
+        isLocal: () => true,
+        originGetter: () => routerOrigin,
+      ),
+    );
+  });
+
+  /// Captures the [Uri] passed to a mocked verb without hitting the network.
+  Uri capturedGetUri() =>
+      verify(() => mockHttp.get(captureAny(), headers: any(named: 'headers')))
+          .captured
+          .single as Uri;
+
+  Uri capturedPostUri() =>
+      verify(() => mockHttp.post(captureAny(), headers: any(named: 'headers')))
+          .captured
+          .single as Uri;
+
+  Uri capturedDeleteUri() => verify(
+          () => mockHttp.delete(captureAny(), headers: any(named: 'headers')))
+      .captured
+      .single as Uri;
+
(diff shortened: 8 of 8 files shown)


AustinChangLinksys (commented):
🤖 **Automated Review — Round 1** · `a384421..0dedfc8` (full)

**Verdict: 💬 Self-review (comment only)** — Self-authored PR; review provided as reference. Two warnings worth addressing before merge.

#### Standards

|---|---|---|---|
| ⚠️ | High | `lib/core/utils/ip_getter/ip_getter.dart` | Barrel uses `dart.library.html` conditional export for web impl; WASM Flutter Web target does not define `dart.library.html` → falls back to stub, proxy never activates on WASM |
| ⚠️ | High | `lib/core/cloud/cloud_host_resolver.dart:38` | `_cloudBase` is injected for `originGetter`/`isLocal` but not for `cloudBase`; `CloudHostResolver._cloudBase` still reads global `cloudEnvironmentConfig` — harder to test multi-env scenarios |
| 💡 | Med | `lib/core/cloud/guardian_api_client.dart:_request` | `forCA` derived from `sessionToken != null` — implicit coupling; if a future CA method doesn't use session token, routing silently regresses |
| 💡 | Low | `lib/core/utils/ip_getter/web_get_local_ip.dart` | `window.location.origin` returns string `"null"` (not empty) in some cross-origin iframe contexts; `isEmpty` guard won't catch it |
| 💡 | Low | `lib/core/cloud/cloud_host_resolver.dart:38` | `https://` hardcoded before `cloudEnvironmentConfig[kCloudBase]`; if config value ever includes scheme, double-scheme results — add a comment |
| 💡 | High | `test/core/cloud/guardian_api_client_test.dart` | `fetchDeviceToken` proxy URL path (`forCA: false`) not tested; other 4 client-side methods are |

#### Spec (Issue #1179)

|---|---|---|---|
| ⚠️ | High | `ip_getter.dart` barrel | AC5 (scheme preserved) + entire proxy feature: WASM target silently regresses via dart.library.html fallback — see W-1 |
| 💡 | High | `cloud_host_resolver_test.dart` + `guardian_api_client_test.dart` | AC coverage otherwise solid — 8 resolver tests + 6 client tests cover all AC scenarios |

*Confidence: High = code-verified · Med = located + reasoned · Low = speculative.*

<details>

**W-1 — `dart.library.html` barrel export will miss WASM web target (High)**
`lib/core/utils/ip_getter/ip_getter.dart`. The conditional export:
```dart
export 'get_local_ip.dart'
    if (dart.library.io) 'mobile_get_local_ip.dart'
    if (dart.library.html) 'web_get_local_ip.dart';
```
`web_get_local_ip.dart` uses `package:web` (js_interop), but the barrel still tests `dart.library.html`. Flutter WASM compilation does NOT define `dart.library.html`; it defines `dart.library.js_interop`. On a WASM web build, the barrel exports the stub (`get_local_ip.dart`), `getCloudOrigin()` always returns `''`, and the entire proxy feature silently falls back to direct cloud — defeating the purpose of this PR.
**Fix:** `if (dart.library.js_interop) 'web_get_local_ip.dart'`

---

**W-2 — `CloudHostResolver._cloudBase` still coupled to global state (High)**
`lib/core/cloud/cloud_host_resolver.dart:38`: `'https://${cloudEnvironmentConfig[kCloudBase]}'`. `originGetter` and `isLocal` are injectable, but `cloudBase` reads a global that depends on async `BuildConfig.load()`. Tests use `final cloudBase = 'https://${cloudEnvironmentConfig[kCloudBase]}'` which is stable in the test environment, but makes multi-environment unit testing require global mutation.
**Fix:** Add an optional `cloudBaseGetter` injection point; default to current global read.

</details>

<details>

- **`CloudHostResolver` design:** clean injection of `originGetter` and `isLocal` closures makes the resolver genuinely unit-testable for the main cases.
- **`forCA` derived from `sessionToken != null`** is a correct invariant for the current API surface — lock-step with auth-header branch, no drift possible.
- **`kProxyPrefix` constant:** single point of change, well-documented.
- **Empty origin fallback:** resolver returns `_cloudBase` when `origin.isEmpty` — safe fallback on non-web / unresolvable origin.
- **Platform stubs:** `mobile_get_local_ip.dart` and `get_local_ip.dart` both return `''`, correct for non-browser environments.
- **Test matrix:** resolver (8 tests) + client URL assembly (6 tests) cover all AC scenarios comprehensively.

</details>

lib/core/cloud/guardian_api_client.dart
```diff
@@ -23,12 +27,25 @@ final guardianApiClientProvider = Provider((ref) => GuardianApiClient());
...
   final LinksysHttpClient _http;
+  final CloudHostResolver _hostResolver;
 
-  GuardianApiClient({LinksysHttpClient? httpClient})
-      : _http = httpClient ?? LinksysHttpClient();
+  GuardianApiClient({
+    LinksysHttpClient? httpClient,
+    CloudHostResolver? hostResolver,
+  })  : _http = httpClient ?? LinksysHttpClient(),
+        _hostResolver = hostResolver ?? CloudHostResolver();
```
PeterJhongLinksys, line 36: ⚠️ **Silent fail-open default.** `CloudHostResolver()`'s default `originGetter` is `_emptyOrigin` (`cloud_host_resolver.dart:32,35`), so any `GuardianApiClient()` constructed without an explicit `hostResolver` disables the proxy entirely and reintroduces the CORS bug — with no log or assert. The correct behaviour depends solely on line 16 remembering `originGetter: getCloudOrigin`.

Suggest defaulting to the platform getter here (`hostResolver ?? CloudHostResolver(originGetter: getCloudOrigin)`), or making `getCloudOrigin` the resolver's own default — the non-web stub already returns `''`, so unit tests stay deterministic.

lib/core/cloud/guardian_api_client.dart
```diff
@@ -213,7 +232,11 @@ class GuardianApiClient {
     String? serialNumber,
     String? sessionToken,
   }) async {
-    final url = Uri.parse(_buildUrl(endpoint, args: args));
+    // CA-side requests authenticate with a session token and must always hit
+    // the cloud host directly; deriving forCA here keeps it in lock-step with
+    // the auth-header branch below (no drift possible).
+    final forCA = sessionToken != null;
```
PeterJhongLinksys, line 238: ⚠️ **Implicit contract (Temporal Coupling).** The derivation is correct today — all six `_request` call sites were verified (`getSessions:122`, `getSessionInfo:138`, `createPin:157`, `deleteSession:183` pass only `linksysToken + serialNumber`; `getSessionInfoForCA:201`, `deleteSessionForCA:215` pass only `sessionToken`), so nothing drifts right now.

But all three token parameters are optional and nullable, and the "all null" combination is legal: it yields `forCA == false` (proxy base) **and** matches neither auth branch at `:242-248`, i.e. an unauthenticated request. "No drift possible" rests on an assumption nothing enforces.

Suggest making the caller's intent explicit (`required bool forCA` per public method, or a small sealed caller type) plus `assert(forCA == (sessionToken != null))`.

lib/core/cloud/guardian_api_client.dart
```diff
@@ -65,7 +82,9 @@ class GuardianApiClient {
     }
 
     logger.d('[Guardian] Fetching new device token');
-    final url = Uri.parse(_buildUrl(kDeviceToken)).replace(queryParameters: {
+    // Client-side (device owner) request → may proxy through the router.
+    final url = Uri.parse(_buildUrl(kDeviceToken, forCA: false))
```
PeterJhongLinksys, line 86: ⚠️ **Only path bypassing `_validateResponse`, and the only `_buildUrl` caller without a test.** Requests via `_request` are screened by `_validateResponse` (`:265-285`); this one decodes directly at `:94`.

`kProxyPrefix` (`lib/constants/cloud_const.dart:55`) is a compile-time constant with no firmware capability probe. On a router whose FW lacks the `/cloud` reverse proxy, this request typically returns the SPA `index.html` with HTTP 200 → `jsonDecode` throws `FormatException` → `remote_assistance_service.dart:36 catch (e)` → `_mapError` (`:150-169`) finds no `ErrorResponse` → generic `UnexpectedError`. RA becomes unusable with an error that does not point at the root cause, and the `_cloudBase` fallback only triggers on an *empty origin*, never on proxy failure.

Suggest validating before decoding and translating `FormatException`/404 into an explicit `PROXY_UNAVAILABLE` error (or probing the proxy once and falling back to direct cloud), plus a test for this URL — proxy base with the query parameters preserved through `Uri.replace`.

lib/core/cloud/cloud_host_resolver.dart
```diff
@@ -0,0 +1,47 @@
...
+  static String _emptyOrigin() => '';
+
+  /// Direct cloud base — always `https://<cloudBase>`.
+  String get _cloudBase => 'https://${cloudEnvironmentConfig[kCloudBase]}';
+
+  /// URL base to prepend to a Guardian endpoint.
+  String resolve({required bool forCA}) {
+    if (forCA || !_isLocal()) return _cloudBase;
+    final origin = _originGetter();
+    if (origin.isEmpty) return _cloudBase; // fallback: never break on no origin
```
PeterJhongLinksys, line 44: ⚠️ **`isEmpty` is not a sufficient guard.** `web_get_local_ip.dart:9` returns `window.location.origin`, and per the HTML spec an *opaque* origin serializes to the literal string `"null"` — not an empty string (sandboxed iframe without `allow-same-origin`, `file:`/`data:` documents). Line 45 then produces `"null/cloud"`, and `Uri.parse('null/cloud/v1/guardians/...')` at `guardian_api_client.dart:239` is a scheme-less *relative* URI, which `http` rejects at send time (`ArgumentError`) → surfaces as an opaque `UnexpectedError`.

Defensive hardening rather than a mainstream user path, but the fix is cheap — validate positively:
```dart
final uri = Uri.tryParse(origin);
if (uri == null || (!uri.isScheme('https') && !uri.isScheme('http')) || uri.host.isEmpty) {
  return _cloudBase;
}
```
and add an `originGetter: () => 'null'` test case.

PeterJhongLinksys (approved):
🤖 **Automated Review — Round 1** · `a384421..0dedfc8` (full)

**Verdict: ✅ APPROVE** — Sound architectural direction, no constitution hard-rule violations and no Critical issues; the CORS fix is correctly scoped and fails closed to the cloud base, but two maintainability traps (silent fail-open default resolver, implicit `forCA` derivation) and one uncovered/unvalidated path (`fetchDeviceToken`) are worth addressing before or shortly after merge.

|---|---|---|---|
| ⚠️ | 🟢 | `lib/core/cloud/guardian_api_client.dart:36` | Default `CloudHostResolver()` uses `_emptyOrigin` → any `GuardianApiClient()` built without an explicit `hostResolver` silently disables the proxy and reintroduces the CORS bug, with no log or assert. |
| ⚠️ | 🟢 | `lib/core/cloud/guardian_api_client.dart:238` | [Temporal Coupling] `forCA = sessionToken != null` is an implicit contract; the "all tokens null" combination is legal, yields `forCA == false` (proxy) **and** no auth header at all. |
| ⚠️ | 🟢 | `lib/core/cloud/guardian_api_client.dart:86-95` | `fetchDeviceToken` is the only path bypassing `_request`/`_validateResponse`, calls `jsonDecode` directly, and is the only `_buildUrl` caller with **no test** — a router without the `/cloud` proxy (SPA fallback HTML 200) degrades to an opaque `UnexpectedError`. |
| ⚠️ | 🟢 | `lib/core/cloud/guardian_api_client.dart:15-17` | [Art. VI §6.2] `CloudHostResolver` is `new`-ed inside the provider body instead of exposed as its own `Provider`, so widget/provider-level tests cannot override host resolution. |
| ⚠️ | 🟡 | `lib/core/cloud/cloud_host_resolver.dart:44` | Guard only rejects `isEmpty`; an opaque origin serializes to the literal string `"null"` (sandboxed iframe, `file:`), producing the scheme-less URI `null/cloud/...` → `ArgumentError` at send time. |
| ⚠️ | 🟡 | `lib/core/cloud/cloud_host_resolver.dart:45` | `http://` origins are preserved by design (locked in by a test), so `X-Linksys-Token` / RA PIN traffic could travel in cleartext over the LAN — a downgrade from the previous end-to-end TLS. |
| ⚠️ | 🟡 | `lib/core/utils/ip_getter/ip_getter.dart:21` | Conditional export still keys on `dart.library.html`; under dart2wasm both `io` and `html` are false → stub returns `''` and the fix silently no-ops. 20 other sites in the repo already use `dart.library.js_interop`. |
| ⚠️ | 🟡 | `lib/core/cloud/cloud_host_resolver.dart:38` | [Duplicated Code] Third copy of `'https://${cloudEnvironmentConfig[kCloudBase]}'`, and `_buildUrl` duplicates `LinksysHttpClient.combineUrl` — a second, parallel host-override seam next to the existing `LinksysHttpClient(getHost:)`. |
| ⚠️ | ⚪ | `lib/core/cloud/guardian_api_client.dart:48-50` | Path-variable substitution is not URL-encoded; with a proxy prefix in play, `../` in a session id could dot-segment-escape `/cloud` and carry the token to a router-local path. Pre-existing pattern, amplified here; current client-side ids come from server responses. |
| ⚠️ | ⚪ | `lib/core/cloud/cloud_host_resolver.dart:42-45` | No allowlist on the origin: `_isLocal()` is a compile-time `--dart-define` constant, not proof of running on the router, so the same bundle served from any host would send client-side RA requests (with token headers) there. No existing secret leaks (secure storage is origin-scoped), so this is a missing origin-binding hardening rather than an exploitable path. |
| 💡 | 🟢 | `test/core/cloud/cloud_host_resolver_test.dart` | The documented `kProxyPrefix` contract ("must NOT end with a trailing slash") and origin-with-trailing-slash behaviour have no executable guard — add `expect(kProxyPrefix.endsWith('/'), isFalse)` and a normalization case. |
| 💡 | 🟢 | `lib/core/utils/ip_getter/web_get_local_ip.dart:9` | [Art. X §10.1] The only implementation that returns a non-empty value is the only one without DartDoc. |
| 💡 | 🟡 | `test/core/cloud/guardian_api_client_test.dart:56-131` | Assertions capture only the URI (`headers: any(...)` is ignored); the very invariant at risk (host ⟺ auth header pairing) is not locked in. |
| 💡 | 🟡 | `test/core/cloud/cloud_host_resolver_test.dart:70-77` | The "deterministic" default-constructor case depends on the mutable static `BuildConfig.forceCommandType`; `--dart-define=force=local` in CI (or test ordering) turns it flaky. Prefer explicit `isLocal: () => false`. |
| 💡 | 🟡 | `lib/core/cloud/cloud_host_resolver.dart:41` | [Flag Argument] `forCA` is a pure pass-through boolean read in the negative at call sites; an enum (`GuardianCallSide.client/ca`) or two named methods would be self-documenting. |
| 💡 | 🟡 | `lib/core/utils/ip_getter/get_local_ip.dart:27` | The fallback stub returns `''` while its neighbours `throw UnsupportedError`; the "should never happen" branch now fails silently instead of loudly. Document the intent or align with the neighbours. |
| 💡 | 🟡 | `lib/core/cloud/cloud_host_resolver.dart:21` | Placement/naming: `lib/core/cloud/` already has `services/`, `providers/`, `http/`; a host-selection policy object arguably belongs in `services/` (and `Resolver` is not in the Art. III suffix set). Low-confidence style call. |
| 💡 | 🟡 | `test/core/cloud/guardian_api_client_test.dart:11` | [Duplicated Code] `MockLinksysHttpClient` is redefined here and in `test/page/firmware_update/services/firmware_ota_check_service_test.dart:8`; worth hoisting into a shared `test/mocks/`. |
| 💡 | 🟡 | `lib/core/cloud/http/linksys_http_client.dart:284` | Pre-existing: `logger.i` prints full headers (including `X-Linksys-Token` / `Authorization`) and survives release builds. RA traffic now newly flows through this client. `guardian_api_client.dart:251` gets this right by logging only `headers.keys`. |
| 💡 | ⚪ | `lib/core/cloud/cloud_host_resolver.dart:42` | CA-side on a local build still goes direct to cloud and will therefore always fail CORS; the asymmetry lives only in a comment. A warning log for `isLocal() && forCA` would make this known-bad combination diagnosable in the field. |
| 💡 | ⚪ | `lib/core/cloud/http/linksys_http_client.dart:329-344` | Retry policy retries 404/401 three times, which amplifies the cost of a misconfigured proxy (three token-bearing requests + UI stall per call). Pre-existing, but worth revisiting now that a proxy hop exists. |
| 💡 | ⚪ | `test/core/cloud/*_test.dart` | [Art. VIII] `'https://192.168.1.1'` and its port/scheme variants are inlined 6+ times; `RemoteAssistanceTestData` is already imported and is the natural home. |

<sub>Confidence: 🟢 High = verified against head code with evidence · 🟡 Med = file:line + reasoning, not fully substantiated · ⚪ Low = speculative / defensive hardening.</sub>
<sub>`[X]` = raised by 1 agent, `[XX]` = raised by both. Both agents independently raised the `fetchDeviceToken` coverage gap and the `BuildConfig`-dependent default-constructor test.</sub>

<details>

**W1 — Silent fail-open default resolver** · `lib/core/cloud/guardian_api_client.dart:36`, `lib/core/cloud/cloud_host_resolver.dart:32,35`
```dart
// guardian_api_client.dart:33-36
})  : _http = httpClient ?? LinksysHttpClient(),
      _hostResolver = hostResolver ?? CloudHostResolver();
// cloud_host_resolver.dart:32,35
: _originGetter = originGetter ?? _emptyOrigin,
static String _emptyOrigin() => '';
```
The proxy behaviour depends entirely on the single site at `guardian_api_client.dart:15-17` remembering `originGetter: getCloudOrigin`. Any future `GuardianApiClient()` (new provider, demo, another feature) reverts to the pre-fix behaviour with no signal.
*Fix:* default to the platform getter — `hostResolver ?? CloudHostResolver(originGetter: getCloudOrigin)`, or make `getCloudOrigin` the resolver's own default (the non-web stub already returns `''`, so tests stay deterministic).

**W2 — Implicit `forCA` contract** · `lib/core/cloud/guardian_api_client.dart:235-248`
```dart
final forCA = sessionToken != null;
final url = Uri.parse(_buildUrl(endpoint, args: args, forCA: forCA));
...
if (sessionToken != null) { ... } else if (linksysToken != null && serialNumber != null) { ... }
```
All three token parameters are optional and nullable. The "all null" combination compiles, selects the proxy base, and matches neither auth branch — an unauthenticated request. The "no drift possible" claim holds only under an assumption that nothing enforces.
*Verified:* the derivation is correct **today** — all six `_request` call sites were checked (`getSessions:122`, `getSessionInfo:138`, `createPin:157`, `deleteSession:183` pass only `linksysToken + serialNumber`; `getSessionInfoForCA:201`, `deleteSessionForCA:215` pass only `sessionToken`). No method passes both or neither.
*Fix:* make the caller's intent explicit (`required bool forCA` per public method, or a small sealed caller type), and `assert(forCA == (sessionToken != null))`.

**W3 — `fetchDeviceToken`: no response validation, no test** · `lib/core/cloud/guardian_api_client.dart:86-95`
```dart
final url = Uri.parse(_buildUrl(kDeviceToken, forCA: false))
    .replace(queryParameters: { ... });
final response = await _http.get(url, headers: _defaultHeaders);
final body = jsonDecode(response.body);
final token = body['linksysToken'] as String?;
```
Requests routed via `_request` are screened by `_validateResponse` (`:265-285`); this path is not. `kProxyPrefix` (`lib/constants/cloud_const.dart:55`) is a compile-time constant with no firmware capability probe, so on a router whose FW lacks the reverse proxy the request typically returns the SPA `index.html` with HTTP 200 → `jsonDecode` throws `FormatException` → `remote_assistance_service.dart:36 catch (e)` → `_mapError` (`:150-169`) finds no `ErrorResponse` → generic `UnexpectedError`. RA becomes unusable with an error that does not point at the root cause; the `_cloudBase` fallback only triggers on an *empty origin*, never on proxy failure.
*Fix:* validate before decoding and translate `FormatException` / 404 into an explicit `PROXY_UNAVAILABLE` error, or probe the proxy once and fall back to the direct cloud base. Add a test covering this URL (proxy base + preserved query parameters).

**W4 — Resolver not provider-injected** · `lib/core/cloud/guardian_api_client.dart:15-17`
```dart
final guardianApiClientProvider = Provider((ref) => GuardianApiClient(
      hostResolver: CloudHostResolver(originGetter: getCloudOrigin),
    ));
```
`remote_assistance_service.dart:9` shows the repo's own idiom (`ref.watch(guardianApiClientProvider)`). A `cloudHostResolverProvider` would allow `ProviderScope.overrides` in widget/provider tests.

**W5 — Opaque origin serializes to `"null"`** · `lib/core/cloud/cloud_host_resolver.dart:41-46`
```dart
final origin = _originGetter();
if (origin.isEmpty) return _cloudBase; // fallback: never break on no origin
return '$origin$kProxyPrefix';
```
`web_get_local_ip.dart:9` returns `window.location.origin`, which per the HTML spec serializes an opaque origin as the string `"null"` (sandboxed iframe without `allow-same-origin`, `file:`/`data:` documents) — not empty. Result: `Uri.parse('null/cloud/v1/...')` is a relative URI and `http` throws on send, surfacing as `UnexpectedError`. Defensive hardening rather than a mainstream user path.
*Fix:* validate positively — `Uri.tryParse`, require an http/https scheme and a non-empty host, otherwise fall back; add a `originGetter: () => 'null'` test case.

**W6 — `http://` scheme deliberately preserved** · `lib/core/cloud/cloud_host_resolver.dart:45`, test `:61-66`
The test `http origin scheme is preserved (not forced to https)` freezes this as a specification. Client-side requests carry `X-Linksys-Token` / `X-Linksys-SN` (`guardian_api_client.dart:245-247`) and `fetchDeviceToken`'s response body itself contains the device token (`:93-95`), so on an http-served router UI those credentials (and the RA PIN) move in cleartext over the LAN — previously they were always TLS-protected to `guardian.tools`. Not marked Critical: nothing in the repo substantiates that local builds are actually served over http (`firmware_local_upload_service.dart:38` uses `wss://`, suggesting https).
*Fix:* decide explicitly — either accept only https origins (fall back otherwise), or document the "local build is always https" guarantee and rename the test to say so; the current name reads like intentional cleartext acceptance.

**W7 — wasm-unsafe conditional export** · `lib/core/utils/ip_getter/ip_getter.dart:19-21`
```dart
export 'get_local_ip.dart'
    if (dart.library.io) 'mobile_get_local_ip.dart'
    if (dart.library.html) 'web_get_local_ip.dart';
```
Under dart2wasm both conditions are false → the stub `getCloudOrigin() => ''` wins → resolver falls back to cloud → CORS regresses silently (and `getLocalIp` would throw `UnsupportedError`). Not currently reachable (`build_web.sh:24` has no `--wasm`), and it fails closed, but it is inconsistent with ~20 other sites already on `dart.library.js_interop`, while `web_get_local_ip.dart:1` already imports the wasm-friendly `package:web/web.dart`.
*Fix:* switch to `if (dart.library.js_interop)`.

**W8 — Third copy of the cloud base / parallel host seam** · `lib/core/cloud/cloud_host_resolver.dart:38` vs `lib/core/cloud/http/linksys_http_client.dart:93-107`
```dart
// linksys_http_client.dart:93-94
String getHost() => _getHost?.call() ?? 'https://${cloudEnvironmentConfig[kCloudBase]}';
// :99-107 combineUrl(endpoint, args:) ≈ GuardianApiClient._buildUrl
```
`'https://${cloudEnvironmentConfig[kCloudBase]}'` now appears in three places (`cloud_host_resolver.dart:38`, `linksys_http_client.dart:94`, `remote_assistance_confirm_view.dart:201`), and two overlapping host-override seams coexist. Please either reuse `LinksysHttpClient(getHost:)` or note in the PR description why a separate seam is preferable, and extract the base into one shared expression.

**W9 — Unencoded path substitution + dot-segment escape** · `lib/core/cloud/guardian_api_client.dart:48-50`
```dart
String url = '${_hostResolver.resolve(forCA: forCA)}$endpoint';
args?.forEach((key, value) => url = url.replaceFirst(key, value));
```
The upstream-most `sessionId` is an untrusted query parameter (`lib/route/route_remote_assistance.dart:9`: `state.uri.queryParameters['session'] ?? ''`), inserted verbatim; `?`, `#` or `../` can rewrite the query, truncate the path, or — now that a `/cloud` prefix exists — escape it. Low confidence: client-side call sites derive `sessionId` from server responses (`remote_client_provider.dart:115/194/252/316`), and the CA path is attacker-supplied on both sides (no privilege gain), and the host cannot be changed.
*Fix:* `Uri.encodeComponent` the substituted values, then `Uri.parse(url).normalizePath()` and assert the path still starts with `kProxyPrefix`.

**W10 — No origin allowlist** · `lib/core/cloud/cloud_host_resolver.dart:42-45`, `lib/constants/build_config.dart:44-45,127-129`
`_isLocal()` derives from `const String.fromEnvironment('force')`, i.e. a value baked into the bundle — not evidence of actually running on the router. A local build served from any host would send every client-side RA request (with token headers) to that host. Kept low: in that scenario `fetchDeviceToken` also goes through the same base, so the token originates from that host, and cached tokens live in origin-scoped `FlutterSecureStorage` — no pre-existing secret is exposed.
*Fix:* assert the origin host matches an expected pattern (private range / `myrouter.*`), or build the base from the router IP already known via USP/JNAP rather than the browser origin.

</details>

<details>

- **`kProxyPrefix` contract untested** (`cloud_const.dart:55` documents "must NOT end with a trailing slash"; `cloud_host_resolver.dart:45` concatenates without normalization). A trailing slash on either side yields `//cloud`. Add `expect(kProxyPrefix.endsWith('/'), isFalse)` plus a trailing-slash-origin case so the "single point of change" promise is enforceable.
- **Missing DartDoc on the web implementation** (`web_get_local_ip.dart:9`) while the stub and mobile variants are documented — note that the returned origin includes the scheme and excludes the path.
- **Tests assert URL only** (`guardian_api_client_test.dart:56-131`): capture and assert headers too, so `client → proxy + X-Linksys-Token` and `CA → cloud + Authorization` are truly locked in lock-step. Today, inverting the `forCA` derivation could still pass.
- **Default-constructor test depends on a mutable global** (`cloud_host_resolver_test.dart:70-77` → `build_config.dart:44,115-124`). Use `isLocal: () => false` explicitly, and assert "default originGetter returns empty" separately.
- **Flag argument** (`cloud_host_resolver.dart:41`, `guardian_api_client.dart:46-48`): `forCA` is forwarded verbatim and read in the negative; an enum or `resolveForClient()` / `resolveForCa()` removes the double negative and closes off W2's implicit derivation.
- **Stub inconsistency** (`get_local_ip.dart:24-27`): neighbours throw `UnsupportedError`; the new function returns `''`. If the silence is deliberate ("fail-safe"), say so in the DartDoc — otherwise the unknown-platform branch degrades invisibly.
- **Placement/naming** (`cloud_host_resolver.dart:21`): `services/` (or `http/`) would fit a host-selection policy better than the package root, and Art. III's suffix set does not include `Resolver`. Style call; owner's decision.
- **Duplicated mock** (`guardian_api_client_test.dart:11` vs `firmware_ota_check_service_test.dart:8`): `LinksysHttpClient` is shared infrastructure; hoist the mock (and the `registerFallbackValue(FakeUri())` boilerplate) into `test/mocks/` before a third copy appears — `test/core/usp/mocks.dart` is the precedent.
- **Header logging** (`linksys_http_client.dart:284`, `logger.i('HEADERS: ${request.headers}')`): pre-existing, and `logger.i` is not filtered in release (`logger.dart:37-40`), but RA traffic newly flows here. Apply `masking_utils` at the client layer as `guardian_api_client.dart:251` already does by logging `headers.keys` only.
- **CA-on-local-build is a known-dead combination** (`cloud_host_resolver.dart:42`): `router_provider.dart:196-202` still routes `?session=` on local builds into the CA confirm flow, which then goes direct to cloud and is blocked by CORS. A warning log makes it diagnosable.
- **Retry amplification** (`linksys_http_client.dart:329-344`): 404/401 retried three times with backoff, each re-attaching the token — noisy and slow when the proxy is misconfigured.
- **Inline test data** (`cloud_host_resolver_test.dart:14,25,35,45,54,64`, `guardian_api_client_test.dart:24`): move `testRouterOrigin` / `testRouterOriginWithPort` into `RemoteAssistanceTestData` (already imported).

</details>

<details>

- **Correct layering.** All new logic sits in `lib/core/` infrastructure with a symmetric `test/core/` layout. No `generated/` import leaks into `providers/` or `views/` (Art. V §5.4), no upward dependencies, and no drive-by changes to the USP layer (Art. I §1.3 scope discipline). Art. IV's four USP hard rules and Art. XV UI Kit rules simply do not apply to this change.
- **`forCA` derivation is verified correct today.** All six `_request` call sites plus `fetchDeviceToken` (the only direct `_buildUrl` caller, explicitly `forCA: false`) were checked one by one: no method passes both token kinds and none passes neither, so host selection and the auth-header branch at `:242-248` really do share a single source.
- **Fallback direction is right (fail-closed to cloud).** When the origin is unavailable, `resolve()` returns `_cloudBase` (`:44`) — the worst case is the pre-change behaviour, never a half-built base.
- **`_isLocal` stored as a function reference, not a construction-time value.** `BuildConfig.forceCommandType` is a mutable static rewritten asynchronously by `BuildConfig.load()` on non-web (`build_config.dart:115-124`); holding a closure avoids freezing `false` when the provider is created before `load()`. `_cloudBase` as a getter likewise picks up runtime environment switches.
- **Platform stubs are consistent and semantically correct** (`get_local_ip.dart:27`, `mobile_get_local_ip.dart:12` → `''`; `web_get_local_ip.dart:9` → `window.location.origin`). Notably the stub does *not* use `Uri.base.origin`, which would throw `StateError` on a `file:` URI, and `cloud_host_resolver_test.dart:79-95` locks that constraint in.
- **Injection style matches repo convention.** `String Function()` / `bool Function()` constructor injection mirrors `LinksysHttpClient({String? Function()? getHost})` (`linksys_http_client.dart:34`) instead of inventing an abstract class + provider — consistent with Art. VII's stance against over-abstraction.
- **Coverage from zero.** Neither `GuardianApiClient` nor `CloudHostResolver` had any test; this PR adds 246 lines using the standard Mocktail idiom (`registerFallbackValue(FakeUri())`, semantic group names, no `TC001`-style numbering — Art. VIII).
- **Tests do not hardcode the domain.** `cloudBase` is derived from `cloudEnvironmentConfig[kCloudBase]` in both files, so switching QA/Prod cannot cause spurious failures.
- **Real edge cases were anticipated**: scheme preservation, origin with port, empty-origin fallback, and non-web stub returning `''` without throwing (`cloud_host_resolver_test.dart:30-39,49-58,61-67,79-95`).
- **No hardcoded secrets.** Device tokens continue to flow through `FlutterSecureStorage` (`guardian_api_client.dart:71-73,104-108`); test tokens live in `test/mocks/test_data/`. `guardian_api_client.dart:251` logs `headers.keys.toList()` only — a good example.
- **Documentation quality is above average.** `cloud_host_resolver.dart:3-20` explains the CORS background, the selection rule, why injection exists, and the fallback semantics; `cloud_const.dart:50-54` records the firmware-prefix verification scope. `kProxyPrefix` is correctly placed alongside the other cloud endpoint constants and referenced from both lib and test (no magic strings).
- **Blast radius is well controlled.** `resolve()`'s first branch (`:42`) sends every CA-side and every non-local build straight to the cloud, keeping the proxy's reach as small as possible.

</details>

PR #1167, opened 2026-07-22
feat(usp): E2E testing initiative — transport seam, demo unification, E2E_MOCK, a11y labels

AustinChangLinksys:
## Summary

App-side work for the PrivacyGUI E2E testing initiative (Phases 1–3 of the
three-repo plan). Extracts a transport seam so demo mode runs the real
`UspClient`, adds an `E2E_MOCK` build flag, removes dead JNAP code, and adds
`semanticLabel`s + `identifier` test hooks that make dialogs/toolbars locatable
(for accessibility and E2E, via a generated single source of truth). Also fixes
a theme-resolution bug that blocked E2E from selecting a flat, effect-free theme.

## What's included

**USP transport seam (P1/P2)**
- Extract `abstract interface class UspTransport` (`lib/core/usp/transport/`);
  `UspClientWeb` + stub implement it; add `UspClient.withTransport(...)`.
  108 call-sites of `UspClient` are unchanged.

**Demo mode = production build, swapped data source (P3/P4/E6)**
- `DemoUspTransport implements UspTransport` (replaces the old
  `DemoUspClient extends UspClient`); demo now runs the real `UspClient`.
- Demo fetches TR-181 fixture data at runtime + a scenario picker; drop the
  demo-only FAB so demo UI matches the production build.

**E2E_MOCK build flag (P5)**
- `bool.fromEnvironment('E2E_MOCK')` in `build_config.dart`; guards SSE connect
  and the first-run onboarding auto-popup. Default false → production unchanged.

**Dead-code removal (P6) + boot contract (P9)**
- Remove dead JNAP transitional code (demo cache data/json, error list).
- Add `doc/usp/e2e-boot-contract.md` (the E2E ↔ app boot contract spec).

**Accessibility / semanticLabels (P8 + follow-ups)**
- Bump ui_kit to v2.30.1 (input `semanticLabel` + `identifier` params); remove
  the login page's manual `Semantics` workaround.
- Add `semanticLabel`s to dashboard toolbar icon buttons and the port-forwarding
  add/edit dialog fields — real a11y improvement, and lets E2E drive those flows.

**E2E test-hook identifiers — single source of truth (SSOT)**
- Wire `identifier:`/`semanticLabel:` test hooks into the widgets the E2E suite
  targets, so tests locate them via a generated single source of truth instead
  of hard-coded strings — an app-side rename becomes a compile error, not a
  silent count-0 timeout. See #1172.
[...]

PR #1201, opened 2026-08-05
feat(ai): persist Bedrock credentials, keep user data out of release logs (#1193, #1194)

HankYuLinksys:
## Summary

Fixes #1193 and #1194. Two independent AI-assistant changes, both confined to the module and reviewable separately (one commit each).

- **#1193** — Bedrock credentials were held only in memory, so the user re-typed an access key and secret key on every app launch. They are now saved with `flutter_secure_storage` and restored on startup.
- **#1194** — The assistant logged its full router context, conversation text, and per-device names, IPs, MACs and SSIDs unconditionally. That is the user's household in a shipped log file.

## #1193 — Persist credentials

Startup resolves in three steps: environment config → saved credentials → the manual config screen. "Change configuration" clears storage, since leaving the record behind would silently restore on the next launch and undo what the user just asked for.

The store follows `RouterFingerprintService`: a `Provider<T>` over `const FlutterSecureStorage()`, constructor injection, `store` / `read` / `clear`.

Two properties the implementation has to hold, both of which cost a bug before they were in place:

**Ordering.** `store()` and `clear()` are fired without awaiting, because a storage failure must not break a working session. Without serialisation the writes of a connect can land *after* the deletes of a "change configuration" that followed it — resurrecting credentials the user had just discarded. Every operation therefore runs through one chain:

```dart
Future<T> _serialize<T>(Future<T> Function() operation) {
  final result = _pending.then((_) => operation());
  _pending = result.then((_) {}, onError: (_) {});   // a failure must not stall the queue
  return result;
}
```

**Atomicity.** The record is a single JSON value under one key rather than three keys. With separate keys a failed *re-save* leaves a new access key next to an old secret — which `read()` cannot detect, and which surfaces as an opaque SigV4 failure rather than "not configured". Legacy keys from the three-key layout are removed on write and clear.

The restore is asynchronous, so the config screen defends against it:

| Guard | Why |
|---|---|
| `_isRestoring` | Fields are read-only and Connect is disabled while the read is in flight |
| `_userTookOver` | A late restore stands down if the user has already typed something |
| `_controller != null` | A late restore never replaces a live session |

Without these, a restore landing after the user started typing would overwrite their input and connect with *different* credentials than the ones they entered; or replace a live controller, orphaning its listener (`RouterChatController` has no `dispose()`) and discarding the conversation.

Also: values are trimmed on both write and read so whitespace cannot reach the form; an unreadable record is treated as absent; changing the model after a restore updates the stored record instead of being forgotten; and clearing resets the selected model so the next credentials are not silently connected on the previous account's choice.

**Platform limit, recorded on the class:** on web this is obfuscation rather than encryption — `flutter_secure_storage_web` keeps the AES key in the same `localStorage` as the ciphertext. The real fix is for the app never to hold AWS credentials at all (a backend proxy would sign the requests), which is out of scope here and noted in the issue.

## #1194 — Keep user data out of release logs

`lib/ai/ai_logging.dart` adds two functions, so which lines carry user data is visible **at the call site** rather than depending on each caller remembering a guard:

- `aiLog(String)` — structural diagnostics: loop progress, tool names, token counts, device *counts*, CPU/memory percentages, success/failure. Kept in release; this is what makes a field report useful.
- `aiLogSensitive(String Function(), {String Function()? orElse})` — content. Compiled out of release builds, and takes a callback so the string is never even built there.

`orElse` exists because several lines carry both a keepable fact and an identifier that is not. Without it the only option was two log calls for one event, which reads as duplicate output in debug logs:

```dart
aiLogSensitive(
  () => 'buildRouterContext: wan.isUp=${m.isUp}, ip=${m.ipAddress}',
  orElse: () => 'buildRouterContext: wan.isUp=${m.isUp}',
);
```

Redacted: the router context dump, user and assistant message text, the A2UI action payload, per-device name/IP/connection/signal, extender name/MAC, WAN IP, per-band SSID, router model/firmware.

Also removes log lines that only restated control flow (`skipping tool_result`, `Rendering A2UIResponseRenderer`, `no textContent`), moves the credential store onto the same helpers, and collapses an extender loop that logged once per node into a single call.

**Not covered:** `generative_ui`'s `AwsContentGenerator` prints the full response body on the error path (`aws_content_generator.dart:294`). Its other 46 `debugPrint`s only emit lengths and booleans. That one line needs a UI kit change and is a different order of magnitude from this — better folded into a later kit release than a cross-repo round trip of its own.

## Verification

| Check | Result |
|---|---|
| `flutter test test/ai/ test/page/ai_assistant/` | 86/86 |
| `./run_tests.sh` (full suite) | **3421/3421** |
| `flutter analyze` on changed files | clean |
| `fvm dart format` | 7 files, 0 changed |

New tests: 18 for the credential store (including the store/clear race, a failed re-save, unreadable records, whitespace) and 3 for the logging helpers.

## Notes for reviewers

- **`router_assistant_view.dart` has no widget test**, so the three restore guards above are enforced by logic but not covered by a test. Called out deliberately rather than left to be discovered — each is reproducible in a short widget test if you would prefer them covered before this lands.
- Region stays hardcoded to `us-west-2`. Letting the user type it adds a way to get stuck with no diagnostic, and there is no evidence anyone needs a different one; the real answer is the backend proxy.
- The two commits are independent — #1194 touches only logging, #1193 only the credential path — so they can be split if you would rather review them apart.
- `dev-2.7.0` already carries `flutter_secure_storage: ^10.3.1` from #1167, so this adds no dependency.

### DIFF
--- a/lib/page/ai_assistant/services/aws_credentials_store.dart
+++ b/lib/page/ai_assistant/services/aws_credentials_store.dart
@@ -12,13 +12,12 @@ import 'package:privacy_gui/ai/ai_logging.dart';
 /// failure rather than "not configured".
 const _kRecordKey = 'ai_assistant_aws_credentials';
 
-/// Keys written by the previous three-key layout, removed on write and clear so
-/// stale halves cannot linger after an upgrade.
-const _kLegacyKeys = [
-  'ai_assistant_aws_access_key_id',
-  'ai_assistant_aws_secret_access_key',
-  'ai_assistant_bedrock_model_id',
-];
+/// How long a single storage operation may take before the chain gives up.
+///
+/// Without this, one hung `write` blocks every operation queued behind it — the
+/// user's `clear()` included, so "change configuration" would never take effect
+/// and the config screen would stay disabled with no error.
+const _kOperationTimeout = Duration(seconds: 10);
 
 final awsCredentialsStoreProvider = Provider<AwsCredentialsStore>((ref) {
   return AwsCredentialsStore(const FlutterSecureStorage());
@@ -42,6 +41,13 @@ final awsCredentialsStoreProvider = Provider<AwsCredentialsStore>((ref) {
 /// without serialisation the writes of a connect could overtake the deletes of
 /// the "change configuration" that followed it — resurrecting credentials the
 /// user had just discarded.
+///
+/// Each operation is bounded by [_kOperationTimeout] so one that never settles
+/// cannot hold the queue — and with it the user's `clear()` — indefinitely.
+///
+/// The ordering guarantee holds because [awsCredentialsStoreProvider] is a
+/// plain [Provider], i.e. one instance for the session. Making it `autoDispose`
+/// would hand out fresh chains and revive the race invisibly.
 class AwsCredentialsStore {
   AwsCredentialsStore(this._storage);
 
@@ -54,9 +60,12 @@ class AwsCredentialsStore {
   ///
   /// The chain is never broken by a failure: the tail swallows errors so a
   /// rejected write cannot stop later operations from running, while the
-  /// returned future still reports the failure to this caller.
+  /// returned future still reports the failure to this caller. A timeout counts
+  /// as a failure, which is what stops a stalled operation from blocking the
+  /// queue forever.
   Future<T> _serialize<T>(Future<T> Function() operation) {
-    final result = _pending.then((_) => operation());
+    final result =
+        _pending.then((_) => operation().timeout(_kOperationTimeout));
     _pending = result.then((_) {}, onError: (_) {});
     return result;
   }
@@ -78,7 +87,6 @@ class AwsCredentialsStore {
           'modelId': modelId,
         }),
       );
-      await _deleteLegacyKeys();
       aiLog('[Credentials] Stored');
     });
   }
@@ -90,34 +98,21 @@ class AwsCredentialsStore {
   /// handles it instead of a signing error later.
   Future<StoredAwsCredentials?> read() {
     return _serialize(() async {
-      final raw = await _storage.read(key: _kRecordKey);
-      if (raw == null || raw.isEmpty) return null;
+      final record = await _readRecord();
+      if (record == null) return null;
 
-      final Map<String, dynamic> record;
-      try {
-        final decoded = jsonDecode(raw);
-        if (decoded is! Map<String, dynamic>) return null;
-        record = decoded;
-      } catch (e) {
-        // Corrupted or externally tampered value; treat as not configured.
-        aiLog('[Credentials] Discarding unreadable record');
-        return null;
-      }
-
-      final accessKeyId = (record['accessKeyId'] as String?)?.trim() ?? '';
-      final secretAccessKey =
-          (record['secretAccessKey'] as String?)?.trim() ?? '';
+      final accessKeyId = _stringOrNull(record['accessKeyId']) ?? '';
+      final secretAccessKey = _stringOrNull(record['secretAccessKey']) ?? '';
       // Blank is as unusable as absent, and whitespace reaching the form would
       // overwrite what the user typed with something that cannot connect.
       if (accessKeyId.isEmpty || secretAccessKey.isEmpty) return null;
 
-      final modelId = (record['modelId'] as String?)?.trim();
       return StoredAwsCredentials(
         accessKeyId: accessKeyId,
         secretAccessKey: secretAccessKey,
         // A model may be missing if it was never chosen; the caller falls back
         // to its own default rather than being forced to re-enter everything.
-        modelId: (modelId == null || modelId.isEmpty) ? null : modelId,
+        modelId: _stringOrNull(record['modelId']),
       );
     });
   }
@@ -128,17 +123,8 @@ class AwsCredentialsStore {
   /// meaning to a model preference without credentials to use it with.
   Future<void> storeModelId(String modelId) {
     return _serialize(() async {
-      final raw = await _storage.read(key: _kRecordKey);
-      if (raw == null || raw.isEmpty) return;
-
-      final Map<String, dynamic> record;
-      try {
-        final decoded = jsonDecode(raw);
-        if (decoded is! Map<String, dynamic>) return;
-        record = decoded;
-      } catch (e) {
-        return;
-      }
+      final record = await _readRecord();
+      if (record == null) return;
 
       record['modelId'] = modelId;
       await _storage.write(key: _kRecordKey, value: jsonEncode(record));
@@ -149,15 +135,42 @@ class AwsCredentialsStore {
   Future<void> clear() {
     return _serialize(() async {
       await _storage.delete(key: _kRecordKey);
-      await _deleteLegacyKeys();
       aiLog('[Credentials] Cleared');
     });
   }
 
-  Future<void> _deleteLegacyKeys() async {
-    for (final key in _kLegacyKeys) {
-      await _storage.delete(key: key);
+  /// The stored record as a map, or null when there is nothing usable.
+  ///
+  /// Never throws on a bad value: absent, blank, non-JSON and non-object all
+  /// collapse to null, so both callers can treat "unusable" as "not
+  /// configured" rather than having to catch.
+  Future<Map<String, dynamic>?> _readRecord() async {
+    final raw = await _storage.read(key: _kRecordKey);
+    if (raw == null || raw.isEmpty) return null;
+
+    try {
+      final decoded = jsonDecode(raw);
+      if (decoded is! Map<String, dynamic>) {
+        aiLog('[Credentials] Discarding unreadable record');
+        return null;
+      }
+      return decoded;
+    } catch (_) {
+      // Corrupted or externally tampered value; treat as not configured.
+      aiLog('[Credentials] Discarding unreadable record');
+      return null;
     }
+  }
+
+  /// A trimmed non-empty String, or null for anything else.
+  ///
+  /// A tampered record can hold a number, bool or list where a String belongs.
+  /// Returning null keeps [read]'s "unusable means not configured" contract
+  /// instead of throwing a cast error at the caller.
+  static String? _stringOrNull(Object? value) {
+    if (value is! String) return null;
+    final trimmed = value.trim();
+    return trimmed.isEmpty ? null : trimmed;
   }
 }
 
--- a/lib/page/ai_assistant/services/aws_credentials_store.dart
+++ b/lib/page/ai_assistant/services/aws_credentials_store.dart
@@ -1,0 +1,177 @@
+import 'dart:convert';
+
+import 'package:flutter_riverpod/flutter_riverpod.dart';
+import 'package:flutter_secure_storage/flutter_secure_storage.dart';
+import 'package:privacy_gui/core/utils/logger.dart';
+
+/// Single key holding the whole record.
+///
+/// One key rather than three is what makes a write atomic from a reader's point
+/// of view: a failed write leaves the previous record intact instead of mixing a
+/// new access key with an old secret, which would surface as an opaque signing
+/// failure rather than "not configured".
+const _kRecordKey = 'ai_assistant_aws_credentials';
+
+/// Keys written by the previous three-key layout, removed on write and clear so
+/// stale halves cannot linger after an upgrade.
+const _kLegacyKeys = [
+  'ai_assistant_aws_access_key_id',
+  'ai_assistant_aws_secret_access_key',
+  'ai_assistant_bedrock_model_id',
+];
+
+final awsCredentialsStoreProvider = Provider<AwsCredentialsStore>((ref) {
+  return AwsCredentialsStore(const FlutterSecureStorage());
+});
+
+/// Bedrock credentials entered by the user, held in encrypted storage.
+///
+/// Only exists so the credentials survive an app restart — without it the user
+/// re-types an access key and secret key on every launch. It is deliberately
+/// not a general settings store.
+///
+/// The proper long-term answer is for the app never to hold AWS credentials at
+/// all (a backend proxy would sign the requests), so nothing here should grow
+/// into an abstraction that makes the current arrangement look permanent.
+///
+/// ## Ordering
+///
+/// Every operation runs through [_serialize], so a `store()` can never land
+/// after a `clear()` that the user triggered later. Callers commonly fire these
+/// without awaiting (a storage failure must not break a working session), and
+/// without serialisation the writes of a connect could overtake the deletes of
+/// the "change configuration" that followed it — resurrecting credentials the
+/// user had just discarded.
+class AwsCredentialsStore {
+  AwsCredentialsStore(this._storage);
+
+  final FlutterSecureStorage _storage;
+
+  /// Tail of the operation chain; each new operation is appended to it.
+  Future<void> _pending = Future.value();
+
+  /// Run [operation] after every operation requested before it.
+  ///
+  /// The chain is never broken by a failure: the tail swallows errors so a
+  /// rejected write cannot stop later operations from running, while the
+  /// returned future still reports the failure to this caller.
+  Future<T> _serialize<T>(Future<T> Function() operation) {
+    final result = _pending.then((_) => operation());
+    _pending = result.then((_) {}, onError: (_) {});
+    return result;
+  }
+
+  /// Persist the credentials and the selected model.
+  ///
+  /// Values are never logged.
+  Future<void> store({
+    required String accessKeyId,
+    required String secretAccessKey,
+    required String modelId,
+  }) {
+    return _serialize(() async {
+      await _storage.write(
+        key: _kRecordKey,
+        value: jsonEncode({
+          'accessKeyId': accessKeyId,
+          'secretAccessKey': secretAccessKey,
+          'modelId': modelId,
+        }),
+      );
+      await _deleteLegacyKeys();
+      logger.d('[AI][Credentials] Stored');
+    });
+  }
+
+  /// The stored credentials, or null when nothing usable is saved.
+  ///
+  /// Returns null rather than a half-built record for anything unusable —
+  /// absent, blank, or unparseable — so the caller's "not configured" path
+  /// handles it instead of a signing error later.
+  Future<StoredAwsCredentials?> read() {
+    return _serialize(() async {
+      final raw = await _storage.read(key: _kRecordKey);
+      if (raw == null || raw.isEmpty) return null;
+
+      final Map<String, dynamic> record;
+      try {
+        final decoded = jsonDecode(raw);
+        if (decoded is! Map<String, dynamic>) return null;
+        record = decoded;
+      } catch (e) {
+        // Corrupted or externally tampered value; treat as not configured.
+        logger.d('[AI][Credentials] Discarding unreadable record');
+        return null;
+      }
+
+      final accessKeyId = (record['accessKeyId'] as String?)?.trim() ?? '';
+      final secretAccessKey =
+          (record['secretAccessKey'] as String?)?.trim() ?? '';
+      // Blank is as unusable as absent, and whitespace reaching the form would
+      // overwrite what the user typed with something that cannot connect.
+      if (accessKeyId.isEmpty || secretAccessKey.isEmpty) return null;
+
+      final modelId = (record['modelId'] as String?)?.trim();
+      return StoredAwsCredentials(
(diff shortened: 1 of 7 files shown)


lib/page/ai_assistant/services/aws_credentials_store.dart
```diff
@@ -0,0 +1,177 @@
...
+/// failure rather than "not configured".
+const _kRecordKey = 'ai_assistant_aws_credentials';
+
+/// Keys written by the previous three-key layout, removed on write and clear so
+/// stale halves cannot linger after an upgrade.
+const _kLegacyKeys = [
+  'ai_assistant_aws_access_key_id',
+  'ai_assistant_aws_secret_access_key',
+  'ai_assistant_bedrock_model_id',
+];
```
PeterJhongLinksys: ⚠️ **Warning** 🟢High — `git log --all -S"ai_assistant_aws_secret_access_key" -- lib` returns only `161b4fde`, i.e. this PR's own first commit, which already used the single-key layout. No released version ever wrote these three keys, so this is dead code that costs 3 `delete` calls on every `store()`/`clear()`. Worse, the block is *delete-only* (there is no `_storage.read(key: legacyKey)` anywhere) and never runs on the `read()` path — so a user who upgrades without re-entering credentials keeps a long-lived AWS secret in storage indefinitely. Either remove it (plus the two tests at 107-120 / 252-264), or make `read()` migrate-then-delete and document a removal target.

lib/page/ai_assistant/services/aws_credentials_store.dart
```diff
@@ -0,0 +1,177 @@
...
+  /// Run [operation] after every operation requested before it.
+  ///
+  /// The chain is never broken by a failure: the tail swallows errors so a
+  /// rejected write cannot stop later operations from running, while the
+  /// returned future still reports the failure to this caller.
+  Future<T> _serialize<T>(Future<T> Function() operation) {
+    final result = _pending.then((_) => operation());
+    _pending = result.then((_) {}, onError: (_) {});
+    return result;
+  }
```
PeterJhongLinksys, line 71: ⚠️ **Warning** 🟡Med — [Reinvented Wheel] The repo already has a general mutual-exclusion helper at `lib/core/usp/providers/usp_mutation_lock.dart:28` (`withLock<T>`), which includes timeout protection and stale-lock recovery. Two parallel concurrency primitives will drift apart. Concretely, this one has **no timeout**: a hung `_storage.write` blocks every later operation on the chain forever — including the `clear()` the user triggers by pressing "Change configuration". Consider extracting a shared `serial_queue.dart`, or at minimum adding a timeout here.

lib/page/ai_assistant/services/aws_credentials_store.dart
```diff
@@ -0,0 +1,177 @@
...
+      if (raw == null || raw.isEmpty) return;
+
+      final Map<String, dynamic> record;
+      try {
+        final decoded = jsonDecode(raw);
+        if (decoded is! Map<String, dynamic>) return;
+        record = decoded;
+      } catch (e) {
+        return;
+      }
```
PeterJhongLinksys: ⚠️ **Warning** 🟡Med — This swallows a corrupt record silently (`e` is unused), while the same failure in `read()` (101-105) logs `Discarding unreadable record`. Inconsistent handling of one failure class means a silent `storeModelId` failure leaves no diagnostic trace. Also note the read → blank-check → `jsonDecode` → `is! Map` sequence here duplicates 93-105; extracting a shared `Future<Map<String, dynamic>?> _readRecord()` would fix both.

test/ai/ai_logging_test.dart
```diff
@@ -0,0 +1,41 @@
...
+        return 'secret';
+      });
+
+      // In a debug test run kDebugMode is true, so the callback does run. The
+      // property under test is that the *callback* is what gates the work: a
+      // release build skips it entirely, so no sensitive string is ever
+      // interpolated or held. Asserting on kDebugMode keeps this honest in
+      // whichever mode the suite is run.
+      expect(built, kDebugMode ? 1 : 0);
+    });
```
PeterJhongLinksys, line 38: ⚠️ **Warning** 🟢High — This assertion derives its expected value from the condition under test, so it passes in every build mode and can never fail. Replacing `kDebugMode` with `kProfileMode` in `ai_logging.dart`, or removing the guard entirely, would leave this suite green — yet that guard is the PR's central security claim. The release `orElse` branch (`ai_logging.dart:38-40`), which six call sites depend on, has no coverage at all. Suggest introducing an injectable seam (a `bool Function()` gate or `@visibleForTesting` override) so the test can actually assert "in release, `message()` is not called and `orElse()` is".

PeterJhongLinksys (approved):
🤖 **Automated Review — Round 1** · `0b97fab..a7f3b0a` (full)

**Verdict: ✅ APPROVE** — No Critical issues found; the credential-persistence design (single atomic record, serialised operations) and the sensitive/structural log split are both sound. Approving with a set of Warnings worth addressing, most notably the ServiceError boundary, the public API naming, and two tests whose names promise more than they assert.

|---|---|---|---|
| ⚠️ | 🟢 | `lib/page/ai_assistant/services/aws_credentials_store.dart:15-21,81,152` | [XX] Legacy-key "migration" is delete-only, never runs on the `read()` path, and `git log -S` shows the three legacy keys first appear **in this PR** — so it is dead code, and for any user who upgrades without re-entering, a long-lived AWS secret would linger forever. |
| ⚠️ | 🟢 | `aws_credentials_store.dart:67-84,129-147,149-155` ↔ `router_assistant_view.dart:113-117,200-202,468-470,540-542` | Art. XIII violated: Service has no try/catch, so raw `PlatformException` escapes to the View, which catches it directly. `StorageError` already exists in `lib/core/errors/service_error.dart:178` and is unused. |
| ⚠️ | 🟢 | `router_assistant_view.dart:100,206,282-291` | Art. XIII §13.6: raw `e.toString()` is stored in `_configError` and rendered to the user; `_localizeConfigError` fallback `return error;` surfaces technical strings. Also uses string-prefix + `substring` where a sealed `ServiceError` belongs. |
| ⚠️ | 🟢 | `test/ai/ai_logging_test.dart:8-20` | The PR's core security claim is untested: `expect(built, kDebugMode ? 1 : 0)` derives the expectation from the condition under test, so it passes in every mode. The release `orElse` branch (`ai_logging.dart:38-40`) has zero coverage. |
| ⚠️ | 🟢 | `test/page/ai_assistant/services/aws_credentials_store_test.dart:317-349` | Test named `never logs credential values` makes no assertion about logging at all — it only checks the JSON written to storage. Gives false safety confidence. |
| ⚠️ | 🟢 | `aws_credentials_store.dart:45,165-177` | Art. III/XI naming: `AwsCredentialsStore` should be `[Feature]Service`; `StoredAwsCredentials` is a UI-facing model but neither ends in `UIModel` nor implements `Equatable`. Cheapest to rename now. |
| ⚠️ | 🟢 | `router_assistant_view.dart:42-210,519-564` | Art. V §5.4: the View owns a 7-field state machine, async I/O orchestration, race arbitration and persistence — i.e. Notifier responsibilities. Consequence: none of it is testable, so the documented race protections are unverified. |
| ⚠️ | 🟢 | `router_assistant_view.dart:61,109-127` + `aws_credentials_store.dart:91-93` | `_isRestoring` has no timeout and is only cleared after `await store.read()`; if that never completes the user is permanently locked out of the config screen (text fields `readOnly`, dropdown and Connect disabled — no escape hatch, no error shown). |
| ⚠️ | 🟢 | `router_assistant_view.dart:65,124,146-149` | [Speculative Generality] `_userTookOver` can never be `true` at its only read site (124): every input path is already sealed by `_isRestoring` (396/410/431/457). Two overlapping guards, the second is a dead condition. |
| ⚠️ | 🟡 | `router_assistant_view.dart:537-542` | `clear()` failure is only logged, then the UI reports success; on next launch `_restoreSavedCredentials()` silently restores and auto-connects with credentials the user believed they had revoked. |
| ⚠️ | 🟡 | `router_assistant_view.dart:543-558` | [XX] "Change configuration" nulls `_controller` without disposing it (a `ChangeNotifier`); an in-flight `_processConversation()` loop keeps calling Bedrock with the credentials the user just replaced. |
| ⚠️ | 🟡 | `lib/ai/orchestrator/router_chat_controller.dart:262,424` | Generic `catch (e)` routes through the always-on `aiLog`; if the `generative_ui` Bedrock exception message carries the request body (system prompt = full `routerContext`), exactly the data this PR removes leaks back into release logs. Endpoint is an external dep — not confirmable, hence not Critical. |
| ⚠️ | 🟡 | `router_assistant_view.dart:753-756` | Release branch logs raw `data['action']` (model output, arbitrary string). `logger.dart:113` `_logTagRegex` parses it into tags, so an `action` containing `\n[USPClient]:...` can forge sections of the support log. |
| ⚠️ | 🟡 | `aws_credentials_store.dart:23-25` (+ web target) | This app ships to web (`kIsWeb` used in 33 places); on web `flutter_secure_storage` degrades to `localStorage`, so "held in encrypted storage" over-promises for a long-lived AWS access key + secret. |
| ⚠️ | 🟡 | `router_assistant_view.dart:109-143` + `lib/main.dart:97-107` | No logout/router-switch hook clears the credentials — only `checkFirstLaunch()`'s `deleteAll()` (first install only). Next user of the same device/browser auto-connects with the previous user's AWS credentials. |
| ⚠️ | 🟡 | `lib/ai/ai_logging.dart:1-41` | New file sits at `lib/ai/` root, outside all five sub-modules and not exported by `_ai.dart:11-15`, forcing `router_assistant_view.dart` to import both the barrel and a deep path. A cross-cutting logging concern's home in this repo is `lib/core/utils/`. |
| ⚠️ | 🟡 | `aws_credentials_store.dart:93-105` vs `131-141` | [Duplicated Code] read-raw → blank check → `jsonDecode` → `is! Map` is written twice, differing only in the failure return. Extract `_readRecord()`. |
| ⚠️ | 🟡 | `aws_credentials_store.dart:51-62` | [Reinvented Wheel] `_pending`/`_serialize` is a hand-rolled mutex while `lib/core/usp/providers/usp_mutation_lock.dart:28` already provides `withLock<T>` with timeout + stale-lock recovery. This version has no timeout: one hung `write` blocks every later operation, including the user's `clear()`. |
| ⚠️ | 🟡 | `router_assistant_view.dart:85-90` vs `179-184` | [Duplicated Code] controller construction + `addListener` duplicated; line 179 overwrites `_controller` without `removeListener` on the old instance (line 544 does). Only safe via the implicit "must be null here" invariant. |
| ⚠️ | 🟡 | `aws_credentials_store.dart:139-141` | `storeModelId` swallows a corrupt record entirely (`catch (e) { return; }`, `e` unused), while `read():101-105` logs `Discarding unreadable record`. Inconsistent handling of the same failure. |
| ⚠️ | 🟡 | `router_assistant_view.dart:14-28` | `BedrockModel` is a domain model living in a View file with no `Equatable`; `AppDropdown<BedrockModel>` comparison works today only because values are const instances from `BedrockModel.models`. |
| ⚠️ | 🟡 | `aws_credentials_store_test.dart` (whole file) | No Test Data Builder — the `AKIAEXAMPLE`/`secret`/`model-a` triple is inlined in 10+ places. Art. VIII asks for `test/mocks/test_data/[feature]_test_data.dart`. |
| ⚠️ | 🟡 | `lib/ai/providers/usp_command_provider.dart:4-17` ↔ `lib/page/ai_assistant/providers/router_command_provider.dart:2` | Pre-existing dependency cycle across layers: `lib/ai` is imported by the page layer and imports 14 page-layer L1 providers. Touched by this PR; worth a follow-up via `lib/ai/abstraction/`. |
| 💡 | 🟢 | `aws_credentials_store.dart:81,152` | `_deleteLegacyKeys()` runs after the record write with no try/catch: a failed legacy delete rejects `store()`, so the caller reports "Could not save credentials" even though the record was saved; a mid-loop throw also leaves some legacy keys behind. |
| 💡 | 🟢 | `aws_credentials_store_test.dart:14,35,38,56,71` | [Dead Code] `FakeSecureStorage.log` is written in five places and asserted on nowhere. Either delete it or use it to pin the ordering contract. |
| 💡 | 🟡 | `router_assistant_view.dart:110` / `aws_credentials_store.dart:23` | The `_serialize` ordering guarantee silently depends on the provider being a non-`autoDispose` singleton. Correct today, but undocumented and untested — switching to `.autoDispose` would revive the race invisibly. |
| 💡 | 🟡 | `router_assistant_view.dart:136-142` | If `_initControllerWithManualConfig(persist: false)` throws, the screen stays on config with the restored access key already filled into a non-obscured `AppTextField` and `_userTookOver` still `false`. |
| 💡 | 🟡 | `router_assistant_view.dart:94-100,206` | `aiLog('...Config error: $e')` and `_configError` unconditionally record the raw exception; AWS SDK signing errors commonly echo the access key id. Prefer `e.runtimeType` + known messages. |
| 💡 | 🟡 | `lib/ai/ai_logging.dart` vs `lib/core/utils/logger.dart:66-84` | [Divergent Change] the repo already has centralised redaction (`MaskingUtils` applied in `CustomOutput.output`); this PR adds a second, call-site-based mechanism. Answering "can field X reach a release log?" now needs both rule sets. Consider adding IP/SSID/hostname to `MaskingUtils`. |
| 💡 | 🟡 | `usp_command_provider.dart:312,341,469,490,527,766` | Pre-existing: `.valueOrNull` swallows `AsyncValue.error` into a fixed string, so no `mapUspErrorToServiceError()` translation and the model cannot tell "unsupported" from "request failed". |
| 💡 | 🟡 | `router_assistant_view.dart` (22 × `SizedBox`) | Art. XV: no `AppGap.*` anywhere in the file despite `APPGAP_MAPPING.md`. Mostly pre-existing; please use `AppGap` for newly added spacing. |
| 💡 | ⚪ | `router_assistant_view.dart:72-77` | If env config succeeds, previously saved credentials are never read *and* never cleared — they just sit in storage. |
| 💡 | ⚪ | `router_assistant_view.dart:396,410` | `readOnly: _isRestoring` leaves fields looking editable while silently discarding input; a disabled/loading state would be honest. |
| 💡 | ⚪ | `router_assistant_view.dart:20-27,175` | `region: 'us-west-2'` and five Bedrock model ids hard-coded in the Presentation layer (Magic Values). |
| 💡 | ⚪ | `lib/ai/providers/usp_command_provider.dart:19`, `router_chat_controller.dart:95` | [Middle Man] `void _log(String m) => aiLog(m);` is a pure pass-through kept at 60+ call sites; understandable to keep the diff small, but it is a permanent extra hop. |
| 💡 | ⚪ | `usp_command_provider.dart:673` | `_getTrafficStats` still logs up/down rates via `aiLog` (release). By this PR's own structural-vs-user-data split, traffic rates describe user network behaviour. |
| 💡 | ⚪ | `ai_logging.dart:35-40` | `orElse` makes debug and release emit *different text* for the same call site, so field logs cannot be diffed line-by-line against a dev run. Consider making the release text a strict prefix or tagging it `[redacted]`. |

<sub>Confidence: 🟢High = read the head-version code and can quote evidence · 🟡Med = file:line + reasoning, not fully proven · ⚪Low = speculative.</sub>
<sub>`[X]` = raised by 1 agent · `[XX]` = independently raised by both agents (Critical/notable convergence only).</sub>

<details>
<summary>🔴 Critical — none</summary>

Neither reviewer could establish a complete evidence chain for a Critical. Two candidates were deliberately downgraded to Warning under the evidence rules:

1. **`router_chat_controller.dart:424` — exception text may carry the full system prompt into release logs.**
```dart
} catch (e) {
  _log('_processConversation: unexpected error: $e');   // aiLog → written in release
  _state = _state.withError('An unexpected error occurred: $e');
```
`logger.dart:37-40` shows the release minimum level is `Level.debug`, and `CustomOutput` (L61-70) writes debug lines to the log file / web cache. Whether the leak is real depends on how `generative_ui` composes its Bedrock exception messages — an external git dependency whose source could not be read, so the endpoint is unproven. **Warning 🟡**, not Critical.

2. **`router_assistant_view.dart:753-756` — log injection via model output.** Forging tag sections in a support log is misleading-content, not a privilege bypass. **Warning 🟡**.
</details>

<details>
<summary>⚠️ Warning — evidence details</summary>

**Legacy-key migration is dead code and incomplete (both agents).** `_kLegacyKeys` is consumed only by:
```dart
Future<void> _deleteLegacyKeys() async {
  for (final key in _kLegacyKeys) { await _storage.delete(key: key); }
}
```
There is no `_storage.read(key: legacyKey)` anywhere in the file, so this is deletion, not migration. `git log --all -S"ai_assistant_aws_secret_access_key" -- lib` returns only `161b4fde` — **this PR's own first commit**, which already used the single-key layout. Two consequences: (a) if a three-key layout ever did exist in a dev build, the user's credentials are silently destroyed rather than migrated; (b) because cleanup only fires from `store()`/`clear()`, a user who upgrades and never re-enters credentials takes the `read()` path, leaving `ai_assistant_aws_secret_access_key` — a long-lived AWS secret — in storage indefinitely. Either delete the block (and its two tests at 107-120 / 252-264), or make `read()` migrate-then-delete and document the removal target.

**No escape from the restore window.** `_isRestoring = false` appears only after `await store.read()` (L125, L137) and in the Change dialog (L556) — but the Change button lives on the chat screen (`_needsConfig == false`), which is unreachable while `_needsConfig == true`. Meanwhile:
```dart
readOnly: _isRestoring,                                // L396, L410
onChanged: _isRestoring ? null : ...,                  // L457
onTap: (_isConfiguring || _isRestoring) ? null : ...,  // L431
```
A `read()` that never completes therefore locks the user out with no error message. Add `.timeout(...)` or clear the flag in a `finally`.

**ServiceError boundary.** Art. XIII §13.1 makes the Service layer the only place allowed to `catch (e)` and the UI layer see `ServiceError` only. `store()`, `clear()` and `storeModelId()` contain no try/catch, so `FlutterSecureStorage`'s `PlatformException` reaches four `.catchError((Object e) {...})` sites plus a `try/catch` in the View. `StorageError` (`lib/core/errors/service_error.dart:178`) and its localization (`service_error_localizations.dart:49`) already exist. Note the repo's USP convention when wrapping: use `catch (e)`, never `on Exception catch`.

**Tautological / mislabelled tests.** `expect(built, kDebugMode ? 1 : 0)` cannot fail; replacing `kDebugMode` with `kProfileMode` in `ai_logging.dart` would keep the suite green. Likewise `never logs credential values` asserts only on the stored JSON. Introducing an injectable gate (`bool Function()` or `@visibleForTesting`) plus an observable log sink would let both tests actually hold their contracts — including the currently uncovered `orElse` release branch used by six call sites.
</details>

<details>

- **USP Provider hard rules all pass**: L1 reads use `_ref.read(...)` throughout (312, 341, 398, 424, 469, 490, 527…), no new L1 provider and no `.autoDispose`, no mutations added so no missing `withLock`, no `PreservableContract` redefinition.
- **Generated-model isolation is clean**: zero `generated/` imports under `lib/ai/` and `lib/page/ai_assistant/` (Art. V §5.4).
- **The single-key atomic record is the right call**, and it is pinned by a test that explains *why* (`a failed re-save leaves the previous record intact`) — a failed write can no longer pair a new access key with an old secret and surface as an opaque signing error.
- **The `operation ordering` test group (267-323) is the strongest part of this PR**: it uses a latency-controllable fake to reproduce "an un-awaited `store()` overtaking a later `clear()`" and asserts that deleted credentials are never resurrected. `_serialize`'s error isolation (`_pending = result.then((_) {}, onError: (_) {})`) is also correct — failures don't stall the queue yet still reach the caller.
- **`read()`'s defensive normalisation is thorough**: null / empty / whitespace-only / non-JSON / non-Map / missing modelId all collapse to "not configured" instead of deferring the failure to request signing, and each case has a test.
- **`aiLogSensitive` taking a callback rather than a String** is the right shape — in release the interpolation never happens at all, rather than being built and thrown away. `orElse` preserving counts/status keeps field diagnosability without emitting two lines per event.
- **`catch (e)` rather than `on Exception catch`** throughout `lib/ai/` and `lib/page/ai_assistant/`, correct for USP's raw-String errors.
- **Credential teardown in the change branch (543-558) is thorough** — controller, both fields, model selection and storage — with a comment explaining why a stale model id would silently carry over to the next account.
- Eight `logger.d` calls that would have printed conversation content and A2UI payloads into release logs are removed, which is exactly the stated goal.
- Service dependencies are constructor-injected and exposed via `Provider<T>`, no navigation, no service-to-service calls (Art. VI).
- The "why, not what" comment quality is consistently high across the whole PR.

</details>

PR #1205, opened 2026-08-05
feat(e2e): add stable identifiers to dialog/CRUD controls (#1172)

AustinChangLinksys:
## What

Group 2 (write-proof follow-up) of #1172. The list-row trigger hooks
(`pf-add/edit/delete-*`, `wifi-quick-setup`) already shipped in dev-2.7.0, but the
**dialog-internal controls** those rows open still lacked stable E2E anchors, so the
write-proof tests (PrivacyGUI-USP-E2E#9) could only reach them via fragile localized
text. This adds the remaining identifiers.

## Changes

| identifier | control | file |
|---|---|---|
| `instant-privacy-enable-confirm` | Instant Privacy enable-confirm button | `instant_privacy_view.dart` |
| `pf-delete-confirm` | PF delete-confirm button | `usp_single_port_tab.dart` |
| `pf-single-cancel` / `pf-single-submit` | PF single-port dialog actions | `port_forwarding_dialog.dart` |
| `wifi-ssid-name-input` | SSID rename dialogs (network + quick-setup) | `wifi_network_card.dart`, `wifi_quick_setup_card.dart` |
| `admin-timezone-item-${timeZoneID}` | timezone option rows | `timezone_edit_dialog.dart` |

Notes:
- The PF single-port dialog's actions were Material `TextButton`/`FilledButton`
  (can't carry an `identifier`), so they were converted to `AppButton.text` — matching
  the port-range dialog. `pf-single-cancel` added alongside `-submit` for symmetry.
- The SSID rename `_editSsid` exists in **two** components (regular network + quick
  setup); both got the same `wifi-ssid-name-input`.
- Timezone rows keyed by the stable `timeZoneID` (e.g. `admin-timezone-item-PST8`),
  not the localized label or row index (Article XVI §16.3).

## No ui_kit change needed

An earlier concern (issue-body comment) suggested `AppTextField` needed
`excludeSemantics` for typing to reach the controller. Investigated and **retracted**
(linksys/privacyGUI-UI-kit#7, closed): CanvasKit typing into the controller is gated by
the engine's 500ms `GestureMode` debounce and is solved **test-side** (`el.focus()` +
`activeElement` poll + `keyboard.type()`, already in the E2E suite's
`fillInputById`/`focusAndType`). The `identifier` passthrough shipped in ui_kit v2.31.0
is sufficient; no node-placement change is required.

## Verification

- `flutter analyze` on all 6 changed files: clean
[...]

PR #1206, opened 2026-08-05
refactor(ui): migrate port/service rule dialogs to AppDialog (#1166)

### LINKED ISSUE #1166: refactor: migrate port/service rule dialogs from AlertDialog to UI Kit dialog pattern
## Summary

Three port/service rule dialogs render with raw Material `AlertDialog` + `Text()`
instead of the UI Kit dialog pattern (`showSimpleAppDialog` / UI Kit widgets).
This is **pre-existing tech debt** — surfaced by the pre-PR readiness review on
`feat/e2e-usp-transport` (the counts of `AlertDialog`/`Text(` are identical in
base and head, so that branch did **not** introduce it).

## Files

- `lib/page/dashboard/views/dialogs/port_forwarding_dialog.dart`
- `lib/page/ipv6_port_service/views/dialogs/ipv6_port_service_rule_dialog.dart`
- `lib/page/port_forwarding/views/dialogs/port_range_forwarding_dialog.dart`

All three are `StatefulWidget` dialogs with their own `TextEditingController`s,
`FocusNode`s, and focus-loss validation.

## Why it wasn't fixed in the transport/E2E PR

Converting these to `showSimpleAppDialog` / UI Kit means restructuring each
StatefulWidget dialog into the UI Kit builder pattern — a sizeable, higher-risk
refactor. Those dialogs were just modified (validation moved to focus-loss +
`semanticLabel`s added for E2E), so stacking a structural rewrite on top would
compound risk. Kept out of that PR deliberately; tracked here for a focused
follow-up.

## Scope / acceptance

- Replace `AlertDialog` with the UI Kit dialog entry point (`showSimpleAppDialog`
  or equivalent), and raw `Text()` with `AppText`, in all three dialogs.
- **Preserve behavior**: focus-loss validation, `_isFormValid`/`_hasRequiredInput`
  submit gating, and the `semanticLabel`s (`port-forwarding-*` etc.) that E2E
  relies on must all still work.
- Keep the golden tests green (regenerate baselines only if the visual change is
  intentional).
- Verify the E2E add/edit-rule flows still pass (they locate fields by the inner
  hintText node — see the ui_kit `AppTextField` `excludeSemantics` follow-up
  noted in the E2E repo's `docs/test-inventory.md`).

## Related

- Found during the E2E testing initiative (`feat/e2e-usp-transport`).
- See also the ui_kit follow-up: `AppTextField` wraps its inner `TextField` in
  `Semantics(textField, label)` without `excludeSemantics`, producing two textbox
  nodes for a labelled field.

AustinChangLinksys:
## Summary

Migrates the port-forwarding and IPv6 port-service rule dialogs from raw
Material `AlertDialog` to the UI Kit `AppDialog` pattern, closing the
tech-debt tracked in #1166.

## Stacking note

> ⚠️ **This PR is stacked on #1205** (`feat/1172-e2e-dialog-identifiers`),
> not on `dev-2.7.0` directly. The `port_forwarding_dialog` action buttons
> were already migrated to `AppButton.text` in #1205; basing on that branch
> avoids touching the same lines and conflicting.
>
> **After #1205 merges into `dev-2.7.0`, re-target this PR's base to
> `dev-2.7.0`.** GitHub will recompute the diff automatically.

## Changes

- **`port_forwarding_dialog.dart`**: `AlertDialog` → `AppDialog`, title
  `Text()` → `AppText.titleLarge`, dropped `SingleChildScrollView` in favor
  of `scrollable: true`. Action buttons were already `AppButton.text`
  (from #1205), left untouched.
- **`ipv6_port_service_rule_dialog.dart`**: same dialog-shell migration,
  plus action buttons `TextButton`/`FilledButton` → `AppButton.text` (new
  identifiers `ipv6-rule-cancel` / `ipv6-rule-submit`).
- The third dialog from the original issue (`port_range_forwarding_dialog`)
  already used `AppDialog` and is **not** part of this PR.

## Behavior preserved

- Focus-loss validation, `_isFormValid` / `_hasRequiredInput` submit gating,
  and all E2E `identifier`s remain intact.
- Callers already used `showAppDialog` — no caller changes needed.
- Remaining raw `Text()` is only `SegmentedButton.ButtonSegment.label`
  (Material API requires a `Widget`).

## Issue description drift (heads-up)

Issue #1166's acceptance criteria are partly stale:
- The **third dialog** (`port_range_forwarding_dialog`) is already done.
- **`semanticLabel`s were replaced by `identifier`s** in #1172/#1205, so the
  "preserve semanticLabels" criterion should read "preserve identifiers".

## E2E SSOT follow-up (Article XVI §16.4)

Two new identifiers (`ipv6-rule-cancel`, `ipv6-rule-submit`) are added. The
E2E selector map is derived from app source — regenerate it (`npm run gen:ids`
in the e2e repo) and migrate any spec that located the IPv6 dialog cancel/submit
via `getByRole({name})` to `byIdentifier()`.

## Testing

- `flutter analyze`: no issues.
- Golden tests (both pages): 44/44 pass; baselines regenerated locally, all
  interaction flows (open / input / validate / submit) render without error.
  Baselines are gitignored, so no golden files are committed.
- Functional provider/service tests: 94/94 pass.

Closes #1166

### DIFF
--- a/lib/page/dashboard/views/dialogs/port_forwarding_dialog.dart
+++ b/lib/page/dashboard/views/dialogs/port_forwarding_dialog.dart
@@ -270,13 +270,15 @@ class _PortForwardingDialogState extends State<PortForwardingDialog> {
         ),
       ),
       actions: [
-        TextButton(
-          onPressed: () => Navigator.of(context).pop(),
-          child: Text(loc(context).cancel),
+        AppButton.text(
+          identifier: 'pf-single-cancel',
+          label: loc(context).cancel,
+          onTap: () => Navigator.of(context).pop(),
         ),
-        FilledButton(
-          onPressed: _isFormValid(context) ? _submit : null,
-          child: Text(_isEdit ? loc(context).save : loc(context).add),
+        AppButton.text(
+          identifier: 'pf-single-submit',
+          label: _isEdit ? loc(context).save : loc(context).add,
+          onTap: _isFormValid(context) ? _submit : null,
         ),
       ],
     );
--- a/lib/page/dashboard/views/dialogs/port_forwarding_dialog.dart
+++ b/lib/page/dashboard/views/dialogs/port_forwarding_dialog.dart
@@ -186,88 +186,85 @@ class _PortForwardingDialogState extends State<PortForwardingDialog> {
 
   @override
   Widget build(BuildContext context) {
-    return AlertDialog(
-      title: Text(_isEdit
+    return AppDialog(
+      title: AppText.titleLarge(_isEdit
           ? loc(context).editPortForwarding
           : loc(context).addPortForwarding),
-      content: SingleChildScrollView(
-        child: Column(
-          mainAxisSize: MainAxisSize.min,
-          children: [
-            AppTextField(
-              controller: _descController,
-              focusNode: _descFocus,
-              identifier: 'pf-single-description',
-              hintText: loc(context).description,
-              errorText: _errors['description'],
-              onChanged: (_) => _onInputChanged(),
-            ),
-            AppGap.lg(),
-            AppTextField(
-              controller: _extPortController,
-              focusNode: _extPortFocus,
-              identifier: 'pf-single-external-port',
-              hintText: loc(context).externalPort,
-              keyboardType: TextInputType.number,
-              errorText: _errors['externalPort'],
-              onChanged: (_) => _onInputChanged(),
-            ),
-            AppGap.lg(),
-            AppTextField(
-              controller: _intPortController,
-              focusNode: _intPortFocus,
-              identifier: 'pf-single-internal-port',
-              hintText: loc(context).internalPort,
-              keyboardType: TextInputType.number,
-              errorText: _errors['internalPort'],
-              onChanged: (_) => _onInputChanged(),
-            ),
-            AppGap.lg(),
-            AppSelectAutoComplete(
-              options: widget.deviceOptions,
+      scrollable: true,
+      content: Column(
+        mainAxisSize: MainAxisSize.min,
+        children: [
+          AppTextField(
+            controller: _descController,
+            focusNode: _descFocus,
+            identifier: 'pf-single-description',
+            hintText: loc(context).description,
+            errorText: _errors['description'],
+            onChanged: (_) => _onInputChanged(),
+          ),
+          AppGap.lg(),
+          AppTextField(
+            controller: _extPortController,
+            focusNode: _extPortFocus,
+            identifier: 'pf-single-external-port',
+            hintText: loc(context).externalPort,
+            keyboardType: TextInputType.number,
+            errorText: _errors['externalPort'],
+            onChanged: (_) => _onInputChanged(),
+          ),
+          AppGap.lg(),
+          AppTextField(
+            controller: _intPortController,
+            focusNode: _intPortFocus,
+            identifier: 'pf-single-internal-port',
+            hintText: loc(context).internalPort,
+            keyboardType: TextInputType.number,
+            errorText: _errors['internalPort'],
+            onChanged: (_) => _onInputChanged(),
+          ),
+          AppGap.lg(),
+          AppSelectAutoComplete(
+            options: widget.deviceOptions,
+            controller: _intClientController,
+            onSelected: (_) => _validate(context),
+            child: AppTextField(
               controller: _intClientController,
-              onSelected: (_) => _validate(context),
-              child: AppTextField(
-                controller: _intClientController,
-                focusNode: _intClientFocus,
-                identifier: 'pf-single-internal-ip',
-                hintText: loc(context).internalIpHint,
-                errorText: _errors['internalClient'],
-                onChanged: (_) => _onInputChanged(),
+              focusNode: _intClientFocus,
+              identifier: 'pf-single-internal-ip',
+              hintText: loc(context).internalIpHint,
+              errorText: _errors['internalClient'],
+              onChanged: (_) => _onInputChanged(),
+            ),
+          ),
+          AppGap.lg(),
+          Row(
+            mainAxisAlignment: MainAxisAlignment.spaceBetween,
+            children: [
+              AppText.bodyMedium(loc(context).protocol),
+              SegmentedButton<String>(
+                segments: [
+                  ButtonSegment(value: 'TCP', label: Text(loc(context).tcp)),
+                  ButtonSegment(value: 'UDP', label: Text(loc(context).udp)),
+                  ButtonSegment(value: 'Both', label: Text(loc(context).both)),
+                ],
+                selected: {_protocol},
+                onSelectionChanged: (v) => setState(() => _protocol = v.first),
               ),
-            ),
-            AppGap.lg(),
-            Row(
-              mainAxisAlignment: MainAxisAlignment.spaceBetween,
-              children: [
-                AppText.bodyMedium(loc(context).protocol),
-                SegmentedButton<String>(
-                  segments: [
-                    ButtonSegment(value: 'TCP', label: Text(loc(context).tcp)),
-                    ButtonSegment(value: 'UDP', label: Text(loc(context).udp)),
-                    ButtonSegment(
-                        value: 'Both', label: Text(loc(context).both)),
-                  ],
-                  selected: {_protocol},
-                  onSelectionChanged: (v) =>
-                      setState(() => _protocol = v.first),
-                ),
-              ],
-            ),
-            AppGap.lg(),
-            Row(
-              mainAxisAlignment: MainAxisAlignment.spaceBetween,
-              children: [
-                AppText.bodyMedium(loc(context).enabled),
-                AppSwitch(
-                  identifier: 'pf-single-enabled',
-                  value: _enabled,
-                  onChanged: (value) => setState(() => _enabled = value),
-                ),
-              ],
-            ),
-          ],
-        ),
+            ],
+          ),
+          AppGap.lg(),
+          Row(
+            mainAxisAlignment: MainAxisAlignment.spaceBetween,
+            children: [
+              AppText.bodyMedium(loc(context).enabled),
+              AppSwitch(
+                identifier: 'pf-single-enabled',
+                value: _enabled,
+                onChanged: (value) => setState(() => _enabled = value),
+              ),
+            ],
+          ),
+        ],
       ),
       actions: [
         AppButton.text(
--- a/lib/page/ipv6_port_service/views/dialogs/ipv6_port_service_rule_dialog.dart
+++ b/lib/page/ipv6_port_service/views/dialogs/ipv6_port_service_rule_dialog.dart
@@ -129,86 +129,87 @@ class _Ipv6PortServiceRuleDialogState extends State<Ipv6PortServiceRuleDialog> {
 
   @override
   Widget build(BuildContext context) {
-    return AlertDialog(
-      title: Text(_isEdit ? loc(context).editRule : loc(context).addRule),
-      content: SingleChildScrollView(
-        child: Column(
-          mainAxisSize: MainAxisSize.min,
-          children: [
-            AppTextField(
-              controller: _descriptionController,
-              focusNode: _descriptionFocus,
-              identifier: 'ipv6-rule-description',
-              hintText: loc(context).ruleName,
-              errorText: _errors['description'],
+    return AppDialog(
+      title: AppText.titleLarge(
+          _isEdit ? loc(context).editRule : loc(context).addRule),
+      scrollable: true,
+      content: Column(
+        mainAxisSize: MainAxisSize.min,
+        children: [
+          AppTextField(
+            controller: _descriptionController,
+            focusNode: _descriptionFocus,
+            identifier: 'ipv6-rule-description',
+            hintText: loc(context).ruleName,
+            errorText: _errors['description'],
+            onChanged: (_) => _onInputChanged(),
+          ),
+          AppGap.lg(),
+          AppSelectAutoComplete(
+            options: widget.deviceOptions,
+            controller: _ipv6Controller,
+            onSelected: (_) => _validate(),
+            child: AppTextField(
+              controller: _ipv6Controller,
+              focusNode: _ipv6Focus,
+              identifier: 'ipv6-rule-address',
+              hintText: loc(context).ipv6AddressSearchHint,
+              errorText: _errors['ipv6Address'],
               onChanged: (_) => _onInputChanged(),
             ),
-            AppGap.lg(),
-            AppSelectAutoComplete(
-              options: widget.deviceOptions,
-              controller: _ipv6Controller,
-              onSelected: (_) => _validate(),
-              child: AppTextField(
-                controller: _ipv6Controller,
-                focusNode: _ipv6Focus,
-                identifier: 'ipv6-rule-address',
-                hintText: loc(context).ipv6AddressSearchHint,
-                errorText: _errors['ipv6Address'],
-                onChanged: (_) => _onInputChanged(),
+          ),
+          AppGap.lg(),
+          Row(
+            mainAxisAlignment: MainAxisAlignment.spaceBetween,
+            children: [
+              AppText.bodyMedium(loc(context).protocol),
+              SegmentedButton<String>(
+                segments: UspIpv6PortServiceService.protocolOptions
+                    .map(
+                        (name) => ButtonSegment(value: name, label: Text(name)))
+                    .toList(),
+                selected: {_protocol},
+                onSelectionChanged: (v) => setState(() => _protocol = v.first),
               ),
-            ),
-            AppGap.lg(),
-            Row(
-              mainAxisAlignment: MainAxisAlignment.spaceBetween,
-              children: [
-                AppText.bodyMedium(loc(context).protocol),
-                SegmentedButton<String>(
-                  segments: UspIpv6PortServiceService.protocolOptions
-                      .map((name) =>
-                          ButtonSegment(value: name, label: Text(name)))
-                      .toList(),
-                  selected: {_protocol},
-                  onSelectionChanged: (v) =>
-                      setState(() => _protocol = v.first),
-                ),
-              ],
-            ),
-            AppGap.lg(),
-            AppRangeInput(
-              startController: _startPortController,
-              endController: _endPortController,
-              startFocusNode: _startPortFocus,
-              endFocusNode: _endPortFocus,
-              startLabel: loc(context).startPort,
-              endLabel: loc(context).endPort,
-              startIdentifier: 'ipv6-rule-start-port',
-              endIdentifier: 'ipv6-rule-end-port',
-              errorText: _errors['startPort'] ?? _errors['endPort'],
-              onChanged: (_, __) => _onInputChanged(),
-            ),
-            AppGap.lg(),
-            Row(
-              mainAxisAlignment: MainAxisAlignment.spaceBetween,
-              children: [
-                AppText.bodyMedium(loc(context).enabled),
-                AppSwitch(
-                  identifier: 'ipv6-rule-enabled',
-                  value: _enabled,
-                  onChanged: (value) => setState(() => _enabled = value),
-                ),
-              ],
-            ),
-          ],
-        ),
+            ],
+          ),
+          AppGap.lg(),
+          AppRangeInput(
+            startController: _startPortController,
+            endController: _endPortController,
+            startFocusNode: _startPortFocus,
+            endFocusNode: _endPortFocus,
+            startLabel: loc(context).startPort,
+            endLabel: loc(context).endPort,
+            startIdentifier: 'ipv6-rule-start-port',
+            endIdentifier: 'ipv6-rule-end-port',
+            errorText: _errors['startPort'] ?? _errors['endPort'],
+            onChanged: (_, __) => _onInputChanged(),
+          ),
+          AppGap.lg(),
+          Row(
+            mainAxisAlignment: MainAxisAlignment.spaceBetween,
+            children: [
+              AppText.bodyMedium(loc(context).enabled),
+              AppSwitch(
+                identifier: 'ipv6-rule-enabled',
+                value: _enabled,
+                onChanged: (value) => setState(() => _enabled = value),
+              ),
+            ],
+          ),
+        ],
       ),
       actions: [
-        TextButton(
-          onPressed: () => Navigator.of(context).pop(),
-          child: Text(loc(context).cancel),
+        AppButton.text(
+          identifier: 'ipv6-rule-cancel',
+          label: loc(context).cancel,
+          onTap: () => Navigator.of(context).pop(),
         ),
-        FilledButton(
-          onPressed: _isFormValid ? _submit : null,
-          child: Text(_isEdit ? loc(context).save : loc(context).add),
+        AppButton.text(
+          identifier: 'ipv6-rule-submit',
+          label: _isEdit ? loc(context).save : loc(context).add,
+          onTap: _isFormValid ? _submit : null,
         ),
       ],
     );
--- a/lib/page/wifi_settings/views/components/wifi_quick_setup_card.dart
+++ b/lib/page/wifi_settings/views/components/wifi_quick_setup_card.dart
@@ -113,6 +113,7 @@ class WifiQuickSetupCard extends ConsumerWidget {
       title: loc(context).name,
       contentBuilder: (ctx, setState, onSubmit) => AppTextFormField(
         controller: controller,
+        identifier: 'wifi-ssid-name-input',
         label: loc(context).name,
         onChanged: (_) => setState(() {}),
       ),
--- a/lib/page/admin/views/dialogs/timezone_edit_dialog.dart
+++ b/lib/page/admin/views/dialogs/timezone_edit_dialog.dart
@@ -148,6 +148,7 @@ class _TimezoneListTile extends StatelessWidget {
   @override
   Widget build(BuildContext context) {
     return Semantics(
+      identifier: 'admin-timezone-item-${timezone.timeZoneID}',
       label: '${timezone.friendlyName}, ${timezone.offsetDisplayText}',
       selected: isSelected,
       button: true,
--- a/lib/page/instant_privacy/views/instant_privacy_view.dart
+++ b/lib/page/instant_privacy/views/instant_privacy_view.dart
@@ -322,6 +322,7 @@ class InstantPrivacyView extends ConsumerWidget {
             onTap: () => Navigator.of(ctx).pop(false),
           ),
           AppButton.primary(
+            identifier: 'instant-privacy-enable-confirm',
             label: loc(context).enable,
             onTap: () => Navigator.of(ctx).pop(true),
           ),
--- a/lib/page/port_forwarding/views/components/usp_single_port_tab.dart
+++ b/lib/page/port_forwarding/views/components/usp_single_port_tab.dart
@@ -171,6 +171,7 @@ class UspSinglePortTab extends ConsumerWidget {
           onTap: () => context.pop(),
         ),
         AppButton.dangerText(
+          identifier: 'pf-delete-confirm',
           label: loc(context).delete,
           onTap: () => context.pop(true),
         ),
--- a/lib/page/wifi_settings/views/components/wifi_network_card.dart
+++ b/lib/page/wifi_settings/views/components/wifi_network_card.dart
@@ -158,6 +158,7 @@ class WifiNetworkCard extends ConsumerWidget {
       title: loc(context).name,
       contentBuilder: (ctx, setState, onSubmit) => AppTextFormField(
         controller: controller,
+        identifier: 'wifi-ssid-name-input',
         label: loc(context).name,
         onChanged: (_) => setState(() {}),
       ),


AustinChangLinksys (commented):
:robot: **Automated Review — Round 1** · `520e910..86ec0cc` (full)

**Verdict: :speech_balloon: Self-review (comment only)** — Austin's own PR; auto-approve suppressed per self-review rule. Findings below for your reference.

#### Standards

|---|---|---|---|
| :warning: | Med | `ipv6_port_service_rule_dialog.dart:120-128` | `_isFormValid` getter calls `validateRule()` on every build — double-invocation per cycle alongside `_validate()` on focus-loss |
| :warning: | Med | `ipv6_port_service_rule_dialog.dart:219-220` | `_submit()` fallback `?? 0` / `?? startPort` bypasses type safety — unreachable via gate but defensive gap |
| :warning: | High | `ipv6_port_service/services/usp_ipv6_port_service_service.dart:228-276` | `validateRule()` returns hardcoded English strings (pre-existing, not introduced here) — surfaced by this dialog migration |
| :warning: | Med | `port_forwarding_dialog.dart:189`, `ipv6_port_service_rule_dialog.dart:132` | [Duplicated Code] Both dialogs share identical StatefulWidget boilerplate (controllers + FocusNodes + `_isEdit` + `_validate`/`_isFormValid`) — 3rd instance without extraction |
| :warning: | High | `test/page/dashboard/views/dialogs/` (absent) | No dialog-level widget tests for either migrated dialog; comparable `DhcpReservationEditDialog` has `dhcp_reservation_edit_dialog_test.dart` |
| :bulb: | Low | `port_forwarding_dialog.dart:273`, `ipv6_port_service_rule_dialog.dart:207` | Uses `Navigator.of(context).pop()` while newer sibling dialogs use `context.pop()` (go_router) — style inconsistency (pre-existing in pf dialog) |
| :bulb: | Low | `ipv6_port_service_rule_dialog.dart:166-173` | `SegmentedButton<String>` is raw Material widget inside `AppDialog` — no `AppSegmentedButton` in UI kit; consistent with `port_range_forwarding_dialog.dart` |

*Confidence: High = code-verified / Med = located + reasoned, not fully confirmed / Low = speculative, please double-check.*

#### Spec

|---|---|---|---|
| :white_check_mark: | High | both files | Both targeted dialogs migrated (`AlertDialog` -> `AppDialog`, `Text` -> `AppText.titleLarge`, `SingleChildScrollView` removed for `scrollable: true`) |
| :white_check_mark: | High | `ipv6_port_service_rule_dialog.dart:204-213` | Action buttons migrated: `TextButton`/`FilledButton` -> `AppButton.text` with new identifiers `ipv6-rule-cancel` / `ipv6-rule-submit` |
| :white_check_mark: | High | PR description | Third dialog (`port_range_forwarding_dialog`) correctly noted as already migrated and excluded |
| :bulb: | Low | PR description | Issue #1166 says "use `showSimpleAppDialog` or equivalent"; PR returns `AppDialog` directly (called via `showAppDialog` at call sites). Functionally equivalent, minor spec drift |

<details>

**`_isFormValid` double-validation (Med)**
`ipv6_port_service_rule_dialog.dart:120-128`:
```dart
bool get _isFormValid {
  final errors = UspIpv6PortServiceService.validateRule(
    description: _descriptionController.text.trim(),
    ipv6Address: _ipv6Controller.text.trim(),
    startPort: _startPortController.text.trim(),
    endPort: _endPortController.text.trim(),
  );
  return errors.isEmpty;
}
```
Called in `build()` (line 212: `onTap: _isFormValid ? _submit : null`). `validateRule()` also runs in `_validate()` on focus-loss. Every rebuild triggers two full validation passes. Contrast with `port_forwarding_dialog.dart` which uses a cached `_errors` map and calls `_validateFields()` only in `_validate()` / `_isFormValid()`.
**Fix:** Cache result in the `_errors` map or guard on `_errors.isEmpty && _hasRequiredInput` analogous to the port_forwarding dialog pattern.

---

**`_submit()` fallback `?? 0` / `?? startPort` (Med)**
`ipv6_port_service_rule_dialog.dart:219-220`:
```dart
final startPort = int.tryParse(_startPortController.text.trim()) ?? 0;
final endPort = int.tryParse(_endPortController.text.trim()) ?? startPort;
```
The `_isFormValid` gate ensures valid integers before `_submit` is callable, so these fallbacks are unreachable normally. However, if the gate is ever bypassed (test harness, programmatic call), a rule with `startPort=0` could be submitted silently. The `port_forwarding_dialog._submit()` (lines 285-288) does an explicit null-guard and early-return by contrast.
**Fix:** Apply the same explicit null-guard pattern for consistency and defensive correctness.

---

**Hardcoded English error strings in `validateRule()` (High -- pre-existing)**
`usp_ipv6_port_service_service.dart:228-276`:
```dart
errors['description'] = 'Name is required';
errors['description'] = 'Name must not have leading or trailing spaces';
errors['startPort'] = 'Start port is required';
```
These surface as `errorText` in dialog fields. Contrast with `port_forwarding_dialog.dart` which uses `loc(context).portMustBe1To65535` etc. (fully localized). Pre-existing in the service layer, not introduced by this PR, but made more visible by this migration.
**Fix:** Move validation logic into the dialog (like port_forwarding_dialog) or have `validateRule()` return localization keys for the caller to resolve.

---

**[Duplicated Code] StatefulWidget dialog boilerplate (Med)**
Both dialogs independently implement: late `TextEditingController` init with `r?.field ?? ''`, `FocusNode` array + `f.addListener(() { if (!f.hasFocus && mounted) _validate...})` loop, `_isEdit` getter, `_onInputChanged()` -> `setState(() {})`, and `dispose()` teardown. `port_range_forwarding_dialog.dart` is a third instance. This PR adds a second instance without extracting the shared state logic.
**Fix:** Consider a base mixin or `DialogFormState` abstraction for the focus-loss/validation lifecycle.

---

**Missing dialog-level widget tests (High)**
`test/page/dashboard/views/dialogs/` only contains `wifi_channel_dialog_test.dart`; no test for `PortForwardingDialog`. No dialog test exists for `Ipv6PortServiceRuleDialog` either. `DhcpReservationEditDialog` (comparable migrated dialog) has `dhcp_reservation_edit_dialog_test.dart` covering submit-gate and validation flows. PR reports 44/44 golden tests and 94/94 functional tests, but these are page-level; dialog-level validation behavior (empty form -> submit disabled, valid input -> submit enabled, focus-loss error display) is untested.
**Fix:** Add `port_forwarding_dialog_test.dart` and `ipv6_port_service_rule_dialog_test.dart` analogous to `dhcp_reservation_edit_dialog_test.dart`.

</details>

<details>

- Clean mechanical migration: `AlertDialog` -> `AppDialog`, `Text` -> `AppText.titleLarge`, `SingleChildScrollView` wrapper removed in favor of `scrollable: true` (confirmed: `AppDialog.scrollable` wraps content in `SingleChildScrollView` internally -- functionally equivalent).
- `AppButton.text(identifier:)` is valid at the pinned v2.31.0 of `ui_kit_library` (pub.lock resolved-ref `efd55408`); the local clone at `~/dev/privacyGUI-UI-kit` is an older snapshot and does not reflect the live dependency. No compile issue.
- All E2E identifiers preserved or upgraded: existing `pf-single-*` identifiers untouched; new `ipv6-rule-cancel` / `ipv6-rule-submit` identifiers added and documented in PR.
- Focus-loss validation pattern, `_isFormValid`/`_hasRequiredInput` submit gating, and all controller/FocusNode lifecycle (`initState`/`dispose`) carry over intact.
- `_submit()` in `port_forwarding_dialog.dart` retains the null-guard + early-return pattern correctly.
- Callers already use `showAppDialog` -- no caller changes required.
- `SegmentedButton` raw Material usage is consistent with project pattern (same in `port_range_forwarding_dialog.dart`); no `AppSegmentedButton` exists in the kit.
- Stacking note and E2E SSOT follow-up clearly documented in PR description.

</details>

*Cross-reviewed by two independent agents (security+correctness / architecture+maintainability). Automated -- please sanity-check before merge.*

lib/page/dashboard/views/dialogs/port_forwarding_dialog.dart
```diff
@@ -186,97 +186,96 @@ class _PortForwardingDialogState extends State<PortForwardingDialog> {
...
+          label: loc(context).cancel,
+          onTap: () => Navigator.of(context).pop(),
         ),
-        FilledButton(
-          onPressed: _isFormValid(context) ? _submit : null,
-          child: Text(_isEdit ? loc(context).save : loc(context).add),
+        AppButton.text(
+          identifier: 'pf-single-submit',
+          label: _isEdit ? loc(context).save : loc(context).add,
+          onTap: _isFormValid(context) ? _submit : null,
```
PeterJhongLinksys, line 278: ⚠️ 🟢 Primary action demoted: `AppButton.text` renders identically to the Cancel button next to it (both resolve to `contentColor: colorScheme.primary` on a transparent surface via `ButtonColorResolver`). `ui_kit_migration.md` maps `FilledButton → AppButton (filled variant)`, and `AppButton.primary` is available. Also note `_isFormValid(context)` re-runs the full rule set (3 regexes) on every keystroke because `_onInputChanged()` calls `setState`; consider caching the errors in state.

lib/page/ipv6_port_service/views/dialogs/ipv6_port_service_rule_dialog.dart
```diff
@@ -129,86 +129,87 @@ class _Ipv6PortServiceRuleDialogState extends State<Ipv6PortServiceRuleDialog> {
...
+      scrollable: true,
+      content: Column(
+        mainAxisSize: MainAxisSize.min,
+        children: [
+          AppTextField(
+            controller: _descriptionController,
+            focusNode: _descriptionFocus,
+            identifier: 'ipv6-rule-description',
+            hintText: loc(context).ruleName,
+            errorText: _errors['description'],
```
PeterJhongLinksys, line 144: ⚠️ 🟢 `_errors` comes from `UspIpv6PortServiceService.validateRule()` (`usp_ipv6_port_service_service.dart:229-243`), which returns hard-coded English strings (`Name is required`, `Invalid IPv6 address format`, ...). These are rendered directly, so every locale sees English. Sibling dialogs (`port_range_forwarding_dialog.dart:166-174`, `dhcp_reservation_edit_dialog.dart:131-141`) return error *keys* and map them via `_localizeError()` — worth aligning.

lib/page/wifi_settings/views/components/wifi_quick_setup_card.dart
```diff
@@ -113,6 +113,7 @@ class WifiQuickSetupCard extends ConsumerWidget {
       title: loc(context).name,
       contentBuilder: (ctx, setState, onSubmit) => AppTextFormField(
         controller: controller,
+        identifier: 'wifi-ssid-name-input',
```
PeterJhongLinksys, line 116: ⚠️ 🟢 Duplicate identifier: `wifi-ssid-name-input` is also used in `wifi_network_card.dart:161`. Verified not a live collision today (`wifi_list_tab.dart:122-129` makes the two cards mutually exclusive and each sits inside its own modal), but the anchor is ambiguous and the two write paths differ (`updateQuickSetupField` vs `updateNetworkField`). Suggest `wifi-quick-setup-ssid-input` / `wifi-network-ssid-input`.

PeterJhongLinksys (approved):
🤖 **Automated Review — Round 1** · `b8894c4..86ec0cc` (full)

**Verdict: ✅ APPROVE** — Solid, faithful `AlertDialog` → `AppDialog` migration with correct `scrollable:` usage and no security or USP/JNAP regressions; approving with a set of non-blocking Warnings, two of which (soft-keyboard avoidance and actions overflow) are real behavioural risks that live in `ui_kit` and deserve a maintainer decision.

|---|---|---|---|
| ⚠️ | 🟡 | `port_forwarding_dialog.dart:189-193`, `ipv6_port_service_rule_dialog.dart:132-135` | `AppDialog` has no `MediaQuery.viewInsets` handling (unlike `AlertDialog`/`Dialog`), so the soft keyboard can cover the lower text fields in these input-heavy forms. |
| ⚠️ | 🟡 | `port_forwarding_dialog.dart:269-280`, `ipv6_port_service_rule_dialog.dart:203-214` | `AppDialog` lays out `actions` in a plain `Row` (no `OverflowBar`/`Wrap`), so long localized labels or large text scale can overflow the 400px max width. |
| ⚠️ | 🟢 | `port_forwarding_dialog.dart:189-192`, `ipv6_port_service_rule_dialog.dart:132-134` | Passing `title:` (widget) instead of `titleText:` leaves `AppDialog`'s `Semantics(label:)` null — the dialog container loses its accessible name. |
| ⚠️ | 🟢 | `port_forwarding_dialog.dart:275-279`, `ipv6_port_service_rule_dialog.dart:209-213` | [XX] Primary action downgraded from `FilledButton` to `AppButton.text`, making Save/Add visually identical to Cancel; `ui_kit_migration.md` maps `FilledButton → AppButton (filled)` and `AppButton.primary` exists. |
| ⚠️ | 🟢 | `ipv6_port_service_rule_dialog.dart:144,157,186` | [XX] `errorText` renders hard-coded English strings from `UspIpv6PortServiceService.validateRule()`; sibling dialogs use error keys + `_localizeError()`. |
| ⚠️ | 🟢 | `port_forwarding_dialog.dart:180-181,278` | [XX] [Duplicated Code] Full form validation (3 regex rules) runs inside `build()` on every keystroke, and `_submit()` re-checks a narrower condition than the enable gate. |
| ⚠️ | 🟢 | `port_forwarding_dialog.dart:105-181` | [Duplicated Code] Business validation kept in the View while the IPv6 dialog in the same PR delegates to its Service; rules duplicate `port_range_forwarding_dialog.dart:121-165` and are untestable at the Service layer. |
| ⚠️ | 🟢 | `usp_single_port_tab.dart:174`, `instant_privacy_view.dart:311` | Identifier coverage is partial: `usp_port_triggering_tab.dart:173`, `usp_port_range_tab.dart:157` and the Instant Privacy *disable* confirm button still have none. |
| ⚠️ | 🟢 | `wifi_network_card.dart:161`, `wifi_quick_setup_card.dart:116` | [XX] Both widgets use the same identifier `'wifi-ssid-name-input'`; verified mutually exclusive today (`wifi_list_tab.dart:122-129`), so not a live collision, but the anchor is ambiguous. |
| ⚠️ | 🟢 | `lib/page/dashboard/views/dialogs/port_forwarding_dialog.dart` | [Feature Envy] File lives under `page/dashboard/` but both consumers are in `page/port_forwarding/`; sibling dialogs already live there. |
| ⚠️ | 🟢 | `port_triggering_dialog.dart:91`, `static_route_dialog.dart:130` | Migration is half-done: 12 files still use `AlertDialog(`, including `port_triggering_dialog` which sits in the same 3-tab page as the migrated ones → two dialog styles side by side. |
| ⚠️ | 🟡 | `port_forwarding_dialog.dart:244-252`, `ipv6_port_service_rule_dialog.dart:166-173` | [Duplicated Code] Content still uses raw Flutter `SegmentedButton<String>` + `Text` (5 hand-rolled copies in repo); could not verify whether `ui_kit` offers an equivalent (git dependency, not vendored). |
| ⚠️ | 🟢 | `ipv6_port_service_rule_dialog.dart:166-170` vs `port_forwarding_dialog.dart:245-249` | Protocol labels have three different behaviours across dialogs: always-English, fully localized, and half-localized. |
| ⚠️ | 🟢 | `test/page/port_forwarding/`, `test/page/ipv6_port_service/` | No `views/` widget tests for either heavily-rewritten dialog and no tests in this PR; submit-enable gate and focus-loss validation have zero regression cover (precedent exists: `dhcp_reservation_edit_dialog_test.dart`). |
| ⚠️ | 🟡 | PR-wide (8 new/changed identifiers) | Art. XVI §16.4: identifier changes are a contract change; no evidence of `identifiers.generated.ts` regeneration / spec migration (cannot verify from this repo). |
| 💡 | 🟢 | `port_forwarding_dialog.dart:190-192`, `ipv6_port_service_rule_dialog.dart:133-134` | Two `AppDialog` title APIs mixed across the repo (`title: AppText.titleLarge(...)` vs `titleText:`); the manual variant drifts when ui_kit retunes dialog typography. |
| 💡 | 🟢 | `port_forwarding_dialog.dart:8-24`, `ipv6_port_service_rule_dialog.dart:8-24` | `*DialogResult` are bare classes without `Equatable` (Art. XI) and duplicate the existing `*RuleUIModel` field sets minus `instancePath` (Art. VII). |
| 💡 | 🟡 | `ipv6_port_service_rule_dialog.dart:4,106,121,167` | View imports and statically calls the Service, bypassing the Provider layer, so the validator cannot be substituted in tests (Art. V §5.4 / Art. VI). |
| 💡 | 🟢 | `usp_single_port_tab.dart:169-177`, `instant_privacy_view.dart:322-326` | Only the confirm buttons got identifiers; the Cancel path of a destructive action is arguably more important for negative E2E tests. |
| 💡 | 🟢 | `timezone_edit_dialog.dart:151` | Identifier embeds `timeZoneID`, some of which contain `:` (e.g. `NST03:30`); IDs are unique (41/41) and stable, but colons need escaping in some selector syntaxes. |
| 💡 | 🟡 | `port_forwarding_dialog.dart:273`, `ipv6_port_service_rule_dialog.dart:207` | Uses `Navigator.of(context).pop()` while sibling dialogs use `context.pop()`; correct today (`showAppDialog` is `useRootNavigator: true`) but inconsistent. |
| 💡 | ⚪ | `port_forwarding_dialog.dart:193`, `ipv6_port_service_rule_dialog.dart:135` | `scrollable: true` adopted here but not in the longer `dhcp_reservation_edit_dialog.dart:145-149` — worth making consistent across the four form dialogs. |
| 💡 | 🟢 | PR scope | Five pure identifier patches (timezone / instant privacy / two wifi cards / single port tab) are unrelated to the AppDialog migration; splitting them out would improve review and revert granularity (Art. I §1.3). |

<sub>Confidence: 🟢 High = head code read, evidence quoted · 🟡 Med = file:line + rationale, not fully substantiated · ⚪ Low = speculative.</sub>
<sub>`[X]` = raised by 1 agent · `[XX]` = independently raised by both agents. Unmarked Warning/Suggestion rows are single-agent findings.</sub>

<details>
<summary>🔴 Critical — none this round</summary>

Neither reviewer could substantiate a complete evidence chain for a user-triggerable correctness or security defect. Notably, the suspected "`AppButton.text` with `onTap: null` does not disable" issue was **disproved** by reading `ui_kit_library v2.31.0`:

```dart
// app_button.dart:402
bool get _isEnabled => widget.onTap != null && !widget.isLoading;
// 502-520: AppInteractionSensor(onTap: ...) is only wired when _isEnabled
// 523-530: Semantics(enabled: _isEnabled, ...)
```

So `AppButton.text(onTap: _isFormValid(context) ? _submit : null)` is behaviourally equivalent to the previous `FilledButton(onPressed: null)` — not clickable, not submittable, and correctly announced as disabled.
</details>

<details>

**1. Soft-keyboard avoidance lost (🟡 Med)**

`AppDialog` never consults `viewInsets`:

```dart
// ui_kit v2.31.0 lib/src/molecules/dialogs/app_dialog.dart:151-181
if (scrollable) {
    constrainedBody = ConstrainedBox(
      constraints: BoxConstraints(
        maxHeight: MediaQuery.of(context).size.height * 0.85,   // size.height, not minus viewInsets
      ), ...
return Center(
    child: ConstrainedBox(
      constraints: BoxConstraints(maxWidth: ... 400.0),
      child: Semantics(... child: AppSurface(padding: ... EdgeInsets.all(24.0), child: constrainedBody)),
```

`showAppDialog` (same file, 205-286) uses `showGeneralDialog` and its `pageBuilder` only wraps `Theme/Portal/Material` — no inset padding either. Both migrated dialogs are keyboard-heavy forms (`port_forwarding_dialog.dart:197-238`: description, external port, internal port, internal IP). `AlertDialog`'s internal `Dialog` adds `MediaQuery.viewInsetsOf(context)` to its inset padding, so the pre-migration dialog shifted up with the keyboard.
Trigger: touch/mobile viewport, editing the lower fields. Confidence is Med because the `AppDialog` side is fully substantiated but the `AlertDialog` behaviour is Flutter SDK code we could not quote locally.
Fix: add `Padding(padding: MediaQuery.viewInsetsOf(context))` in `AppDialog`/`showAppDialog` in ui_kit, or wrap locally; at minimum verify on a mobile viewport before merge.

**2. Actions row can overflow (🟡 Med)**

```dart
// app_dialog.dart:94-109
if (actions != null && actions!.isNotEmpty) {
    actionsWidget = Row(                       // plain Row, no Wrap / OverflowBar
      mainAxisAlignment: effectiveDialogStyle?.buttonAlignment ?? .end,
      children: actions!.asMap().entries.map((entry) { ... }).toList(),
    );
```

Each `AppButton` shrink-wraps (`app_button.dart:491-498`: `ConstrainedBox(minWidth: 44)` + `Center(widthFactor: 1.0)`); the label ellipsizes only *inside* the button, and the dialog is capped at 400px (`app_dialog.dart:167`). Long-locale labels or increased `textScaleFactor` can therefore exceed `400 - 48` and produce a RenderFlex overflow (debug) / clipping (release). `AlertDialog` used `OverflowBar`, which stacks vertically instead.

**3. Dialog loses its accessible name (🟢 High)**

```dart
// app_dialog.dart:169-171
child: Semantics(
    label: semanticLabel ?? titleText,   // a `title` widget does not participate → null
    container: true,
```

Both call sites pass `title:`:

```dart
return AppDialog(
    title: AppText.titleLarge(_isEdit ? loc(context).editPortForwarding : loc(context).addPortForwarding),
    scrollable: true,
```

`AlertDialog` supplied a default (`MaterialLocalizations.alertDialogLabel`). Given that this PR is explicitly about improving automation/a11y anchors, this is a regression in the opposite direction. Fix: use `titleText:`, or keep `title:` and add `semanticLabel:`.

**4. Primary action visual demotion (🟢 High, both agents)**

```dart
// app_button.dart:162-179
const AppButton.text({...}) : variant = SurfaceVariant.highlight,
                              styleVariant = ButtonStyleVariant.text;
// app_button.dart:67-84
const AppButton.primary({...}) : variant = SurfaceVariant.highlight,
                                 styleVariant = ButtonStyleVariant.filled;
```

Via `ButtonColorResolver` (`button_color_resolver.dart:154-161`) both Cancel and Save resolve to `contentColor: colorScheme.primary` on a transparent surface — i.e. identical. `ui_kit_migration.md` maps `FilledButton → AppButton (filled variant)`. Note `dhcp_reservation_edit_dialog.dart:208-217` and `port_range_forwarding_dialog.dart:273-282` already have the same problem, so this PR is *consistent with an existing mistake*; a follow-up sweep to `AppButton.primary` for all submit buttons would be the clean fix.

**5. Unlocalized validation messages (🟢 High, both agents)**

```dart
// lib/page/ipv6_port_service/services/usp_ipv6_port_service_service.dart:229-243
if (description.isEmpty) { errors['description'] = 'Name is required'; }
...
errors['ipv6Address'] = 'Invalid IPv6 address format';
```

```dart
// ipv6_port_service_rule_dialog.dart:144
errorText: _errors['description'],
```

Compare `dhcp_reservation_edit_dialog.dart:131-141` / `port_range_forwarding_dialog.dart:166-174`, which return error **keys** and map them through `_localizeError(key)` → `loc(context)`. Every locale currently sees English here.

**6. Validation inside `build()` (🟢 High, both agents)**

```dart
180|  bool _isFormValid(BuildContext context) =>
181|      _hasRequiredInput && _validateFields(context).isEmpty;
...
278|          onTap: _isFormValid(context) ? _submit : null,
```

`_onInputChanged()` is `setState(() {})` (112-114), so every keystroke re-runs `_validateFields`, including `IpAddressRule` / `IpAddressNoReservedRule` regexes. Functionally correct but a validation-in-build anti-pattern. Additionally `_submit()` (284-288) only re-checks `extPort/intPort/client` and not the description rules, so it is not a safety net if the enable gate ever regresses — `dhcp_reservation_edit_dialog.dart:222-224` uses the safer `if (!valid) return;` guard.

**7. Wrong module (🟢 High)**

`port_forwarding/cards/usp_port_forwarding_card.dart:10` and `port_forwarding/views/components/usp_single_port_tab.dart:11` both `import 'package:privacy_gui/page/dashboard/views/dialogs/port_forwarding_dialog.dart'`; nothing under `page/dashboard/` uses it, while `port_range_forwarding_dialog.dart` and `port_triggering_dialog.dart` already live under `port_forwarding/views/dialogs/`. This PR rewrites the file, which makes it the cheapest moment to relocate it.

**8. Test gap (🟢 High)**

```
$ ls test/page/port_forwarding test/page/ipv6_port_service
providers  services          # no views/
$ grep -rn "PortForwardingDialog\|Ipv6PortServiceRuleDialog\|pf-single-\|ipv6-rule-" test/   → 0 hits
$ grep -rn "AlertDialog" test/                                                               → 0 hits
```

Good news: no existing test breaks from the `AlertDialog` → `AppDialog` finder change. Bad news: the submit-enable gate, focus-loss validation and protocol defaults of both rewritten dialogs have no coverage at all, despite `test/page/dhcp/views/dialogs/dhcp_reservation_edit_dialog_test.dart` providing a ready-made template (`showAppDialog` + `find.byType(AppTextField)`).
</details>

<details>

- **`scrollable: true` is the correct replacement for the removed `SingleChildScrollView`.** `app_dialog.dart:116-121` wraps content in `Flexible(SingleChildScrollView(...))` when scrollable, and 159-162 only wraps the whole body when *not* scrollable — the two paths are mutually exclusive, so there is no double scroll view. Padding comes from a single source (`AppSurface(padding: EdgeInsets.all(24))`). Title and actions now stay pinned while only the content scrolls, which is better than the previous behaviour.
- **USP/JNAP error handling is clean and untouched.** `grep -rn "on Exception catch" lib` → zero hits. `usp_ipv6_port_service_service.dart:154-157` follows the required pattern: `} catch (e) { if (e is ServiceError) rethrow; throw mapUspErrorToServiceError(e); }`. UI side (`usp_ipv6_port_service_view.dart:260-264`) uses `catch (e)` + `context.mounted` + `localizeServiceError`.
- **Layering is correct.** `grep 'generated/'` across the 7 changed files → zero hits; views only touch `*UIModel`. No USP Provider hard-rule surface is touched (no `autoDispose` on L1, no `ref.watch` in fetch, no missing `withLock`, no `PreservableContract` redefinition).
- **No lifecycle bugs.** Focus listeners guard on `mounted` (`port_forwarding_dialog.dart:82-85`, `ipv6_port_service_rule_dialog.dart:84-87`) and `dispose()` releases all four controllers and four focus nodes in both files.
- **Async races are guarded.** Every `await showAppDialog` call site (`usp_single_port_tab.dart:127,148,180`, `usp_ipv6_port_service_view.dart:212,233`) checks `result == null || !context.mounted` before proceeding.
- **Security surface is clean.** No hard-coded secrets/tokens, no `flutter_secure_storage` changes, no string-concatenation injection surface (protocol values come from the `UspIpv6PortServiceService.protocolOptions` whitelist with a `?? 255` fallback in `mapDisplayToIana`). Notably the new identifiers were added to the SSID *name* field only — the password field was deliberately left without a stable anchor.
- **`Semantics(identifier:)` usage is exactly right** in `timezone_edit_dialog.dart:151`: `identifier:` used as the test hook (not `label:`), instance key derived from data (`timeZoneID`) rather than row index, and `label:` preserved for genuine a11y text.
- **Per-instance key derivation is pure and unit-tested**: `ruleIdentifierKey` (`_shared/models/port_forwarding_rule_ui_model.dart:10`) with 8 cases in `test/page/_shared/models/port_forwarding_rule_identifier_test.dart`.
- **The focus-loss validation comments were preserved through the refactor** (`port_forwarding_dialog.dart:48-51,109-111,122-125`), including the CanvasKit `<input>` rebuild root cause and the cross-reference to `usp_local_network_view`. Valuable context, good that it survived.
- Both `*RuleUIModel` classes `extend Equatable` and provide `copyWith` (Art. XI).
</details>

---

HankYuLinksys (approved):
### Correction: my finding #1 was a test-harness artifact — there is no 1280px overflow

@AustinChangLinksys Thanks for pushing back rather than taking my numbers on trust. I re-ran the measurement properly and **my finding #1 was wrong**. It was an artifact of my own probe, not a defect in this PR.

**What I got wrong.** My throwaway probe lived at `test/` root. There is no `flutter_test_config.dart` there, so **no real fonts were loaded** and every glyph fell back to the engine's built-in monospace test font. `test/helpers/load_app_test_fonts.dart` documents exactly this trap:

> Flutter's test runner registers no real fonts: every glyph falls back to the engine's built-in monospace test font, which is **1.8-2.7x wider** than NeueHaasGrotTextRound […] Tests that measure layout therefore need the real fonts, or they measure a font no user ever sees.

I measured a font no user ever sees. The golden suite calls `loadAppTestFonts()` from `test/golden_test/flutter_test_config.dart`, which is precisely why **your** golden run was the trustworthy one and mine was not.

**Same probe, fonts toggled, nothing else changed** (this PR's code, 1280x800, flat):

| Locale / dialog | fonts **not** loaded (my original error) | real app fonts loaded |
|---|---|---|
| pf / en | overflow **51px** | ✅ **NONE** |
| ipv6 / en | overflow **93px** | ✅ **NONE** |
| pf / el | overflow **285px** | ✅ **NONE** |
| ipv6 / el | overflow **121px** | ✅ **NONE** |

The overflow numbers I reported are reproduced *exactly* by simply omitting the font load — which is what identifies the harness, not the code, as the cause.

**Actual headroom at 400px `maxWidth`.** I checked this isn't marginal-and-lucky, measuring `label + SegmentedButton` against the 350px row across all 16 shipping locales:

- Widest: **el** (`Πρωτόκολλο`, 70.0px label) — needs 333.2px of 350px, **16.8px slack**
- Then **cs** 330.6px, **it** 326.8px, **es**/**pt** 323.7px, **fr** 323.4px
- en needs 304.7px, 45.3px slack

Every locale fits, tightest margin 16.8px. So the 400px clamp is genuinely adequate and the `Row` → `Column` stacking change is **not** needed — I withdraw that suggestion, and #1223 should close as not-a-bug.

The inverse of my claim is actually true: at real font metrics the **base** (`AlertDialog`) is what overflows on narrow viewports — 375px gives 58px (pf) / 68px (ipv6), 390px gives 43px / 53px, because `IntrinsicWidth` lets the dialog grow past the screen. This PR's fixed clamp **removes** those. That's an improvement I missed and should have credited.

### Finding #2 (`viewInsets`) — still real, but my numbers were inflated too

Same font bug, plus a second flaw: I rendered the dialog into a `Scaffold` body, where `resizeToAvoidBottomInset` does the shifting — so I was partly measuring `Scaffold`, not `AppDialog`. Re-measured through the real entry point (`showAppDialog`, as `usp_single_port_tab.dart` calls it) with fonts loaded:

| Device | Keyboard | base (`AlertDialog`) | this PR (`AppDialog`) |
|---|---|---|---|
| iPhone SE 375x667 | 260 | ✅ reachable | ❌ clipped **158.5px** (I said 184) |
| iPhone 14 390x844 | 336 | ✅ reachable | ❌ clipped **146.0px** (I said 172) |
| Pixel 7 412x915 | 300 | ✅ reachable | ❌ clipped **74.5px** (I said 100) |
| iPad mini 744x1133 | 370 | ✅ reachable | ❌ clipped **35.5px** (I said 61) |
| Desktop 1280x800 | 0 | ✅ reachable | ✅ reachable |

Direction and mechanism hold — `AppDialog` centres against full `size.height` and never consults `viewInsets`, while `AlertDialog` adds `MediaQuery.viewInsetsOf(context)` to its padding — but every figure is ~26px smaller than I reported. iPad mini is 35.5px, not 61px. **linksys/privacyGUI-UI-kit#12 stands**, and treating it as non-blocking for a desktop-web-primary target looks right.

### One correction in the other direction, on Peter's first nit

You classified the demoted primary action as "came in via #1205 (not touched by this PR)". That holds for `port_forwarding_dialog` — its actions row is two `AppButton.text` on `dev-2.7.0` already, untouched here. But it does **not** hold for `ipv6_port_service_rule_dialog`, where this PR is the change:

```
- FilledButton(          →  + AppButton.text(
```

So @PeterJhongLinksys's nit is PR-introduced for the ipv6 dialog: its submit was a `FilledButton` on base and becomes visually identical to Cancel here. Not a blocker — cosmetic, and the kit has `AppButton.primary` — but worth routing the follow-up as "fix ipv6 (regression) + align pf/port_range (#1205 debt)" rather than filing all three as pre-existing, so the ipv6 one doesn't get deprioritised as inherited debt.

### Takeaway beyond this PR

Any test that measures layout without loading real fonts produces confident, precise, wrong numbers — mine were self-consistent across two dialogs, four locales and two revisions, which is exactly what made them convincing. This got caught only because you checked the live app and the golden suite instead of accepting my table. Worth a note in the testing docs; happy to send that PR.

Sorry for the detour on #1223.

PR #1207, opened 2026-08-06
fix(ai): map storage and config failures to ServiceError instead of surfacing
  raw exceptions (#1203)

HankYuLinksys:
## Summary

Fixes #1203. The AI assistant's credential and configuration paths surfaced raw exception text to the user. `StorageError` already existed for this, with localization in place, and was unused across the whole repo.

## `AwsCredentialsStore` — errors

Conversion happens in `_serialize`, which every method already routes through, so `PlatformException` cannot leave the class and no method body carries error handling. The platform text goes on `originalError`, never into `detail` — `detail` is the field a localizer may surface (`localizeServiceError` does for `UnexpectedError`), and a keychain error code is not something a user can act on. `mapUspErrorToServiceError` makes the same call for an unparseable error.

Storage failures are logged rather than shown, which the class doc now states as a contract: persistence is a convenience, a failed save leaves a working session untouched, and the user has nothing to act on. `originalError` is logged explicitly, because `ServiceError.toString()` renders only the type name — a storage failure would otherwise log the uninformative `Storage` and nothing about the cause.

## `AwsCredentialsStore` — ordering

`Future.timeout` does not cancel the operation it wraps, it only stops waiting. Timing out the **queue** therefore does not remove a stalled platform call; it lets the next operation start while that call is still in flight, and whichever finishes last wins at the storage layer. Three concrete losses, each verified with a per-call-latency fake:

- a stalled `clear()` deleting credentials the user entered afterwards, while their `store()` reported success
- a stalled `store()` restoring credentials a later `clear()` removed
- a stalled `storeModelId()` — a read-modify-write — rewriting the whole record from its stale snapshot over newer credentials, with no `clear()` involved

A "has a clear happened since?" guard cannot close these: it answers a different question than the one that matters, which is whether the record in storage is still the one this operation wrote. The platform offers no cancellation and `write` is not conditional, so the queue now waits for genuine completion and only the caller's future is bounded. `_kCallerTimeout` is named for what it bounds.

**Accepted cost, documented on the class:** one hung platform call blocks later operations, a `clear()` included. That is a real downside, taken because the alternative silently loses or resurrects credentials, and because the caller is still told promptly.

`readWithin` exists so a caller that needs a shorter bound — the config screen disables itself during a restore — does not wrap `read()` in its own `.timeout()`, which can surface a bare `dart:async` `TimeoutException` and break the only-`ServiceError` guarantee.

## `router_assistant_view.dart`

**`_configError` is a sealed type.** It was a `String?` carrying a hand-rolled tagged union decoded by string prefix and `substring`, ending in `return error;` — so any failure that was not one of two known prefixes was rendered to the user verbatim. The localizing `switch` is now exhaustive by compiler, so a new kind cannot be added without deciding how it reads.

Splitting it separated three things that had been conflated:

| Kind | Handling |
|---|---|
| Field validation | Own l10n key — a separate line from `ServiceError` per the guide §1.4 |
| A real failure | Carries a `ServiceError`, goes through `localizeServiceError` |
| No environment credentials | **Not an error.** Logged only |

That third case is the usual one. `assets/agents/.env` is gitignored, so `dotenv.load` in `main.dart` fails and `AWSConfig.fromEnvironment()` throws dotenv's `NotInitializedError` — which extends `Error`, so it matches neither `on ConfigurationException` nor `on Exception`. It is how the app discovers it should show the manual form; displaying it reported the app's own configuration as the user's problem, on a screen that already explains what to do.

`_configFromEnvironment()` therefore recognises exactly two modes — `ConfigurationException` and `NotInitializedError` — and rethrows anything else into the reported path, so a future `generative_ui` adding a check or a cast leaves a genuine fault visible rather than silently showing the manual form.

**A failed revocation is shown.** "Change configuration" is the user asking for the credentials to be gone. A failed `clear()` was logged and nothing else, so the record would restore on the next launch while the user believed it removed. A failed *save* still stays silent — the session works and the only cost is re-typing next launch.

## l10n

Removes `failedToInitialize` from all 26 locales. Its message was `"Failed to initialize: {error}"` — a localized wrapper interpolating an untranslated exception, the shape the guide's what-to-display rule exists to prevent. The prefix also carried no information, since the user is looking at the configuration screen.

Verified before removal: unused in Dart, tests, and non-Dart sources; absent from `tools/translated.json`, so `update_translations.py` cannot resurrect it; `tools/check_l10n.dart` reports orphan-keys and duplicate-keys clean, and its own rule is that a genuinely unreferenced key must be deleted rather than allowlisted. ARB diffs are pure deletions — 0 insertions across all 26 files, with key order, encoding and trailing newlines intact.

Not verified: the E2E repo is not checked out locally. The removed key's message was never displayed, so a selector could not have depended on it.

## Out of scope

`router_chat_controller.dart` puts unlocalized English into the chat error banner (`withError(e.message)` plus two hardcoded strings). Credentials are only validated on the first send, so a wrong access key surfaces there rather than on the config screen — which makes that the most likely error path in practice. Four `withError` calls already exist on `dev-2.7.0`; the correct output there is a structured failure a model can act on, not a localized string for a human, so it belongs with the output-reliability work that touches those lines. Tracked on #1191's follow-up list.

Also unchanged: wrapping the config state machine in a Notifier (which is why the restore guards have no test), and clearing credentials on logout or router switch.

## Verification

| Check | Result |
|---|---|
| `aws_credentials_store_test.dart` | 32/32 |
| `aws_environment_config_test.dart` | 3/3 |
| `flutter test test/ai/ test/page/ai_assistant/` | 105/105 |
| `./run_tests.sh` (full suite) | **3463/3463** |
| `flutter analyze` | 0 errors, one fewer issue than before |
| `fvm dart format` | clean |

Mutation-checked: putting the timeout back on the queue fails three tests; removing the `ServiceError` conversion fails six; restoring the unsafe JSON cast fails one; weakening the debug-logging guard to `kProfileMode` fails two.

## Notes for reviewers

- **The three commits are sequential, not parallel.** Each round of review found that the previous fix had left a reachable path to the same class of failure. Squashing on merge is fine; the history is kept because the second and third commits document why the shape of the final fix is what it is.
- **`router_assistant_view.dart` has no widget test.** The restore guards and the `_configError` lifecycle are enforced by logic but uncovered. That is a Notifier-shaped problem, called out rather than left to be discovered.
- `readWithin`'s `onError` is defensive: `_asServiceError` passes a `ServiceError` through unchanged, so it is harmless when the inner bound fired first. I could not construct a case where removing it changes behaviour, and kept it rather than rely on that.
- `aws_credentials_store.dart` does not end in `_service.dart`. Deliberate — it is not a USP feature service — and discussed on #1201.

### DIFF
--- a/lib/page/ai_assistant/services/aws_credentials_store.dart
+++ b/lib/page/ai_assistant/services/aws_credentials_store.dart
@@ -99,10 +99,11 @@ class AwsCredentialsStore {
   /// the platform gives no way to cancel and `write` is not conditional, the
   /// only sound answer is to never let two operations overlap.
   ///
-  /// The cost is that one hung platform call blocks later operations — including
-  /// a `clear()` — for as long as it hangs. That is a genuine downside, accepted
-  /// because the alternative silently loses or resurrects credentials, and
-  /// because the caller is still told promptly via [_kCallerTimeout].
+  /// The cost is that one hung platform call blocks the operations behind it for
+  /// as long as it hangs, and since the platform never reports back, that can be
+  /// the rest of the session. Accepted for writes, whose worst case is a stale
+  /// record. NOT accepted for [clear] — see [_forceClear], because a revocation
+  /// that can never be applied is the one failure this class must not have.
   Future<T> _serialize<T>(Future<T> Function() operation) {
     final settled = _pending.then((_) => operation());
     // Chained to the real completion, so nothing starts early. This also
@@ -161,23 +162,31 @@ class AwsCredentialsStore {
     });
   }
 
-  /// The stored credentials, or null when nothing usable is saved.
-  ///
-  /// Returns null rather than a half-built record for anything unusable —
-  /// absent, blank, or unparseable — so the caller's "not configured" path
-  /// handles it instead of a signing error later. A bad record is therefore not
-  /// an error; only storage itself failing is ([StorageError], [TimeoutError]).
-  /// As [read], but giving up after [within] instead of [_kCallerTimeout].
+  /// As [read], but giving up after `min(within, _kCallerTimeout)`.
   ///
   /// For a caller that is holding UI hostage while it waits — the config screen
   /// disables itself during a restore — and would rather show an empty form
   /// sooner than the default allows. Wrapping `read()` in `.timeout()` at the
   /// call site instead would throw a bare `TimeoutException`, breaking this
   /// class's guarantee that callers only ever see a [ServiceError].
-  Future<StoredAwsCredentials?> readWithin(Duration within) => read()
-      .timeout(within)
-      .onError<Object>((e, _) => throw _asServiceError(e));
+  ///
+  /// The bound is a minimum, not a replacement: `read()` is already bounded by
+  /// [_kCallerTimeout], so a [within] longer than that has no effect. Asserted
+  /// rather than left as a trap for a future caller.
+  Future<StoredAwsCredentials?> readWithin(Duration within) {
+    assert(within <= _kCallerTimeout,
+        'readWithin cannot extend the bound past _kCallerTimeout');
+    return read()
+        .timeout(within)
+        .onError<Object>((e, _) => throw _asServiceError(e));
+  }
 
+  /// The stored credentials, or null when nothing usable is saved.
+  ///
+  /// Returns null rather than a half-built record for anything unusable —
+  /// absent, blank, or unparseable — so the caller's "not configured" path
+  /// handles it instead of a signing error later. A bad record is therefore not
+  /// an error; only storage itself failing is ([StorageError], [TimeoutError]).
   Future<StoredAwsCredentials?> read() {
     return _serialize(() async {
       final record = await _readRecord();
@@ -217,14 +226,70 @@ class AwsCredentialsStore {
   /// Remove the stored credentials.
   ///
   /// This is the revocation path — the user pressing "change configuration"
-  /// means they want these credentials gone. A [TimeoutError] here means the
-  /// platform has not answered yet, not that the delete was dropped: it stays
-  /// queued, and nothing requested after it can be applied before it.
+  /// means they want these credentials gone.
+  ///
+  /// Ordinarily queued like everything else, so it cannot be undone by a write
+  /// requested before it. But a queue held by a platform call that never returns
+  /// would mean a revocation is never applied and the record silently restores
+  /// on the next launch, so this one operation escapes that: see [_forceClear].
   Future<void> clear() {
-    return _serialize(() async {
+    final queued = _serialize(() async {
       await _storage.delete(key: _kRecordKey);
       aiLog('[Credentials] Cleared');
     });
+    return queued
+        .onError<TimeoutError>((_, __) => _clearOutOfBand().then((_) {}));
+  }
+
+  /// Delete outside the queue, unless the queued delete already landed.
+  ///
+  /// The check first, because a slow-but-moving queue often applies its delete
+  /// while the caller's bound is elapsing — and skipping the queue is a
+  /// concession that should not be made when it is not needed.
+  Future<void> _clearOutOfBand() async {
+    if (await _isRecordAbsent()) {
+      aiLog('[Credentials] Queued clear landed after its caller gave up');
+      return;
+    }
+    return _forceClear();
+  }
+
+  Future<bool> _isRecordAbsent() async {
+    try {
+      final raw =
+          await _storage.read(key: _kRecordKey).timeout(_kCallerTimeout);
+      return raw == null || raw.isEmpty;
+    } catch (_) {
+      // Cannot tell; assume it is still there so the delete is attempted.
+      return false;
+    }
+  }
+
+  /// Delete the record without waiting for the queue.
+  ///
+  /// Reached only when a queued [clear] timed out, which means some earlier
+  /// platform call has not returned. Skipping the queue reintroduces the overlap
+  /// [_serialize] exists to prevent — deliberately, and only here:
+  ///
+  /// * the racing operation can only be a write, and the user has since asked
+  ///   for the record to be gone, so if the delete loses the race the outcome is
+  ///   the same stale record we already had, not a worse one
+  /// * whereas not trying at all means the revocation is never applied and the
+  ///   credentials come back on the next launch
+  ///
+  /// The queued attempt is left running: if it eventually completes, it deletes
+  /// an already-deleted key, which is a no-op.
+  Future<void> _forceClear() async {
+    aiLog('[Credentials] Queue stalled; deleting outside it');
+    try {
+      await _storage.delete(key: _kRecordKey).timeout(_kCallerTimeout);
+      aiLog('[Credentials] Cleared out of band');
+    } catch (e) {
+      // The caller shows this one, unlike other storage failures: the user
+      // believes these credentials are gone and they are not.
+      aiLog('[Credentials] Out-of-band clear failed: ${e.runtimeType}');
+      throw _asServiceError(e);
+    }
   }
 
   /// The stored record as a map, or null when there is nothing usable.
--- a/lib/page/ai_assistant/services/aws_credentials_store.dart
+++ b/lib/page/ai_assistant/services/aws_credentials_store.dart
@@ -14,12 +14,13 @@ import 'package:privacy_gui/core/errors/service_error.dart';
 /// failure rather than "not configured".
 const _kRecordKey = 'ai_assistant_aws_credentials';
 
-/// How long a single storage operation may take before the chain gives up.
+/// How long a caller waits for its own operation before being told it failed.
 ///
-/// Without this, one hung `write` blocks every operation queued behind it — the
-/// user's `clear()` included, so "change configuration" would never take effect
-/// and the config screen would stay disabled with no error.
-const _kOperationTimeout = Duration(seconds: 10);
+/// This bounds the *reported* wait, not the operation. A stalled platform call
+/// keeps its place in the queue — see [AwsCredentialsStore._serialize] for why
+/// abandoning that place is unsafe — so this only stops a caller (a `View`
+/// awaiting a restore, say) from hanging on a keychain that never answers.
+const _kCallerTimeout = Duration(seconds: 10);
 
 final awsCredentialsStoreProvider = Provider<AwsCredentialsStore>((ref) {
   return AwsCredentialsStore(const FlutterSecureStorage());
@@ -44,8 +45,10 @@ final awsCredentialsStoreProvider = Provider<AwsCredentialsStore>((ref) {
 /// the "change configuration" that followed it — resurrecting credentials the
 /// user had just discarded.
 ///
-/// Each operation is bounded by [_kOperationTimeout] so one that never settles
-/// cannot hold the queue — and with it the user's `clear()` — indefinitely.
+/// Operations never overlap, even when one stalls: [_kCallerTimeout] bounds what
+/// a caller waits for, not the operation itself. See [_serialize] — abandoning a
+/// stalled operation's place in the queue is what makes credentials go missing
+/// or come back.
 ///
 /// The ordering guarantee holds because [awsCredentialsStoreProvider] is a
 /// plain [Provider], i.e. one instance for the session. Making it `autoDispose`
@@ -55,7 +58,7 @@ final awsCredentialsStoreProvider = Provider<AwsCredentialsStore>((ref) {
 ///
 /// Every method throws only [ServiceError] subtypes — [StorageError] for a
 /// platform storage failure, [TimeoutError] when an operation exceeds
-/// [_kOperationTimeout]. `FlutterSecureStorage`'s `PlatformException` never
+/// [_kCallerTimeout]. `FlutterSecureStorage`'s `PlatformException` never
 /// leaves this class, so no caller has to reason about platform-specific error
 /// shapes (constitution Art. XIII §13.1).
 ///
@@ -70,47 +73,49 @@ class AwsCredentialsStore {
   /// Tail of the operation chain; each new operation is appended to it.
   Future<void> _pending = Future.value();
 
-  /// Counts [clear] calls, so a write can tell whether one overtook it.
-  ///
-  /// `Future.timeout` does not cancel the operation it wraps — it only stops
-  /// waiting. A write that exceeds [_kOperationTimeout] therefore keeps running
-  /// after the queue has moved on, and can reach storage *after* a later
-  /// `clear()` has deleted the record — resurrecting credentials the user
-  /// revoked, which is the failure serialising exists to prevent. Timing the
-  /// queue out is what makes that reachable, and not timing it out lets one
-  /// hung write block the `clear()` entirely: same outcome, different route.
-  ///
-  /// So the timeout stays, and this counter gives `clear()` the last word
-  /// instead. See [_undoIfCleared].
-  int _clearGeneration = 0;
-
   /// Run [operation] after every operation requested before it.
   ///
-  /// The chain is never broken by a failure: the tail swallows errors so a
-  /// rejected write cannot stop later operations from running, while the
-  /// returned future still reports the failure to this caller. A timeout counts
-  /// as a failure, which is what stops a stalled operation from blocking the
-  /// queue forever.
+  /// The queue waits for each operation to genuinely settle; only the future
+  /// handed back to the caller is time-bounded.
   ///
-  /// This is also the single place platform errors are converted, so each
-  /// method body can stay free of error handling.
+  /// ## Why the timeout must not apply to the queue
+  ///
+  /// `Future.timeout` does not cancel the operation it wraps — it only stops
+  /// waiting. So timing out the *queue* does not remove a stalled platform call;
+  /// it lets the next operation start while that call is still in flight, and
+  /// whichever finishes last wins at the storage layer. Every ordering
+  /// guarantee this class exists to provide is lost at that moment:
+  ///
+  /// * a stalled `clear()` can delete credentials the user entered afterwards,
+  ///   while their `store()` reports success
+  /// * a stalled `store()` can restore credentials a later `clear()` removed
+  /// * a stalled `storeModelId()` — a read-modify-write — can rewrite the whole
+  ///   record from its stale snapshot over newer credentials, with no `clear()`
+  ///   involved at all
+  ///
+  /// Guarding writes with a "has a clear happened since?" counter cannot fix
+  /// this: it answers a different question than the one that matters, which is
+  /// whether the record in storage is still the one this operation wrote. Since
+  /// the platform gives no way to cancel and `write` is not conditional, the
+  /// only sound answer is to never let two operations overlap.
+  ///
+  /// The cost is that one hung platform call blocks later operations — including
+  /// a `clear()` — for as long as it hangs. That is a genuine downside, accepted
+  /// because the alternative silently loses or resurrects credentials, and
+  /// because the caller is still told promptly via [_kCallerTimeout].
   Future<T> _serialize<T>(Future<T> Function() operation) {
-    final result = _pending
-        .then((_) => operation().timeout(_kOperationTimeout))
+    final settled = _pending.then((_) => operation());
+    // Chained to the real completion, so nothing starts early. This also
+    // handles a late failure: without a listener, an operation that fails after
+    // its caller's timeout has stopped watching becomes an unhandled async
+    // error.
+    _pending = settled.then((_) {}, onError: (Object e) {
+      aiLog('[Credentials] Operation failed after its caller stopped waiting: '
+          '${e.runtimeType}');
+    });
+    return settled
+        .timeout(_kCallerTimeout)
         .onError<Object>((e, _) => throw _asServiceError(e));
-    _pending = result.then((_) {}, onError: (_) {});
-    return result;
-  }
-
-  /// Delete the record if a [clear] happened since [generationAtStart].
-  ///
-  /// Covers the write that lost its place in the queue by timing out: it cannot
-  /// be cancelled, so the record it may have just written is removed instead.
-  /// Called after every write, and cheap when nothing raced.
-  Future<void> _undoIfCleared(int generationAtStart) async {
-    if (_clearGeneration == generationAtStart) return;
-    await _storage.delete(key: _kRecordKey);
-    aiLog('[Credentials] Discarded a write that a clear had superseded');
   }
 
   /// Convert anything storage throws into a [ServiceError].
@@ -135,16 +140,14 @@ class AwsCredentialsStore {
   /// Persist the credentials and the selected model.
   ///
   /// Values are never logged. Throws [StorageError] or [TimeoutError].
+  ///
+  /// A [TimeoutError] means only that the platform has not answered yet — the
+  /// write is still queued and will still be applied in order.
   Future<void> store({
     required String accessKeyId,
     required String secretAccessKey,
     required String modelId,
   }) {
-    // Captured here, synchronously, NOT inside the queued body: the body does
-    // not run until the queue reaches it, by which point a `clear()` issued
-    // right after this call would already have bumped the counter and the
-    // comparison would see no race.
-    final generation = _clearGeneration;
     return _serialize(() async {
       await _storage.write(
         key: _kRecordKey,
@@ -155,7 +158,6 @@ class AwsCredentialsStore {
         }),
       );
       aiLog('[Credentials] Stored');
-      await _undoIfCleared(generation);
     });
   }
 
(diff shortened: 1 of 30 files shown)


AustinChangLinksys (approved):
🤖 **Automated Review — Round 1** · `264b874..64ce201` (full)

**Verdict: ✅ APPROVE** — Correct fix for #1203; error boundary well-designed with good test coverage. Minor documentation and test-coverage gaps noted.

#### Standards

|---|---|---|---|
| ⚠️ | 🟢High | `aws_credentials_store.dart:164–177` | Doc block for `readWithin()` opens with `read()`'s return-value semantics; `read()` itself has no doc comment |
| ⚠️ | 🟡Med | `router_assistant_view.dart:614–620` | `clear()` failure path (`_configError` set asynchronously) has no widget test; logic is correct but fragile to refactoring |
| ⚠️ | 🟢High | `router_assistant_view.dart` (no widget test file) | New view paths (`_restoreSavedCredentials`, `_configFromEnvironment` rethrow path, `clear()` failure display) uncovered by widget tests |
| 💡 | 🟡Med | `router_assistant_view.dart:278–285` | [Feature Envy] `_logStorageFailure` deconstructs `StorageError.originalError`; belongs closer to `StorageError` or logging util |
| 💡 | ⚪Low | `aws_credentials_store.dart:177–179` | Double-timeout stacking (`readWithin` on top of `_serialize`'s inner timeout) lacks inline comment explaining intentional layering |
| 💡 | ⚪Low | `router_assistant_view.dart:33–1010` | [Large Class] View conflates config-screen and chat-screen stateful logic (1010 lines); pre-existing but grows with each PR |

#### Spec

|---|---|---|---|
| ✅ | 🟢High | All criteria | All #1203 acceptance criteria met (StorageError mapping, no raw text to UI, exhaustive switch, failedToInitialize removed, test coverage, 3463/3463 pass) |

<details>

**W-1: Doc comment misplacement — `aws_credentials_store.dart:164–177`** 🟢High

```dart
// Lines 164–177 — the entire doc block sits before readWithin()
/// The stored credentials, or null when nothing usable is saved.
///
/// Returns null rather than a half-built record for anything unusable —
/// absent, blank, or unparseable — so the caller's "not configured" path
/// handles it instead of a signing error later. A bad record is therefore not
/// an error; only storage itself failing is ([StorageError], [TimeoutError]).
/// As [read], but giving up after [within] instead of [_kCallerTimeout].
///
/// For a caller that is holding UI hostage...
Future<StoredAwsCredentials?> readWithin(Duration within) => read()  // line 177

Future<StoredAwsCredentials?> read() {   // line 181 — NO doc comment
```

The first five lines of this block describe `read()`'s return-value contract (null semantics, error types), not `readWithin()`'s. `dart doc` and IDE hover attribute the entire block to `readWithin()`; `read()` at line 181 shows as undocumented. **Fix:** Move the null/error semantics paragraph to `read()`; leave `readWithin()` doc starting at "As `[read]`, but giving up after `[within]`...".

---

**W-2: `clear()` failure display — missing widget test** 🟡Med (`router_assistant_view.dart:614–636`)

```dart
// Line 612–636
ref.read(awsCredentialsStoreProvider).clear().catchError((Object e) {
  _logStorageFailure('clear credentials')(e);
  if (!mounted) return;
  setState(() => _configError = _ConfigError.failure(...));  // async, T+X
});
setState(() {   // sync T+0
  ...
  _configError = null;   // clears first, then error appears later — correct
  ...
});
```

Logic is correct: the synchronous `setState` clears the screen first, then the async `catchError` sets the error message. However this ordering is not covered by any test, making it fragile to refactoring (e.g., accidentally `await`-ing `clear()` or reordering the setStates). The 10-second `_kCallerTimeout` means users could see a blank config screen for up to 10 seconds before the error appears — this known trade-off should be commented inline.

---

**W-3: View layer new paths uncovered** 🟢High

The following new stateful paths in `RouterAssistantView` have no widget tests:
- `_restoreSavedCredentials()` — `_isRestoring` flag locking form during async read
- `_configFromEnvironment()` rethrow path (non-ConfigurationException/NotInitializedError errors)
- `clear()` failure → error display on config screen
- `_initControllerWithManualConfig(persist: false)` restore path

`aws_credentials_store_test.dart` and `aws_environment_config_test.dart` cover service-layer behaviour well. The PR note acknowledges "router_assistant_view.dart has no widget test" — this warning tracks the gap for follow-up.

</details>

<details>

- **`_serialize()` timeout semantics**: The fix correctly places `timeout()` on the returned caller future, not on `_pending`. The comment in `_serialize()` exhaustively explains why timing out the queue breaks ordering. Test suite with per-call-latency fakes validates all three ordering scenarios (stall-then-clear, clear-while-write-in-flight, slow-write-alone).
- **`_asServiceError()` pass-through**: `ServiceError` instances pass through unchanged, preventing double-wrapping. `TimeoutException` → `TimeoutError`, all else → `StorageError`. Detail deliberately placed on `originalError` not `detail` (prevents leak into UI via `localizeServiceError`).
- **Sealed `_ConfigError`**: Replaces `String?` hand-rolled tagged union. Exhaustive `switch` in `_localizeConfigError` prevents new error kinds from silently falling through to raw string display — precisely what the old `return error;` fallback did.
- **`readWithin()` correctness**: With `_restoreTimeout` (5s) < `_kCallerTimeout` (10s), outer timeout always fires first. Inner `TimeoutError` from `_serialize` also passes through `_asServiceError` as a ServiceError pass-through — no double-wrapping.
- **`aws_environment_config_test.dart`**: Pins the third-party `NotInitializedError is not Exception` assumption — will fail loudly if `flutter_dotenv` changes its error hierarchy.
- **`awsCredentialsStoreProvider` as plain `Provider`**: Correctly not `autoDispose`; class doc explains why (fresh instances would revive the ordering race invisibly).
- **`failedToInitialize` removal**: Verified unused in Dart, tests, and non-Dart sources before removal. ARB diffs are pure deletions across all 26 locales.

</details>

lib/page/ai_assistant/services/aws_credentials_store.dart
```diff
@@ -58,21 +75,74 @@ class AwsCredentialsStore {
...
-    final result =
-        _pending.then((_) => operation().timeout(_kOperationTimeout));
-    _pending = result.then((_) {}, onError: (_) {});
-    return result;
+    final settled = _pending.then((_) => operation());
+    // Chained to the real completion, so nothing starts early. This also
+    // handles a late failure: without a listener, an operation that fails after
+    // its caller's timeout has stopped watching becomes an unhandled async
+    // error.
+    _pending = settled.then((_) {}, onError: (Object e) {
```
PeterJhongLinksys, line 113: ⚠️ **Warning** 🟢 — This `onError` listener fires on *every* failure, not only on failures that arrive after the caller's timeout. A write that fails in 10ms — one the caller received as a `StorageError` through the returned future — still logs "failed after its caller stopped waiting", which will mislead field triage into thinking an error went unhandled. Either track an `abandoned` flag set when the caller-side timeout fires, or make the message neutral: `'[Credentials] Operation failed: ${e.runtimeType}'`.

lib/page/ai_assistant/services/aws_credentials_store.dart
```diff
@@ -95,7 +165,19 @@ class AwsCredentialsStore {
   ///
   /// Returns null rather than a half-built record for anything unusable —
   /// absent, blank, or unparseable — so the caller's "not configured" path
-  /// handles it instead of a signing error later.
+  /// handles it instead of a signing error later. A bad record is therefore not
+  /// an error; only storage itself failing is ([StorageError], [TimeoutError]).
+  /// As [read], but giving up after [within] instead of [_kCallerTimeout].
```
PeterJhongLinksys: ⚠️ **Warning** 🟢 — Two things collided here. (1) Lines 164-169 are `read()`'s DartDoc (they describe the null semantics), but they now sit above `readWithin`, leaving `read()` at :181 with no doc at all; this line starts `readWithin`'s own description mid-paragraph. (2) "instead of [_kCallerTimeout]" is not what happens: `read()` already goes through `_serialize`, which applies `.timeout(_kCallerTimeout)` at :117, so the effective bound is `min(within, _kCallerTimeout)`. The current 5s caller hides it, but a future `readWithin(Duration(seconds: 30))` would silently give up at 10s. Suggest threading the bound through `_serialize` (e.g. `_serialize(op, {Duration? callerTimeout})`) so there is exactly one timeout, and splitting the doc back into two blocks.

lib/page/ai_assistant/views/router_assistant_view.dart
```diff
@@ -1,10 +1,13 @@
 import 'package:flutter/material.dart';
+import 'package:flutter_dotenv/flutter_dotenv.dart' show NotInitializedError;
```
PeterJhongLinksys, line 2: ⚠️ **Warning** 🟢 — Importing `NotInitializedError` from `flutter_dotenv` puts a third-party infrastructure error type into the Presentation layer, and the classification at :154-163 (`e is! ConfigurationException && e is! NotInitializedError`) means the View knows the config is backed by dotenv. Art. V §5.4 / Art. XIII want that knowledge below the boundary: moving `_configFromEnvironment()` into `services/` would keep both the dotenv and `generative_ui` imports there and leave the View with a plain nullable/boolean UI model — and, as a side benefit, make this logic unit-testable, which it currently is not.

lib/page/ai_assistant/views/router_assistant_view.dart
```diff
@@ -531,11 +605,18 @@ class _RouterAssistantViewState extends ConsumerState<RouterAssistantView> {
...
+              // config screen they are about to see.
               ref
                   .read(awsCredentialsStoreProvider)
                   .clear()
                   .catchError((Object e) {
-                aiLog('RouterAssistantView: Could not clear credentials: $e');
+                _logStorageFailure('clear credentials')(e);
+                if (!mounted) return;
+                setState(() => _configError = _ConfigError.failure(
+                    e is ServiceError ? e : UnexpectedError(originalError: e)));
```
PeterJhongLinksys, line 619: ⚠️ **Warning** 🟢 — Three issues converge on this handler. (1) The View is constructing a `ServiceError` itself, duplicating `AwsCredentialsStore._asServiceError`; and it contradicts `_logStorageFailure`'s own comment at :275 ("The store throws only ServiceErrors, so there is nothing to map here") — the same contract is trusted at :182 and defended against here. (2) `TimeoutError` should not be shown at all: `clear()`'s doc at aws_credentials_store.dart:217-222 states a timeout means the delete is still queued and will be applied in order, so surfacing it tells the user the revocation failed when it will in fact happen. (3) The comment above argues a failed revocation must be shown because the user would otherwise believe credentials were removed — but `StorageError` renders as `errorUnexpected` ("Something went wrong. Please try again.") and `TimeoutError` as `errorTimeout`, neither of which conveys "the credentials may still be on this device". Suggest: show only for `StorageError`, with a dedicated l10n key. Related: `service_error_localizations.dart:48` still documents `StorageError` as "never reaches UI" — that contract is now broken and should be updated.

PeterJhongLinksys (approved):
🤖 **Automated Review — Round 1** · `264b874..64ce201` (full)

**Verdict: ✅ APPROVE** — The `ServiceError` boundary in the store is correctly built and the real information leak (`failedToInitialize:$e` shown to the user) is genuinely fixed; remaining items are non-blocking Warnings, mostly doc/implementation drift and layering.

|---|---|---|---|
| ⚠️ | 🟢 | `services/aws_credentials_store.dart:106-119` | A platform call that never settles poisons `_pending` for the whole session — every later `store`/`read`/`clear` only ever gets `TimeoutError`, so a revocation can never be applied and credentials silently restore next launch (accepted trade-off in the doc, but there is no recovery path and no user-visible signal). |
| ⚠️ | 🟢 | `services/aws_credentials_store.dart:170,177-179` | `readWithin` doc says "instead of `_kCallerTimeout`" but `read()` is already `.timeout(_kCallerTimeout)`, so the effective bound is `min(within, 10s)` — any future `readWithin(30s)` silently fails at 10s. |
| ⚠️ | 🟢 | `services/aws_credentials_store.dart:164-181` | Doc comment misplacement: `read()`'s DartDoc has been absorbed into `readWithin`'s block (line 170 starts mid-paragraph) and `read()` is left with no doc at all (Art. X §10.1). |
| ⚠️ | 🟢 | `services/aws_credentials_store.dart:112-115` | The `_pending` onError listener fires on *every* failure, so a fast failure the caller did receive still logs "failed after its caller stopped waiting" — misleading for field triage. |
| ⚠️ | 🟢 | `services/aws_credentials_store.dart:59-61,106-118` | Class doc says `TimeoutError` means "an operation exceeds `_kCallerTimeout`", but the timeout is attached at enqueue time so it covers queueing + operation; the store's own test shows a caller getting `TimeoutError` for a write that ultimately succeeds. |
| ⚠️ | 🟢 | `views/router_assistant_view.dart:125-134,262-268,619` | View catches raw throwables and constructs `UnexpectedError(originalError: e)` itself — Art. XIII puts the `ServiceError` conversion point in the service layer only; the leak was replaced by "View acts as Service". |
| ⚠️ | 🟢 | `views/router_assistant_view.dart:2,154-163` | Presentation layer imports third-party infrastructure error types (`NotInitializedError` from `flutter_dotenv`, plus `ConfigurationException`) and does the classification inline — swapping dotenv would require editing a View (Art. V §5.4). |
| ⚠️ | 🟢 | `views/router_assistant_view.dart:608-620` vs `localizations/service_error_localizations.dart:48` | The mapper explicitly documents `StorageError` as "never reaches UI", but the failed-revocation path now sends it to `localizeServiceError` — a cross-file contract broken without updating it. |
| ⚠️ | 🟢 | `views/router_assistant_view.dart:608-620` | The 4-line comment argues a failed revocation MUST be shown, yet the rendered text is `errorUnexpected` / `errorTimeout` — neither conveys "your credentials may still be on this device", which was the entire point. |
| ⚠️ | 🟢 | `views/router_assistant_view.dart:615-620` vs `services/aws_credentials_store.dart:217-222` | `clear()`'s `TimeoutError` is shown as a failure although the store documents it as "still queued, not dropped" — the user is told the revocation failed when it will in fact be applied. |
| ⚠️ | 🟡 | `views/router_assistant_view.dart:279-285` | `_logStorageFailure` interpolates the raw `PlatformException` (`originalError`) into `aiLog`, which is release-on by contract (`ai/ai_logging.dart:9-14` reserves content for `aiLogSensitive`); `_serialize` already logs only `runtimeType`, so the two are inconsistent. |
| ⚠️ | 🟢 | `views/router_assistant_view.dart:619` vs `:182,279` | `e is ServiceError ? e : UnexpectedError(...)` duplicates `_asServiceError` and contradicts `_logStorageFailure`'s own comment that the store throws only `ServiceError`s — the same contract is trusted in one place and defended against in another. |
| ⚠️ | 🟢 | `test/page/ai_assistant/aws_environment_config_test.dart:1-69` | [Fragile Test] The file exercises only `generative_ui`/`dotenv`, no repo code, and depends on execution order ("the first test must run before anything loads it") — it would break under randomized ordering or an inserted test. |
| ⚠️ | 🟢 | `views/router_assistant_view.dart` (1010 lines) | [Large Class] No Application layer for this page (`providers/` holds no Notifier), so all state, restore, persistence and error mapping live in the StatefulWidget — which is why this PR's new error handling has no unit coverage and the Provider ≥85% bar cannot apply. |
| ⚠️ | 🟢 | `test/.../aws_credentials_store_test.dart:139-149,308-400,466-580` | [Duplicated Code] No Test Data Builder for AWS credentials (`test/mocks/test_data/` has none) — the same literal triple and the same 3-line `fakeAsync` store setup are repeated across the new tests. |
| 💡 | 🟢 | `views/router_assistant_view.dart:262-268` | Removing `failedToInitialize` collapses a user-fixable failure (mistyped key) into the generic `errorUnexpected`; consider mapping recognizable validation failures to `InvalidInputError` to keep actionability. |
| 💡 | 🟢 | `views/router_assistant_view.dart:279-285,182,616` | `_logStorageFailure(action)(e)` double-call reads awkwardly; a plain `_logStorageFailure(action, e)` avoids allocating a closure per call site. |
| 💡 | 🟢 | `services/aws_credentials_store.dart:177-179` vs `:116-118` | [Duplicated Code] The `.timeout(...).onError(_asServiceError)` wrapper exists twice; `_serialize(op, {Duration? callerTimeout})` + `read({Duration? within})` would remove the extra public API and the ambiguous double bound. |
| 💡 | 🟢 | `views/router_assistant_view.dart:88` vs `services/aws_credentials_store.dart:23` | The 5s restore bound and the store's 10s caller bound describe the same policy in two layers, and the 10s is dead on that path; consider exposing the restore bound as a named constant on the store. |
| 💡 | 🟢 | `views/router_assistant_view.dart:992-1010` | `_ConfigError` / `_MissingFields` / `_Failure` sit at the tail of a 1010-line View; they carry state semantics and would be easier to find (and to test) in `providers/` or a `models/` file. |
| 💡 | 🟡 | `views/router_assistant_view.dart:992-1010` | `_ConfigError` subtypes do not implement `Equatable` (Art. XI); harmless while View-private, but it becomes redundant rebuilds and awkward expectations the moment this moves into a Notifier state. |
| 💡 | 🟡 | `services/aws_credentials_store.dart:132-138` | `_asServiceError` wraps any `Error` (`TypeError`, `StateError`) as `StorageError`, so a genuine code bug is reported as an environment/storage failure in logs and telemetry; `UnexpectedError` would classify it honestly. |
| 💡 | 🟡 | `services/aws_credentials_store.dart:135` | `TimeoutError(detail: 'credential storage')` is safe only because `localizeServiceError` currently ignores `TimeoutError.detail` (`service_error_localizations.dart:39`); if it ever adopts the `detail ?? …` pattern used for `UnexpectedError`, this English string reaches the UI. |
| 💡 | 🟡 | `views/router_assistant_view.dart:107-134` | The `try` block also wraps `ref.read(routerCommandProviderProvider)` and `buildRouterContext`, so a Riverpod wiring failure degrades to a generic "Something went wrong" with one log line. |
| 💡 | 🟡 | `test/.../aws_credentials_store_test.dart:13,27-72,649-669` | Two doubling strategies coexist (Mocktail `MockSecureStorage` used by one test, hand-written `FakeSecureStorage` by the rest); the Fake earns its place via the `log` ordering assertions, but the split deserves a note at the top of the file. |
| 💡 | ⚪ | `services/aws_credentials_store.dart:25-26` | Default `const FlutterSecureStorage()` with no `WebOptions`: on the web target this is obfuscation rather than protection for a long-lived AWS secret. The class doc already concedes a backend signing proxy is the right answer — worth an issue-numbered TODO rather than a comment. |
| 💡 | ⚪ | `views/router_assistant_view.dart:249-261` | The fire-and-forget `store()` failure is intentionally log-only, which is defensible — but it sits in tension with the "a failed revocation IS shown" rule and the two policies should be stated together. |

<sub>Confidence: 🟢High = read head code, evidence quoted · 🟡Med = file:line + reasoning, not fully proven · ⚪Low = speculative.</sub>
<sub>`[X]` = raised by 1 agent, `[XX]` = raised by both. No Critical findings this round.</sub>

<details>

**1. A hung platform call poisons the queue for the whole session** — `services/aws_credentials_store.dart:106-119` (🟢, both agents)
```dart
Future<T> _serialize<T>(Future<T> Function() operation) {
  final settled = _pending.then((_) => operation());
  _pending = settled.then((_) {}, onError: (Object e) { ... });
  return settled
      .timeout(_kCallerTimeout)
      .onError<Object>((e, _) => throw _asServiceError(e));
}
```
`_pending` is now chained to the *real* completion. If `operation()` never settles, `_pending` never settles, so every subsequent `store()` / `read()` / `storeModelId()` / `clear()` on this instance can only ever end in `TimeoutError`. `awsCredentialsStoreProvider` is a plain `Provider` (`:25`), one instance per session, so there is no recovery path. Sequence: a hung `store()` → the user presses "change configuration" → `clear()` is queued behind it forever → the UI shows the generic timeout text → next launch `_restoreSavedCredentials` (`views/router_assistant_view.dart:175-207`) reads the old record back and auto-connects with credentials the user believes they revoked. Re-entering credentials in the same session does not help either, because the new `store()` sits in the same queue. The reasoning for not timing out the queue is sound and the test at `test/.../aws_credentials_store_test.dart:556-585` pins the behaviour deliberately — the gap is that there is no poison detection and no user-visible consequence. Consider bounding each platform call *inside* `operation()` plus a record-generation marker (a different mechanism from the clear-counter the doc rules out), or marking the store unavailable and refusing auto-restore on the next launch.

**2. `readWithin` is a double timeout, contradicting its own doc** — `services/aws_credentials_store.dart:170,177-179` (🟢)
```dart
/// As [read], but giving up after [within] instead of [_kCallerTimeout].
Future<StoredAwsCredentials?> readWithin(Duration within) => read()
    .timeout(within)
    .onError<Object>((e, _) => throw _asServiceError(e));
```
`read()` already goes through `_serialize`, i.e. `settled.timeout(_kCallerTimeout)` at `:117`. The effective bound is `min(within, 10s)`, not "instead of". The only caller uses 5s (`views/router_assistant_view.dart:88`) so it is currently invisible; the test at `:676-678` even concedes "which of the two bounds fired is an implementation detail". Passing a bound longer than 10s would silently fail at 10s. Either thread the bound through `_serialize`, or document it as `min(...)` and assert.

**3. View performs the `ServiceError` conversion** — `views/router_assistant_view.dart:125-134`, `:262-268`, `:619` (🟢)
```dart
} catch (e) {
  aiLog('RouterAssistantView: could not start a session from environment config: $e');
  _needsConfig = true;
  _configError = _ConfigError.failure(UnexpectedError(originalError: e));
}
```
and
```dart
setState(() => _configError = _ConfigError.failure(
    e is ServiceError ? e : UnexpectedError(originalError: e)));
```
Art. XIII puts the conversion point in the service layer so that Provider/UI only ever *see* a `ServiceError`. Here the View decides which `ServiceError` a failed `AWSConfig.fromEnvironment()` / `AwsContentGenerator` / `RouterChatController` construction becomes, and `:619` re-implements `_asServiceError`'s guard clause. Extracting an `AiSessionService` (owning `_configFromEnvironment` + controller construction, and the dotenv/`generative_ui` imports) would put both this and the `NotInitializedError` import back below the boundary — and would make the new error handling unit-testable, which it currently is not.

**4. Failed-revocation UX loses the message it was added to convey** — `views/router_assistant_view.dart:608-620` (🟢)
The comment states a failed revocation must be shown because "the user would believe credentials were removed that are still there". The actual rendering is `StorageError() => l.errorUnexpected` ("Something went wrong. Please try again.") or `TimeoutError() => l.errorTimeout` — neither mentions the stored record. Worse, `TimeoutError` here is documented by the store itself (`:217-222`) as "still queued, not dropped", so showing it as a failure is actively wrong. Suggest a dedicated l10n key (and a dedicated `_ConfigError` case), shown only for `StorageError`.

**5. Raw platform error interpolated into an always-on log** — `views/router_assistant_view.dart:279-285` (🟡)
```dart
void Function(Object) _logStorageFailure(String action) {
  return (Object e) {
    final cause = e is StorageError ? e.originalError : null;
    aiLog('RouterAssistantView: could not $action: $e'
        '${cause != null ? ' ($cause)' : ''}');
  };
}
```
`originalError` is the original `PlatformException`, whose `details` is `dynamic` and platform-defined (on web it comes from browser storage). `ai/ai_logging.dart:9-14` reserves `aiLog` for structural diagnostics and routes content through debug-only `aiLogSensitive`. The same pattern appears at `:130-131` and `:263`, and `:263` wraps the construction of `AWSConfig(accessKeyId:…, secretAccessKey:…)` — if any dependency ever throws an `ArgumentError.value`-style error carrying the value, it lands in a release log. The UI side is safe (`detail` stays null), so this is a defensive hardening, not a demonstrated leak: `${e.runtimeType}` in `aiLog` plus `aiLogSensitive` for the cause would match what `_serialize` already does at `:113-114`.

</details>

<details>

- **The real leak is fixed.** The old `_configError = 'failedToInitialize:$e'` fed the raw exception text into `loc(context).failedToInitialize(detail)` and straight onto the screen. Now `detail` stays null and `localizeServiceError` falls back to a generic message, with the l10n key removed from all 26 locale files — `failedToInitialize` has zero remaining hits across `lib/` and `test/` (grepped on the head worktree).
- **The `detail` vs `originalError` split is judged precisely.** `service_error_localizations.dart:47` does surface `UnexpectedError.detail` to the user, and the comment at `aws_credentials_store.dart:123-131` shows the author read that before deciding where the keychain code goes. `test/.../aws_credentials_store_test.dart:715-727` pins it with `expect(error.detail, isNull)` — a high-quality regression guard.
- **`_localizeConfigError` moved from string sniffing to a sealed exhaustive switch**, deleting the old `error.startsWith('failedToInitialize:')` branch and the `return error;` fallback that painted arbitrary strings onto the screen. The note on binding `_Failure(error: final serviceError)` rather than `:final error` (to avoid shadowing a parameter that would still compile against an `Object` argument) is exactly the right level of care.
- **`_configFromEnvironment` uses `catch (e)` + type tests rather than `on Exception catch`**, and `aws_environment_config_test.dart:542-558` pins the reason (`NotInitializedError` extends `Error`, so an `on Exception` clause would let it through). This is the same class of defect as the USP raw-`String` error trap, handled proactively.
- **The serialization argument is correct and non-obvious** (`:81-105`: `Future.timeout` does not cancel the operation, so timing out the queue would let two operations race at the storage layer), and it is backed by three complementary tests — stalled write then clear, clear during an in-flight write, and a negative case ensuring a slow write with nothing behind it is simply applied. `:843` honestly asserts `cleared, isFalse` for the accepted trade-off instead of hiding it.
- **`readWithin` exists for the right reason**: `.timeout()` at the call site would throw a bare `TimeoutException` and break the only-`ServiceError` contract; `:680-681` asserts `isNot(isA<TimeoutException>())`.
- **`Provider` rather than `autoDispose` is justified in writing** (`:53-55`), matching Art. IV Rule 1.
- No `generated/` imports in `providers/` or `views/` under `lib/page/ai_assistant/` (Art. V §5.4 clean), and no USP mutation paths touched, so Art. IV Rules 3–4 do not apply.

</details>

PR #1209, opened 2026-08-06
feat(golden-report): report where and why a golden overflowed (#1197)

PeterJhongLinksys:
Closes #1197

## Problem

The golden reports only showed an `OVERFLOW` badge. The runner recorded the
golden name and Flutter's one-line message and nothing else, so neither report
could say which widget overflowed, where, or by how much — the badge told you a
run had a problem but not what to open.

Worse, the full diagnostics dump was already being computed to resolve the
location and then thrown away. That left the cases where the badge is set but no
overflow stripe is visible in the image with no diagnostic trail at all.

## What this branch does

**Records the detail.** `overflow_diagnostics.dart` extracts the overflowing
widget, file, line, direction and pixel amount from Flutter's error, and keeps
the whole diagnostics dump alongside them. Absolute paths are rewritten to
repo-relative form (and pub-cache paths to `<package>/<path>`), so a report built
on CI does not embed the runner's workspace.

**One shared loader.** Both generators now go through
`test_scripts/overflow_details.dart`, which collapses Flutter's
per-`RenderObject` duplicates into sites with an occurrence count. Previously the
two reports each did their own thing and could show different counts for the same
run.

**Raw log button.** Every site names the culprit on the gallery card / verify
test row, with a `raw log` button that opens the full dump in a modal with a copy
button. Shown for sites whose location did not resolve too — that is exactly when
the dump is the only lead. The UI stays one line per site; the dump is behind the
button.

**Stable, non-duplicated storage.** Per-run object ids (`RenderFlex#4195b`) are
stripped, so the same overflow produces the same text on every run — without this
one culprit reported in 24 goldens stayed 24 "distinct" logs. Dumps are stored
once in an indexed table referenced by `logIndex`: inline storage made 76% of the
record file duplicated text (~8MB extrapolated to a full run); the dhcp+fr run
went 56KB → 14KB. The loader still reads the older flat-list format.

**Two fixes found on the way:**
- The detail now renders on *passing* rows. An overflow does not fail a test, and
  all 33 overflows in a manual `fr` run landed on passing tests — the original
  code only rendered detail for failures, so none of them would have shown it.
- `run_golden_verify.sh` clears `goldens/overflow_warnings.json` before running.
  The runner appends (each suite writes at its own `tearDownAll`), so stale
  records from a previous run were being attributed to the current one.

## Out of scope: CI golden stage retired

`.github/workflows/ci.yml` Stage B is commented out. Golden baselines are owned
by the `linksys/PrivacyGUI-golden-ci` orchestration repo, which clones this repo
and runs the golden scripts itself — and the job never produced anything anyway:
it uploaded `snapshots/`, a path the golden runner does not write. Its
`workflow_dispatch` trigger and the now-jobless `push: main` trigger go with it.
Commented rather than deleted so the SSH / env-template preamble stays on hand.

Unrelated to #1197 — flagging it here rather than hiding it in the diff.

## Verification

- `flutter test test/test_scripts/overflow_details_test.dart` +
  `test/test_scripts/overflow_diagnostics_test.dart` — 60 tests passed
- `fvm dart analyze` clean; `fvm dart format --set-exit-if-changed` 0 changed
- End-to-end over dhcp/admin/local_network in `fr` at 480px: the written record
  file is `{records, logs}` with 30 records collapsing to 4 distinct logs, all 30
  `logIndex` values in range, no inline `log` key left behind
- Both reports regenerate and carry the modal and log table; no absolute path and
  no per-run object id in the `overflowSites` / `overflowLogs` data of either

The parser tests assert against a real diagnostics dump rather than a fixture
string: Flutter's dump format is an implementation detail, not an API contract, so
an SDK upgrade that changes it should fail loudly in one place instead of
silently emptying the reports.

## Notes for review

- Nothing here changes app code or any golden image — it is all test tooling and
  report generation.
- Pre-existing and **not** touched by this branch: `DATA.tests[].messages` (from
  `test_result_parser.dart`) carries raw test stdout, including absolute paths and
  object ids from stack traces. Worth a separate issue if we want the report free
  of machine-specific text end to end.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,19 +1,26 @@
 name: PrivacyGUI CI/CD
 
 on:
-  push:
-    branches: [ "main" ]
   pull_request:
     branches: [ "main", "dev-*" ]
-  workflow_dispatch:
-    inputs:
-      job:
-        description: 'Select job to run'
-        required: true
-        default: 'main_snapshot_baseline'
-        type: choice
-        options:
-          - main_snapshot_baseline
+  # Every job below is gated on `pull_request`, so a push to main produced a
+  # workflow run with no jobs in it. Retired with STAGE B, which was the only job
+  # that ever ran on it.
+  # push:
+  #   branches: [ "main" ]
+  # Golden snapshot generation moved to the linksys/PrivacyGUI-golden-ci
+  # orchestration repo, which owns the baseline PNGs this repo cannot commit
+  # (.gitignore: test/**/goldens/*, snapshots/*). The manual trigger and its
+  # only choice went with it — see STAGE B below.
+  # workflow_dispatch:
+  #   inputs:
+  #     job:
+  #       description: 'Select job to run'
+  #       required: true
+  #       default: 'main_snapshot_baseline'
+  #       type: choice
+  #       options:
+  #         - main_snapshot_baseline
   # schedule:
   #   - cron: '0 0 * * *' # Nightly run at midnight UTC
 
@@ -81,57 +88,67 @@ jobs:
         run: flutter build web --release --no-pub
 
   # ========================================================================
-  # STAGE B: Main Branch Baseline
+  # STAGE B: Main Branch Baseline (Retired)
   # ------------------------------------------------------------------------
-  # Focus: Generate Goldens for record (No Diff Verification)
-  # Trigger: Push to Main, or Manual Dispatch
+  # Golden baselines are owned by the linksys/PrivacyGUI-golden-ci
+  # orchestration repo, which clones this repo and runs
+  # run_generate_loc_snapshots.sh / run_golden_verify.sh itself. Generating a
+  # baseline is a human judgement ("this rendering is correct — freeze it"), so
+  # it lives on a manual trigger there rather than firing on every push to main.
+  #
+  # This job also never produced anything: it uploaded snapshots/, a path
+  # run_generate_loc_snapshots.sh does not write (the golden runner writes
+  # test/golden_test/**/goldens/ instead).
+  #
+  # Retired rather than deleted so the SSH / env-template / pub-get preamble
+  # stays on hand if a golden job is ever brought back in-repo.
   # ========================================================================
-  main_snapshot_baseline:
-    name: 📸 Generate Baseline Snapshots (Main)
-    if: |
-      (github.event_name == 'push' && github.ref == 'refs/heads/main') ||
-      (github.event_name == 'workflow_dispatch' && github.event.inputs.job == 'main_snapshot_baseline')
-    # Note: If pixel-perfect consistency with local Mac dev is required, 
-    # change this to 'macos-latest'. Keeping ubuntu-latest for cost efficiency.
-    runs-on: ubuntu-latest
-    timeout-minutes: 45
-
-    steps:
-      - uses: actions/checkout@v4
-
-      - name: Setup Flutter
-        uses: subosito/flutter-action@v2
-        with:
-          channel: 'stable'
-          cache: true
-
-      # Setup SSH for private dependencies (Deploy Key approach)
-      - name: 🔑 Setup SSH for Private Deps
-        uses: webfactory/ssh-agent@v0.9.0
-        with:
-          ssh-private-key: ${{ secrets.UI_KIT_DEPLOY_KEY }}
-
-      # Force git to use SSH instead of HTTPS
-      - name: 🔄 Configure Git to use SSH
-        run: git config --global url."git@github.com:".insteadOf "https://github.com/"
-
-      - name: 📋 Setup Environment File
-        run: cp assets/agents/env.template assets/agents/.env
-
-      - name: Install Dependencies
-        run: flutter pub get
-
-      # Generate new snapshots (No failure on diff, just generation)
-      - name: 📸 Generate Localization Snapshots
-        run: sh run_generate_loc_snapshots.sh
-
-      # Save the generated snapshots as artifacts
-      - name: 💾 Upload Snapshot Artifacts
-        uses: actions/upload-artifact@v4
-        with:
-          name: privacygui-goldens-main
-          path: snapshots/
-          retention-days: 7
+  # main_snapshot_baseline:
+  #   name: 📸 Generate Baseline Snapshots (Main)
+  #   if: |
+  #     (github.event_name == 'push' && github.ref == 'refs/heads/main') ||
+  #     (github.event_name == 'workflow_dispatch' && github.event.inputs.job == 'main_snapshot_baseline')
+  #   # Note: If pixel-perfect consistency with local Mac dev is required,
+  #   # change this to 'macos-latest'. Keeping ubuntu-latest for cost efficiency.
+  #   runs-on: ubuntu-latest
+  #   timeout-minutes: 45
+  #
+  #   steps:
+  #     - uses: actions/checkout@v4
+  #
+  #     - name: Setup Flutter
+  #       uses: subosito/flutter-action@v2
+  #       with:
+  #         channel: 'stable'
+  #         cache: true
+  #
+  #     # Setup SSH for private dependencies (Deploy Key approach)
+  #     - name: 🔑 Setup SSH for Private Deps
+  #       uses: webfactory/ssh-agent@v0.9.0
+  #       with:
+  #         ssh-private-key: ${{ secrets.UI_KIT_DEPLOY_KEY }}
+  #
+  #     # Force git to use SSH instead of HTTPS
+  #     - name: 🔄 Configure Git to use SSH
+  #       run: git config --global url."git@github.com:".insteadOf "https://github.com/"
+  #
+  #     - name: 📋 Setup Environment File
+  #       run: cp assets/agents/env.template assets/agents/.env
+  #
+  #     - name: Install Dependencies
+  #       run: flutter pub get
+  #
+  #     # Generate new snapshots (No failure on diff, just generation)
+  #     - name: 📸 Generate Localization Snapshots
+  #       run: sh run_generate_loc_snapshots.sh
+  #
+  #     # Save the generated snapshots as artifacts
+  #     - name: 💾 Upload Snapshot Artifacts
+  #       uses: actions/upload-artifact@v4
+  #       with:
+  #         name: privacygui-goldens-main
+  #         path: snapshots/
+  #         retention-days: 7
 
   # ========================================================================
   # STAGE C: Nightly Deep Check (Commented Out)
--- a/run_golden_verify.sh
+++ b/run_golden_verify.sh
@@ -37,6 +37,13 @@ echo "Locales: $locales"
 echo "Screens: $screens"
 echo "Version: $version"
 
+# The golden runner appends to this file, so a previous run's records would be
+# attributed to this one — reporting overflows at line numbers that have since
+# moved, or on goldens this run never touched. Cleared before, not after, so the
+# file is still readable for debugging once the run ends. Placed ahead of the
+# branch below because both paths generate a report.
+rm -f goldens/overflow_warnings.json
+
 if [ -z "$file" ]; then
   IFS=',' read -ra LOCS <<< "$locales"
   for locale in "${LOCS[@]}"; do
--- a/test/golden_test/golden_framework/golden_runner.dart
+++ b/test/golden_test/golden_framework/golden_runner.dart
@@ -15,6 +15,7 @@ import 'package:privacy_gui/theme/theme_json_config.dart';
 import 'golden_interactions.dart';
 import 'golden_test_config.dart';
 import 'mocks/mock_common.dart';
+import 'overflow_diagnostics.dart';
 
 // Re-export so every test file that imports golden_runner.dart gets the shared
 // interaction helpers (switchToTab, settleWithTimeout) without a separate line.
@@ -343,7 +344,12 @@ Widget _buildGoldenWidget(
 /// Tracks which golden file is currently being rendered.
 String _currentGoldenName = '';
 
-/// Collected overflow warnings: golden filename → error message.
+/// Collected overflow records, one per reported error.
+///
+/// Shape is defined by [buildOverflowRecord]: the golden name, the raw message,
+/// and the parsed side/pixels/widget/file/line. Flutter reports overflow per
+/// RenderObject, so sibling rows built from a list produce duplicate records;
+/// the report generators collapse them.
 final List<Map<String, String>> _overflowWarnings = [];
 
 /// Suppresses RenderFlex overflow errors during golden tests but records them.
@@ -357,10 +363,15 @@ void _suppressOverflowErrors() {
   FlutterError.onError = (details) {
     final isOverflow = details.exceptionAsString().contains('overflowed');
     if (isOverflow) {
-      _overflowWarnings.add({
-        'golden': _currentGoldenName,
-        'message': details.exceptionAsString(),
-      });
+      // Record the direction, amount and source location alongside the raw
+      // message so the reports can say where and by how much, not just that it
+      // happened (#1197). The test process runs from the app root, which is why
+      // the relative paths in _writeOverflowReport resolve.
+      _overflowWarnings.add(buildOverflowRecord(
+        goldenName: _currentGoldenName,
+        details: details,
+        runDirectory: Directory.current.path,
+      ));
       return;
     }
     originalHandler?.call(details);
--- a/test/golden_test/golden_framework/golden_runner.dart
+++ b/test/golden_test/golden_framework/golden_runner.dart
@@ -388,6 +388,19 @@ void _suppressOverflowErrors() {
 /// Appends, because each test suite writes at its own tearDownAll. A file left
 /// by a run predating the log table is read back in its flat-list form so the
 /// records already collected are not dropped.
+///
+/// The read-modify-write is not atomic and not locked, and `flutter test` runs
+/// suites concurrently — so in principle two suites can both read the same
+/// snapshot and the second write can drop the first's records, taking its
+/// `logIndex` values with it. Measured rather than assumed: four full runs of the
+/// 32 golden suites (three concurrent, one `--concurrency=1`) all produced the
+/// same 40 records over 21 goldens with every `logIndex` in range, and no golden
+/// present in the serial run was missing from a concurrent one. The window is
+/// narrow because only 7 suites write at all and the write is one small
+/// `writeAsStringSync`. Left as-is deliberately: this is a diagnostic aside, and
+/// a lost record costs a line in a report rather than a wrong test result. If it
+/// ever does show up, the fix is per-suite shard files merged by the loader, or a
+/// lock plus temp-and-rename.
 void _writeOverflowReport() {
   if (_overflowWarnings.isEmpty) return;
   final dir = Directory('goldens');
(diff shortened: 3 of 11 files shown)


PeterJhongLinksys (commented):
🤖 **Automated Review — Round 1** · `b8894c4..4472a3d` (full)

**Verdict: 💬 Self-review (comment only)** — No Critical found; solid, well-tested diagnostics work, but two shared-state issues (concurrent report writes, missing reset on the gallery path) can make the new report attribute overflows to the wrong golden or the wrong dump.

|---|---|---|---|
| ⚠️ | 🟢 | `test/golden_test/golden_framework/golden_runner.dart:391-435` | Concurrent suites read-modify-write the same `goldens/overflow_warnings.json` with no lock/atomic rename; lost records, and the new positional `logIndex` can point at another suite's dump. |
| ⚠️ | 🟢 | `run_generate_loc_snapshots.sh:36-51` | Missing the new `rm -f goldens/overflow_warnings.json`; this is the path CI Stage B runs (`.github/workflows/ci.yml:126`), so stale records still get attributed to the current gallery run — exactly what `run_golden_verify.sh:45` set out to fix. |
| ⚠️ | 🟢 | `test_scripts/overflow_details.dart:170,213-220` | `as String?` casts sit outside the `try` that returns `OverflowReport.empty`; a type-mismatched field throws `TypeError` and takes the whole report generator down, contradicting the doc comment at `:135-136`. |
| ⚠️ | 🟢 | `test_scripts/overflow_details.dart:143-208` vs `golden_runner.dart:397-434` | Envelope format (`{records, logs}`, legacy list, log interning) is implemented twice — writer in `test/`, reader in `test_scripts/` — and only the reader has tests; writer-side drift would fail silently under `catch (_)`. |
| ⚠️ | 🟢 | `test_scripts/html_generate_functions.dart:206-243,426-435,757-800` vs `generate_gallery_report.dart:353-393,647-656,677-716` | [Duplicated Code] ~115 lines of raw-log viewer CSS/markup/JS copied verbatim into both generators; `escapeAttr` vs `esc` are already identical bodies under different names. |
| ⚠️ | 🟢 | `test/golden_test/golden_framework/overflow_diagnostics_test.dart:1` | `@Tags(['ui'])` is excluded by `run_tests.sh:80` (`--exclude-tags="golden||loc||ui"`), so the blocking CI stage never runs these tests — the "pin the shapes this code depends on" guarantee does not hold on PRs. |
| ⚠️ | 🟡 | `test/golden_test/golden_framework/golden_runner.dart:401-419` | `catch (_)` does not reset `records`/`logs`/`logIndexes`, so a partial-parse failure leaves old records with `logIndex` values misaligned against a freshly-numbered `logs` table; comment says "start the report over" but code does not. |
| ⚠️ | 🟢/⚪ | `test_scripts/generate_gallery_report.dart:659,661`, `html_generate_functions.dart:438` | `jsonEncode` does not escape `<` or U+2028/2029, so a literal `</script>` or line separator inside a dump/message breaks out of the `<script>` block. Mechanism is certain (🟢); reachability today is low (⚪ — sources are `l10n/*.arb` + mocks, currently clean), but this PR greatly widens what raw text gets embedded. |
| 💡 | 🟡 | `test_scripts/generate_gallery_report.dart:474,559-560` | Overflow summary count uses `overflowDetails.length` (all goldens in JSON) while badges come from PNG scan, so the number can exceed the visibly marked cards. |
| 💡 | 🟡 | `test_scripts/combine_results.dart:175-181` | Reconstructed golden key omits `-dark` (`golden_runner.dart:229-233` appends it) and `test_result_parser.dart:220` regex won't match dark test names; details will silently vanish once dark theme is enabled. |
| 💡 | 🟢 | `test/golden_test/golden_framework/overflow_diagnostics.dart:124-126` | [XX] `RegExp(r'(?<=\w)#[0-9a-f]{5}\b')` also strips real content like `Device#12345`/`SSID#abcde`; anchor on known prefixes (`RenderFlex#`, `GlobalKey#`) or require at least one `a-f`. |
| 💡 | 🟡 | `test/golden_test/golden_framework/overflow_diagnostics.dart:26-27` | `_locationPattern` uses `[^\s:]+` for the path, so `file:///C:/...` never matches on Windows → no location, and the untouched absolute path (incl. account name) stays in the dump. |
| 💡 | 🟡 | `test/golden_test/golden_framework/overflow_diagnostics.dart:59-88` | `normalizeSourcePath` returns unrecognized absolute paths unchanged, so `/Users/<name>/...` can reach the JSON/HTML report; `overflow_diagnostics_test.dart:176-177` asserts the opposite but only covers pub-cache. |
| 💡 | 🟢 | `test_scripts/overflow_details.dart:110-121` | `_siteKey` uses `List.join`, which renders `null` as `"null"`, so `file == null` and `file == "null"` collide as the same identity. |
| 💡 | 🟢 | `test_scripts/overflow_details.dart:200-208` | `if (index is int)` silently drops a JSON `1.0`; prefer `(record['logIndex'] as num?)?.toInt()`. |
| 💡 | 🟢 | `test/golden_test/golden_framework/overflow_diagnostics_test.dart:118-139,271-296` | `FlutterError.onError` is restored on the happy path only; if `pumpWidget` throws, the handler leaks into later tests and can swallow errors. Use `addTearDown`. |
| 💡 | 🟢 | `golden_runner.dart:395`, `overflow_details.dart:128,138`, `run_golden_verify.sh:45`, `clear_goldens.sh:18` | `goldens/overflow_warnings.json` is hardcoded in five places — the root cause of the missed reset above. |
| 💡 | 🟢 | `test_scripts/generate_gallery_report.dart:182-1057`, `html_generate_functions.dart:3-1018` | [Long Method] ~875 / ~1015-line generators mixing Dart, CSS, HTML and JS; the new block is cleanly extractable (see duplication row). |
| 💡 | 🟢 | `html_generate_functions.dart:689,759`, `generate_gallery_report.dart:595,679,844` | Magic strings `'location unresolved'` / `'Overflow raw log'` duplicated 3-4× each; one edit yields three wordings in the same report. |
| 💡 | 🟡 | `test/golden_test/golden_framework/overflow_diagnostics.dart:182`, `golden_runner.dart:370` | `toStringDeep()` runs unconditionally per overflow and accumulates in memory until `tearDownAll`; consider de-duplicating by site key at record time. |
| 💡 | 🟢 | `generate_gallery_report.dart:162-166,669-673`, `html_generate_functions.dart:734-747` | Four escaper implementations with slightly different semantics (`escapeHtml` doesn't quote-escape, `escapeAttr` does, neither handles `'`). Safe today because all attributes use double quotes — worth consolidating and documenting. |
| 💡 | 🟢 | `run_golden_verify.sh:45,51-59` | New `rm -f` relies on CWD being repo root (silent no-op elsewhere) and the script has no `set -u`/quoted `$REPORT_DIR`; anchor with `cd "$(git rev-parse --show-toplevel)"`. |
| 💡 | 🟡 | `overflow_diagnostics.dart` vs `test_scripts/overflow_details.dart` | Names don't distinguish producer from consumer, and the field contract (`golden/message/pixels/side/widget/file/line/log/logIndex`) is spelled out in both without shared key constants. |
| 💡 | ⚪ | `test_scripts/overflow_details.dart:15,94` | `OverflowDetail`/`OverflowReport` are not `Equatable`; Art. XI arguably doesn't apply to tooling DTOs, but value equality would simplify the tests' field-by-field assertions. |
| 💡 | ⚪ | `test/test_scripts/overflow_details_test.dart:6` | Relative import reaching out of `test/` into `test_scripts/`; works, but the unit under test falls outside package coverage accounting. |
| 💡 | 🟡 | `html_generate_functions.dart:431`, `generate_gallery_report.dart:652` | Modal close control is a `<span onclick>` (not focusable, no `role`/`aria-label`), and the dialog has no `aria-modal`/focus trap. Escape handling is already correct. |

<sub>Confidence: 🟢 High = read head code, evidence quoted · 🟡 Med = file:line + rationale, not fully proven · ⚪ Low = speculative.</sub>
<sub>`[X]` = raised by 1 agent · `[XX]` = independently raised by both agents. Unmarked rows are single-agent Warnings/Suggestions.</sub>

<details>
<summary>Warning evidence (top items)</summary>

**1. Concurrent read-modify-write of the shared report file**

`golden_runner.dart:72` / `391-435`:
```dart
tearDownAll(() => _writeOverflowReport());
...
final file = File('goldens/overflow_warnings.json');
if (file.existsSync()) {                       // read
  final decoded = jsonDecode(file.readAsStringSync());
  ...
record['logIndex'] = logIndexes.putIfAbsent(log, () { logs.add(log); return logs.length - 1; });
...
file.writeAsStringSync(                        // full overwrite
    JsonEncoder.withIndent('  ').convert({'records': records, 'logs': logs}));
```
`run_golden_verify.sh:51` runs `flutter test test/golden_test/` with default concurrency (no `concurrency` in `dart_test.yaml`), so every golden test file is a separate suite whose `tearDownAll` writes the same file. Sequence: A reads `logs.length == 3`, B reads 3, A writes, B overwrites from its stale snapshot → A's records vanish. Worse for this PR: `logIndex` is a *positional* index into that snapshot, so after an overwrite a surviving record can point at a different suite's dump. `overflow_details.dart:203` only bounds-checks `index >= 0 && index < writtenLogs.length`, which cannot catch an in-range-but-wrong index.
Fix: write per-suite files (`goldens/overflow_warnings/<pid>-<suite>.json`) and merge in the loader, or take a lockfile + temp-and-rename; if the shared file stays, make `logIndex` a content hash rather than a position.

**2. Gallery path never resets the report**

`run_golden_verify.sh:45` (new):
```bash
rm -f goldens/overflow_warnings.json
```
`run_generate_loc_snapshots.sh:36-51` (unchanged) runs `flutter test test/golden_test/ --update-goldens` then `dart run test_scripts/generate_gallery_report.dart "$version"` with no reset — and `.github/workflows/ci.yml:126` invokes exactly this script. So the failure mode the new comment describes ("a previous run's records would be attributed to this one — reporting overflows at line numbers that have since moved") is still live on the gallery path, which is precisely where the new detail UI surfaces. `clear_goldens.sh:18` is a manual step and cannot be relied on.
Fix: add the same `rm -f` before the test run, or extract a single `reset_overflow_report` helper used by both scripts.

**3. Loader can crash the report generators**

`overflow_details.dart:159-161` closes the `try` right after `jsonDecode`, but the casts happen later:
```dart
  } catch (_) {
    return OverflowReport.empty;
  }
...
  final golden = record['golden'] as String?;
...
        line: site.record['line'] as String?,
        pixels: site.record['pixels'] as String?,
```
A syntactically valid file with `"line": 77` throws `TypeError` outside the guard, so `dart run test_scripts/combine_results.dart` / `generate_gallery_report.dart` exit non-zero and `run_golden_verify.sh:59` only sets `FAILED=1` — no HTML report at all. This contradicts `:135-136` ("a diagnostic aside must never take a report generator down").
Fix: use `record['line']?.toString()` style coercion, or extend the `try` over the grouping/mapping and return `empty`.

</details>

<details>

- Escaping was taken seriously: raw log content goes through `textContent` (`generate_gallery_report.dart:687`, `html_generate_functions.dart:765`) rather than `innerHTML`, and the title travels as a `data-` attribute instead of inline JS.
- Hand-concatenated JSON (`{"feature":"${e.feature}"...}`) replaced with `jsonEncode` — that closed a real quoting hole.
- The diagnostics path is fail-soft end to end (`overflow_diagnostics.dart:181-190`): a diagnostic aside can never turn a passing golden red.
- `logs` de-duplication + `logIndex` is a real win on report size, with bounds checking and orphan-log pruning on the read side.
- Sharing one loader collapses the long-standing discrepancy where the two reports disagreed on the overflow count (`combine_results.dart:197` vs `generate_gallery_report.dart:474` now agree), and `overflowCount` correctly switched from raw records to affected goldens.
- `stripEphemeralIds` + `normalizeSourcePath` make records reproducible across runs and CI checkout names (`overflow_diagnostics_test.dart:67-101` even covers a differently-named CI directory), which is what makes dump de-duplication effective.
- Backward compatibility with the legacy list format is handled on both sides and explicitly tested (`overflow_details_test.dart:400-407`).
- Test naming/narrative quality is high: groups map to units under test, each test states why the behaviour is worth protecting, and `overflow_diagnostics_test.dart` pins Flutter's message/dump shape with a real `pumpWidget`.
- Scope discipline is good — no unrelated refactors, no drive-by lint fixes; no `lib/` changes, so the three-layer, generated-model isolation, USP provider and ServiceError rules are untouched.

</details>

AustinChangLinksys (approved):
🤖 **Automated Review — Round 1** · `b8894c40..4472a3db` (full)

**Verdict: ✅ APPROVE** — Clean spec implementation; two warnings worth addressing before merge.

#### Standards

|---|---|---|---|
| ⚠️ | 🟢High | `overflow_diagnostics.dart:25` | `_locationPattern` regex truncates paths containing spaces, amplified by new `normalizeDumpPaths` |
| ⚠️ | 🟡Med | `golden_runner.dart:391–434` | `_writeOverflowReport` is a non-atomic read-modify-write; concurrent `tearDownAll` across parallel suites has a race window (pre-existing, but larger with new log dumps) |
| ⚠️ | 🟢High | `overflow_details.dart:127–130` | [Middle Man] `loadOverflowDetails` is dead code — no production caller, just wraps `loadOverflowReport().byGolden` |
| ⚠️ | 🟡Med | `overflow_diagnostics.dart:124` | `_objectIdPattern` hardcodes `{5}` hex chars (Flutter `shortHash()` private detail) with no doc comment; silent failure if Flutter changes the length |
| 💡 | 🟢High | `overflow_diagnostics.dart:19–21` | `[\d.]+` in `_amountPattern` accepts multiple decimal points; prefer `\d+(?:\.\d+)?` for semantic precision |
| 💡 | 🟢High | `overflow_details.dart:113–120` | `_siteKey` with all-null fields yields identical keys for unrelated records (degenerate edge case) |
| 💡 | 🟡Med | `overflow_details_test.dart:54` | Test group named `loadOverflowDetails` but exercises `loadOverflowReport().byGolden`; rename when removing the dead wrapper |
| 💡 | 🟡Med | `golden_runner.dart:391–436` | Append-to-existing-file logic in `_writeOverflowReport` has no integration-level round-trip test |
| 💡 | ⚪Low | `overflow_diagnostics.dart:80–83` | `lastIndexOf('-')` in `normalizeSourcePath` can leave version number in name for semver pre-release pub-cache paths (rare in practice) |

#### Spec

|---|---|---|---|
| ✅ | 🟢High | `overflow_diagnostics.dart:166–191` | Spec §1 FULFILLED — side/pixels/widget/file/line captured, paths repo-relative, failure-tolerant |
| ✅ | 🟢High | `overflow_details.dart:110–177` | Spec §2 FULFILLED — deduplication on correct 5-field key, occurrences count, shared loader |
| ✅ | 🟢High | `generate_gallery_report.dart:580–598` | Spec §3 FULFILLED — always-visible (no click gate), keyed by golden name |
| ✅ | 🟢High | `combine_results.dart:197` + `html_generate_functions.dart:690` | Spec §4 FULFILLED — attached to test row, distinct-golden count, escapeHtml |
| ✅ | 🟢High | `run_golden_verify.sh:45` | Spec §5 FULFILLED — stale-record clear before test branch |

<details>

**W1 — `_locationPattern` truncates space-containing paths** 🟢High
`overflow_diagnostics.dart:25`
```dart
final RegExp _locationPattern =
    RegExp(r'([A-Za-z_]\w*):file://([^\s:]+):(\d+):(\d+)');
```
`[^\s:]+` stops at the first whitespace, so a developer home directory like `/Users/John Smith/dev/...` will be truncated at the space. This regex is pre-existing (commit 38f70187), but this PR's new `normalizeDumpPaths` (`overflow_diagnostics.dart:131–138`) runs `replaceAllMapped` over the full dump text, meaning every match in the dump gets corrupted partial-replacement text written into the JSON and displayed in the log modal.
**Fix:** Use `Uri.decodeFull` at the `normalizeSourcePath` entry point, or relax the character class to allow `%20`-encoded spaces, e.g. `([^\s:]+(?:%20[^\s:]+)*)`.

---

**W2 — Non-atomic read-modify-write in `_writeOverflowReport`** 🟡Med
`golden_runner.dart:391–434`
```dart
if (file.existsSync()) {
  try {
    final decoded = jsonDecode(file.readAsStringSync()); // read
    // … merge records and logs …
  } catch (_) {
    // A corrupt file must not take the run down; start the report over.
  }
}
// …
file.writeAsStringSync(…);  // write
```
`flutter test` runs suites in parallel (per-CPU-core Isolates by default). Each suite's `tearDownAll` triggers `_writeOverflowReport`, performing a non-atomic read–merge–write. Two suites can interleave: A reads (2 records), B reads (2 records), A writes (2+N), B writes (2+M) — silently discarding A's N records. With this PR's larger payloads (2–4 KB dumps per record), the race window is wider and a torn write is more likely to trigger the `catch (_)` that silently resets the file.
Pre-existing issue, but the surface area is larger now.
**Fix (minimal):** Add `--concurrency=1` to `run_golden_verify.sh` to serialize suites, or switch to an append-only `.ndjson` shard-per-suite format that `loadOverflowReport` merges at generation time.

---

**W3 — [Middle Man] `loadOverflowDetails` is dead code** 🟢High
`overflow_details.dart:127–130`
```dart
Map<String, List<OverflowDetail>> loadOverflowDetails({
  String path = 'goldens/overflow_warnings.json',
}) =>
    loadOverflowReport(path: path).byGolden;
```
No production caller uses this function; both `generate_gallery_report.dart` and `combine_results.dart` call `loadOverflowReport()` directly (they need `logs` too). The only caller is `overflow_details_test.dart:54`, which tests behaviour, not usage. This is a Fowler Middle Man (⑪) + Speculative Generality (⑨) smell.
**Fix:** Delete `loadOverflowDetails`; rename its test group to `'loadOverflowReport – byGolden'`.

---

**W4 — `_objectIdPattern` hardcodes Flutter-internal `shortHash()` length** 🟡Med
`overflow_diagnostics.dart:124`
```dart
final RegExp _objectIdPattern = RegExp(r'(?<=\w)#[0-9a-f]{5}\b');
```
The `{5}` comes from Flutter's private `shortHash()` (`hashCode.toUnsigned(20).toRadixString(16).padLeft(5,'0')`). If Flutter ever changes the hash width, this pattern silently fails: object ids are no longer stripped, the same overflow produces different log text each run, and the deduplication designed around stable log text breaks.
**Fix:** Add a doc comment explaining the `{5}` dependency on Flutter's `shortHash()`. Optionally broaden to `[0-9a-f]{4,8}` and pin a regression test with explicit format note.

</details>

<details>

- **Full spec compliance (issue #1197):** all five required changes implemented correctly and verified against the source.
- **Failure-tolerant extraction:** every parsing step is independently guarded; a parse miss in one field never corrupts others or fails a passing test run.
- **HTML output safety:** Dart side uses `_escapeHtml`/JSON encoding; JS side uses `esc()`, `escapeHtml()`, `escapeAttr()`, and `textContent` (not `innerHTML`) for the raw log modal — no XSS surface.
- **Log deduplication design:** `stripEphemeralIds` + `normalizeDumpPaths` makes the same overflow produce identical log text across runs and goldens — the set-based dedup is well-founded.
- **Indexed log table:** switching from inline to `logIndex`-referenced `logs` table cuts JSON size ~76% (56 KB → 14 KB for a dhcp run); backward-compatible reader for older flat-list format included.
- **Shared loader eliminates report divergence:** both generators previously had their own `_loadOverflowWarnings()` that could disagree; now they share `loadOverflowReport()`, eliminating the inconsistency.
- **Deep modules:** `overflow_diagnostics.dart` and `overflow_details.dart` each hide complex parsing behind a minimal public API; complexity passes the "deletion test".
- **Test coverage:** 52 tests covering parsers with a real Flutter diagnostics dump, not a fixture string — will break loudly if an SDK upgrade changes the format.

</details>

PR #1211, opened 2026-08-07
feat(e2e): stable topology node identifiers for E2E location (#1208)

AustinChangLinksys:
## Summary

Populate a stable, data-derived Semantics `identifier` on every topology `MeshNode` so the E2E suite can locate nodes via `byIdentifier()` instead of the fragile display-label heuristic.

Closes #1208. Depends on UI-kit#9 (`MeshNode.identifier` + node-dispatch seam), landed in `ui_kit_library` **v2.34.2**.

## Changes

- Bump `ui_kit_library` + `generative_ui` ref `v2.31.0` → `v2.34.2`.
- Add `node_identifier.dart`: a pure, unit-tested helper (constitution Article XVI §16.3) computing the shortest MAC suffix that stays unique within each node group; it grows group-wide (`4 → 6 → full`) on collision and is order-independent.
- `UspTopologyBuilder` sets `identifier` at all three `MeshNode` construction sites:
  | Node | Identifier | Source |
  |------|-----------|--------|
  | master | `topology-node-master` | fixed, single per topology |
  | slave | `topology-node-slave-<suffix>` | `deviceId` |
  | client | `topology-node-client-<suffix>` | `mac` |
- Additive only: the human label / name / a11y text are unchanged.

## Why the ui_kit v2.34.2 dependency

From v2.34.2, identifier ownership sits on the **node-dispatch seam** rather than per-renderer, so every node type carries its identifier in both graph and tree view regardless of which renderer the registry dispatches — including tree-view clients, which the `.unified` registry otherwise renders without a dedicated tree builder.

## Identifier scheme (Article XVI §16.3)

Per-instance keys are **data-derived, not positional** — the MAC suffix means an identifier stays stable across node reordering and signal-quality changes, satisfying the issue's "stable across node order" requirement. The derivation helper is pure and unit-tested.

## Tests

- `node_identifier_test`: normalization, last-4 default, collision extension, uniqueness, order-independence, genuine-duplicate fallback.
- `usp_topology_builder_test`: identifier value, cross-graph uniqueness, label decoupling, stability across signal quality.
- `topology_node_identifier_widget_test`: `bySemanticsIdentifier` locates master, every slave and every client in **both** graph and tree view.

## Verification

- `dart format --set-exit-if-changed` — clean
- `flutter analyze` — 0 error / 0 warning on changed files
- `./run_tests.sh` — 3491 tests passing
- `flutter test --tags ui` (widget test) — passing

### DIFF
--- a/lib/page/topology/helpers/node_identifier.dart
+++ b/lib/page/topology/helpers/node_identifier.dart
@@ -1,0 +1,73 @@
+/// Pure helpers that build stable, data-derived E2E `identifier` values for
+/// topology nodes (constitution Article XVI §16.3).
+///
+/// Node identifiers are consumed by the E2E suite via `byIdentifier()` against
+/// the CanvasKit Semantics tree, so they MUST be stable across
+/// quality% / renderer / view-mode / node order and independent of the human
+/// label. Per-instance keys are derived from the node's MAC (never a row
+/// index), reduced to the shortest suffix that stays unique within the graph.
+library;
+
+/// Identifier for the master / gateway node. There is exactly one per topology,
+/// so it carries no per-instance key.
+const String kTopologyMasterIdentifier = 'topology-node-master';
+
+/// Composes the slave / extender node identifier from a MAC suffix key.
+String topologySlaveIdentifier(String macSuffix) =>
+    'topology-node-slave-$macSuffix';
+
+/// Composes the client node identifier from a MAC suffix key.
+String topologyClientIdentifier(String macSuffix) =>
+    'topology-node-client-$macSuffix';
+
+/// Normalizes a MAC / device id to uppercase hex with every separator removed.
+///
+/// `aa:bb:cc:dd:ee:ff` → `AABBCCDDEEFF`. Any non-hex character (`:`, `-`, `.`,
+/// whitespace, …) is stripped, so callers may pass raw `deviceId` / `mac`
+/// values without pre-cleaning them.
+String normalizeMac(String raw) =>
+    raw.toUpperCase().replaceAll(RegExp('[^0-9A-F]'), '');
+
+/// Maps each input MAC / id to the shortest hex suffix that is unique within
+/// the group, keeping the derived identifier short yet collision-free.
+///
+/// - Inputs are normalized via [normalizeMac] (separators stripped, uppercased).
+/// - A single group-wide suffix length is used, starting at [minLength]
+///   (default 4) and growing in steps of [step] (default 2, MAC-octet aligned)
+///   until every suffix in the group is distinct. One shared length keeps the
+///   result deterministic and independent of input order.
+/// - Genuine duplicates (two inputs that normalize identically) cannot be
+///   disambiguated; both fall back to the full normalized value.
+///
+/// The returned map is keyed by the ORIGINAL input string (exactly as passed),
+/// so callers can look up by `slave.deviceId` / `client.mac` directly. Empty
+/// and duplicate inputs are tolerated.
+Map<String, String> shortestUniqueMacSuffixes(
+  Iterable<String> macs, {
+  int minLength = 4,
+  int step = 2,
+}) {
+  final originals = macs.toList();
+  if (originals.isEmpty) return const {};
+
+  final normalized = {for (final m in originals) m: normalizeMac(m)};
+  final maxLen =
+      normalized.values.fold<int>(0, (a, s) => s.length > a ? s.length : a);
+
+  String suffixOf(String s, int len) =>
+      s.length <= len ? s : s.substring(s.length - len);
+
+  // Grow a single group-wide length until all suffixes are distinct.
+  for (var len = minLength; len < maxLen; len += step) {
+    final suffixes = normalized.values.map((s) => suffixOf(s, len)).toList();
+    if (suffixes.toSet().length == suffixes.length) {
+      return {
+        for (final entry in normalized.entries)
+          entry.key: suffixOf(entry.value, len),
+      };
+    }
+  }
+
+  // Full normalized value: covers the capped length and genuine-duplicate case.
+  return normalized;
+}
--- a/lib/page/topology/helpers/usp_topology_builder.dart
+++ b/lib/page/topology/helpers/usp_topology_builder.dart
@@ -7,6 +7,7 @@ import 'package:privacy_gui/page/_shared/models/client_device.dart'
     hide ConnectionType;
 import 'package:privacy_gui/page/_shared/models/mesh_network.dart';
 import 'package:privacy_gui/page/_shared/models/system_info_ui_model.dart';
+import 'package:privacy_gui/page/topology/helpers/node_identifier.dart';
 import 'package:ui_kit_library/ui_kit.dart';
 
 /// Builds a [MeshTopology] from USP dashboard state for [AppTopology] widget.
@@ -35,6 +36,7 @@ class UspTopologyBuilder {
     );
     nodes.add(MeshNode(
       id: gatewayId,
+      identifier: kTopologyMasterIdentifier,
       name:
           master.displayName.isNotEmpty ? master.displayName : info.gatewayName,
       type: MeshNodeType.gateway,
@@ -95,6 +97,12 @@ class UspTopologyBuilder {
           'backhaulParentDeviceId: ${slave.backhaul.parentNodeId}');
     }
 
+    // Stable, data-derived E2E identifier keys (Article XVI §16.3): shortest
+    // MAC suffix that stays unique within each node group, independent of the
+    // display label / node order.
+    final slaveIdKeys =
+        shortestUniqueMacSuffixes(meshNetwork.slaves.map((s) => s.deviceId));
+
     // Slave nodes
     for (final slave in meshNetwork.slaves) {
       final extenderId = 'extender-${slave.deviceId}';
@@ -111,6 +119,7 @@ class UspTopologyBuilder {
       final extenderIconName = routerIconTestByModel(modelNumber: slave.model);
       nodes.add(MeshNode(
         id: extenderId,
+        identifier: topologySlaveIdentifier(slaveIdKeys[slave.deviceId] ?? ''),
         name: slave.displayName,
         type: MeshNodeType.extender,
         status: MeshNodeStatus.online,
@@ -148,6 +157,8 @@ class UspTopologyBuilder {
     }
 
     // Client devices — use allClients which includes master + slave clients
+    final clientIdKeys =
+        shortestUniqueMacSuffixes(meshNetwork.allClients.map((c) => c.mac));
     for (final client in meshNetwork.allClients) {
       final clientId = 'client-${client.mac}';
       final isEthernet = !client.isWifi;
@@ -178,6 +189,7 @@ class UspTopologyBuilder {
 
       nodes.add(MeshNode(
         id: clientId,
+        identifier: topologyClientIdentifier(clientIdKeys[client.mac] ?? ''),
         name: client.displayName,
         type: MeshNodeType.client,
         status:
--- a/pubspec.yaml
+++ b/pubspec.yaml
@@ -59,11 +59,11 @@ dependencies:
   ui_kit_library:
     git:
       url: https://github.com/linksys/privacyGUI-UI-kit.git
-      ref: v2.31.0
+      ref: v2.34.2
   generative_ui:
     git:
       url: https://github.com/linksys/privacyGUI-UI-kit.git
-      ref: v2.31.0
+      ref: v2.34.2
       path: generative_ui
   flutter_blue_plus: ^1.4.0
   crypto: ^3.0.2
--- a/test/page/topology/helpers/node_identifier_test.dart
+++ b/test/page/topology/helpers/node_identifier_test.dart
@@ -1,0 +1,110 @@
+import 'package:flutter_test/flutter_test.dart';
+import 'package:privacy_gui/page/topology/helpers/node_identifier.dart';
+
+void main() {
+  group('normalizeMac', () {
+    test('strips colons and uppercases', () {
+      expect(normalizeMac('aa:bb:cc:dd:ee:ff'), 'AABBCCDDEEFF');
+    });
+
+    test('strips dashes, dots and whitespace', () {
+      expect(normalizeMac('aa-bb-cc-dd-ee-ff'), 'AABBCCDDEEFF');
+      expect(normalizeMac('aabb.ccdd.eeff'), 'AABBCCDDEEFF');
+      expect(normalizeMac(' AA BB CC DD EE FF '), 'AABBCCDDEEFF');
+    });
+
+    test('drops any non-hex character', () {
+      expect(normalizeMac('AA:BB:CC:DD:EE:ZZ'), 'AABBCCDDEE');
+    });
+
+    test('empty input yields empty', () {
+      expect(normalizeMac(''), '');
+    });
+  });
+
+  group('identifier composers', () {
+    test('master identifier is a fixed, key-less string', () {
+      expect(kTopologyMasterIdentifier, 'topology-node-master');
+    });
+
+    test('slave / client composers prepend the correct prefix', () {
+      expect(topologySlaveIdentifier('EEFF'), 'topology-node-slave-EEFF');
+      expect(topologyClientIdentifier('9C1D'), 'topology-node-client-9C1D');
+    });
+  });
+
+  group('shortestUniqueMacSuffixes', () {
+    test('empty input yields empty map', () {
+      expect(shortestUniqueMacSuffixes(const []), isEmpty);
+    });
+
+    test('uses the last 4 hex chars when that is already unique', () {
+      final result = shortestUniqueMacSuffixes([
+        'AA:BB:CC:DD:E1:11',
+        'AA:BB:CC:DD:E2:22',
+      ]);
+      expect(result['AA:BB:CC:DD:E1:11'], 'E111');
+      expect(result['AA:BB:CC:DD:E2:22'], 'E222');
+    });
+
+    test('map is keyed by the original (un-normalized) input', () {
+      const input = 'aa:bb:cc:dd:ee:ff';
+      final result = shortestUniqueMacSuffixes([input]);
+      expect(result.containsKey(input), isTrue);
+      expect(result[input], 'EEFF');
+    });
+
+    test('extends suffix length group-wide on last-4 collision', () {
+      // Both share suffix "EEFF" at length 4 → must grow to length 6.
+      final result = shortestUniqueMacSuffixes([
+        'AA:BB:CC:11:EE:FF',
+        'AA:BB:CC:22:EE:FF',
+      ]);
+      expect(result['AA:BB:CC:11:EE:FF'], '11EEFF');
+      expect(result['AA:BB:CC:22:EE:FF'], '22EEFF');
+      // Single shared length: every suffix has the same length.
+      final lengths = result.values.map((s) => s.length).toSet();
+      expect(lengths, {6});
+    });
+
+    test('all resulting suffixes are unique for distinct inputs', () {
+      final macs = [
+        'AA:BB:CC:DD:EE:01',
+        'AA:BB:CC:DD:EE:02',
+        'AA:BB:CC:DD:EE:03',
+        '11:22:33:44:55:66',
+      ];
+      final result = shortestUniqueMacSuffixes(macs);
+      expect(result.length, macs.length);
+      expect(result.values.toSet().length, macs.length);
+    });
+
+    test('genuine duplicates collapse to the full normalized value', () {
+      final result = shortestUniqueMacSuffixes([
+        'AA:BB:CC:DD:EE:FF',
+        'aa-bb-cc-dd-ee-ff', // same MAC, different formatting
+      ]);
+      // Cannot be disambiguated → both fall back to full normalized value.
+      expect(result['AA:BB:CC:DD:EE:FF'], 'AABBCCDDEEFF');
+      expect(result['aa-bb-cc-dd-ee-ff'], 'AABBCCDDEEFF');
+    });
+
+    test('result is independent of input order', () {
+      final a = shortestUniqueMacSuffixes([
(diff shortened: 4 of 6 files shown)


AustinChangLinksys (commented):
🤖 **Automated Review — Round 1** · `1c048d2..1880ce6` (full)

**Verdict: 💬 Comment** — Self-review (comment only); review is for reference only, human sign-off required.

#### Standards

|---|---|---|---|
| ⚠️ | High | `node_identifier.dart:47-49` + `usp_topology_builder.dart:122,192` | `?? ''` fallback at both slave/client lookup sites silently produces a suffix-less identifier (`topology-node-slave-`) if key is ever missing; both reviewers flagged independently. |
| ⚠️ | Med | `node_identifier.dart:48-49` | Same-string duplicate `deviceId` inputs cause Map key collapse → single-element list → loop exits at `minLength` → both nodes share the same 4-char suffix (behavior diverges from docstring "fall back to full normalized value"). |
| 💡 | High | `usp_topology_builder.dart:78,104` | Two MAC normalization patterns coexist without explanatory comments; future reader may mistakenly "unify" them and break routing logic. |
| 💡 | Med | `node_identifier.dart:44-48` | `minLength`/`step` named params have zero callers passing custom values; appears speculatively generic — if changing them would break E2E they shouldn't be configurable. |
| 💡 | High | `test/.../usp_topology_builder_test.dart` | No explicit test asserting that cross-type prefix difference (slave vs client) is sufficient to prevent cross-group identifier collision. |

*[X] = 1 agent · [XX] = 2 agents flagged independently, higher confidence (Critical only).*

#### Spec

|---|---|---|---|
| ⚠️ | High | `usp_topology_builder.dart:122,192` | Slave/client identifiers use shortest MAC suffix (e.g. `topology-node-slave-EE01`) not full `deviceId`; spec explicitly left scheme "open to review" but E2E repo (PrivacyGUI-USP-E2E) may still expect the original full-deviceId form — cross-repo alignment needed before merge. |

*[X] = 1 agent · [XX] = 2 agents flagged independently, higher confidence (Critical only).*

<details>

**W-1: `?? ''` silent fallback producing suffix-less identifier** (Confidence: High)
`usp_topology_builder.dart:122` and `usp_topology_builder.dart:192`
```dart
identifier: topologySlaveIdentifier(slaveIdKeys[slave.deviceId] ?? ''),
// => 'topology-node-slave-' (no suffix) if key is somehow missing

identifier: topologyClientIdentifier(clientIdKeys[client.mac] ?? ''),
// => 'topology-node-client-' (no suffix)
```
In the current code path the keys always exist (map is built from the same collection). But `?? ''` is a silent degradation — if any future refactoring decouples the two iterations, multiple nodes will share the same degenerate identifier and E2E locators will fail non-obviously. Flagged by both Reviewer A and Reviewer B.
Fix: Replace `?? ''` with `!` (or add `assert(slaveIdKeys.containsKey(slave.deviceId))` before the lookup) so the bug surfaces immediately in debug/test rather than silently in production.

**W-2: Same-string duplicate `deviceId` causes Map key collapse** (Confidence: Med)
`node_identifier.dart:48-49`
```dart
final normalized = {for (final m in originals) m: normalizeMac(m)};
```
If `originals = ['AA:BB:CC:DD:EE:FF', 'AA:BB:CC:DD:EE:FF']` (identical strings), Dart Map keeps only one key. The loop then sees a 1-element list; set.length == list.length at `minLength=4`, so it returns 4-char suffix for the single key. Both callers looking up that key get the same suffix. The doc comment says "both fall back to the full normalized value" but the actual behavior returns `minLength` suffix — docstring is inaccurate. Trigger: two slaves with identical raw `deviceId` strings (hardware anomaly or test fixture).
Fix: Either add a test case for this edge, or clarify the docstring, or handle with a grouping approach that tolerates duplicate keys.

**C-1 (Spec): Slave/client identifier format not confirmed with E2E repo** (Confidence: High)
`usp_topology_builder.dart:122,192`
Issue #1208 suggested `topology-node-slave-<deviceId>` (full deviceId) but left the final scheme "open to review." This PR uses a MAC suffix instead (e.g. `topology-node-slave-EE01`). The implementation logic is sound and satisfies all stated stability requirements. However, PrivacyGUI-USP-E2E#26 (the downstream consumer) was likely written against the suggested full-deviceId format.
Fix (pre-merge action): Verify PrivacyGUI-USP-E2E#26 uses the MAC-suffix format or update it accordingly. This is a coordination check, not a code bug.

</details>

<details>
<summary>Suggestion Details</summary>

**S-1: Two MAC normalization paths lack isolation comments** (Confidence: High)
`usp_topology_builder.dart:78` (routing, colon-only) vs `node_identifier.dart:31` (identifier, all non-hex stripped). Both look like "normalize a MAC" but serve different semantics. A one-line inline comment at each call site prevents accidental unification.

**S-2: `minLength`/`step` named params appear speculatively generic** (Confidence: Med)
`node_identifier.dart:44-48`: no caller passes non-default values; no test exercises custom values. If changing these values would alter identifier strings (breaking E2E), they shouldn't be exposed as configuration knobs. Consider inlining as constants or at minimum adding a test that passes non-default values.

**S-3: Missing cross-type collision invariant test** (Confidence: High)
`usp_topology_builder_test.dart`: existing uniqueness test uses fixtures where slave and client MACs happen not to collide in their last 4 hex digits. Add an explicit test case where a slave MAC and a client MAC share the same last 4 hex digits — and assert the identifiers are still unique (they are, due to different prefixes) — to document and protect this invariant.

</details>

<details>

- Architecture is clean: `node_identifier.dart` is a pure, deep module — small API surface, hidden collision-resolution algorithm, fully unit-tested.
- Three MeshNode construction sites all covered: master (line 39), slave (line 122), client (line 192) — no site missed.
- Additive: `label`/`name`/a11y fields are completely untouched; this is a strictly additive change.
- Article XVI §16.3 compliance: kebab-case identifiers, MAC-derived (not positional), stable across quality%/renderer/view-mode/node order.
- Test coverage is thorough: unit tests cover normalization, collision extension, order-independence, genuine-duplicate fallback, short-MAC edge cases; widget tests validate `bySemanticsIdentifier` locatability in both graph and tree view modes.
- Dependency bump scoped correctly: `ui_kit_library` + `generative_ui` (same git repo, different path) bumped together to v2.34.2 — no half-bump.
- No security issues: MAC suffix in Semantics DOM is not a new exposure (full MAC already in `metadata`); regex `[^0-9A-F]` is linear, no ReDoS.

</details>

lib/page/topology/helpers/node_identifier.dart
```diff
@@ -0,0 +1,73 @@
...
+  // Grow a single group-wide length until all suffixes are distinct.
+  for (var len = minLength; len < maxLen; len += step) {
+    final suffixes = normalized.values.map((s) => suffixOf(s, len)).toList();
+    if (suffixes.toSet().length == suffixes.length) {
+      return {
+        for (final entry in normalized.entries)
+          entry.key: suffixOf(entry.value, len),
+      };
+    }
+  }
```
PeterJhongLinksys, line 69: ⚠️ **Warning [XX] 🟢High — group-wide suffix length makes identifiers unstable.** The chosen length is shared by the whole group, so a node’s identifier changes when *other* nodes join or leave. Two extenders `AA:BB:CC:11:EE:FF` / `AA:BB:CC:22:EE:FF` collide at 4 and both become `topology-node-slave-11EEFF`; once one drops out of `meshNetwork.slaves` the survivor shrinks back to `topology-node-slave-EEFF` — same physical node, different identifier. This contradicts the "MUST be stable" contract at lines 4-8 and is worse for `allClients`, which changes every poll. Suggest using a length that depends only on the node itself (full `normalizeMac()`, or a fixed last-6): identifiers are machine-read, so length costs nothing, and the whole loop plus the two unused tuning params disappear.

lib/page/topology/helpers/usp_topology_builder.dart
```diff
@@ -178,6 +189,7 @@ class UspTopologyBuilder {
 
       nodes.add(MeshNode(
         id: clientId,
+        identifier: topologyClientIdentifier(clientIdKeys[client.mac] ?? ''),
```
PeterJhongLinksys, line 192: ⚠️ **Warning [XX] 🟢High — empty/non-hex MAC produces duplicate identifiers, and `?? ...` is not a real guard.** `normalizeMac` maps a blank or non-hex value to an empty string, and the fallback in `node_identifier.dart:71-72` returns it unchanged, so the identifier becomes `topology-node-client-` (no key). Two such devices get *identical* identifiers, and the ui_kit v2.34.2 seam only skips `null || isEmpty` — this value is non-empty, so it is attached anyway. Empty MACs are reachable: `lib/generated/connected_devices.g.dart:229` does `(response[...PhysAddress] ?? "") as String` with no non-empty check, and `lib/page/_shared/utils/mesh_network_builder.dart:246` only trims/uppercases (compare `usp_wifi_data_service.dart:288`, which does `if (mac.isEmpty) continue;`). Also note the `?? ...` can never fire — the map is keyed by the very strings just passed in — so it only makes the empty-key case *look* handled. Suggest a stable non-colliding fallback plus unit tests for `[""]`/`["",""]`.

PeterJhongLinksys (approved):
🤖 **Automated Review — Round 1** · `1c048d2..1880ce6` (full)

**Verdict: ✅ APPROVE** — Correct approach and unusually strong tests for an E2E-testability change; no Critical issues. Two design points (group-wide suffix length, empty-MAC fallback) are worth a follow-up before the E2E suite hardens around these identifiers.

|---|---|---|---|
| ⚠️ | 🟢 | `lib/page/topology/helpers/node_identifier.dart:61-69` | [XX] Suffix length is group-wide, so a node's identifier changes when *other* nodes join/leave — contradicts the file's own "MUST be stable" contract. |
| ⚠️ | 🟢 | `lib/page/topology/helpers/usp_topology_builder.dart:122,192` | [XX] Empty / non-hex MAC normalizes to `''` → duplicate `topology-node-client-` identifiers; `?? ''` is dead code that only *looks* like a guard. |
| ⚠️ | 🟢 | `lib/page/topology/helpers/node_identifier.dart:16-21,28-29` | Identifier values contain uppercase hex (`...-EE01`); constitution §16.3 mandates kebab-case and existing `ruleIdentifierKey` is all-lowercase. |
| ⚠️ | 🟢 | `lib/page/topology/helpers/node_identifier.dart:28-29` | [Duplicated Code] Third MAC-normalization implementation in the repo; 4 older `.replaceAll(':', '')` sites remain in the *same* modified file. |
| ⚠️ | 🟡 | `pubspec.yaml:62,66` | [XX] ui_kit_library v2.31.0 → v2.34.2 (729 files / 6599 lines upstream) with no golden/regression evidence and `pubspec.lock` gitignored. |
| ⚠️ | 🟢 | `test/page/topology/helpers/topology_node_identifier_widget_test.dart:314-348` | Widget test couples to ui_kit internals (`NodeRendererRegistry.unified`, `preferAnimationNode`) and waits a fixed `pump(500ms)` with animation on → flaky / breaks on upgrade. |
| 💡 | ⚪ | `lib/page/topology/helpers/node_identifier.dart:16-21` | E2E generator scans for literal values next to `identifier:`; prefixes live only inside helper bodies, so they may not land in `identifiers.generated.ts` (§16.4). Please confirm. |
| 💡 | 🟢 | `lib/page/topology/helpers/node_identifier.dart:47-48` | [Speculative Generality] `minLength` / `step` params have no non-default caller. |
| 💡 | 🟢 | `lib/page/topology/helpers/node_identifier.dart:35-40` | DartDoc promises "a single group-wide length", but the fallback path returns per-input full lengths. |
| 💡 | 🟢 | `test/page/topology/helpers/node_identifier_test.dart` | Missing negative cases: `['', '']`, `['N/A','???']`, and a "group membership changed → existing identifier unchanged" stability test. |
| 💡 | 🟢 | `test/page/topology/helpers/topology_node_identifier_widget_test.dart:352,386` | Hard-coded `'topology-node-master'` / `'EE01'` / `'5501'` instead of deriving from `kTopologyMasterIdentifier` and `DevicesTestData` constants. |
| 💡 | 🟢 | `lib/ai/components/sections/topology_section.dart:160,181,240` | Other `MeshNode` construction path (AI topology section) carries no `identifier`, so `byIdentifier()` cannot locate nodes there. |
| 💡 | 🟢 | `lib/page/topology/helpers/node_identifier.dart:28-29` | `RegExp` recompiled on every `normalizeMac` call; hoist to a top-level `final`. |
| 💡 | 🟡 | `lib/page/topology/helpers/usp_topology_builder.dart:122,192` | MAC suffixes reach the production DOM semantics tree unconditionally (no debug/E2E gate). Low risk — same page already shows full MACs — but worth a conscious decision. |

<sub>Confidence: 🟢 High = read head code, evidence quoted · 🟡 Med = file:line + reasoning, not fully proven · ⚪ Low = speculative.</sub>
<sub>`[X]` = raised by 1 agent · `[XX]` = independently raised by both agents.</sub>

<details>

### W1 — Group-wide suffix length makes a node's identifier depend on the rest of the group `[XX]` 🟢

`lib/page/topology/helpers/node_identifier.dart:61-69`
```dart
// Grow a single group-wide length until all suffixes are distinct.
for (var len = minLength; len < maxLen; len += step) {
  final suffixes = normalized.values.map((s) => suffixOf(s, len)).toList();
  if (suffixes.toSet().length == suffixes.length) {
    return {
      for (final entry in normalized.entries)
        entry.key: suffixOf(entry.value, len),
    };
  }
}
```
The chosen length is shared by the whole group. Given extenders `AA:BB:CC:11:EE:FF` and `AA:BB:CC:22:EE:FF` (last-4 collision) both get length 6 → `topology-node-slave-11EEFF`. When one goes offline and drops out of `meshNetwork.slaves`, the survivor's suffix shrinks back to 4 → `topology-node-slave-EEFF`. **The same physical node's identifier changed although nothing about that node changed.**

This directly contradicts the file's own contract at `node_identifier.dart:4-8`:
```dart
/// Node identifiers are consumed by the E2E suite via `byIdentifier()` against
/// the CanvasKit Semantics tree, so they MUST be stable across
/// quality% / renderer / view-mode / node order and independent of the human label.
```
The client group is more exposed than slaves: `meshNetwork.allClients` changes on every poll (roaming, offline clients).

**Trigger:** two MACs in the same group sharing the last 4 hex digits (common for same-model batches), plus any group membership change during a session.

**Suggested fix:** drop the "shortest" optimization and use a length that depends only on the node itself — the full `normalizeMac()` value (12 chars) or a fixed last-6. Identifiers are machine-read, so length has no cost, and this removes the loop entirely (also resolving the two unused tuning params). If the dynamic length is kept intentionally, document the coupling and add a test pinning "adding a colliding node does not change existing identifiers".

### W2 — Empty / non-hex MAC yields duplicate, key-less identifiers; `?? ''` is not a real guard `[XX]` 🟢

`lib/page/topology/helpers/usp_topology_builder.dart:122,192`
```dart
identifier: topologySlaveIdentifier(slaveIdKeys[slave.deviceId] ?? ''),
identifier: topologyClientIdentifier(clientIdKeys[client.mac] ?? ''),
```
`node_identifier.dart:28-29` maps a blank or non-hex input to `''`, and the fallback at `:71-72` returns `normalized` unchanged, so an empty MAC survives as an empty key → `topology-node-client-`. Two such devices produce **identical** identifiers.

The upstream seam only skips `null || isEmpty`, and `'topology-node-client-'` is non-empty, so it is attached anyway (ui_kit v2.34.2 `node_identifier_seam.dart`):
```dart
final identifier = node.identifier;
if (identifier == null || identifier.isEmpty) return child;
return Semantics(identifier: identifier, child: child);
```

The data flow admits empty MACs — `lib/generated/connected_devices.g.dart:229` has no non-empty validation:
```dart
macAddress: (response['${p}PhysAddress'] ?? '') as String,
```
and `lib/page/_shared/utils/mesh_network_builder.dart:246` only trims/uppercases:
```dart
final mac = device.macAddress.trim().toUpperCase();
```
For contrast, `lib/page/wifi_settings/services/usp_wifi_data_service.dart:288` does guard with `if (mac.isEmpty) continue;` — the repo already treats empty MAC as a real possibility.

Separately, note that `?? ''` can never actually fire: the map is keyed by the original strings that were just passed in, so the lookup never misses. Its only effect is to make the empty-key case *look* handled.

**Trigger:** USP `Hosts.Host.*.PhysAddress` returns `''`/`null` or a non-hex placeholder for ≥2 devices that pass the inclusion filter in `mesh_network_builder.dart:41-46`.

**Suggested fix:** give the empty case a stable, non-colliding fallback (e.g. derive from `serialNumber`, or emit `topology-node-client-unknown-<stable key>`), make the intent explicit with an `assert`/`logger.w` instead of `?? ''`, and add unit tests for `['', '']` and for the genuine-duplicate path.

### W3 — Identifier values use uppercase hex, against §16.3 kebab-case and existing repo convention `[X]` 🟢

`node_identifier.dart:16-21` composes `topology-node-slave-$macSuffix` where `macSuffix` comes from `normalizeMac`'s `toUpperCase()` (`:28-29`), and `node_identifier_test.dart:178-179` pins the uppercase form. `constitution.md` §16.3 states identifier values MUST be kebab-case, and §16.5's checklist repeats it; the existing `lib/page/_shared/models/port_forwarding_rule_ui_model.dart:11-15` `ruleIdentifierKey` is a lowercase slug. Mixing `pf-edit-web-server` with `topology-node-slave-EE01` in one selector map means E2E authors must remember which feature uses which casing, and case-sensitive mismatches fail opaquely.

**Suggested fix:** keep uppercase internally for comparison but lowercase the emitted identifier (`topology-node-slave-ee01`), or add an explicit hex exception to §16.3.

### W4 — Third MAC-normalization implementation; 4 older sites left untouched in the same file `[X]` 🟢

`node_identifier.dart:28-29` duplicates `lib/core/utils/oui_lookup.dart:83,93` verbatim (`mac.toUpperCase().replaceAll(RegExp(r'[^0-9A-F]'), '')`), and collides by name with `lib/page/instant_privacy/services/instant_privacy_service.dart:318` `normalizeMac`. Meanwhile the file this PR modifies still hand-rolls a weaker variant at `usp_topology_builder.dart:80,87,115,170` (`.toUpperCase().replaceAll(':', '')`), so **one file now holds two different definitions of "normalized MAC"**. If DataElements ever returns `AA-BB-...`, the parent matching at lines 115/170 would fail while identifiers stayed fine — a hard bug to locate.

**Suggested fix:** at minimum switch those 4 in-file sites to `normalizeMac` in this PR; ideally hoist a single MAC utility into `lib/core/utils/` and converge `oui_lookup.dart` + `instant_privacy_service.dart` in a follow-up.

### W5 — ui_kit_library v2.31.0 → v2.34.2 with no regression evidence `[XX]` 🟡→🟢 (version choice verified)

`pubspec.yaml:62,66`. The version choice itself is **correct and minimal** — checking upstream tags, `MeshNode.identifier` first appears in v2.34.1 but `node_identifier_seam.dart` only exists in v2.34.2, so v2.34.2 is the earliest viable ref (there is no v2.33.x). The concern is the payload: `git diff --stat v2.31.0 v2.34.2` is 729 files / 6599 insertions, including topology internals (`link_renderer.dart`, `orbit_node.dart`, `topology_tree_view.dart`) and new design themes. This PR's tests cover only topology identifiers, golden/screenshot coverage is deferred, and `.gitignore:76` excludes `/pubspec.lock`, so the dependency is pinned only by a movable git tag.

**Suggested fix:** state the "v2.34.2 is minimal" rationale in the PR description with the tag evidence, list the manual regression scope covered (dashboard / topology / port-forwarding at least), confirm `flutter analyze` is clean on the new kit, and consider pinning by commit SHA.

### W6 — Widget test couples to ui_kit internals and to a magic animation delay `[X]` 🟢

`test/page/topology/helpers/topology_node_identifier_widget_test.dart:314-331` hand-copies the production `AppTopology` configuration (`nodeRendererRegistry: NodeRendererRegistry.unified`, `preferAnimationNode: true`, `enableAnimation: true`), and `:347-348` waits `await tester.pump(const Duration(milliseconds: 500))` before asserting. Two problems: (a) it asserts third-party projection behaviour, so a rename or default change in ui_kit turns it red for reasons unrelated to our identifiers, and the copied config silently drifts from `usp_topology_view.dart`; (b) a fixed 500ms wait with animation enabled is timing-fragile and fails as `findsNothing`, which hides the real cause. The identifier contract does not depend on animation.

**Suggested fix:** `enableAnimation: false` + `pumpAndSettle()`, share one topology-config builder between the production view and the test, and keep the real contract assertions in `usp_topology_builder_test.dart` (which correctly does not depend on rendering). The second test case (`reads back verbatim`) is mode-independent and can move out of the graph/tree loop.

</details>

<details>

- **Right architectural shape for the requirement.** The derivation is a pure, side-effect-free, fully unit-tested function, exactly as constitution §16.3 mandates ("MUST be a pure, unit-tested function"). No I/O, no state.
- **Identity source chosen correctly** — MAC, not row index and not the display label.
- **Tests assert the intent, not the implementation.** `usp_topology_builder_test.dart:453-475` (two identically-named nodes still get distinct identifiers) and `:477-506` (identifier unchanged across signal quality) target the actual §16.3 goal and the real source of E2E flakiness.
- **Clean three-layer test split**: pure algorithm → builder contract → Semantics-tree smoke test, with a comment honestly explaining why no web round is needed.
- **Map keyed by the original string** (`node_identifier.dart:53`) lets callers pass `slave.deviceId` / `client.mac` directly, avoiding re-normalization drift at call sites, and the contract is documented at `:42-44`.
- **Prefix partitioning** (master/slave/client) makes cross-group collisions structurally impossible; uniqueness only needs to hold within a group.
- **Purely additive change** — `name` and a11y text untouched, so screen-reader behaviour is unaffected, matching §16.1's split between silent automation hooks and human labels.
- **All test data goes through `DevicesTestData` builders** (no ad-hoc inline fixtures), test names describe behaviour, and all hard-coded suffixes were verified to match the fixture constants.
- **Good docs and commit message**, including an accurate cross-repo note that identifier ownership sits on the ui_kit v2.34.2 node-dispatch seam.
- **Layering checks are clean / not applicable**: no reverse dependencies, no `generated/` imports outside `services/`, no Provider or Service changes (so the USP Provider four hard rules and ServiceError isolation do not apply here), no hand-rolled UI components.

</details>

PR #1214, opened 2026-08-09
feat(e2e): add stable identifiers to Static Routing + DMZ controls (#1172 Group 3)

AustinChangLinksys:
## Summary

Adds E2E `identifier` hooks to the write-proof controls the coverage backlog (PrivacyGUI-USP-E2E#16) needs for **P12 Static Routing** and **P13 DMZ** — the third batch of #1172. Both pages previously exposed no stable per-control anchor for these targets, forcing positional / localized-text selectors.

Addresses the Group 3 items in #1172.

### Static Routing (P12)
- `static-route-add` / `-edit-${key}` / `-delete-${key}` on the row action buttons
- `static-route-submit` / `-cancel` on the dialog actions — native `FilledButton`/`TextButton` swapped to `AppButton` (which forwards `identifier`), matching the shipped `pf-single-submit` precedent and Article XIV (UI-Kit-First)
- `StaticRouteUIModel.identifierKey` reuses the existing `ruleIdentifierKey` helper so edit/delete carry a **stable per-rule key** rather than a row index (Article XVI — row-index selectors are a violation)

### DMZ (P13 — previously zero identifiers)
- `dmz-enabled`, `dmz-dest-ip`, `dmz-source-any` / `-cidr`, `dmz-source-cidr-ip`
- DMZ Save needs no new hook — it uses the page-level bottom bar (`FRAMEWORK_IDS.pageSave`)

No behavior or widget-signature change beyond the button-widget swap.

## Two deviations from the issue's literal spec (intentional)
1. **edit/delete use a stable key, not a bare `static-route-edit`/`-delete` string** — those are per-row buttons; a bare string collides across rows (the exact positional problem this issue exists to eliminate). Follows the PF `pf-edit-${ruleKey}` pattern.
2. **submit button converted to `AppButton`** — native Material buttons do not forward `identifier`; putting the hook on the button body (not an outer `Semantics` wrapper) avoids the outer/inner binding fragility flagged in the issue's Group 2 note.

## Testing
- `dart format` / `flutter analyze` → clean
- Affected functional tests pass (static_routing + dmz notifier/service, 72 tests)
- Golden tests render all widgets without throwing (project does not commit golden baselines, so pixel diffs are local-only noise unrelated to these semantics-only changes)

## Follow-up (E2E repo, not this repo)
Once merged, run `gen:ids` on PrivacyGUI-USP-E2E to regenerate the selector map, unskip the P12 add write-proof, and start the P13 DMZ spec.

### DIFF
--- a/lib/page/static_routing/models/static_routing_ui_model.dart
+++ b/lib/page/static_routing/models/static_routing_ui_model.dart
@@ -1,4 +1,6 @@
 import 'package:equatable/equatable.dart';
+import 'package:privacy_gui/page/_shared/models/port_forwarding_rule_ui_model.dart'
+    show ruleIdentifierKey;
 
 /// UI model for a single static route entry.
 ///
@@ -24,6 +26,12 @@ class StaticRouteUIModel extends Equatable {
     required this.interfaceName,
     this.interfacePath = '',
   });
+
+  /// Stable, kebab-case key for E2E `identifier` hooks (e.g.
+  /// `static-route-edit-<key>`). Derived from the route name ("Web Server" →
+  /// "web-server"); falls back to the saved instance number, then "unnamed",
+  /// so it is always non-empty and never collides across rows.
+  String get identifierKey => ruleIdentifierKey(name, instancePath);
 
   StaticRouteUIModel copyWith({
     String? instancePath,
--- a/lib/page/static_routing/views/usp_static_routing_view.dart
+++ b/lib/page/static_routing/views/usp_static_routing_view.dart
@@ -98,6 +98,7 @@ class UspStaticRoutingView extends ConsumerWidget {
           children: [
             AppText.titleMedium(loc(context).staticRoutes),
             AppIconButton(
+              identifier: 'static-route-add',
               icon: AppIcon.font(Icons.add, size: 20),
               onTap: isSaving ? null : () => _showAddDialog(context, ref),
             ),
@@ -167,12 +168,14 @@ class UspStaticRoutingView extends ConsumerWidget {
               ),
             ),
             AppIconButton(
+              identifier: 'static-route-edit-${route.identifierKey}',
               icon: AppIcon.font(Icons.edit, size: 18),
               onTap: isSaving
                   ? null
                   : () => _showEditDialog(context, ref, index, route),
             ),
             AppIconButton(
+              identifier: 'static-route-delete-${route.identifierKey}',
               icon: AppIcon.font(Icons.delete_outline, size: 18),
               onTap: isSaving
                   ? null
--- a/lib/page/dmz/views/usp_dmz_view.dart
+++ b/lib/page/dmz/views/usp_dmz_view.dart
@@ -192,6 +192,7 @@ class _UspDmzViewState extends ConsumerState<UspDmzView> {
               ),
             ),
             AppSwitch(
+              identifier: 'dmz-enabled',
               value: pending.isEnabled,
               onChanged: disabled
                   ? null
@@ -226,6 +227,7 @@ class _UspDmzViewState extends ConsumerState<UspDmzView> {
             padding: const EdgeInsets.all(AppSpacing.md),
             child: AppIpv4TextField(
               controller: _destIpController,
+              identifier: 'dmz-dest-ip',
               onChanged: (value) {
                 notifier.updateSetting((m) => m.copyWith(destIp: value));
               },
@@ -262,10 +264,12 @@ class _UspDmzViewState extends ConsumerState<UspDmzView> {
               itemHeight: 56,
               items: [
                 AppRadioListItem(
+                  identifier: 'dmz-source-any',
                   title: loc(context).anyAllSources,
                   value: DmzSourceType.any,
                 ),
                 AppRadioListItem(
+                  identifier: 'dmz-source-cidr',
                   title: loc(context).cidrRange,
                   expandedWidget: pending.sourceType == DmzSourceType.cidr
                       ? Container(
@@ -273,6 +277,7 @@ class _UspDmzViewState extends ConsumerState<UspDmzView> {
                           child: AppTextFormField(
                             controller: _cidrController,
                             focusNode: _cidrFocus,
+                            identifier: 'dmz-source-cidr-ip',
                             hintText: 'e.g. 192.168.1.0/24',
                             onChanged: (value) {
                               notifier.updateSetting(
--- a/lib/page/static_routing/views/dialogs/static_route_dialog.dart
+++ b/lib/page/static_routing/views/dialogs/static_route_dialog.dart
@@ -199,13 +199,15 @@ class _StaticRouteDialogState extends State<StaticRouteDialog> {
         ),
       ),
       actions: [
-        TextButton(
-          onPressed: () => Navigator.of(context).pop(),
-          child: Text(loc(context).cancel),
+        AppButton.text(
+          identifier: 'static-route-cancel',
+          label: loc(context).cancel,
+          onTap: () => Navigator.of(context).pop(),
         ),
-        FilledButton(
-          onPressed: _isFormValid ? _submit : null,
-          child: Text(_isEdit ? loc(context).save : loc(context).add),
+        AppButton.primary(
+          identifier: 'static-route-submit',
+          label: _isEdit ? loc(context).save : loc(context).add,
+          onTap: _isFormValid ? _submit : null,
         ),
       ],
     );
--- a/test/demo/demo_usp_transport_composition_test.dart
+++ b/test/demo/demo_usp_transport_composition_test.dart
@@ -71,12 +71,14 @@ void main() {
       expect(r.containsKey('Device.WiFi.Radio.2.Enable'), isTrue);
     });
 
-    test('UspClient back-fills a missing non-wildcard path as null', () async {
+    test('UspClient omits a missing non-wildcard path (no back-fill, #1184)',
+        () async {
       final r = await client.get(['Device.DeviceInfo.DoesNotExist']);
-      // Demo transport omits missing paths; the real UspClient back-fills them
-      // to null so codegen null-casts don't blow up.
-      expect(r.containsKey('Device.DeviceInfo.DoesNotExist'), isTrue);
-      expect(r['Device.DeviceInfo.DoesNotExist'], isNull);
+      // #1184 removed the null back-fill: an absent concrete path is NOT added
+      // to the result (it emits an onMissingPath warning instead). Back-filling
+      // made containsKey() true and silently defeated the codegen 9998
+      // required-leaf check, so the path must now be absent from the map.
+      expect(r.containsKey('Device.DeviceInfo.DoesNotExist'), isFalse);
     });
   });
 


lib/page/static_routing/models/static_routing_ui_model.dart
```diff
@@ -25,6 +27,12 @@ class StaticRouteUIModel extends Equatable {
     this.interfacePath = '',
   });
 
+  /// Stable, kebab-case key for E2E `identifier` hooks (e.g.
+  /// `static-route-edit-<key>`). Derived from the route name ("Web Server" →
+  /// "web-server"); falls back to the saved instance number, then "unnamed",
+  /// so it is always non-empty and never collides across rows.
+  String get identifierKey => ruleIdentifierKey(name, instancePath);
```
AustinChangLinksys, line 34: **W-1 — Same-name identifierKey collision.** Two routes with the same name (e.g. both "Server") will produce duplicate `static-route-edit-server` / `static-route-delete-server` in the widget tree. E2E selector becomes non-unique. Consider appending index or instance number. See Warning details in the review summary.

lib/page/static_routing/models/static_routing_ui_model.dart
```diff
@@ -1,4 +1,6 @@
 import 'package:equatable/equatable.dart';
+import 'package:privacy_gui/page/_shared/models/port_forwarding_rule_ui_model.dart'
+    show ruleIdentifierKey;
```
AustinChangLinksys, line 3: **W-3 — [Feature Envy] Cross-domain import.** `ruleIdentifierKey` is generic slug logic (not PF-specific) but lives in `port_forwarding_rule_ui_model.dart`. Now consumed by 3 unrelated page domains. Suggest extracting to `_shared/utils/identifier_utils.dart`.

lib/page/static_routing/views/usp_static_routing_view.dart
```diff
@@ -167,12 +168,14 @@ class UspStaticRoutingView extends ConsumerWidget {
               ),
             ),
             AppIconButton(
+              identifier: 'static-route-edit-${route.identifierKey}',
```
AustinChangLinksys, line 171: **W-1 ref.** If two routes share the same name, this identifier will collide with the other row. See W-1 in the review summary.

AustinChangLinksys (commented):
🤖 **Automated Review — Round 1** · `d914833..dcd10de` (full)

**Verdict: 💬 Self-review (comment only)** — Austin's own PR; cannot auto-approve. Review provided as a blind-spot check.

#### Standards

|---|---|---|---|
| ⚠️ | 🟢High | `static_routing_ui_model.dart:34` / `usp_static_routing_view.dart:171,178` | Same-name routes produce identical `identifierKey` → duplicate `static-route-edit-<key>` / `-delete-<key>`, E2E selector non-unique |
| ⚠️ | 🟢High | `usp_static_routing_view.dart:137` | Per-row `AppSwitch` (enable/disable route) has no `identifier`; PF equivalent has `pf-rule-enable-${key}` — coverage gap |
| ⚠️ | 🟢High | `static_routing_ui_model.dart:2-3` | [Feature Envy] `StaticRouteUIModel` imports `ruleIdentifierKey` from a `port_forwarding`-named file; third consumer after PF; function is generic, housing is not |
| 💡 | 🟡Med | `static_route_dialog.dart:114` | Error `switch` fallback `_ => key` exposes raw error-key strings to the user (pre-existing; touched by this PR) |
| 💡 | 🟢High | `static_route_dialog.dart:173` | `SegmentedButton<String>` is native Material, not ui_kit; pre-existing pattern across 4+ dialogs — track as batch migration |
| 💡 | 🟢High | `static_routing_ui_model.dart:34` | `StaticRouteUIModel.identifierKey` has no unit test; `PortForwardingRuleUIModel.identifierKey` does — parity gap |

#### Spec

|---|---|---|---|
| 💡 | 🟢High | `static_route_dialog.dart:202-206` | `static-route-cancel` not in Group 3 spec — reasonable but undeclared scope addition; PR body covers the `AppButton` swap rationale but not this specific identifier |
| 💡 | 🟢High | spec#Group3-item3/4 → `usp_static_routing_view.dart:171,178` | edit/delete use `${key}` suffix vs spec's bare strings — intentional deviation, correctly explained in PR body, conformance confirmed |
| 💡 | 🟡Med | `static_routing_ui_model.dart:34` | Spec doesn't specify fallback for same-name routes; `ruleIdentifierKey` inherits PF's limitation — same-name routes produce colliding keys (covered in Standards above) |

<details>

**W-1 — Same-name route identifierKey collision** · 🟢High
`lib/page/static_routing/models/static_routing_ui_model.dart:34` / `lib/page/static_routing/views/usp_static_routing_view.dart:171, 178`

Both Reviewer A and B flagged independently (convergent finding).

```dart
// static_routing_ui_model.dart:34
String get identifierKey => ruleIdentifierKey(name, instancePath);

// usp_static_routing_view.dart:171, 178
identifier: 'static-route-edit-${route.identifierKey}',
identifier: 'static-route-delete-${route.identifierKey}',

// ruleIdentifierKey (port_forwarding_rule_ui_model.dart:10-16)
String ruleIdentifierKey(String description, String? instancePath) {
  final slug = description.trim().toLowerCase()
      .replaceAll(RegExp(r'[^a-z0-9]+'), '-')
      .replaceAll(RegExp(r'^-+|-+$'), '');
  if (slug.isNotEmpty) return slug;
  // falls back to instance number or 'unnamed'
```

Two routes with the same name (e.g. both named "Server") before being saved to the device both get `identifierKey == "server"`, producing duplicate `static-route-edit-server` / `static-route-delete-server` in the same render tree. E2E `[flt-semantics-identifier="static-route-edit-server"]` will match multiple elements — non-deterministic click / test failure.

**Fix (option A — index-aware):**
```dart
identifier: 'static-route-edit-${index}-${route.identifierKey}',
identifier: 'static-route-delete-${index}-${route.identifierKey}',
```
Option B: change `ruleIdentifierKey` to combine name-slug with instance-number whenever instancePath is non-null (already saved rows become unique via device path). Note: if the same limitation is acceptable for PF it may be acceptable here too as a known-gap; the PR comment notes it explicitly.

---

**W-2 — Per-row AppSwitch missing identifier** · 🟢High (flagged by both A and B)
`lib/page/static_routing/views/usp_static_routing_view.dart:137`

Line not in this diff (unchanged), but the surrounding edit/delete buttons were added here and the switch was left unhooked.

```dart
AppSwitch(
  // no identifier
  value: route.enabled,
  scale: 0.8,
  onChanged: isSaving ? null : (value) => ref
      .read(uspStaticRoutingProvider.notifier)
      .toggleRoute(index, value),
),
```

PF equivalent (usp_single_port_tab.dart):
```dart
AppSwitch(
  identifier: 'pf-rule-enable-${rule.identifierKey}',
  value: rule.enabled,
  ...
```

E2E cannot directly target the enable/disable toggle for a specific route by stable identifier. Recommend adding `identifier: 'static-route-toggle-${route.identifierKey}'`.

---

**W-3 — [Feature Envy] ruleIdentifierKey lives in a port-forwarding-named file** · 🟢High
`lib/page/static_routing/models/static_routing_ui_model.dart:2-3`

```dart
import 'package:privacy_gui/page/_shared/models/port_forwarding_rule_ui_model.dart'
    show ruleIdentifierKey;
```

`ruleIdentifierKey` is generic slug logic (name → kebab-case key), not port-forwarding-specific. Now imported by `PortForwardingRuleUIModel`, `PortTriggeringRuleUIModel`, and `StaticRouteUIModel`. If the slug logic ever needs to change (e.g. Unicode folding), the change lands in a file named `port_forwarding_rule_ui_model.dart`, which is surprising for maintainers of the other pages.

**Fix:** Extract `ruleIdentifierKey` to `lib/page/_shared/utils/identifier_utils.dart` (or a semantically neutral shared file). Existing callers continue with `show ruleIdentifierKey` — purely a mechanical refactor.

</details>

<details>

- **ruleIdentifierKey() is always non-empty**: three-tier fallback (name slug → instance number → 'unnamed') prevents blank identifiers.
- **DMZ identifiers are all static strings** (dmz-enabled, dmz-dest-ip, dmz-source-any, dmz-source-cidr, dmz-source-cidr-ip) — no user input interpolation, no injection surface.
- **AppButton swap is correct**: ui_kit_library v2.34.3 AppButton accepts and forwards identifier; `onTap: null` disabled-state semantics match the original `FilledButton(onPressed: null)`.
- **_isFormValid on line 126** intentionally calls `_computeErrors().isEmpty` (not cached `_errors`) to ensure Submit is correctly disabled in add-mode with an empty form — design is deliberate and correct.
- **_submit() trims input** (static_route_dialog.dart:216-221) before sending — no whitespace-only names reach the model.
- **Group 3 spec coverage is complete**: all 9 required identifiers (static-route-add, -submit, -cancel, -edit-${key}, -delete-${key}; dmz-enabled, -dest-ip, -source-any, -source-cidr, -source-cidr-ip) are present at the correct call sites.
- **dart format / flutter analyze clean** and 72 functional tests pass (as declared in PR body).

</details>

AustinChangLinksys (commented):
🤖 **Automated Review — Round 2** · `dcd10de..628062a` (incremental)

**Verdict: 💬 Self-review (comment only)** — Austin's own PR; cannot auto-approve. Review provided as a blind-spot check.

#### Standards

|---|---|---|---|
| ⚠️ | 🟢High | `demo_usp_transport_composition_test.dart:10` | File-level docstring still says "wildcard back-fill" — misleading after #1184 removed concrete-path back-fill; should read "wildcard expansion" |
| 💡 | 🟡Med | `demo_usp_transport_composition_test.dart:74` | Test name mixes behavior description + historical reference (`#1184`) — minor naming style |

#### Spec

|---|---|---|---|
| ⚠️ | 🟢High | spec#1184-contract → `demo_usp_transport_composition_test.dart:10` | Docstring "wildcard back-fill" contradicts #1184 contract ("deliberately do NOT back-fill") — same finding as Standards W-1 above |
| 💡 | 🟡Med | spec#1184-contract → `demo_usp_transport_composition_test.dart:74–85` | `onMissingPath` callback trigger not verified at composition layer; unit-layer coverage exists (`usp_client_normalize_response_test.dart:80–95`) — enhancement only |

<details>

**W-1 — Stale "wildcard back-fill" in file-level docstring** · 🟢High
`test/demo/demo_usp_transport_composition_test.dart:10`

Flagged independently by both Reviewer B (Standards) and Reviewer C (Spec).

```dart
// Line 10 (head, 628062a)
/// UspClient's real coercion / wildcard back-fill and that demo's SET/ADD/
```

After #1184, `UspClient.normalizeGetResponse()` (`usp_client.dart:334–349`) deliberately does **not** back-fill absent concrete paths — it calls `onMissingPath?.call(path)` and omits the key from the result. The `normalizeGetResponse` doc comment itself says "deliberately do NOT back-fill it." The docstring at line 10 still says "wildcard back-fill," which now implies the removed null-fill behaviour rather than wildcard-expansion semantics that remain correct.

This is a test-only file with no production impact, but the misleading term could cause a future maintainer to misread the intent of this composition suite.

**Fix:** Change line 10 from `/ wildcard back-fill` to `/ wildcard expansion (no concrete-path back-fill since #1184)` (or similar), aligned with the `normalizeGetResponse` documentation.

</details>

<details>

- **Core assertion is correct**: `expect(r.containsKey('Device.DeviceInfo.DoesNotExist'), isFalse)` correctly mirrors `UspClient.normalizeGetResponse()` — absent concrete path is omitted from result, not back-filled. Code-verified against `usp_client.dart:334–349`.
- **Second assertion correctly removed**: The deleted `expect(r['...DoesNotExist'], isNull)` was testing back-fill behaviour that no longer exists; keeping it would be misleading (Dart's Map returns `null` for absent keys regardless).
- **Alignment with unit test**: This composition test and `usp_client_normalize_response_test.dart:18–32` (`'absent concrete path is NOT added to the result'`) now agree — dual-layer coverage with proper separation of concerns.
- **Security posture improved**: The old `containsKey → true` assertion was inadvertently confirming a behaviour that silenced the codegen 9998 required-field check. The new assertion removes that misconfirmation.
- **Round 1 Warnings (W-1 identifierKey collision, W-2 AppSwitch missing identifier, W-3 Feature Envy) not addressed in this incremental commit** — these remain open and will carry forward; this incremental diff is scoped to the CI-unblocking test fix only.

</details>

AustinChangLinksys (commented):
🤖 **Automated Review — Round 2 · Reviewer B** · `dcd10de..628062a` (incremental)

**Verdict: 💬 Self-review (comment only)** — Austin's own PR; cannot auto-approve.

---

### Incremental scope

Single commit `628062a` · 1 file · 12 lines changed:
`test/demo/demo_usp_transport_composition_test.dart` — aligns stale CI assertion with the #1184 no-back-fill contract.

---

### Architecture & Maintainability findings

#### Standards (deep-module / design rubric)

| | Conf. | Where | Issue |
|---|---|---|---|
| ⚠️ | 🟢High | `demo_usp_transport_composition_test.dart:10` | File-level docstring still reads `"wildcard back-fill"` — misleading now that concrete-path back-fill was removed by #1184; should read `"wildcard expansion"` |
| 💡 | 🟡Med | `demo_usp_transport_composition_test.dart:74` | Test name embeds historical PR reference (`#1184`) — couples test naming to ticket history rather than behaviour; minor style |

*Confidence: 🟢High = code-verified · 🟡Med = located + reasoned · ⚪Low = speculative.*

---

<details>

**W-1 — Stale "wildcard back-fill" in file-level docstring** · 🟢High
`test/demo/demo_usp_transport_composition_test.dart:10`

```dart
// Current (head 628062a, line 10):
/// UspClient's real coercion / wildcard back-fill and that demo's SET/ADD/
```

`usp_client.dart:322–325` (the canonical source of truth for this contract) now reads:

```
/// 2. Warn (via [onMissingPath]) for each requested non-wildcard path absent
///    from the response — but deliberately do NOT back-fill it.
```

The docstring at line 10 of the composition test file still says *"wildcard back-fill"*, which is now a misnomer that conflates two distinct concepts:
- **wildcard expansion** (still present ✅) — `Device.WiFi.Radio.*.Enable` → two concrete keys
- **concrete-path back-fill** (removed by #1184 ❌) — absent key silently added as `null`

A future maintainer reading line 10 could reasonably assume the null-fill is intentional and revert the correct `isFalse` assertion.

**Fix:** Change `/ wildcard back-fill` → `/ wildcard expansion (concrete-path back-fill removed by #1184)` or similar, matching the language in `usp_client.dart:323`.

---

**S-1 — Test name contains PR-ticket reference (style)** · 🟡Med
`test/demo/demo_usp_transport_composition_test.dart:74`

```dart
// Current:
test('UspClient omits a missing non-wildcard path (no back-fill, #1184)', () async {
```

Coupling a test name to a PR/issue number is a mild form of informational coupling — the ticket may be closed, renumbered, or on a different tracker. The parenthetical `(#1184)` is not discoverable via `flutter test --name` filtering by behaviour. Prefer a purely behaviour-centric name:

```dart
test('absent concrete path is omitted from result map (not back-filled)', () async {
```

This aligns with how `usp_client_normalize_response_test.dart` names the equivalent unit test (`'absent concrete path is NOT added to the result'`). Consistency between the unit and composition layer is a small but real maintenance gain.

</details>

---

<details>

**Core logic is correct and complete:**
- `expect(r.containsKey('Device.DeviceInfo.DoesNotExist'), isFalse)` — correctly mirrors `normalizeGetResponse()` behaviour (line 346 of `usp_client.dart`): absent concrete path triggers `onMissingPath?.call(path)` and the key is **not** added to the result map.
- The second deleted assertion (`expect(r['...DoesNotExist'], isNull)`) was testing the removed back-fill behaviour; removing it is correct. (Dart `Map[]` returns `null` for missing keys regardless, so it would have passed vacuously — keeping it would have been silently misleading.)

**Test-layer module shape is good (deep-module check passed):**
- Deletion test: removing this composition test would push the "absent path is omitted" contract verification to `usp_client_normalize_response_test.dart` alone — but that unit test operates at the normalizeGetResponse seam with a raw Map, not through the full UspClient/transport stack. The composition test earns its existence because it verifies the contract holds end-to-end through `DemoUspTransport → UspClient → normalizeGetResponse`. ✅ Deep enough to keep.

**Dual-layer coverage alignment:**
- Unit layer (`usp_client_normalize_response_test.dart:18–32`): `'absent concrete path is NOT added to the result'` — verifies normalizeGetResponse in isolation.
- Composition layer (this file, line 74): verifies the same contract is observable through the full wiring. Two adapters at different granularity — not redundant, complementary (correct multi-layer test design).

**No production impact:** test-only file; no changes to any `lib/` path.

**Round 1 open items correctly not in scope:** W-1 (identifierKey collision), W-2 (AppSwitch missing identifier), W-3 (Feature Envy) remain open from Round 1; this incremental commit correctly scopes to CI unblocking only.

</details>

---

### Round 1 open items status

| Round 1 item | Status after this commit |
|---|---|
| W-1 Same-name identifierKey collision | 🔴 Still open |
| W-2 AppSwitch per-row missing identifier | 🔴 Still open |
| W-3 Feature Envy (ruleIdentifierKey location) | 🔴 Still open |
| S-1 static-route-cancel undeclared in spec | 🔴 Still open (informational) |

None of these are expected to be addressed in this CI-fix commit. Carrying forward.

---

*Reviewer B — Architecture & Maintainability · Automated · please sanity-check before merge.*

AustinChangLinksys (commented):
Reviewer C (Spec) - Round 2 incremental review - dcd10de..628062a

## Round 2 Prior Issue Tracking

| Status | Prior Item | Notes |
|---|---|---|
| NOT FIXED | demo_usp_transport_composition_test.dart:10 - "wildcard back-fill" in docstring | See C-1 below |
| PARTIAL | :74-85 - onMissingPath callback not verified at composition layer | See C-2 below |

## Incremental Diff Summary

This round adds exactly 1 commit (628062a), touching only `test/demo/demo_usp_transport_composition_test.dart` (12 lines). The core change: test at line 74 updated from `containsKey -> true + value null` (old back-fill behavior) to `containsKey -> false` (the #1184 no-back-fill contract).

## Open Issues

### C-1 - Warning / Confidence: High
**File-level docstring line 10 still says "wildcard back-fill" - contradicts #1184 contract**

Spec reference (#1184 contract): "normalizeGetResponse doc says deliberately do NOT back-fill it. absent concrete path is now OMITTED from the result (containsKey -> false), NOT back-filled."

Actual code at head (demo_usp_transport_composition_test.dart:9-11):
```
/// is VM-testable for the first time. They assert that demo data flows through
/// UspClient's real coercion / wildcard back-fill and that demo's SET/ADD/
/// DELETE return the WASM-unified shape the real parser expects
```

The phrase "wildcard back-fill" at line 10 describes UspClient coercion behavior, but #1184 removed concrete-path back-fill. Future readers will be misled into thinking absent concrete paths are still null-filled. This commit fixed the test assertion but left the docstring stale.

Suggested fix - change line 10 to:
```
/// UspClient's real coercion / wildcard expansion and that demo's SET/ADD/
```
Or append: "concrete-path back-fill removed by #1184; see usp_client_normalize_response_test.dart"

### C-2 - Suggestion / Confidence: Medium
**onMissingPath callback still unverified at composition layer**

Spec reference (#1184): "it emits an onMissingPath warning instead" (commit message + test comment line 77-80)

Code at head (demo_usp_transport_composition_test.dart:74-85):
```dart
test('UspClient omits a missing non-wildcard path (no back-fill, #1184)',
    () async {
  final r = await client.get(['Device.DeviceInfo.DoesNotExist']);
  // #1184 removed the null back-fill: an absent concrete path is NOT added
  // to the result (it emits an onMissingPath warning instead).
  expect(r.containsKey('Device.DeviceInfo.DoesNotExist'), isFalse);
});
```

The test comment at line 77-80 itself notes "emits an onMissingPath warning instead" but only asserts `containsKey -> false`. The callback firing is part of the #1184 behavioral contract at the composition layer. Not a core spec gap (the primary #1184 requirement is "no back-fill"), so keeping as Suggestion; can be addressed in a follow-up issue.

Suggested: If UspClient.withTransport supports injecting onMissingPath, attach a listener in setUp and add:
`expect(missingPaths, contains('Device.DeviceInfo.DoesNotExist'));`

## Summary Table

| Item | Severity | Confidence | Status |
|---|---|---|---|
| C-1: docstring line 10 "wildcard back-fill" misleading (unresolved from prior round) | Warning | High | OPEN |
| C-2: onMissingPath callback unverified in composition layer | Suggestion | Medium | OPEN (same as prior round) |

**Spec compliance for #1184**: The test logic change in this diff is CORRECT - `containsKey -> false` properly aligns with the #1184 "absent concrete path is omitted" contract. C-1 is a documentation precision issue only, does not affect behavioral correctness. C-2 is a non-blocking enhancement suggestion.

AustinChangLinksys (commented):
🤖 **Automated Review — Round 2** · `dcd10de..628062a` (incremental)

**Verdict: 💬 Self-review (comment only)** — Austin's own PR (`feat/e2e-identifiers-group3`); cannot auto-approve. Review provided as a blind-spot check.

---

#### Standards

|---|---|---|---|
| ⚠️ | 🟢High | `demo_usp_transport_composition_test.dart:10` | File-level docstring still says "wildcard back-fill" — contradicts #1184 no-back-fill contract; should read "wildcard expansion (concrete-path back-fill removed by #1184)" |
| 💡 | 🟡Med | `demo_usp_transport_composition_test.dart:74` | Test name embeds `(#1184)` reference — minor style point; acceptable convention but slightly couples test identity to ticket |

#### Spec

|---|---|---|---|
| ⚠️ | 🟢High | spec#1184-contract → `demo_usp_transport_composition_test.dart:10` | Docstring "wildcard back-fill" contradicts #1184 contract ("deliberately do NOT back-fill") — same finding as Standards W-1 |
| 💡 | 🟡Med | spec#1184-contract → `demo_usp_transport_composition_test.dart:74-85` | `onMissingPath` callback trigger not verified at composition layer; unit-layer coverage exists in `usp_client_normalize_response_test.dart` — enhancement only |

<details>

**W-1 — Stale "wildcard back-fill" in file-level docstring** · 🟢High
`test/demo/demo_usp_transport_composition_test.dart:10`

Flagged independently by both Reviewer A (Standards/Security) and Reviewer B (Architecture/Maintainability). Also flagged by Reviewer C (Spec axis) as contradicting the #1184 contract.

```dart
// Line 10 (head, 628062a)
/// UspClient's real coercion / wildcard back-fill and that demo's SET/ADD/
```

After #1184, `UspClient.normalizeGetResponse()` deliberately does **not** back-fill absent concrete paths — it calls `onMissingPath?.call(path)` and omits the key from the result. The docstring here still says "wildcard back-fill," which now implies the removed null-fill behaviour rather than the wildcard-expansion semantics that remain correct.

This is a test-only file with no production impact, but the misleading term could cause a future maintainer to misread the intent of this composition suite.

**Fix:** Change line 10 from `/ wildcard back-fill` to `/ wildcard expansion (no concrete-path back-fill since #1184)`, aligned with the `normalizeGetResponse` documentation.

**Note:** This carry-over Warning from Round 1 was not addressed in the Round 2 incremental commit (which scoped to CI-unblocking assertion fix only). The new assertion (`isFalse`) is correct; only the docstring line remains stale.

</details>

<details>

- **Core assertion is correct**: `expect(r.containsKey('Device.DeviceInfo.DoesNotExist'), isFalse)` properly mirrors `UspClient.normalizeGetResponse()` — absent concrete path is omitted from result, not back-filled. Code-verified against `usp_client.dart:334–349`.
- **Second assertion correctly removed**: The deleted `expect(r['...DoesNotExist'], isNull)` was testing back-fill behaviour that no longer exists; keeping it would be misleading.
- **Test name updated**: Line 74 now reads `'UspClient omits a missing non-wildcard path (no back-fill, #1184)'` — descriptive and behavior-focused.
- **Alignment with unit test**: This composition test and `usp_client_normalize_response_test.dart` now agree — dual-layer coverage with proper separation of concerns.
- **Round 1 Warnings (W-1 identifierKey collision, W-2 AppSwitch missing identifier, W-3 Feature Envy) remain open** — not in scope for this CI-fix commit; will carry forward.

</details>

AustinChangLinksys (commented):
🤖 **Automated Review — Round 2** · `dcd10de..628062a` (incremental)

**Verdict: 💬 Self-review (comment only)** — Austin's own PR (`feat/e2e-identifiers-group3`); cannot auto-approve. Review provided as a blind-spot check.

---

#### Standards

|---|---|---|---|
| ⚠️ | 🟢High | `test/demo/demo_usp_transport_composition_test.dart:10` | File-level docstring still says "wildcard back-fill" — contradicts #1184 no-back-fill contract; this commit's stated purpose is aligning with #1184 yet line 10 was not touched |
| 💡 | 🟡Med | `test/demo/demo_usp_transport_composition_test.dart:74–80` | `onMissingPath` callback invocation not verified at composition layer; unit-layer coverage exists in `usp_client_normalize_response_test.dart` — enhancement only, not a blocker |

#### Spec

|---|---|---|---|
| ⚠️ | 🟡Med | spec#1184-contract → `test/demo/demo_usp_transport_composition_test.dart:10` | `wildcard back-fill` phrasing semantically ambiguous — may refer to wildcard expansion (still valid) or null back-fill (#1184 abolished); suggest `wildcard expansion` to eliminate ambiguity |

<details>

**W-1 — Stale "wildcard back-fill" in file-level docstring (carried forward, persists across rounds)** · 🟢High  
`test/demo/demo_usp_transport_composition_test.dart:10`

Flagged independently by both Reviewer A (Standards/Security) and Reviewer B (Architecture/Maintainability). Also flagged by Reviewer C (Spec axis) as potentially ambiguous with respect to the #1184 contract.

```dart
// Line 10 (head, 628062a)
/// UspClient's real coercion / wildcard back-fill and that demo's SET/ADD/
```

This commit's purpose is to align the demo composition test with #1184's no-back-fill contract (commit message: `"test(usp): update demo composition test to no-back-fill contract (#1184)"`). The test function body and test name were correctly updated, but the file-level docstring on line 10 still reads "wildcard back-fill". The production contract at `lib/core/usp/services/usp_client.dart:322–325` is unambiguous: `"deliberately do NOT back-fill it"`.

**Note (Reviewer C reconciliation):** The phrase "wildcard back-fill" on line 10 _could_ refer to wildcard path expansion (a legitimate behaviour that remains) rather than null back-fill. However, given the surrounding context — this suite was just updated to assert the _removal_ of null back-fill — the ambiguous phrasing poses a real risk of misleading future maintainers. Changing to "wildcard expansion" eliminates both readings.

**Fix:** Change line 10:
```dart
// Before:
/// UspClient's real coercion / wildcard back-fill and that demo's SET/ADD/
// After:
/// UspClient's real coercion / wildcard expansion (no null back-fill since #1184) and that demo's SET/ADD/
```

</details>

<details>

- **Core assertion is correct**: `expect(r.containsKey('Device.DeviceInfo.DoesNotExist'), isFalse)` properly mirrors `UspClient.normalizeGetResponse()` — absent concrete path is omitted from result, not back-filled. Code-verified against `usp_client.dart` `normalizeGetResponse` implementation.
- **Second assertion correctly removed**: The deleted `expect(r['...DoesNotExist'], isNull)` was testing back-fill behaviour that no longer exists; keeping it would be misleading.
- **Test name updated correctly**: Line 74 now reads `'UspClient omits a missing non-wildcard path (no back-fill, #1184)'` — descriptive and behavior-focused.
- **Alignment with unit test**: This composition test and `usp_client_normalize_response_test.dart` now agree — dual-layer coverage with proper separation of concerns.
- **S-1 (onMissingPath) resolved at unit layer**: `usp_client_normalize_response_test.dart` already contains dedicated coverage (`absent concrete path invokes onMissingPath; present one does not`), so the composition test does not need to duplicate this. Prior S-1 concern is effectively closed.
- **CI fix scope is correct**: This CI-fix commit is scoped tightly to what #1184 broke; it does not widen scope into #1172 Group 3 identifier work.

</details>

PeterJhongLinksys (approved):
🤖 **Automated Review — Round 1** · `d914833..628062a` (full)

**Verdict: ✅ APPROVE** — Clean, data-derived E2E identifier work with no architectural violations; merging is safe, but the `identifierKey` fallback chain does not actually work for Static Routing (all unnamed routes collapse to `unnamed`), so please fix the regex before relying on these hooks in tests.

|---|---|---|---|
| ⚠️ | 🟢 | `_shared/models/port_forwarding_rule_ui_model.dart:19` | Instance-number fallback regex `(\d+)$` never matches real `instancePath` values, which end with a dot — empty-name routes all become `unnamed`. |
| ⚠️ | 🟢 | `test/page/_shared/models/port_forwarding_rule_identifier_test.dart:50,55` | Existing unit test feeds dot-less `instancePath` shapes that never occur in production, so it green-lights the broken fallback above. |
| ⚠️ | 🟢 | `static_routing/models/static_routing_ui_model.dart:33` | DartDoc claims "never collides across rows", but route names have no uniqueness validation — two same-named routes produce identical identifiers. |
| ⚠️ | 🟢 | `static_routing/views/usp_static_routing_view.dart:137` | Row `AppSwitch` (enable toggle) has no `identifier`, while edit/delete in the same row got one — inconsistent with `pf-rule-enable-*` in port forwarding. |
| ⚠️ | 🟢 | `static_routing/models/static_routing_ui_model.dart:34` | New `identifierKey` getter has no unit test; `test/page/static_routing/models/` does not exist. |
| 💡 | 🟢 | `_shared/models/port_forwarding_rule_ui_model.dart:10` | [Feature Envy / Misplaced Responsibility] `ruleIdentifierKey` lives in a port-forwarding-named model file but is now shared by 3 features. |
| 💡 | 🟡 | `static_routing/views/usp_static_routing_view.dart:68` · `dmz/views/usp_dmz_view.dart:119` | `UiKitBottomBarConfig` Save/Revert has no identifier anywhere in the repo — the "write-proof" goal is not closed for the actual save action. |
| 💡 | 🟢 | `static_routing/views/dialogs/static_route_dialog.dart:173` | Raw `SegmentedButton` (interface selector) still un-migrated and has no identifier, though it drives the gateway validation branch. |
| 💡 | 🟢 | `static_routing/models/static_routing_ui_model.dart:30-33` | [Duplicated Code] Fallback-chain spec restated in 3 doc comments; will drift. |
| 💡 | 🟢 | `test/demo/demo_usp_transport_composition_test.dart:74` | Test-only `#1184` contract fix is correct but is a second concern versus the PR title — worth calling out in the description. |
| 💡 | 🟢 | `test/golden_test/page/static_routing/localizations/usp_static_routing_view_test.dart:63` | Golden test still locates the add button via `find.byType(AppIconButton).first`; `static-route-add` now exists and could replace it. |

<sub>Confidence: 🟢 High = verified against head code with a quoted snippet · 🟡 Med = file:line + reasoning, not fully substantiated · ⚪ Low = speculative.</sub>
<sub>`[X]` = raised by 1 agent · `[XX]` = raised by both agents independently. No Critical findings this round.</sub>

<details>

**W1 — `identifierKey` instance-number fallback is dead code for Static Routing (and PF/PT) [XX] 🟢**

Helper, `lib/page/_shared/models/port_forwarding_rule_ui_model.dart:10-21`:
```dart
String ruleIdentifierKey(String description, String? instancePath) {
  final slug = description.trim().toLowerCase()
      .replaceAll(RegExp(r'[^a-z0-9]+'), '-')
      .replaceAll(RegExp(r'^-+|-+$'), '');
  if (slug.isNotEmpty) return slug;
  final instance = instancePath == null
      ? null
      : RegExp(r'(\d+)$').firstMatch(instancePath)?.group(1);
  return instance ?? 'unnamed';
}
```
But the codegen builds `instancePath` **with a trailing dot**, `lib/generated/static_routing.g.dart:87,131`:
```dart
final p = '$basePath$id.';   // e.g. 'Device.Routing.Router.1.IPv4Forwarding.3.'
...
instancePath: p,
```
and the service passes it through verbatim (`usp_static_routing_service.dart:193` `instancePath: r.instancePath`). `(\d+)$` anchors at end-of-string, so it does not match `...IPv4Forwarding.3.` — the second fallback tier is unreachable and the result is always `'unnamed'`.

Why this is reachable: an empty route name is a state the code already accommodates — `usp_static_routing_view.dart:152` renders `route.name.isNotEmpty ? route.name : loc(context).unnamed`, and codegen reads `Alias` as `(response['${p}Alias'] ?? '') as String` (`static_routing.g.dart:140`), so a device returning an empty Alias is normal. Non-ASCII names (e.g. `辦公室`) are also fully stripped by `[^a-z0-9]+`.

Trigger + impact: a device with ≥2 routes whose names are empty / symbol-only / non-ASCII yields two rows with identical `static-route-edit-unnamed` and `static-route-delete-unnamed`. Playwright strict mode fails, or worse, a non-strict locator deletes the wrong route.

Fix: tolerate the trailing separator, e.g. `RegExp(r'(\d+)\.?$')`, and keep the change in the shared helper so PF/PT benefit too.

---

**W2 — Existing helper test uses a production-impossible `instancePath` shape [X] 🟢**

`test/page/_shared/models/port_forwarding_rule_identifier_test.dart:50,55`:
```dart
expect(ruleIdentifierKey('', 'Device.NAT.PortMapping.2'), '2');
expect(ruleIdentifierKey('---', 'Device.NAT.PortTrigger.7'), '7');
```
Real values are dot-terminated (`lib/generated/port_forwarding.g.dart:89` `final p = '$basePath$id.'`; `port_triggering.g.dart` likewise). The test therefore passes while the production path fails — this is exactly why W1 went unnoticed. Since this PR makes a third feature depend on that fallback, please fix the fixtures to the real shape in the same change.

---

**W3 — "never collides across rows" is not enforced [XX] 🟢**

`lib/page/static_routing/models/static_routing_ui_model.dart:30-34`:
```dart
/// ... falls back to the saved instance number, then "unnamed",
/// so it is always non-empty and never collides across rows.
String get identifierKey => ruleIdentifierKey(name, instancePath);
```
`UspStaticRoutingService.validateRoute()` (`usp_static_routing_service.dart:219-268`) validates name presence, `length > 32`, IP, subnet mask, and gateway subnet — there is no cross-route uniqueness check; `UspStaticRoutingNotifier.addRoute()` (`usp_static_routing_notifier.dart:126-137`) appends unconditionally. Two routes named "Office" with different destinations is a legal configuration and produces duplicate identifiers.

Fix (pick one): soften the doc to "stable when names are unique"; or make the key composite (name slug + instance number, e.g. `web-server-3`) — note that appending the row `index` would reintroduce a positional selector and is not an option; or add uniqueness validation in `validateRoute` (which also has standalone UX value).

---

**W4 — Row enable toggle left without an identifier [XX] 🟢**

`lib/page/static_routing/views/usp_static_routing_view.dart:137-145`:
```dart
AppSwitch(
  value: route.enabled,
  scale: 0.8,
  onChanged: isSaving ? null : (value) =>
      ref.read(uspStaticRoutingProvider.notifier).toggleRoute(index, value),
),   // no identifier
```
Edit and delete in the same row both received keys (`:171`, `:178`). Port forwarding already establishes the pattern for exactly this control: `identifier: 'pf-rule-enable-${rule.identifierKey}'` in `usp_single_port_tab.dart:63`, `usp_port_range_tab.dart:62`, `usp_port_triggering_tab.dart:62`. This switch has no label and no stable visible text, so the role+accessible-name exemption does not apply and tests must fall back to `.nth()`. Suggest `identifier: 'static-route-enable-${route.identifierKey}'` — note it must be distinct from the dialog's existing `static-route-enabled` (`static_route_dialog.dart:192`).

---

**W5 — No test for the new getter [XX] 🟢**

`test/page/_shared/models/port_forwarding_rule_identifier_test.dart` has dedicated groups for `PortForwardingRuleUIModel.identifierKey` (:65) and `PortTriggeringRuleUIModel.identifierKey` (:79). The third binding added here has none, and `test/page/static_routing/models/` does not exist (only `providers/` and `services/`). The new decision this PR makes — feeding `name` rather than `description` — is precisely what is untested; a field rename would break the hooks silently. Suggested cases: name slug, empty name + real dot-terminated `instancePath`, non-ASCII name, two same-named rows.

</details>

<details>

**S1 — Helper ownership.** `ruleIdentifierKey` is defined in `port_forwarding_rule_ui_model.dart:10`, a file named after (and documented as belonging to) port forwarding, yet it now has three consumers across three features: `port_triggering_rule_ui_model.dart:93`, `static_routing_ui_model.dart:34`, and its own model at `:90`. The `show ruleIdentifierKey` clause hides the coupling in the source but not in the import graph. Suggest extracting to `lib/page/_shared/models/rule_identifier_key.dart` (renaming the test file to match) and making the DartDoc feature-neutral.

**S2 — Save/Revert has no anchor.** Both `usp_static_routing_view.dart:68-73` and `usp_dmz_view.dart:119-128` build `UiKitBottomBarConfig` with only `positiveLabel: loc(context).save` / `onPositiveTap`; a repo-wide grep finds no `identifier` on any `UiKitBottomBarConfig`. Any write-path E2E must press Save, which currently means matching a localized label. We could not read `ui_kit_library` v2.34.3 sources locally, so we cannot confirm whether an identifier passthrough already exists — please confirm, and either add it here or open the ui_kit follow-up.

**S3 — Un-migrated `SegmentedButton`.** `static_route_dialog.dart:173-183` still uses the raw Material `SegmentedButton<String>` for the interface selector, with no identifier, even though interface (LAN/Internet) selects the gateway validation branch in `validateRoute` and is therefore on the write path. The same pattern exists in several other dialogs (pre-existing debt), so a full sweep is out of scope; wrapping this one in `Semantics(identifier: 'static-route-interface-<value>')` would be cheap. Related pre-existing items: raw `TextButton`/`FilledButton` still present in `port_triggering_dialog.dart:205,209` and `ipv6_port_service_rule_dialog.dart:205,209` — good follow-up now that this PR sets the `AppButton` precedent.

**S4 — Duplicated doc spec.** The "slug → instance number → unnamed" chain is restated in `static_routing_ui_model.dart:30-33`, `port_forwarding_rule_ui_model.dart:4-9`, and `:86-89`. `port_triggering_rule_ui_model.dart:91-92` already uses the better form (`/// See [ruleIdentifierKey].`); copying that keeps the spec in one place.

**S5 — Mixed concerns.** The `#1184` contract fix in `test/demo/demo_usp_transport_composition_test.dart` is correct and likely needed to keep CI green after the base merge, but it is unrelated to "add stable identifiers (#1172 Group 3)". Worth one line in the PR description so reviewers do not misjudge scope and a revert does not drag it along.

**S6 — Golden test could adopt the new hook.** `test/golden_test/page/static_routing/localizations/usp_static_routing_view_test.dart:63-73` still does `find.byType(AppIconButton)` + `.first` and `EditableText.at(baseIndex)`. Not introduced here, but `static-route-add` now exists (`usp_static_routing_view.dart:101`), making the fix near-zero cost.

</details>

<details>

- **Data-derived keys, not row indices.** `static-route-edit-${route.identifierKey}` / `static-route-delete-${route.identifierKey}` deliberately avoid the `index` parameter that `_buildRouteCard` already has in hand (`usp_static_routing_view.dart:127`) — exactly the right call, and it avoids positional selectors dressed up as identifiers.
- **The `#1184` test correction is right, and we verified every hop.** `usp_client.dart:334-350` `normalizeGetResponse` only coerces keys already present in the raw map and merely calls `onMissingPath?.call(path)` for absent ones without writing back, so `containsKey` is necessarily `false`; `demo_usp_data_loader.dart:168-169` (`final val = _data[path]; if (val != null) result[path] = val;`) omits missing paths as well. The old assertion was wrong. Removing the `expect(r[...], isNull)` line is also an improvement — after the back-fill removal it was a vacuous always-true assertion. The "warning is emitted" behaviour is already covered by `test/core/usp/services/usp_client_normalize_response_test.dart:80-93`, so not repeating it at the composition layer is correct.
- **Genuine Art. XV fix, not just an identifier vehicle.** Replacing raw `TextButton`/`FilledButton` with `AppButton.text` / `AppButton.primary` (`static_route_dialog.dart:202-211`) removes a pre-existing UI-Kit-First violation, and using `primary` for submit is arguably better than the `AppButton.text`-for-both convention in `port_range_forwarding_dialog.dart:278`.
- **Reuse over reinvention.** Calling the already unit-tested pure `ruleIdentifierKey` rather than writing a new slugifier satisfies the "pure, unit-tested derivation helper" rule; the ownership nit (S1) does not diminish the decision.
- **Naming is consistent with the ~113 existing identifiers**: kebab-case, `{feature}-{control}[-{key}]`; `dmz-` / `static-route-` sit naturally alongside `pf-` / `local-network-` / `admin-` / `dashboard-`.
- **`identifierKey` correctly stays out of `props`** (`static_routing_ui_model.dart:59-68`) — putting derived getters into Equatable props is a common misstep, avoided here.
- **Incidental output sanitisation.** `replaceAll(RegExp(r'[^a-z0-9]+'), '-')` narrows an untrusted device-provided `Alias` to `[a-z0-9-]`, so quotes / `<script>` / control characters cannot leak into the Flutter Web `flt-semantics-identifier` DOM attribute.
- **DMZ identifiers are attached at the right nodes**: `dmz-source-cidr-ip` on `AppTextFormField` rather than the wrapping `Container` (`usp_dmz_view.dart:280`), and `dmz-source-any` / `dmz-source-cidr` on the `AppRadioListItem` bodies (`:267`, `:272`); all use `identifier:` rather than a `semanticLabel:` workaround.
- **Zero architectural violations.** No generated imports leaked into `views/` or `providers/`; no provider `autoDispose` / `watch` / `withLock` semantics touched; no new `catch` clauses, so `ServiceError` isolation is intact. Existing USP error handling in `usp_static_routing_service.dart:56-57,178-180` and `usp_dmz_service.dart:40-41,82-84,128-130` correctly uses `catch (e)` with `mapUspErrorToServiceError(e)` and remains untouched.

</details>

AustinChangLinksys (commented):
🤖 **Automated Review — Round 2** · `dcd10de..628062a` (incremental)

**Verdict: 💬 Self-review (comment only)** — Austin's own PR (`feat/e2e-identifiers-group3`); cannot auto-approve. Review provided as a blind-spot check.

---

#### Standards

|---|---|---|---|
| ⚠️ | 🟢High | `test/demo/demo_usp_transport_composition_test.dart:10` | [Carried forward R1→R2] File-level docstring still reads `wildcard back-fill` — contradicts #1184 no-back-fill contract; test body was fixed but docstring was not touched |
| 💡 | 🟡Med | `test/demo/demo_usp_transport_composition_test.dart:74–85` | No `expect(r, isEmpty)` to fully assert the result map contains nothing — `isFalse` on containsKey is sufficient but complementary isEmpty tightens the safety net |
| 💡 | 🟡Med | `test/demo/demo_usp_transport_composition_test.dart:74–85` | Test note mentions `onMissingPath` path but no cross-reference to unit-layer coverage in `usp_client_normalize_response_test.dart:80` — add one-liner pointer for future readers |

#### Spec

|---|---|---|---|
| 💡 | 🟡Med | spec#1184-contract → `test/demo/demo_usp_transport_composition_test.dart:10` | `wildcard back-fill` in file docstring is ambiguous post-#1184 — `wildcard expansion` would remove the dual-reading entirely |

<details>

**W-1 — Stale `wildcard back-fill` in file-level docstring (third consecutive round)** · 🟢High
`test/demo/demo_usp_transport_composition_test.dart:10`

Flagged independently by both Reviewer A (Standards/Security) and Reviewer B (Architecture/Maintainability).

```dart
// Line 10 (head, 628062a) — unchanged across all rounds
/// UspClient's real coercion / wildcard back-fill and that demo's SET/ADD/
```

This commit correctly updates the test name (line 74) and assertions (lines 78-81) to reflect the #1184 no-back-fill contract, but the file-level docstring on line 10 was not touched. `wildcard back-fill` in the #1184 context strongly connotes "null back-fill (abolished)" — even though wildcard _expansion_ (`.*.` path expansion to concrete instances) is a different and still-valid behaviour. The ambiguity misleads future readers of the file header.

Cross-reference: `lib/core/usp/services/usp_client.dart:322-325` (head) is unambiguous: "deliberately do NOT back-fill it".

**Fix (one-line, no logic change):**
```dart
// Before (line 10):
/// UspClient's real coercion / wildcard back-fill and that demo's SET/ADD/
// After:
/// UspClient's real coercion / wildcard expansion and that demo's SET/ADD/
```

</details>

<details>

- **Core assertion flip is correct**: `expect(r.containsKey('Device.DeviceInfo.DoesNotExist'), isFalse)` properly mirrors `UspClient.normalizeGetResponse()` — absent concrete path is omitted from result, not back-filled. Code-verified against `usp_client.dart` `normalizeGetResponse` implementation.
- **Old back-fill assertion removed cleanly**: `expect(r['...DoesNotExist'], isNull)` deleted — no longer tests removed behaviour, no misleading assertion remains.
- **Test name updated correctly**: Line 74 now reads `'UspClient omits a missing non-wildcard path (no back-fill, #1184)'` — descriptive and behavior-focused.
- **Wildcard expansion test unaffected**: `test('wildcard query expands to concrete instances', ...)` at line 68 remains correct and orthogonal — this is wildcard expansion, not null back-fill; the two concepts are correctly kept separate.
- **Scope tightly limited**: Commit touches only the one stale test; no production code changed, no scope creep into #1172 identifier work.
- **Alignment with unit tests**: `usp_client_normalize_response_test.dart` and this composition test now agree — dual-layer coverage with correct separation of concerns.

</details>

PR #1215, opened 2026-08-10
feat(ai): emit the A2UI protocol version in generated messages (requires ui_kit v2.34.4)

HankYuLinksys:
Fixes #1216.

## Summary

From the #1202 evaluation. Our A2UI output omits the `version` field the spec puts on every message, and the reference implementation does not treat it as optional:

```dart
// a2ui_core-0.1.0/lib/src/core/messages.dart:11-21
if (rawVersion is! String) throw A2uiValidationError(...);
if (rawVersion != 'v0.9') throw A2uiValidationError(...);
```

Nothing breaks today, because our own renderer never reads it. The cost is deferred rather than absent: any move to a conformant parser would begin by rejecting **100%** of our model output, so it would have to start with a prompt change and a fresh round of verifying how reliably the model complies. Paying it now, while it is one field, removes that from every later decision — including the re-evaluation this PR's parent issue schedules for when `genui` reaches 1.0.

## Requires ui_kit v2.34.4

This bump is not cosmetic — the PR does not work without it.

The renderer's message-start pattern required the message key to sit immediately after the opening brace:

```
{"version":"v0.9","updateComponents":…}   → no match
```

`containsA2UI()` would have returned false, the view would have taken its plain-text branch, and **the user would have seen raw JSONL instead of cards**. Nothing throws; there is no test spanning both repos.

Found by reviewing this change before pushing it, and fixed in ui_kit `v2.34.4` first, which is safe to land alone because the tolerance is purely additive. Verified both directions: the old pattern does fail on the new format, and the new pattern accepts messages with and without the field.

## The prompt change

Every whole-message example carries the field, and the format rules plus the v0.9 section state the requirement — models improvise shapes the examples do not cover, so it has to appear as a rule and not only as a demonstration.

**Component-schema snippets are untouched.** They are entries inside a message's `components` array and correctly carry no version of their own. Getting this wrong is easy: my first pass flagged 11 of them, which is what led to the test selecting by message-type key rather than by "line starts with `{`".

**The deliberately-wrong multi-line sample also gained the field**, which sharpens it. It now differs from the correct form in exactly one way — a newline inside the JSON, the mistake it exists to show — instead of two.

## Verification

End-to-end rather than by inspection. All six message examples from the live prompt were run through the real `A2UIResponseRenderer` with the real `RouterComponentRegistry`:

```
E2E examples found: 6
E2E detected=true  (×6)
E2E parse error shown: false
```

and rendering reached our actual components (`AppText` via `ai_info_row.dart`), so detection, parsing and widget construction all complete.

| Check | Result |
|---|---|
| `test/ai/prompts/router_system_prompt_test.dart` | 3/3 |
| `flutter test test/ai/ test/page/ai_assistant/` | 110/110 |
| `./run_tests.sh` (full suite) | **3574/3574** |
| `flutter analyze` | 0 issues |
| ui_kit `generative_ui` package tests | 196/196 |

Mutation-checked in both repos: dropping the field from one prompt example fails a test here; restoring the old regex fails five tests in ui_kit, across both the detection and extraction paths.

## Notes for reviewers

- **The extraction path in ui_kit had no test at all** before v2.34.4 — only `containsA2UI` was covered. That is precisely why the brace-adjacency assumption went unnoticed, and it is now closed with three widget tests that render through the real widget.
- **The test guards the envelope, not the wording.** It selects lines carrying a message-type key, so adding or rewording examples will not make it fail spuriously, and it distinguishes whole messages from component fragments.
- This is forward-compatibility work with **no behavioural change today**. The reason to take it now rather than at migration time is sequencing, not urgency.
- The other actionable item from the #1202 evaluation is tracked as #1217 (streaming). It touches the same prompt and renderer, so this one landing first avoids editing the same lines twice.

### DIFF
--- a/lib/ai/prompts/router_system_prompt.dart
+++ b/lib/ai/prompts/router_system_prompt.dart
@@ -96,6 +96,7 @@ CRITICAL FORMAT RULES:
 1. Each JSON message MUST be on a SINGLE LINE (no line breaks inside JSON)
 2. NO pretty-printing, NO indentation, NO newlines within the JSON object
 3. Two JSON messages total: updateComponents (line 1) + createSurface (line 2)
+4. EVERY message MUST start with "version":"v0.9" as its first field
 
 FORBIDDEN:
 - ❌ Multi-line / pretty-printed JSON
@@ -104,14 +105,14 @@ FORBIDDEN:
 - ❌ Code blocks (```json)
 
 CRITICAL ROOT RULE:
-4. The FIRST component MUST have id="root" — this is MANDATORY for A2UI v0.9
+5. The FIRST component MUST have id="root" — this is MANDATORY for A2UI v0.9
 
 CORRECT FORMAT (exactly 2 lines):
-{"updateComponents":{"surfaceId":"main","components":[{"id":"root","component":"AppCard","childIds":["c"]},{"id":"c","component":"WanSection","wanStatus":"Connected"}]}}
-{"createSurface":{"surfaceId":"main","catalogId":"a2ui://router-assistant/v1"}}
+{"version":"v0.9","updateComponents":{"surfaceId":"main","components":[{"id":"root","component":"AppCard","childIds":["c"]},{"id":"c","component":"WanSection","wanStatus":"Connected"}]}}
+{"version":"v0.9","createSurface":{"surfaceId":"main","catalogId":"a2ui://router-assistant/v1"}}
 
 WRONG FORMAT (multi-line JSON breaks the parser):
-{"updateComponents":{
+{"version":"v0.9","updateComponents":{
   "surfaceId":"main",
   "components":[...]
 }}
@@ -120,10 +121,10 @@ WRONG FORMAT (multi-line JSON breaks the parser):
   /// Reminder at the end of the prompt.
   static const _formatReminder = '''
 ⚠️ FINAL FORMAT CHECK:
-- Line 1: {"updateComponents":...} (SINGLE LINE, no newlines inside)
-- Line 2: {"createSurface":...} (SINGLE LINE)
+- Line 1: {"version":"v0.9","updateComponents":...} (SINGLE LINE, no newlines inside)
+- Line 2: {"version":"v0.9","createSurface":...} (SINGLE LINE)
 - NO pretty-printing, NO indentation
-- Start NOW with {"updateComponents":
+- Start NOW with {"version":"v0.9","updateComponents":
 ''';
 
   static const _basePrompt = '''
@@ -138,8 +139,8 @@ When a user asks "how's my network?", you DO NOT write:
   "Your network looks healthy. Internet is connected..."
 
 Instead, you MUST output A2UI JSONL (SINGLE LINE per message, root id="root"):
-{"updateComponents":{"surfaceId":"main","components":[{"id":"root","component":"AppCard","childIds":["c"]},{"id":"c","component":"Column","childIds":["h","wan"]},{"id":"h","component":"SectionHeader","title":"Network"},{"id":"wan","component":"WanSection","wanStatus":"Connected","connectedDevices":3}]}}
-{"createSurface":{"surfaceId":"main","catalogId":"a2ui://router-assistant/v1"}}
+{"version":"v0.9","updateComponents":{"surfaceId":"main","components":[{"id":"root","component":"AppCard","childIds":["c"]},{"id":"c","component":"Column","childIds":["h","wan"]},{"id":"h","component":"SectionHeader","title":"Network"},{"id":"wan","component":"WanSection","wanStatus":"Connected","connectedDevices":3}]}}
+{"version":"v0.9","createSurface":{"surfaceId":"main","catalogId":"a2ui://router-assistant/v1"}}
 
 ## Behavior Guidelines
 
@@ -307,6 +308,7 @@ IMPORTANT FOR MULTI-TURN CONVERSATIONS:
 - The client will REPLACE the entire UI with your new updateComponents
 
 ### v0.9 Message Format Changes (from v0.8):
+- EVERY message carries `"version":"v0.9"` as its first field
 - `surfaceUpdate` → `updateComponents`
 - `dataModelUpdate` → `updateDataModel`
 - `beginRendering` → `createSurface` (with required `catalogId`)
@@ -362,16 +364,16 @@ IMPORTANT FOR MULTI-TURN CONVERSATIONS:
 ### Examples (CRITICAL: Each JSON must be on ONE line, first component MUST have id="root")
 
 #### Example 1: Network Summary
-{"updateComponents":{"surfaceId":"main","components":[{"id":"root","component":"AppCard","childIds":["content"]},{"id":"content","component":"Column","childIds":["h1","wan","d1","h2","lan"]},{"id":"h1","component":"SectionHeader","title":"WAN"},{"id":"wan","component":"WanSection","wanStatus":"Connected","connectedDevices":8},{"id":"d1","component":"AppDivider"},{"id":"h2","component":"SectionHeader","title":"LAN"},{"id":"lan","component":"LanSection","ipAddress":"192.168.1.1","subnetMask":"255.255.255.0"}]}}
-{"createSurface":{"surfaceId":"main","catalogId":"a2ui://router-assistant/v1"}}
+{"version":"v0.9","updateComponents":{"surfaceId":"main","components":[{"id":"root","component":"AppCard","childIds":["content"]},{"id":"content","component":"Column","childIds":["h1","wan","d1","h2","lan"]},{"id":"h1","component":"SectionHeader","title":"WAN"},{"id":"wan","component":"WanSection","wanStatus":"Connected","connectedDevices":8},{"id":"d1","component":"AppDivider"},{"id":"h2","component":"SectionHeader","title":"LAN"},{"id":"lan","component":"LanSection","ipAddress":"192.168.1.1","subnetMask":"255.255.255.0"}]}}
+{"version":"v0.9","createSurface":{"surfaceId":"main","catalogId":"a2ui://router-assistant/v1"}}
 
 #### Example 2: Single Section
-{"updateComponents":{"surfaceId":"main","components":[{"id":"root","component":"AppCard","childIds":["wan"]},{"id":"wan","component":"WanSection","wanStatus":"Connected","connectedDevices":5}]}}
-{"createSurface":{"surfaceId":"main","catalogId":"a2ui://router-assistant/v1"}}
+{"version":"v0.9","updateComponents":{"surfaceId":"main","components":[{"id":"root","component":"AppCard","childIds":["wan"]},{"id":"wan","component":"WanSection","wanStatus":"Connected","connectedDevices":5}]}}
+{"version":"v0.9","createSurface":{"surfaceId":"main","catalogId":"a2ui://router-assistant/v1"}}
 
 #### Example 3: Device List
-{"updateComponents":{"surfaceId":"main","components":[{"id":"root","component":"DeviceListView","devices":[{"name":"iPhone","ip":"192.168.1.101","connectionType":"WiFi 5GHz"}]}]}}
-{"createSurface":{"surfaceId":"main","catalogId":"a2ui://router-assistant/v1"}}
+{"version":"v0.9","updateComponents":{"surfaceId":"main","components":[{"id":"root","component":"DeviceListView","devices":[{"name":"iPhone","ip":"192.168.1.101","connectionType":"WiFi 5GHz"}]}]}}
+{"version":"v0.9","createSurface":{"surfaceId":"main","catalogId":"a2ui://router-assistant/v1"}}
 
 ⚠️ CRITICAL: Always INLINE data directly into components. Do NOT use boundPath references.
 ''';
--- a/lib/ai/prompts/router_system_prompt.dart
+++ b/lib/ai/prompts/router_system_prompt.dart
@@ -3,6 +3,19 @@ import 'package:generative_ui/generative_ui.dart';
 /// System prompt template for the Router AI Assistant.
 ///
 /// This defines how the AI should behave and what components it can generate.
+///
+/// ## Upgrading the A2UI protocol version
+///
+/// `v0.9` appears throughout the prompt below — in the message examples, in the
+/// format rules, and in section headings. It is written out rather than
+/// interpolated because these are `const` strings: interpolating would make
+/// every one of them `static final`, losing the compile-time constant that lets
+/// them be assembled without runtime work on each request.
+///
+/// So a protocol upgrade is a deliberate find-and-replace across this file, and
+/// `router_system_prompt_test.dart` is the safety net — it asserts the version
+/// appears on every whole-message example and that the renderer recognises the
+/// shape, so a partial replacement fails rather than silently shipping a mix.
 class RouterSystemPrompt {
   RouterSystemPrompt._();
 
--- a/test/ai/prompts/router_system_prompt_test.dart
+++ b/test/ai/prompts/router_system_prompt_test.dart
@@ -1,4 +1,5 @@
 import 'package:flutter_test/flutter_test.dart';
+import 'package:generative_ui/generative_ui.dart';
 import 'package:privacy_gui/ai/prompts/router_system_prompt.dart';
 
 /// Guards the A2UI envelope the prompt teaches the model to emit.
@@ -14,7 +15,10 @@ void main() {
     late String prompt;
 
     setUp(() {
-      prompt = RouterSystemPrompt.staticPrompt;
+      // `build()` rather than `staticPrompt`: the latter omits the trailing
+      // format reminder, which IS sent to the model and repeats the message
+      // shape. Asserting on the smaller string would leave that copy unguarded.
+      prompt = RouterSystemPrompt.build();
     });
 
     /// Whole-message example lines, i.e. those carrying a message-type key.
@@ -23,6 +27,11 @@ void main() {
     /// are entries inside a message's `components` array and correctly carry no
     /// version of their own.
     List<String> messageExamples(String text) {
+      // The v0.9 message types. Mirrors the set in ui_kit's `_messageStart`
+      // pattern; if the protocol gains a message type, both need it — but the
+      // consequence of drift here is only that a new example goes unchecked,
+      // and the renderer-recognition test below would still catch a shape it
+      // cannot parse.
       const messageKeys = [
         'updateComponents',
         'createSurface',
@@ -56,6 +65,26 @@ void main() {
       // has to appear as a rule.
       expect(prompt, contains('"version":"v0.9"'));
       expect(prompt, contains('EVERY message'));
+    });
+
+    test('the closing reminder shows the versioned shape too', () {
+      // The reminder is the last thing the model reads before generating, so a
+      // stale shape here competes with the rules above it. It is not part of
+      // `staticPrompt`, which is why this asserts against `build()`.
+      expect(prompt, contains('Start NOW with {"version":"v0.9"'),
+          reason: 'the final instruction must not demonstrate the old shape');
+    });
+
+    test('the renderer recognises every example the prompt teaches', () {
+      // The string assertions above cannot catch the failure that matters: the
+      // renderer detects a message by pattern, so a shape the prompt teaches but
+      // the renderer does not recognise is displayed to the user as raw JSONL.
+      // This is the cross-repo contract, asserted here rather than trusted —
+      // it is what makes the ui_kit ref bump verifiable in CI.
+      for (final example in messageExamples(prompt)) {
+        expect(A2UIResponseRenderer.containsA2UI(example), isTrue,
+            reason: 'renderer would treat this as plain text: $example');
+      }
     });
 
     test('each correct-format example stays on one line', () {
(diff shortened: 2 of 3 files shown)


AustinChangLinksys (approved):
🤖 **Automated Review — Round 1** · `1175273..5fe4aec` (full)

**Verdict: ✅ APPROVE** — Core logic is correct and well-evidenced; 3 warnings to note before and after merge.

#### Standards

|---|---|---|---|
| ⚠️ | 🟢High | `pubspec.yaml:62,67` | Prompt format and renderer version are tightly coupled — PR must be merged atomically; cherry-picking only the prompt half silently degrades the UI to raw JSONL output |
| ⚠️ | 🟢High | `lib/ai/prompts/router_system_prompt.dart` (multiple) | `"v0.9"` version literal is hardcoded in ≥12 places across 5 `static const` fields — a future protocol upgrade requires a shotgun-style find-replace [Shotgun Surgery] |
| 💡 | 🟢High | `test/ai/prompts/router_system_prompt_test.dart:18` | Test only covers `staticPrompt` (which omits `_formatReminder`); the version prefix added to `_formatReminder` is not asserted by any test |
| 💡 | 🟡Med | `test/ai/prompts/router_system_prompt_test.dart:26-40` | `messageExamples()` key list duplicates the key set in `ui_kit`'s `_messageStart` regex — needs a sync comment [Duplicated Code] |
| 💡 | 🟢High | `pubspec.yaml:59-67` | Two `git:` blocks point to the same monorepo ref; no inline comment warns that both `ref:` lines must always be bumped together |

#### Spec

|---|---|---|---|
| ⚠️ | 🟡Med | `test/ai/prompts/router_system_prompt_test.dart` (whole file) | spec#AC-6 → test file: AC-6 ("end-to-end check via real renderer") is met only by manual / ui_kit-side testing; no automated assertion in this repo pipes a prompt example into `A2UIResponseRenderer` |

<details>

**W-1 — Prompt/renderer version coupling (pubspec.yaml:62,67)** 🟢High

`Fix:` This is a deployment-order risk, not a code defect — the PR is correctly assembled (both the prompt change and the `v2.34.4` bump are included). The practical risk is if someone cherry-picks only the prompt changes. Recommend adding a merge checklist note: *"Must be merged as a unit — do not cherry-pick the prompt change without the pubspec bump."* CI's `flutter pub get` will validate that `v2.34.4` is resolvable at merge time.

Evidence chain: v2.34.3 `containsA2UI()` regex requires the message key immediately after `{`; Python-verified that the new prompt format `{"version":"v0.9","updateComponents":...}` does **not** match (returns `False`) — renderer falls back to plain-text branch showing raw JSONL. v2.34.4 adds an optional `version` group; tag SHA `267c1373` confirmed present in remote.

**W-2 — `"v0.9"` scattered across ≥12 literal occurrences (router_system_prompt.dart, multiple)** 🟢High

`Fix:` Add a single `static const _kA2uiVersion = 'v0.9';` anchor constant at the top of `RouterSystemPrompt`, with a doc comment: `/// A2UI protocol version emitted in generated messages. Update here when the protocol version changes.` Note: Dart string interpolation prevents the interpolated strings from being `const`; fields would become `static final String`. If `const` is required downstream, alternative is a block comment grouping the relevant lines: `// Protocol version: "v0.9" — update all occurrences here when upgrading`.

**W-3 (Spec AC-6) — End-to-end test gap (test/ai/prompts/router_system_prompt_test.dart)** 🟡Med

The PR's `router_system_prompt_test.dart` only validates at the string level. The commit message states "all six message examples from the live prompt were run through the real `A2UIResponseRenderer` with the real `RouterComponentRegistry`" — but no corresponding automated test exists in this diff. The extraction-path widget tests live in `ui_kit v2.34.4`'s `a2ui_renderer_test.dart`, using a `Probe` component rather than `RouterComponentRegistry`, and they don't use the actual prompt text.

`Fix:` Add a test assertion such as:
```dart
test('prompt examples are detected by A2UIResponseRenderer', () {
  for (final example in messageExamples(RouterSystemPrompt.staticPrompt)) {
    expect(A2UIResponseRenderer.containsA2UI(example), isTrue,
        reason: 'Example not detected: $example');
  }
});
```
This would elevate the end-to-end guarantee from manual verification to CI-enforced.

</details>

<details>

- **Core correctness verified**: v2.34.4 `_messageStart` regex correctly makes `version` an optional leading field; both old and new formats match. v2.34.4 `A2UIParser.parseJson()` dispatches on message-type keys, ignoring the `version` key harmlessly.
- **All 11 whole-message example lines updated**: Every message-format line in `_formatConstraint`, `_basePrompt`, and all three `_a2uiGuide` examples now begins with `"version":"v0.9"`.
- **Rule stated explicitly**: Both `_formatConstraint` rule #4 and the `### v0.9 Message Format Changes` section explicitly state the requirement — not only demonstrated in examples.
- **Component-schema snippets untouched**: All component-schema fragments in `_componentGuide` and `_a2uiGuide` correctly have no `version` field, per spec.
- **Intentionally-wrong example sharpened**: The multi-line WRONG FORMAT example now has `version` added, isolating exactly one dimension of error (newline inside JSON) instead of two.
- **Test envelope design is sound**: `messageExamples()` selects by message-type key membership, not fixed positions — editing prompt wording won't break tests spuriously.
- **Pattern deduplication in ui_kit v2.34.4**: `_parseContent`'s `startRegex` now references the shared `_messageStart` static, eliminating prior divergence risk.
- **No security issues**: No secrets, no dynamic interpolation, no auth/permission logic touched.
- **pubspec dual-ref bump is correct and necessary**: Both `ui_kit_library` and `generative_ui` are from the same monorepo and must track the same ref.

</details>

test/ai/prompts/router_system_prompt_test.dart
```diff
@@ -0,0 +1,84 @@
...
+/// defaulting. Our own renderer tolerates its absence, so nothing in the app
+/// fails if the prompt stops asking for it — the cost would appear only as a
+/// wholesale rejection the first time our output meets a conformant parser.
+/// Hence a test rather than a comment.
+void main() {
+  group('A2UI envelope', () {
+    late String prompt;
+
+    setUp(() {
+      prompt = RouterSystemPrompt.staticPrompt;
```
PeterJhongLinksys, line 17: ⚠️ **Warning (🟢 High)** — `staticPrompt` does not include `_formatReminder` (see `router_system_prompt.dart:77-87`, which writes only 4 sections, while `build():29` and `buildParts():63,68` append the reminder). Consequently the three version lines this PR changed inside `_formatReminder` (`:124`, `:125`, `:127` — including `Start NOW with {"version":"v0.9","updateComponents":`) are never asserted. A future edit that drops the version from the reminder keeps all three tests green. Suggest asserting against `RouterSystemPrompt.build()` as well.

test/ai/prompts/router_system_prompt_test.dart
```diff
@@ -0,0 +1,84 @@
...
+          reason:
+              'a conformant parser rejects a message with no version, so an '
+              'example without one teaches output that cannot migrate');
+    });
+
+    test('the version requirement is stated, not only demonstrated', () {
+      // Models improvise message shapes the examples do not cover, so the rule
+      // has to appear as a rule.
+      expect(prompt, contains('"version":"v0.9"'));
+      expect(prompt, contains('EVERY message'));
```
PeterJhongLinksys, line 58: ⚠️ **Warning (🟢 High)** — `contains('EVERY message')` asserts the natural-language wording of rule 4 (`router_system_prompt.dart:99`), which contradicts the stated intent of guarding the envelope rather than the wording. Rewording the rule to "Every message must carry…" would turn this red for no behavioural reason. Prefer a structural assertion, e.g. `RegExp(r'(?i)every message .*"version":"v0\.9"')`, or extract the rule text into a const the test references.

test/ai/prompts/router_system_prompt_test.dart
```diff
@@ -0,0 +1,84 @@
...
+      //
+      // The prompt also contains a deliberately-WRONG multi-line sample, shown
+      // so the model can recognise the mistake. Excluded by taking only lines
+      // that close their own object — which is exactly the property under test
+      // for the rest.
+      final wellFormed = messageExamples(prompt)
+          .where((l) => l.trimRight().endsWith('}'))
+          .toList();
+
+      expect(wellFormed.length, greaterThanOrEqualTo(8),
```
PeterJhongLinksys, line 74: 💡 **Suggestion (🟡 Med)** — Magic number. Re-running this filter over the head-version `staticPrompt` yields `examples: 11`, `missing: []`, `wellFormed: 10` (the one non-`}`-terminated line is the deliberate WRONG FORMAT example at `router_system_prompt.dart:115`). A threshold of 8 therefore tolerates silently deleting two examples. `expect(wellFormed.length, examples.length - 1)` states the intent directly.

lib/ai/prompts/router_system_prompt.dart
```diff
@@ -120,10 +121,10 @@ WRONG FORMAT (multi-line JSON breaks the parser):
   /// Reminder at the end of the prompt.
   static const _formatReminder = '''
 ⚠️ FINAL FORMAT CHECK:
-- Line 1: {"updateComponents":...} (SINGLE LINE, no newlines inside)
-- Line 2: {"createSurface":...} (SINGLE LINE)
+- Line 1: {"version":"v0.9","updateComponents":...} (SINGLE LINE, no newlines inside)
```
PeterJhongLinksys, line 124: ⚠️ **Warning (🟢 High)** — `v0.9` is hardcoded ~19× across this file and the new test, yet the dependency already exports the single source of truth: `generative_ui/lib/src/domain/entities/a2ui_messages.dart:9` → `const String a2uiProtocolVersion = 'v0.9';`. `grep -rn a2uiProtocolVersion lib/ test/` returns nothing, so app and renderer now hold two independent copies. Dart allows const interpolation inside `'''` literals, so these can reference one constant; the test should import it rather than retyping the literal (otherwise the test cannot guard a version bump either).

PeterJhongLinksys (approved):
🤖 **Automated Review — Round 1** · `1175273..5fe4aec` (full)

**Verdict: ✅ APPROVE** — Correct, well-scoped change with an honest test rationale; no correctness or security defect found. The remaining points are maintainability debt: the `v0.9` string is hardcoded in ~19 places while `generative_ui` already exports a `a2uiProtocolVersion` constant, and the new test reads `staticPrompt`, which does **not** include `_formatReminder` — so 3 of the newly changed version lines are outside the guard the test was written to provide.

|---|---|---|---|
| ⚠️ | 🟢 | `test/ai/prompts/router_system_prompt_test.dart:17` | Test reads `staticPrompt`, which omits `_formatReminder` → the 3 version lines at `router_system_prompt.dart:124,125,127` are never checked (guard has a blind spot). |
| ⚠️ | 🟢 | `lib/ai/prompts/router_system_prompt.dart` (17 sites) + `test/...:46,57` | [Duplicated Code] `v0.9` hardcoded ~19×; `generative_ui` already exports `const String a2uiProtocolVersion = 'v0.9'` (`src/domain/entities/a2ui_messages.dart:9`) — no single source of truth, and the test hardcodes it too, so it cannot guard an upgrade. |
| ⚠️ | 🟢 | `lib/ai/prompts/router_system_prompt.dart:14-20, 44-50, 79-85` | [Duplicated Code] `build()` / `buildParts()` / `staticPrompt` each re-write the same `writeln` sequence; the divergence (first two append the reminder, the third does not) is exactly what causes the blind spot above. |
| ⚠️ | 🟢 | `test/ai/prompts/router_system_prompt_test.dart:58` | [Fragile Test] `expect(prompt, contains('EVERY message'))` asserts prompt wording, contradicting the stated intent of guarding the envelope, not the wording. |
| 💡 | 🟢 | `test/ai/prompts/router_system_prompt_test.dart:25-37` | [Primitive Obsession] `messageExamples()` parses prompt text by string prefix; `messageKeys` is a second copy of the message-type list that will silently miss a future 5th message type (false green). |
| 💡 | 🟡 | `test/ai/prompts/router_system_prompt_test.dart:74` | [Magic Number] `greaterThanOrEqualTo(8)` — actual counts are examples=11 / wellFormed=10 (verified by executing the filter), so 2 examples could be deleted without turning the test red. |
| 💡 | 🟢 | `lib/ai/prompts/router_system_prompt.dart:96-108` | Numbered rules 1–4 then a `FORBIDDEN:` block, then rule 5 — the renumber 4→5 is correct, but the sequence is now interrupted (no stale references to the old numbering remain; verified by grep). |
| 💡 | 🟡 | `pubspec.yaml:90,95` | The `v2.34.4` bump is a hard requirement of the prompt change (renderer's `containsA2UI()` message-start pattern) but nothing at the pin records why — a future downgrade would not see the constraint. |
| 💡 | 🟡 | `test/ai/prompts/router_system_prompt_test.dart:13` | `group('A2UI envelope')` does not follow the `[ClassName] - [Feature]` convention (existing tests in this directory do the same, hence Suggestion only). |

<sub>Confidence: 🟢 High = read head code, evidence quoted · 🟡 Med = file:line + reasoning, not fully proven · ⚪ Low = speculative.</sub>
<sub>`[X]` = raised by 1 agent, `[XX]` = raised by both agents. **Note:** this round the security/correctness agent exhausted its budget before emitting a report; its verified findings were folded in manually by the synthesizer, so agreement markers are omitted. The architecture/maintainability pass completed normally.</sub>

<details>
<summary>🔴 Critical (0)</summary>

None. No hardcoded credential, injection vector, unhandled error path or async race was found in the changed files; the change touches no USP/JNAP service, provider or `ServiceError` boundary (`lib/ai/prompts/` and `test/ai/prompts/` only). The two maintainability risks below are real but do not produce a runtime defect today, and the change was verified end-to-end against the real renderer, so neither meets the Critical evidence bar.
</details>

<details>

### W1 — `staticPrompt` omits `_formatReminder`, so the new test cannot see 3 of the changed lines
`test/ai/prompts/router_system_prompt_test.dart:17` · `lib/ai/prompts/router_system_prompt.dart:77-87, 122-128` · 🟢 High

```dart
// router_system_prompt.dart:77-87 — staticPrompt composes 4 sections
static String get staticPrompt {
  buffer.writeln(_formatConstraint);
  buffer.writeln(_basePrompt);
  buffer.writeln(_componentGuide);
  buffer.writeln(_a2uiGuide);   // <-- _formatReminder is NOT appended
}
// build():29 and buildParts():63,68 DO append _formatReminder
```
`_formatReminder` contains three lines changed by this PR (`:124`, `:125`, `:127`), including the highest-impact one — `Start NOW with {"version":"v0.9","updateComponents":`. Because `setUp` binds `prompt = RouterSystemPrompt.staticPrompt` (`test:17`), those three lines are permanently outside the assertion.

**When it bites:** someone later edits only `_formatReminder` (drops the version, or moves to `v1.0`); all three tests stay green while the prompt actually sent to the model contradicts itself. That defeats the file's own stated rationale ("Hence a test rather than a comment").

**Fix:** assert against `RouterSystemPrompt.build()` (or add a case for it) so the reminder is covered, and/or extract a shared compose helper (see W3).

### W2 — `v0.9` hardcoded ~19× while the dependency already exports the constant
🟢 High

`lib/ai/prompts/router_system_prompt.dart`: `:99, :108, :111, :112, :115, :124, :125, :127, :142, :143, :291, :293, :310, :311, :367, :368, :371, :372, :375, :376`
`test/ai/prompts/router_system_prompt_test.dart`: `:6` (doc), `:46` (`startsWith('{"version":"v0.9"')`), `:57` (`contains('"version":"v0.9"')`)

`generative_ui` already ships the single source of truth:
```dart
// generative_ui/lib/src/domain/entities/a2ui_messages.dart:8-9
/// Protocol version constant.
const String a2uiProtocolVersion = 'v0.9';
```
Grep confirms this app references it **nowhere** (`grep -rn a2uiProtocolVersion lib/ test/` → no match), so the app and the renderer now carry two independent copies of the protocol version. Dart permits const interpolation inside `'''` literals, so this is mechanically extractable:
```dart
static const _formatConstraint = '''
4. EVERY message MUST start with "version":"$a2uiProtocolVersion" as its first field
{"version":"$a2uiProtocolVersion","updateComponents":{...}}
''';
```
**Fix:** import/reuse `a2uiProtocolVersion` (or define one local const) and have the test import it rather than retyping the literal. Keep `:310` ("from v0.8") as literal prose — that is history, not the current version.

### W3 — Three composition paths repeat the same section sequence
`lib/ai/prompts/router_system_prompt.dart:14-20, 44-50, 79-85` · 🟢 High

`build()`, `buildParts()` and `staticPrompt` each independently re-write `_formatConstraint → _basePrompt → _componentGuide → _a2uiGuide`. The divergence between them is the root cause of W1. **Fix:** extract `static String _composeStatic()` used by all three, and at minimum document on `staticPrompt` (`:76`, "for reference/debugging") that it excludes the format reminder, so the next test author does not repeat the mistake.

### W4 — Test asserts prompt prose while claiming not to
`test/ai/prompts/router_system_prompt_test.dart:58` · 🟢 High

```dart
expect(prompt, contains('EVERY message'));
```
`'EVERY message'` is the natural-language wording of rule 4 at `router_system_prompt.dart:99`. Rewording to "Every message must carry…" turns the test red for no behavioural reason. **Fix:** assert structurally, e.g. `RegExp(r'(?i)every message .*"version":"v0\.9"')`, or extract the rule text into a const the test references (which also removes part of W2).
</details>

<details>

**S1 — `messageExamples()` fragility** (`test:25-37`, 🟢): `split('\n')` → `startsWith('{"')` → `contains('"$k"')`. `messageKeys` (`:26-31`) duplicates the message-type list that lives in the prompt; a future 5th message type example would silently escape the check. Also, treating "line contains `"createSurface"`" as "this is a message" would mis-collect any prose line that quotes the key (none exist today). A sturdier shape: extract envelope construction into a testable function (e.g. `a2uiMessage({required String type, ...})`), generate the prompt examples from it, and assert on the function plus "every example in the prompt originates from it" — that pins version, single-line and first-field-order in one place and subsumes W2. If that is over-engineering for prose, at least annotate `messageKeys` as "must stay in sync with `_a2uiGuide`".

**S2 — magic threshold** (`test:74`, 🟡): executing the filter over the head-version `staticPrompt` yields `examples: 11`, `missing: []`, `wellFormed: 10` (the single non-`}`-terminated line is the deliberate WRONG FORMAT example at `:115`). `greaterThanOrEqualTo(8)` therefore tolerates deleting two examples. Prefer `expect(wellFormed.length, examples.length - 1)`, which states the design intent directly.

**S3 — rule numbering interrupted** (`lib:96-108`, 🟢): `1. 2. 3. 4.` → `FORBIDDEN:` block → `CRITICAL ROOT RULE: 5.`. Consider moving the new rule after `CRITICAL ROOT RULE`, or moving `FORBIDDEN:` below all numbered rules. Verified there are no stale references to the old numbering anywhere in the tree (`test/ai/registry/component_catalog_sync_test.dart:60,80` reads component names only).

**S4 — dependency pinning** (`pubspec.yaml:59-97`, 🟢): `pubspec.lock` is intentionally untracked (`.gitignore:76`), so "lock out of sync" does not apply here. Both git deps are pinned to tag `v2.34.4` consistently. Residual risk (🟡, project-level, not introduced here): git tags are movable and a stale local pub cache will not re-resolve; if CI does not verify the resolved commit, consider adding that check.

**S5 — pubspec rationale** (`pubspec.yaml:90,95`, 🟡): the v2.34.4 requirement is documented only in the PR/commit message. `git blame` on the pin shows just a version number, so a future downgrade for an unrelated fix would not see that the prompt depends on the fixed `containsA2UI()` message-start pattern. Add a one-line comment above the pin or in the `router_system_prompt.dart` header doc.

**S6 — test group naming** (`test:13`, 🟡): `group('A2UI envelope')` vs the `[ClassName] - [Feature]` convention → `RouterSystemPrompt - A2UI envelope`. Existing neighbours (`test/ai/ai_logging_test.dart:7,68`) do the same, so this is a nit.
</details>

<details>

1. **Layering is clean.** Changes are confined to `lib/ai/prompts/` and `test/ai/prompts/`; no cross-layer import, no `generated/` import outside `services/`, no USP provider / mutation-lock / `ServiceError` surface touched. Art. IV / V §5.4 / XIII / XI / XV are not implicated.
2. **No security surface.** No credential, token or secret introduced; nothing reads or writes storage. Only pre-existing illustrative `password` placeholders in the component catalogue prose remain (unchanged by this PR).
3. **Coupling handled honestly.** The `v2.34.4` bump and the prompt change are genuinely bidirectionally coupled (old `containsA2UI()` pattern rejects `{"version":...` → users would see raw JSONL), so shipping them in one PR is the right call rather than two independently-broken commits.
4. **Test lives in the mirrored path** (`test/ai/prompts/router_system_prompt_test.dart`), behaviour-named, no `TC001`-style numbering.
5. **Scope discipline** — 3 files, +103/−17, no drive-by lint churn or unrelated refactor.
6. **The negative example was updated too** (`lib:115`): the WRONG FORMAT sample now isolates exactly one mistake (newlines inside JSON) instead of differing in two ways. Easy detail to miss.
7. **The test documents its own cost model** (`test:4-11`): it states plainly that nothing in this app breaks today and that the cost is deferred, which stops a future reader from dismissing it as a tautology.
8. **Verification was real, not assumed** — mutation check (removing the field from one example turns the test red) and an end-to-end pass through the actual renderer + `RouterComponentRegistry`; the synthesizer independently re-ran the filter logic and reproduced `examples: 11 / missing: [] / wellFormed: 10`.
9. **Version bump is paired** — `ui_kit_library` and `generative_ui` both moved to `v2.34.4`, no drift between the two entries from the same repo.
</details>

Issue #759, opened 2026-03-24
fix: [Internet Settings] connection type incorrectly displays as Bridge after saving MTU, reverts to DHCP on re-entry


PeterJhongLinksys:
## Description

After saving an MTU change on the Internet Settings page, the displayed connection type changes to Bridge. However, navigating back and re-entering the page shows the connection type as DHCP again.

## Root Cause

The issue likely involves the interaction between `_saveWanSettings()` and the connection type detection logic:

1. **Save logic** (`lib/page/internet_settings/services/usp_internet_settings_service.dart:241`): When `typeChanged` is false (only MTU was edited), `addressingType` is sent as `null` (skipped). However, a separate `bridgeEnabled` write may still occur or the device may return an ambiguous state after the MTU update.

2. **Type detection** (`lib/page/internet_settings/models/usp_wan_connection_type.dart:16`): `fromRawFields()` determines Bridge mode when `addressingType` is empty AND `bridgeEnabled` is true:
   ```dart
   if (bridgeEnabled && addressingType.isEmpty) return bridge;
   ```
   If the post-save re-fetch returns an empty `addressingType` with `bridgeEnabled = true`, the type is incorrectly detected as Bridge.

3. **On re-entry**, a fresh fetch retrieves the correct device state showing DHCP, exposing the inconsistency as a transient display issue.

## Expected Behavior

Saving MTU should not alter the displayed connection type. The post-save state should correctly reflect the unchanged connection type.

PeterJhongLinksys:
[Root cause]
After saving non-type fields (e.g. MTU), the post-save re-fetch from the device transiently returns an empty `addressingType`. Combined with `bridgeEnabled` being always `true` on most routers (it controls the LAN-side L2 bridge, not WAN bridge mode), `fromRawFields()` misdetects the connection type as Bridge.

The underlying issue is that the current Bridge Mode detection logic (`addressingType` empty + `bridgeEnabled` true) is fragile, and the Bridge Mode write path is effectively a no-op (writes `addressingType = 'DHCP'` + `bridgeEnabled = true`, both of which are normal DHCP state).

[Solution]
Added a one-shot `_preservedConnectionType` guard in `UspInternetSettingsNotifier`:
- `performSave()`: when connection type was NOT changed, remembers the current type
- `performFetch()`: if the post-save fetch returns a different type (transient misdetection), preserves the known type and logs a warning
- Guard is consumed after one use to avoid affecting subsequent normal fetches

[...]

PR #1219, opened 2026-08-10
fix(usp): resolve #1184 back-fill regressions — conditional tunnel fetch, IPv6 partial SET, table/wildcard log noise

### LINKED ISSUE #1184: USP codegen: required-leaf-missing (9998) → ServiceErrorView is unreachable for concrete-path models (silenced by _rawGet null-backfill)

AustinChangLinksys:
Closes #1184.

## Summary

Follow-up to #1198, which removed `_rawGet`'s global back-fill to make the codegen required-leaf (9998) check reachable again. That correct fix exposed three latent regressions and two long-standing log-noise bugs on live routers. This PR fixes all of them.

### 1. Conditional WAN tunnel fetch — fixes false 9998 on absent GRE/L2TP tunnels
`GreTunnel` / `L2tpTunnel` are generated with **concrete** paths (`Device.GRE.Tunnel.1.*`), so with the back-fill gone their required leaves fire 9998 whenever the tunnel is absent — which is the normal case for DHCP/Static/PPPoE WANs.

Fixed by a two-phase fetch in `UspInternetSettingsService`: phase 1 fetches the always-present sources in parallel and derives the connection type; phase 2 fetches **only** the tunnel that the connection type actually uses (GRE for PPTP, L2TP for L2TP), and nothing for the others.

**Firmware contract dependency (PPTP/L2TP):** the phase-2 tunnel fetch trusts the firmware contract — a PPTP WAN guarantees `Device.GRE.Tunnel.1.*` and an L2TP WAN guarantees `Device.L2TPv2.Tunnel.1.*` exist once `PPP.Interface.LowerLayers` references them. If that contract were ever broken, `GreTunnel`/`L2tpTunnel.fetch` throws 9998 before the form is built (surfaced as a `ServiceError`), rather than being masked. This is a deliberate trade-off: we do not swallow the tunnel-absent case for PPTP/L2TP.

**Cost note:** PPTP/L2TP WANs now pay one extra serialized round trip (phase 2 runs after phase 1 resolves the connection type). DHCP/Static/PPPoE WANs are unaffected — they fetch no tunnel at all.

### 2. IPv6 settings SET now uses `allow_partial=true` — fixes 7005 save failure
IPv6 settings span three USP micro-services (`IP.Interface`, `DHCPv6.Client`, `IPv6rd.InterfaceSetting`). The OBUSPA broker rejects an atomic SET (`allow_partial=false`) that touches more than one service, returning **7005 "Allow partial=false not supported across more than one USP Service"** and failing the entire save.

A cross-service SET cannot be atomic on this router, so the IPv6 save now passes `allowPartial: true`. `_handleSetResult` still surfaces any per-parameter failure as a `UspPartialFailureError`, so nothing is silently swallowed.

### 3. Stop false "missing / empty" GET warnings for table & wildcard queries
Two false-positive log warnings on healthy routers:
- **Empty multi-instance tables** (table-query-only GETs with zero rows) logged `GET response EMPTY` at `warn`. Now classified via `isTableQueryOnlyRequest` and logged at `debug` — a zero-row table is a normal outcome.
- **Trailing-dot object/table GETs** (e.g. `...IPv6Address.`) were reported as missing paths, because the router expands them into instance paths and never echoes the requested key. Now skipped like wildcards. Wildcard (`*`) and trailing-dot (`.`) paths share one `_isTableExpandedPath` predicate.
- Per-path missing warnings are **aggregated** into one `missing N path(s)` line instead of flooding one line per absent leaf.

## Tests
- `usp_client_normalize_response_test.dart`: added `isTableQueryOnlyRequest` classification group (wildcard, trailing-dot, mixed, concrete-negative), trailing-dot skip, and aggregation-collection tests.
- `usp_internet_settings_service_test.dart`:
  - IPv6-carrying SET is issued with `allowPartial: true`.
  - **#1184 regression guard**: a DHCP WAN fetch issues NO GRE/L2TP tunnel GET (asserted directly on the captured GET paths, independent of the mock handler's fail-fast).
  - PPTP/L2TP fetch tests pass only the active tunnel response; the mock fail-fasts if the inactive tunnel is ever fetched.
- `pnp_service_test.dart`: updated GET-call counts for the conditional tunnel fetch (11 → 9 for PPPoE / Static IP); mock fail-fasts on unexpected tunnel GET.
- `flutter analyze` clean; full functional suite green.

## Review follow-ups (PR #1219)
- **Item 6 (DRY):** `connectionType` is derived once in `fetchSettings` and passed into `_buildForm`.
- **Item 1 (HIGH):** removed the misleading `?? ''` dead-code fallback in the PPTP/L2TP branches; see the firmware-contract note above.
- **Item 3:** added the explicit DHCP-WAN "no tunnel GET" regression guard.
- **Item 4:** documented the cross-service `allow_partial` invariant at the `instant_safety_service` single-service SET site.
- **Item 5:** noted 7005's second, broker-level meaning in the `usp_error.dart` fault-code table.

## Notes
The `ipv6settings` YAML contract currently matches live router state, so its concrete `.1.` paths are intentionally left as-is (contract holds; no speculative change).

### DIFF
--- a/lib/core/usp/services/usp_client.dart
+++ b/lib/core/usp/services/usp_client.dart
@@ -297,14 +297,16 @@ class UspClient {
           '${_prettyMap(rawMap)}');
 
       if (rawMap.isEmpty) {
-        if (isWildcardOnlyRequest(paths)) {
-          // A wildcard GET (e.g. Device.Firewall.DMZ.*) is expanded by the
-          // router across the discovered instances of a multi-instance table.
-          // When that table has zero instances the response is legitimately
-          // empty — the generated model iterates the (empty) instance set and
-          // returns an empty list, so this is a normal "no rows" outcome, not
-          // a fault. Log it at debug so it does not drown real warnings.
-          logger.d('$_tag$label GET empty (wildcard table, no instances): '
+        if (isTableQueryOnlyRequest(paths)) {
+          // A table query — a wildcard GET (e.g. Device.Firewall.DMZ.*) or an
+          // object/table path (e.g. Device.IP.Interface.1.IPv6Address.) — is
+          // expanded by the router across the discovered instances of a
+          // multi-instance table. When that table has zero instances the
+          // response is legitimately empty — the generated model iterates the
+          // (empty) instance set and returns an empty list, so this is a normal
+          // "no rows" outcome, not a fault. Log it at debug so it does not
+          // drown real warnings.
+          logger.d('$_tag$label GET empty (table query, no instances): '
               '$paths');
         } else {
           // At least one concrete path was requested — an empty response there
@@ -365,7 +367,7 @@ class UspClient {
       result[entry.key] = _coerceValue(entry.key, entry.value);
     }
     for (final path in paths) {
-      if (path.contains('*') || path.endsWith('.')) continue;
+      if (_isTableExpandedPath(path)) continue;
       if (!result.containsKey(path)) {
         onMissingPath?.call(path);
       }
@@ -373,18 +375,30 @@ class UspClient {
     return result;
   }
 
-  /// Whether every requested path is a wildcard query (contains '*').
+  /// Whether a requested path is one the router expands into concrete instance
+  /// paths, so the requested key itself never appears verbatim in the response.
   ///
-  /// A wildcard GET targets a multi-instance table and the router expands it
-  /// across the discovered instances. When that table has zero instances the
-  /// response is legitimately empty (the generated model returns an empty
-  /// list), so an empty response to a wildcard-only request is a normal
-  /// "no rows" outcome rather than a fault. Callers use this to decide whether
-  /// an empty GET response deserves a warning. An empty [paths] list is not a
-  /// wildcard request.
+  /// Two shapes qualify, and they must be treated identically everywhere:
+  /// - a wildcard search path (contains '*', e.g. Device.Firewall.DMZ.*)
+  /// - an object/table path (ends in '.', e.g. Device.IP.Interface.1.IPv6Address.)
+  ///
+  /// Both target a multi-instance set; an absent requested key is expected, not
+  /// missing, and an empty response means the set has zero rows — not a fault.
+  static bool _isTableExpandedPath(String path) =>
+      path.contains('*') || path.endsWith('.');
+
+  /// Whether every requested path is a table query the router expands
+  /// (see [_isTableExpandedPath]).
+  ///
+  /// When such a query hits a table with zero instances the response is
+  /// legitimately empty (the generated model returns an empty list), so an
+  /// empty response to a table-query-only request is a normal "no rows"
+  /// outcome rather than a fault. Callers use this to decide whether an empty
+  /// GET response deserves a warning. An empty [paths] list is not a table
+  /// query.
   @visibleForTesting
-  static bool isWildcardOnlyRequest(List<String> paths) =>
-      paths.isNotEmpty && paths.every((p) => p.contains('*'));
+  static bool isTableQueryOnlyRequest(List<String> paths) =>
+      paths.isNotEmpty && paths.every(_isTableExpandedPath);
 
   /// Coerce a raw string value from USP into the appropriate Dart type.
   /// - "true" / "false" →bool (any path)
--- a/lib/core/usp/services/usp_client.dart
+++ b/lib/core/usp/services/usp_client.dart
@@ -297,15 +297,37 @@ class UspClient {
           '${_prettyMap(rawMap)}');
 
       if (rawMap.isEmpty) {
-        logger.w('$_tag$label GET response EMPTY for paths: $paths');
+        if (isWildcardOnlyRequest(paths)) {
+          // A wildcard GET (e.g. Device.Firewall.DMZ.*) is expanded by the
+          // router across the discovered instances of a multi-instance table.
+          // When that table has zero instances the response is legitimately
+          // empty — the generated model iterates the (empty) instance set and
+          // returns an empty list, so this is a normal "no rows" outcome, not
+          // a fault. Log it at debug so it does not drown real warnings.
+          logger.d('$_tag$label GET empty (wildcard table, no instances): '
+              '$paths');
+        } else {
+          // At least one concrete path was requested — an empty response there
+          // is genuinely unexpected and worth a warning.
+          logger.w('$_tag$label GET response EMPTY for paths: $paths');
+        }
+        // Every requested path is absent, but the single warning above already
+        // says so. Skip the per-path missing warnings (they would just repeat
+        // the same paths — e.g. 9 lines for lan_network_info's concrete GET).
+        return normalizeGetResponse(paths, rawMap);
       }
 
-      return normalizeGetResponse(
-        paths,
-        rawMap,
-        onMissingPath: (path) =>
-            logger.w('$_tag$label GET missing path in response: "$path"'),
-      );
+      // Non-empty (possibly partial) response: collect any absent concrete
+      // leaves and emit ONE aggregated warning rather than one per path, so a
+      // genuine partial-response warning is not diluted into a wall of lines.
+      final missing = <String>[];
+      final normalized =
+          normalizeGetResponse(paths, rawMap, onMissingPath: missing.add);
+      if (missing.isNotEmpty) {
+        logger.w('$_tag$label GET missing ${missing.length} path(s) in '
+            'response: $missing');
+      }
+      return normalized;
     } catch (e) {
       sw.stop();
       final label = _idLabel(id);
@@ -327,9 +349,11 @@ class UspClient {
   ///    is the `?? ''` on each codegen assignment). Leaving the key absent lets
   ///    the required-leaf contract fire as designed (#1184).
   ///
-  /// Wildcard search paths (containing '*') are expanded by the router into
-  /// concrete instance paths, so the original wildcard path won't appear in the
-  /// response — those are skipped by the missing-path warning.
+  /// Wildcard search paths (containing '*') and object/table paths (ending in
+  /// '.', e.g. Device.IP.Interface.1.IPv6Address.) are expanded by the router
+  /// into concrete instance paths, so the original requested key never appears
+  /// in the response. Both are skipped by the missing-path warning — only a
+  /// requested *concrete leaf* that is absent is genuinely missing.
   @visibleForTesting
   static Map<String, dynamic> normalizeGetResponse(
     List<String> paths,
@@ -341,13 +365,26 @@ class UspClient {
       result[entry.key] = _coerceValue(entry.key, entry.value);
     }
     for (final path in paths) {
-      if (path.contains('*')) continue;
+      if (path.contains('*') || path.endsWith('.')) continue;
       if (!result.containsKey(path)) {
         onMissingPath?.call(path);
       }
     }
     return result;
   }
+
+  /// Whether every requested path is a wildcard query (contains '*').
+  ///
+  /// A wildcard GET targets a multi-instance table and the router expands it
+  /// across the discovered instances. When that table has zero instances the
+  /// response is legitimately empty (the generated model returns an empty
+  /// list), so an empty response to a wildcard-only request is a normal
+  /// "no rows" outcome rather than a fault. Callers use this to decide whether
+  /// an empty GET response deserves a warning. An empty [paths] list is not a
+  /// wildcard request.
+  @visibleForTesting
+  static bool isWildcardOnlyRequest(List<String> paths) =>
+      paths.isNotEmpty && paths.every((p) => p.contains('*'));
 
   /// Coerce a raw string value from USP into the appropriate Dart type.
   /// - "true" / "false" →bool (any path)
--- a/lib/page/internet_settings/services/usp_internet_settings_service.dart
+++ b/lib/page/internet_settings/services/usp_internet_settings_service.dart
@@ -131,15 +131,22 @@ class UspInternetSettingsService {
         .toList();
 
     // Resolve server address from the tunnel that fetchSettings' phase 2 loaded
-    // for this connection type. The tunnel is non-null by contract: pptp always
-    // fetches GRE.Tunnel.1, l2tp always fetches L2TPv2.Tunnel.1 (both concrete
-    // paths the firmware guarantees exist once LowerLayers references them). If
-    // that contract were ever violated, GreTunnel/L2tpTunnel.fetch would already
-    // have thrown 9998 upstream, so `!` here documents the invariant rather than
-    // guarding a reachable null.
+    // for this connection type. gre/l2tp are nullable because DHCP/Static/PPPoE
+    // WANs pass null — but on the pptp/l2tp branches the tunnel is non-null by
+    // contract: phase 2 fetched the matching tunnel, and GreTunnel/L2tpTunnel
+    // .fetch throws 9998 upstream if the concrete row is absent. The throw
+    // documents that invariant with a diagnosable message instead of a bare `!`
+    // NPE (which would surface only as "Null check operator used on a null
+    // value" with no context).
     final serverAddress = switch (connectionType) {
-      UspWanConnectionType.pptp => gre!.remoteEndpoints,
-      UspWanConnectionType.l2tp => l2tp!.remoteEndpoints,
+      UspWanConnectionType.pptp => (gre ??
+              (throw StateError('PPTP WAN but GRE tunnel absent — '
+                  'fetchSettings phase-2 contract violated')))
+          .remoteEndpoints,
+      UspWanConnectionType.l2tp => (l2tp ??
+              (throw StateError('L2TP WAN but L2TPv2 tunnel absent — '
+                  'fetchSettings phase-2 contract violated')))
+          .remoteEndpoints,
       _ => '',
     };
 
--- a/lib/page/internet_settings/services/usp_internet_settings_service.dart
+++ b/lib/page/internet_settings/services/usp_internet_settings_service.dart
@@ -46,28 +46,45 @@ class UspInternetSettingsService {
   // Fetch
   // ---------------------------------------------------------------------------
 
-  /// Fetch WAN, IPv6, PPP, VLAN, and tunnel settings in parallel.
+  /// Fetch WAN, IPv6, PPP, and VLAN settings in parallel, then the tunnel that
+  /// the resolved connection type actually uses.
   Future<InternetSettingsFetchResult> fetchSettings() async {
     try {
+      // Phase 1 — always needed, fetched in parallel.
       final results = await Future.wait([
         WanSettings.fetch(_usp),
         Ipv6Settings.fetch(_usp),
         PppInterface.fetch(_usp),
         VlanTermination.fetch(_usp),
-        GreTunnel.fetch(_usp),
-        L2tpTunnel.fetch(_usp),
         _fetchHostName(),
       ]);
       final wan = results[0] as WanSettings;
       final ipv6 = results[1] as Ipv6Settings;
       final ppp = results[2] as PppInterface;
       final vlan = results[3] as VlanTermination;
-      final gre = results[4] as GreTunnel;
-      final l2tp = results[5] as L2tpTunnel;
-      final hostName = results[6] as String;
+      final hostName = results[4] as String;
 
       final pppInstance = ppp.items.isNotEmpty ? ppp.items.first : null;
       final vlanInstance = vlan.items.isNotEmpty ? vlan.items.first : null;
+
+      // Phase 2 — tunnel is connection-type-specific. A DHCP/PPPoE/Static WAN
+      // has zero GRE/L2TP tunnel instances, so fetching them would trip the
+      // generated required-leaf check (code 9998) on an absent Tunnel.1. Only
+      // fetch the tunnel the current connection type actually uses.
+      final connectionType = UspWanConnectionType.fromRawFields(
+        addressingType: wan.addressingType,
+        lowerLayers: pppInstance?.lowerLayers ?? '',
+      );
+      GreTunnel? gre;
+      L2tpTunnel? l2tp;
+      switch (connectionType) {
+        case UspWanConnectionType.pptp:
+          gre = await GreTunnel.fetch(_usp);
+        case UspWanConnectionType.l2tp:
+          l2tp = await L2tpTunnel.fetch(_usp);
+        default:
+          break;
+      }
 
       return InternetSettingsFetchResult(
         form: _buildForm(wan, ipv6, pppInstance, vlanInstance, gre, l2tp),
@@ -101,8 +118,8 @@ class UspInternetSettingsService {
     Ipv6Settings ipv6,
     PppInterfaceInstance? ppp,
     VlanTerminationInstance? vlan,
-    GreTunnel gre,
-    L2tpTunnel l2tp,
+    GreTunnel? gre,
+    L2tpTunnel? l2tp,
   ) {
     // Split comma-separated DNS into 3 fields
     final dnsParts = wan.dnsServers
@@ -119,8 +136,8 @@ class UspInternetSettingsService {
 
     // Resolve server address from the appropriate tunnel
     final serverAddress = switch (connectionType) {
-      UspWanConnectionType.pptp => gre.remoteEndpoints,
-      UspWanConnectionType.l2tp => l2tp.remoteEndpoints,
+      UspWanConnectionType.pptp => gre?.remoteEndpoints ?? '',
+      UspWanConnectionType.l2tp => l2tp?.remoteEndpoints ?? '',
       _ => '',
     };
 
--- a/lib/page/internet_settings/services/usp_internet_settings_service.dart
+++ b/lib/page/internet_settings/services/usp_internet_settings_service.dart
@@ -87,7 +87,8 @@ class UspInternetSettingsService {
       }
 
       return InternetSettingsFetchResult(
-        form: _buildForm(wan, ipv6, pppInstance, vlanInstance, gre, l2tp),
+        form: _buildForm(
+            wan, ipv6, pppInstance, vlanInstance, gre, l2tp, connectionType),
         readOnlyInfo: _buildReadOnlyInfo(wan, pppInstance, hostName),
         pppInstancePath: pppInstance?.instancePath,
         vlanInstancePath: vlanInstance?.instancePath,
@@ -120,6 +121,7 @@ class UspInternetSettingsService {
     VlanTerminationInstance? vlan,
     GreTunnel? gre,
     L2tpTunnel? l2tp,
+    UspWanConnectionType connectionType,
   ) {
     // Split comma-separated DNS into 3 fields
     final dnsParts = wan.dnsServers
@@ -128,16 +130,16 @@ class UspInternetSettingsService {
         .where((s) => s.isNotEmpty)
         .toList();
 
-    final lowerLayers = ppp?.lowerLayers ?? '';
-    final connectionType = UspWanConnectionType.fromRawFields(
-      addressingType: wan.addressingType,
-      lowerLayers: lowerLayers,
-    );
-
-    // Resolve server address from the appropriate tunnel
+    // Resolve server address from the tunnel that fetchSettings' phase 2 loaded
+    // for this connection type. The tunnel is non-null by contract: pptp always
+    // fetches GRE.Tunnel.1, l2tp always fetches L2TPv2.Tunnel.1 (both concrete
+    // paths the firmware guarantees exist once LowerLayers references them). If
+    // that contract were ever violated, GreTunnel/L2tpTunnel.fetch would already
+    // have thrown 9998 upstream, so `!` here documents the invariant rather than
+    // guarding a reachable null.
     final serverAddress = switch (connectionType) {
-      UspWanConnectionType.pptp => gre?.remoteEndpoints ?? '',
-      UspWanConnectionType.l2tp => l2tp?.remoteEndpoints ?? '',
+      UspWanConnectionType.pptp => gre!.remoteEndpoints,
+      UspWanConnectionType.l2tp => l2tp!.remoteEndpoints,
       _ => '',
     };
 
--- a/lib/page/internet_settings/services/usp_internet_settings_service.dart
+++ b/lib/page/internet_settings/services/usp_internet_settings_service.dart
@@ -134,19 +134,23 @@ class UspInternetSettingsService {
     // for this connection type. gre/l2tp are nullable because DHCP/Static/PPPoE
     // WANs pass null — but on the pptp/l2tp branches the tunnel is non-null by
     // contract: phase 2 fetched the matching tunnel, and GreTunnel/L2tpTunnel
-    // .fetch throws 9998 upstream if the concrete row is absent. The throw
-    // documents that invariant with a diagnosable message instead of a bare `!`
-    // NPE (which would surface only as "Null check operator used on a null
-    // value" with no context).
+    // .fetch throws 9998 upstream (caught in fetchSettings → ServiceErrorView)
+    // if the concrete Tunnel.1 row is absent, so _buildForm is never reached
+    // with a null tunnel on the matching branch. The asserts document that
+    // invariant in dev; the `?? ''` keeps the (release-only, contract-violating)
+    // fallback graceful — an empty serverAddress fails form validation, which
+    // prompts the user to re-enter rather than surfacing a bare NPE.
+    assert(
+      connectionType != UspWanConnectionType.pptp || gre != null,
+      'PPTP WAN but GRE tunnel absent — fetchSettings phase-2 contract violated',
+    );
+    assert(
+      connectionType != UspWanConnectionType.l2tp || l2tp != null,
+      'L2TP WAN but L2TPv2 tunnel absent — fetchSettings phase-2 contract violated',
+    );
     final serverAddress = switch (connectionType) {
-      UspWanConnectionType.pptp => (gre ??
-              (throw StateError('PPTP WAN but GRE tunnel absent — '
-                  'fetchSettings phase-2 contract violated')))
-          .remoteEndpoints,
-      UspWanConnectionType.l2tp => (l2tp ??
-              (throw StateError('L2TP WAN but L2TPv2 tunnel absent — '
-                  'fetchSettings phase-2 contract violated')))
-          .remoteEndpoints,
+      UspWanConnectionType.pptp => gre?.remoteEndpoints ?? '',
+      UspWanConnectionType.l2tp => l2tp?.remoteEndpoints ?? '',
       _ => '',
     };
 
--- a/lib/page/internet_settings/services/usp_internet_settings_service.dart
+++ b/lib/page/internet_settings/services/usp_internet_settings_service.dart
@@ -556,6 +556,13 @@ class UspInternetSettingsService {
     UspInternetSettingsForm original,
     UspInternetSettingsForm edited,
   ) async {
+    // IPv6 settings span three USP services (IP.Interface, DHCPv6.Client,
+    // IPv6rd.InterfaceSetting). The OBUSPA broker rejects an atomic SET
+    // (allow_partial=false) that touches more than one service — it returns
+    // 7005 "Allow partial=false not supported across more than one USP
+    // Service", failing the whole save. A cross-service SET cannot be atomic on
+    // this router, so allow partial application; _handleSetResult still surfaces
+    // any per-parameter failure as a UspPartialFailureError.
     _handleSetResult(await Ipv6Settings.update(
       _usp,
       ipv6Enabled: _diff(original.ipv6Enabled, edited.ipv6Enabled),
@@ -566,6 +573,7 @@ class UspInternetSettingsService {
           _diff(original.ipv6rdIpv4MaskLength, edited.ipv6rdIpv4MaskLength),
       ipv6rdBorderRelay:
           _diff(original.ipv6rdBorderRelay, edited.ipv6rdBorderRelay),
+      allowPartial: true,
     ));
   }
 
--- a/test/page/instant_setup/services/pnp_service_test.dart
+++ b/test/page/instant_setup/services/pnp_service_test.dart
@@ -93,15 +93,14 @@ void main() {
       if (paths.any((p) => p.contains('VLANTermination'))) {
         return vlanResponse;
       }
+      // PnP setup never configures a PPTP/L2TP WAN, so connectionType is never
+      // PPTP/L2TP and Phase-2 fetches no tunnel. A tunnel GET here means an
+      // unconditional fetch regressed — fail fast instead of returning empty.
       if (paths.any((p) => p.contains('GRE.Tunnel'))) {
-        return <String, dynamic>{
-          'Device.GRE.Tunnel.1.RemoteEndpoints': '',
-        };
+        throw StateError('Unexpected GRE.Tunnel GET in PnP fetch');
       }
       if (paths.any((p) => p.contains('L2TPv2.Tunnel'))) {
-        return <String, dynamic>{
-          'Device.L2TPv2.Tunnel.1.RemoteEndpoints': '',
-        };
+        throw StateError('Unexpected L2TPv2.Tunnel GET in PnP fetch');
       }
       return {};
     });
@@ -306,15 +305,13 @@ void main() {
         if (paths.any((p) => p.contains('VLANTermination'))) {
           return vlanExistingResponse;
         }
+        // PPPoE connection → no tunnel fetch in Phase-2; a tunnel GET is a
+        // regression (unconditional fetch).
         if (paths.any((p) => p.contains('GRE.Tunnel'))) {
-          return <String, dynamic>{
-            'Device.GRE.Tunnel.1.RemoteEndpoints': '',
-          };
+          throw StateError('Unexpected GRE.Tunnel GET in PnP PPPoE fetch');
         }
         if (paths.any((p) => p.contains('L2TPv2.Tunnel'))) {
-          return <String, dynamic>{
-            'Device.L2TPv2.Tunnel.1.RemoteEndpoints': '',
-          };
+          throw StateError('Unexpected L2TPv2.Tunnel GET in PnP PPPoE fetch');
         }
         return {};
       });
@@ -377,15 +374,13 @@ void main() {
         if (paths.any((p) => p.contains('VLANTermination'))) {
           return vlanDisabledResponse;
         }
+        // PPPoE connection → no tunnel fetch in Phase-2; a tunnel GET is a
+        // regression (unconditional fetch).
         if (paths.any((p) => p.contains('GRE.Tunnel'))) {
-          return <String, dynamic>{
-            'Device.GRE.Tunnel.1.RemoteEndpoints': '',
-          };
+          throw StateError('Unexpected GRE.Tunnel GET in PnP PPPoE fetch');
         }
         if (paths.any((p) => p.contains('L2TPv2.Tunnel'))) {
-          return <String, dynamic>{
-            'Device.L2TPv2.Tunnel.1.RemoteEndpoints': '',
-          };
+          throw StateError('Unexpected L2TPv2.Tunnel GET in PnP PPPoE fetch');
         }
         return {};
       });
--- a/test/page/instant_setup/services/pnp_service_test.dart
+++ b/test/page/instant_setup/services/pnp_service_test.dart
@@ -202,14 +202,14 @@ void main() {
       // - Ipv6Settings._resolveInstance() + fetch() = 2 calls
       // - PppInterface.fetch() = 1 call
       // - VlanTermination.fetch() = 1 call
-      // - GreTunnel.fetch() = 1 call
-      // - L2tpTunnel.fetch() = 1 call
       // - _fetchHostName() (Device.DeviceInfo.HostName) = 1 call
+      // No GRE/L2TP fetch: tunnels are fetched only for the pptp/l2tp
+      // connection types, not for Static IP.
       // saveAll:
       // - WanStaticIp.updateOrdered() → _resolveInstance() = 1 call
       // - Ipv6Settings.update() → _resolveInstance() = 1 call
-      // Total = 11 get calls
-      verify(() => mockUsp.get(any())).called(11);
+      // Total = 9 get calls
+      verify(() => mockUsp.get(any())).called(9);
 
       // Verify setOrdered was called for Static IP mode switch
       final capturedOrdered = verify(() => mockUsp.setOrdered(captureAny(),
@@ -246,13 +246,14 @@ void main() {
       await service.saveIspSettings(config);
 
       // Verify fetchSettings + saveAll get calls:
-      // fetchSettings: 9 calls (WanSettings x2, Ipv6 x2, PPP, VLAN, GRE, L2TP,
-      //   _fetchHostName = Device.DeviceInfo.HostName)
+      // fetchSettings: 7 calls (WanSettings x2, Ipv6 x2, PPP, VLAN,
+      //   _fetchHostName = Device.DeviceInfo.HostName). No GRE/L2TP fetch:
+      //   tunnels are fetched only for the pptp/l2tp connection types.
       // saveAll:
       // - WanPppoe.update() → _resolveInstance() = 1 call
       // - Ipv6Settings.update() → _resolveInstance() = 1 call
-      // Total = 11 get calls
-      verify(() => mockUsp.get(any())).called(11);
+      // Total = 9 get calls
+      verify(() => mockUsp.get(any())).called(9);
 
       // Verify PppInterface.add was called (new PPP instance created)
       final addCaptures = verify(() => mockUsp.add(captureAny())).captured;
--- a/lib/core/usp/errors/usp_error.dart
+++ b/lib/core/usp/errors/usp_error.dart
@@ -140,7 +140,7 @@ UspError? parseUspError(Object error) {
 /// | code | meaning                          | → ServiceError          |
 /// |------|----------------------------------|-------------------------|
 /// | 7004 | parameter not writable           | InvalidInputError       |
-/// | 7005 | invalid parameter name           | InvalidInputError       |
+/// | 7005 | invalid parameter name¹          | InvalidInputError       |
 /// | 7006 | invalid parameter value          | InvalidInputError       |
 /// | 7026 | parameter (path) not found       | ResourceNotFoundError   |
 /// | 7027 | object not found                 | ResourceNotFoundError   |
@@ -148,6 +148,15 @@ UspError? parseUspError(Object error) {
 /// | 9005 | bbfdm: invalid/unimplemented param | ResourceNotFoundError |
 /// | 9007 | bbfdm: (resource not found)      | ResourceNotFoundError   |
 /// | 9008 | bbfdm: non-writable parameter    | InvalidInputError       |
+///
+/// ¹ 7005 has a second, broker-level meaning on OBUSPA: an atomic SET
+/// (`allow_partial=false`) that spans more than one USP micro-service is
+/// rejected with 7005 "Allow partial=false not supported across more than one
+/// USP Service". This is NOT an invalid parameter name — it is a cross-service
+/// atomicity limit. The fix is on the caller (pass `allowPartial: true` for
+/// multi-service SETs), not this mapping; InvalidInputError remains a reasonable
+/// surface if it ever reaches here. See `_saveIpv6Settings` in
+/// usp_internet_settings_service.dart.
 ///
 /// ### Source 3 — Dart codegen (`lib/generated/*.g.dart`), NOT from Rust
 ///
--- a/lib/page/instant_safety/services/instant_safety_service.dart
+++ b/lib/page/instant_safety/services/instant_safety_service.dart
@@ -46,6 +46,13 @@ class UspInstantSafetyService {
   Future<void> save(SafeBrowsingType type) async {
     try {
       final dnsValue = dnsValueForType(type);
+      // Uses the default allow_partial=false (atomic). Safe here because this
+      // SET touches a single USP micro-service (DHCPv4.Server.Pool DNSServers).
+      // The OBUSPA broker only rejects atomic SETs that span more than one
+      // service with 7005 — see _saveIpv6Settings in
+      // usp_internet_settings_service.dart, which must relax this. Keep this SET
+      // single-service; if it ever grows to touch another service, switch to
+      // allowPartial: true.
       final result = await LanNetworkInfo.update(_usp, dnsServers: dnsValue);
       final parsed = UspResultParser.parseSetResult(result);
       switch (parsed) {
--- a/lib/page/instant_safety/services/instant_safety_service.dart
+++ b/lib/page/instant_safety/services/instant_safety_service.dart
@@ -46,13 +46,10 @@ class UspInstantSafetyService {
   Future<void> save(SafeBrowsingType type) async {
     try {
       final dnsValue = dnsValueForType(type);
-      // Uses the default allow_partial=false (atomic). Safe here because this
-      // SET touches a single USP micro-service (DHCPv4.Server.Pool DNSServers).
-      // The OBUSPA broker only rejects atomic SETs that span more than one
-      // service with 7005 — see _saveIpv6Settings in
-      // usp_internet_settings_service.dart, which must relax this. Keep this SET
-      // single-service; if it ever grows to touch another service, switch to
-      // allowPartial: true.
+      // Single-service SET → the default atomic allow_partial is safe here. If
+      // this ever adds a param from another USP service, switch to
+      // allowPartial: true — OBUSPA rejects cross-service atomic SETs with 7005
+      // (see usp_error.dart).
       final result = await LanNetworkInfo.update(_usp, dnsServers: dnsValue);
       final parsed = UspResultParser.parseSetResult(result);
       switch (parsed) {
--- a/test/core/usp/services/usp_client_normalize_response_test.dart
+++ b/test/core/usp/services/usp_client_normalize_response_test.dart
@@ -77,6 +77,24 @@ void main() {
       expect(missing, isEmpty);
     });
 
+    test(
+        'trailing-dot object/table request paths never trigger a missing-path '
+        'warning', () {
+      final missing = <String>[];
+      // A path ending in '.' is an object/table GET (e.g. IPv6Address.); the
+      // router expands it into instance paths, so the requested key itself is
+      // absent from the response and must NOT be reported as missing.
+      UspClient.normalizeGetResponse(
+        ['Device.IP.Interface.1.IPv6Address.'],
+        <String, String?>{
+          'Device.IP.Interface.1.IPv6Address.1.IPAddress': '::1',
+        },
+        onMissingPath: missing.add,
+      );
+
+      expect(missing, isEmpty);
+    });
+
     test('absent concrete path invokes onMissingPath; present one does not',
         () {
       final missing = <String>[];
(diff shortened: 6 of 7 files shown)


AustinChangLinksys (commented):
🤖 **Automated Review — Round 1** · `1175273..267b9dc` (full)

**Verdict: 💬 Self-review (comment only)** — Austin is the author; GitHub prohibits self-approval. Review is provided as a reference to catch blind spots — 0 Critical, 4 Warnings, 7 Suggestions across both axes.

#### Standards
|---|---|---|---|
| ⚠️ | Med | `usp_internet_settings_service.dart:74–92` | `connectionType` derived twice independently — Phase-2 fetch and `_buildForm` each call `fromRawFields`; stale `lowerLayers` on PPTP/L2TP silently yields empty `serverAddress` |
| ⚠️ | Med | `usp_internet_settings_service.dart:555–577` | `allowPartial: true` cross-service SET: OBUSPA may silently apply partial results without reporting `UspPartialSuccess`; IPv6 state can be half-written with `UspSuccess` returned |
| ⚠️ | High | `usp_internet_settings_service_test.dart:131–138` | [Duplicated Code] Dead GRE/L2TP mock branches in `createFetchMockHandler` — never reached under Phase-2 conditional fetch; weakens regression guard |
| ⚠️ | High | `pnp_service_test.dart:96–105` | [Duplicated Code] Same dead GRE/L2TP mock branches in `setupFetchMocks` — same pattern as above (two test files, one structural smell) |
| 💡 | High | `usp_client.dart:387–391` | `isWildcardOnlyRequest` checks `p.contains('*')` only — trailing-dot paths (e.g. `Device.IP.Interface.1.IPv6Address.`) not classified as wildcard-only, so empty-table GET still emits spurious EMPTY warning |
| 💡 | High | `usp_client.dart` (`isWildcardOnlyRequest`) | `@visibleForTesting` one-liner seam is justified but should document why it exists (compensates for `_rawGet` untestability) |
| 💡 | Med | `usp_internet_settings_service.dart` (`_buildForm`) | Implicit caller invariant: nullable `gre`/`l2tp` silently yields `''` if Phase-2 ever forgets to fetch — consider `assert` or pre-resolved `serverAddress` parameter |
| 💡 | Med | `usp_internet_settings_service.dart` (Phase-2 switch) | `default: break` in tunnel-fetch switch bypasses Dart exhaustiveness check — future new `UspWanConnectionType` case silently skipped |
| 💡 | Low | `usp_internet_settings_service.dart` (`_saveIpv6Settings`) | `allowPartial: true` constraint only in prose comment; future callers of `Ipv6Settings.update` may omit it and re-introduce 7005 |
| 💡 | High | `usp_internet_settings_service.dart:566` | `_resolveInstance` GET fires unconditionally before `params.isEmpty` short-circuit — extra GET on every `saveAll` that doesn't touch IPv6 (pre-existing, now more visible) |

#### Spec
|---|---|---|---|
| ⚠️ | Med | (absence) | Issue #1184 cites a skipped E2E test for `lan_network_info` ServiceErrorView as the motivation; no artifact in this PR re-enables or references it — closure of #1184 may be incomplete |
| 💡 | Low | `usp_internet_settings_service_test.dart:903` | No `verifyNever` assertion for conditional fetch — PPTP test verifies `serverAddress` but not that `L2tpTunnel.fetch` was never called (and vice versa) |

<details>

**W-1 — Double-derived `connectionType` (Med)** · `usp_internet_settings_service.dart:74–92`

`fetchSettings()` calls `UspWanConnectionType.fromRawFields(addressingType, lowerLayers: pppInstance?.lowerLayers ?? '')` in Phase 2 to decide which tunnel to fetch, and `_buildForm` independently re-derives the same value for `serverAddress`. If `lowerLayers` is transiently empty on a PPTP/L2TP device (e.g. `PppInterface` returned before `LowerLayers` is set), both calls resolve to `pppoe` — Phase 2 skips the tunnel fetch, and `_buildForm` falls through to `_ => ''`, rendering `serverAddress` as an empty string silently.

**Fix:** Pass the already-computed `connectionType` into `_buildForm` as a parameter instead of re-deriving it, so both uses agree by construction.

**W-2 — `allowPartial: true` silent-partial OBUSPA gap (Med)** · `usp_internet_settings_service.dart:555–577`

`_handleSetResult` correctly throws on explicit `UspPartialSuccess`. The gap: `allowPartial: true` means the OBUSPA broker _may_ apply some parameters and not surface failures as `UspPartialSuccess` on certain firmware versions. In such cases `_handleSetResult` sees `UspSuccess` and passes, while IPv6 state is inconsistently half-written. This is inherent to the protocol; the PR's fix is correct. Suggest a comment or follow-up monitoring noting that silent partials are accepted as a known limitation of this router's OBUSPA version.

**W-3 — Dead mock branches in `createFetchMockHandler` (High)** · `usp_internet_settings_service_test.dart:131–138`

```dart
// These branches can never be reached when using the default pppResponse
// (pppInstance?.lowerLayers == '' → connectionType resolves to pppoe → no tunnel fetch)
if (paths.any((p) => p.contains('GRE.Tunnel'))) { return greResponse; }
if (paths.any((p) => p.contains('L2TPv2.Tunnel'))) { return l2tpResponse; }
```

A regression re-introducing unconditional tunnel fetches would silently succeed instead of failing.

**Fix:** Remove the dead branches from the default mock handler (keep them only in the PPTP/L2TP override variants). Add a throwing fallback so unexpected tunnel `get()` calls fail fast.

**W-4 — Same dead-mock pattern in `pnp_service_test.dart` (High)** · `pnp_service_test.dart:96–105`

Same dead GRE/L2TP branches in `setupFetchMocks` — call-count assertions were correctly updated (11→9), but the mock handler still silently accepts tunnel calls that can no longer arrive. Two files with the same structural issue = [Duplicated Code] smell worth consolidating.

**Fix:** Same as W-3 — remove the dead branches; add a throwing fallback for unexpected calls.

**Spec W1 — Skipped E2E test for `lan_network_info` not re-enabled (Med)** · (absence)

Issue #1184's stated motivation: a `ServiceErrorView` (9998) E2E test for `lan_network_info` was skipped because the branch was unreachable. This PR closes #1184 but includes no artifact re-enabling that test. The `S06-local-network-states.spec.ts` test likely lives in `PrivacyGUI-USP-E2E` repo — confirm in a comment on #1184 whether (a) it will be re-enabled in a separate PR, (b) deferred with rationale, or (c) tracked elsewhere.

</details>

<details>

- **`_rawGet` no-backfill contract preserved** — `normalizeGetResponse` correctly leaves absent concrete keys out of the result; 9998 path is reachable as designed (core #1184 fix from #1198, verified still intact).
- **Conditional tunnel fetch Phase-2 logic** — `switch (connectionType)` with `GreTunnel.fetch` / `L2tpTunnel.fetch` only for PPTP/L2TP; `gre?.remoteEndpoints ?? ''` null-safe access in `_buildForm` prevents crashes.
- **`allowPartial: true` wire-up end-to-end** — `Ipv6Settings.update` correctly threads `allowPartial` to `client.set()`; `_handleSetResult` throws `UspPartialFailureError` on explicit partial failures — nothing silently swallowed.
- **`normalizeGetResponse` trailing-dot fix** — `path.endsWith('.')` guard correctly suppresses false missing-path warnings for object/table GETs.
- **`isWildcardOnlyRequest` empty-list guard** — `paths.isNotEmpty &&` correctly returns `false` for empty list.
- **Test call-count updates in `pnp_service_test`** — 11 to 9 for Static IP / PPPoE is arithmetically correct for the Phase-2 conditional fetch.
- **IPv6 regression test** — `usp_internet_settings_service_test.dart` correctly captures all `mockUsp.set()` calls and asserts IPv6-carrying SETs use `allowPartial: true`.
- **All three PR-described regressions addressed per spec** — conditional tunnel fetch (fix A), `allow_partial=true` IPv6 (fix B), wildcard/trailing-dot log noise (fix C) all verified in code and tests.
- **Layer / Provider architecture clean** — no layer violations, no Provider hard-rule violations in diff.
</details>

*Cross-reviewed by two independent agents (security+correctness / architecture+maintainability) plus one spec-compliance agent. Automated — please sanity-check before merge.*

AustinChangLinksys (commented):
🤖 **Automated Review — Round 2** · `267b9dc..5c1ed78` (incremental)

**Verdict: 💬 Self-review (comment only)** — Austin is the author; GitHub prohibits self-approval. Review is provided as a reference to catch blind spots — 0 Critical, 1 Warning, 4 Suggestions.

#### Standards
|---|---|---|---|
| ⚠️ | High | `usp_internet_settings_service_test.dart:230–255` | [Duplicated Code] DNS-split test's inline mock duplicates `createFetchMockHandler` dispatch logic; PR added StateError guards to both instead of consolidating into factory |
| 💡 | High | `usp_client_normalize_response_test.dart` | Missing `isTableQueryOnlyRequest` test case for `[trailing-dot, concrete] -> false` (mirror of the renamed function's new shape) |
| 💡 | High | `pnp_service_test.dart:102` vs `312,381` | Minor: first PnP scenario uses `'in PnP fetch'`; PPPoE scenarios use `'in PnP PPPoE fetch'` — inconsistent but diagnostic (acceptable as-is) |
| 💡 | High | `usp_client.dart:387` | `_isTableExpandedPath` is the canonical predicate definition but has no direct test; tested indirectly through two seams — note for future maintainers adding a third shape |

*[X] = 1 agent · [XX] = 2 agents flagged independently -> higher confidence (Critical only).*

#### Spec
|---|---|---|---|
| ⚠️ | High | (absence) `PrivacyGUI-USP-E2E` | spec#Related -> E2E repo: #1184 open question now resolved (option b chosen), but `S06-local-network-states.spec.ts` skip note still not re-enabled/updated per spec's stated intent |
| 💡 | High | issue #1184 | Option (a) vs (b) open question was implicitly resolved by impl; never explicitly closed/commented on the issue |

*[X] = 1 agent · [XX] = 2 agents flagged independently -> higher confidence (Critical only).*

<details>
<summary>Round 1 findings status (carry-over tracking)</summary>

| Round 1 ID | Description | Status in Round 2 |
|---|---|---|
| W-3 | Dead GRE/L2TP mock branches in `pnp_service_test.dart` [Duplicated Code] High | Resolved — replaced with fail-fast `StateError` in all three PnP scenarios |
| W-4 | Dead GRE/L2TP mock branches in `usp_internet_settings_service_test.dart` [Duplicated Code] High | Resolved — `createFetchMockHandler` converted to nullable params + StateError guards; unused fixtures dropped |
| W-1 | Double-derived `connectionType` in `usp_internet_settings_service.dart:74–92` Med | Not in Round 2 diff — pre-existing, carry-over |
| W-2 | `allowPartial: true` silent-partial OBUSPA gap Med | Not in Round 2 diff (fix landed in base) — carry-over |
| Spec-W1 | Skipped E2E test for `lan_network_info` error-state not re-enabled | Persists — see Spec table above |

</details>

<details>

**W-1 (Standards) — [Duplicated Code] DNS-split test duplicates mock dispatch logic (High)**
`test/page/internet_settings/services/usp_internet_settings_service_test.dart:230–255`

The DNS-split scenario uses an inline `thenAnswer` that reimplements the same path-matching if-cascade as `createFetchMockHandler`: alias resolution, tunnel fail-fast guards, AddressingType, PPP.Interface, etc. The Round 2 diff correctly added StateError throws for unexpected GRE.Tunnel / L2TPv2.Tunnel GETs, but did so in **both** the factory helper (`createFetchMockHandler`) **and** this inline handler separately, rather than converting the test to use the factory.

**Why it matters:** If a new fetch path is added to the service in the future, a developer must remember to update both sites. The W-3/W-4 fix was aimed at exactly this class of problem; the inline mock is the same smell left behind.

**Fix:** Refactor the DNS-split test to use `createFetchMockHandler(pppResponse: const {}, ...)` (or equivalent empty/overridden PPP response) so future path additions only need updating in one place.

**Spec-W1 — E2E skip not re-enabled/updated (High)**
`PrivacyGUI-USP-E2E tests/S06-local-network-states.spec.ts` (separate repo, not in diff)

Issue #1184's "Related" section explicitly states: the S06 Local Network error-state test is skipped pending this decision. Skip note in `PrivacyGUI-USP-E2E tests/S06-local-network-states.spec.ts` will be updated to reference this issue.

The open question has now been resolved: option (b) was chosen — `_rawGet`'s `putIfAbsent(path, () => null)` backfill was removed (landed in base branch), and the no-backfill contract is pinned by regression tests in `usp_client_normalize_response_test.dart`. The blocker the skip cited no longer exists. This PR closes #1184 but does not update or re-enable the E2E test.

**Fix:** In `PrivacyGUI-USP-E2E`, either re-enable `S06-local-network-states.spec.ts`'s skipped error-state test or update its skip annotation to note the decision has been made and point to the relevant commits.

</details>

<details>

- **`_isTableExpandedPath` extraction is correct and DRY** — the single predicate (`contains('*') || endsWith('.')`) now used consistently by both `normalizeGetResponse` (missing-path skip) and `isTableQueryOnlyRequest` (empty-GET classifier), eliminating the pre-existing asymmetry that was the root bug of the trailing-dot false warning.
- **`isWildcardOnlyRequest` -> `isTableQueryOnlyRequest` rename** — more accurate, reflects the new trailing-dot classification. No dangling call sites found.
- **Fail-fast StateError mocks** — all three PnP scenarios and the service test's `createFetchMockHandler` now throw loudly on unexpected tunnel GETs. A regression reintroducing unconditional tunnel fetches will fail immediately rather than silently returning empty data.
- **Unused fixture cleanup** — `_greEmptyResponse` / `_l2tpEmptyResponse` correctly dropped from `usp_internet_settings_service_test.dart`.
- **StateError propagation correctness** — throw inside an async body is correctly captured as a rejected Future by Dart's async machinery; behaviorally equivalent to `thenThrow`.
- **Test coverage for new shapes** — `isTableQueryOnlyRequest` gets explicit tests for trailing-dot-only, mixed wildcard+trailing-dot, and the rename; `normalizeGetResponse` trailing-dot skip is tested.
- **No new flutter analyze issues** — all 57 affected tests pass per author's commit message and Reviewer A verification.
- **W-1/W-2 (Round 1) unchanged** — not in this diff's scope; no regression observed in those areas.
</details>

AustinChangLinksys (commented):
🤖 **Automated Review — Round 3** · `267b9dc..5c1ed78` (incremental)

**Verdict: 💬 Self-review (comment only)** — Austin is the author; GitHub prohibits self-approval. Review provided as a reference to catch blind spots.

> **Context note:** HankYuLinksys posted a thorough human review between rounds. This automated review independently verified all 6 of Hank's findings against the actual code; findings below reflect that cross-check. Hank's **Item 2 (trailing-dot warning)** is **not valid** against the current head — the fix is complete and correct (see refutals section). Items 1, 3–6 are confirmed at varying confidence levels.

---

#### Standards

|---|---|---|---|
| ⚠️ | High | `usp_internet_settings_service.dart:80–87` + `gre_tunnel.g.dart:31` | PPTP/L2TP Phase-2 fetch throws 9998 if `Tunnel.1` row absent; `gre?.remoteEndpoints ?? ''` fallback in `_buildForm:139` is structurally dead code |
| ⚠️ | High | `usp_internet_settings_service.dart:99` + `gre_tunnel.g.dart:31` + `usp_error.dart:181` | Absent Tunnel.1 on PPTP/L2TP maps to misleading `InvalidInputError` (9998 "Required fields missing") — firmware fault surfaced as user-input error |
| ⚠️ | High | `usp_internet_settings_service_test.dart` (absent test) | No regression test asserting DHCP/Static/PPPoE `fetchSettings()` never issues any tunnel GET (PnP tests guard it; internet-settings suite does not) |
| ⚠️ | Med | `instant_safety_service.dart:49` + `lan_network_info.g.dart:136` | [Shotgun Surgery] Cross-service SET invariant (must use `allowPartial: true`) lives only in prose comment; no systemic guard — next cross-service call will silently regress with mysterious 7005 |
| ⚠️ | High | `usp_error.dart:143` + `usp_internet_settings_service.dart:562` | [Shotgun Surgery] 7005 gains second OBUSPA meaning ("cross-service allow_partial=false"); error contract table + `service_error_localizations.dart` not updated |
| ⚠️ | High | `usp_internet_settings_service.dart:74,132` | [Duplicate Code] `UspWanConnectionType.fromRawFields(...)` computed twice with identical inputs — `fetchSettings()` then `_buildForm()`; pass computed value in to eliminate re-derivation |
| 💡 | High | `usp_client.dart:387–401` | Hank H-2 REFUTED: `_isTableExpandedPath` / `isTableQueryOnlyRequest` already handle `endsWith('.')` correctly; no action needed |

---

#### Spec

|---|---|---|---|
| ✅ | High | `PrivacyGUI-USP-E2E tests/S06-local-network-states.spec.ts` | Spec-W1 RESOLVED — E2E skip removed in commit `f691d369` (2026-08-08), test re-implemented and active |
| ⚠️ | High | spec#PR description "Closes #1184" → `usp_internet_settings_service.dart:80–87` | PR claims to fix false 9998 on absent GRE/L2TP tunnels; PPTP/L2TP Phase-2 path can still hit 9998 if firmware hasn't provisioned Tunnel.1 — firmware guarantee not confirmed |
| ⚠️ | High | spec → issue #1184 (`state: open`) | Issue #1184 still open; option (b) decision was never recorded as a comment on the issue — will auto-close on merge with the rationale buried in PR #1198 body only |
| 💡 | High | spec#"closes #1184" → `usp_internet_settings_service_test.dart` | No test for PPTP-with-absent-GRE-row path (complement to Hank H-3); spec commitment for PPTP/L2TP defensive handling is untested |
| 💡 | High | `usp_client.dart:387–401` + test file | Trailing-dot `_isTableExpandedPath` extraction is spec-correct; minor gap: no `[trailing-dot, concrete] → false` negative test |

---

<details>

**W-1 — PPTP/L2TP Phase-2 fetch throws 9998 on absent Tunnel.1; `?? ''` fallback is dead code (High) — confirmed by Reviewers A+B+C**

`lib/page/internet_settings/services/usp_internet_settings_service.dart:80–87, 139–140`
`lib/generated/gre_tunnel.g.dart:27–31`

Phase-2 fetch at `usp_internet_settings_service.dart:82`:
```dart
case UspWanConnectionType.pptp:
  gre = await GreTunnel.fetch(_usp);   // throws 9998 if Tunnel.1 absent
```
`gre_tunnel.g.dart:27–31`:
```dart
if (!response.containsKey('Device.GRE.Tunnel.1.RemoteEndpoints')) {
  missing.add('Device.GRE.Tunnel.1.RemoteEndpoints');
}
if (missing.isNotEmpty) {
  throw 'Get failed: Validation error: Required fields missing ... (code: 9998)';
}
```
`_buildForm:139` uses `gre?.remoteEndpoints ?? ''` — but if `GreTunnel.fetch()` throws, the catch at line 99 rethrows before `_buildForm` is ever called; if it succeeds, `gre` is non-null and `remoteEndpoints` is `final String` (never null). The `?? ''` fallback is structurally unreachable for the PPTP branch in both outcomes.

**Fix:** Either (a) confirm with firmware team that `Device.GRE.Tunnel.1` is guaranteed to exist when `LowerLayers` references it and add a comment documenting that assumption, or (b) wrap the Phase-2 fetch in a try/catch that returns `null` on 9998 so `?? ''` does real work and surfaces a user-friendly validation message.

---

**W-2 — Absent Tunnel.1 maps to misleading `InvalidInputError` (High)**

`lib/core/usp/errors/usp_error.dart:181–183`
`lib/page/internet_settings/services/usp_internet_settings_service.dart:99`

```dart
// usp_internet_settings_service.dart:99–101
} catch (e) {
  throw mapUspErrorToServiceError(e);
}
// usp_error.dart:181–183
UspErrorCategory.validation =>
  InvalidInputError(code: parsed.faultCode, detail: parsed.message),
```
A 9998 from an absent Tunnel.1 maps to `InvalidInputError("Required fields missing from response: Device.GRE.Tunnel.1.RemoteEndpoints")`. That is a firmware/provisioning fault, not a user-input error. Surfaced via `service_error_localizations.dart:73` it shows `errorInvalidInput` — misleading.

**Fix:** Catch 9998 specifically in the PPTP/L2TP Phase-2 arms and rethrow as a more appropriate error with a firmware-context message.

---

**W-3 — No internet-settings regression test for non-PPTP/L2TP "never fetches tunnel" invariant (High)**

`test/page/internet_settings/services/usp_internet_settings_service_test.dart` (absent)

PnP tests guard this with fail-fast `StateError`; internet-settings service tests do not include a DHCP/Static/PPPoE `fetchSettings()` call where a GRE/L2TP GET would throw. A regression reintroducing an unconditional Phase-2 fetch for DHCP WAN would not be caught by the internet-settings suite.

**Fix:** Add a test case using the default `createFetchMockHandler()` (null greResponse + null l2tpResponse) with a DHCP-WAN config; success with no StateError pins the invariant.

---

**W-4 — Cross-service SET invariant not encoded; safe by coincidence (Med)**

`lib/page/instant_safety/services/instant_safety_service.dart:49`
`lib/generated/lan_network_info.g.dart:136`

```dart
// instant_safety_service.dart:49 — uses default allowPartial: false
final result = await LanNetworkInfo.update(_usp, dnsServers: dnsValue);

// lan_network_info.g.dart:136
bool allowPartial = false,  // default
```
Currently safe because the call writes one param on one service. If `LanNetworkInfo.update` is ever extended to span a second USP service, the default `allowPartial: false` will silently regress to a 7005.

**Fix:** Add a comment at line 49 explaining why `allowPartial: false` is acceptable here (single-service, single-param). Or encode the invariant in codegen/`UspClient._batchSet`.

---

**W-5 — 7005 error table not updated; Shotgun Surgery risk (High)**

`lib/core/usp/errors/usp_error.dart:143`

```dart
/// | 7005 | invalid parameter name           | InvalidInputError       |
```
This PR introduces a second OBUSPA-specific meaning for 7005: "Allow partial=false not supported across more than one USP Service". The table header explicitly says "keep in sync".

**Fix:** Add note to the 7005 row documenting the OBUSPA cross-service meaning. Mirror in `service_error_localizations.dart`.

---

**W-6 — `connectionType` computed twice; [Duplicate Code] (High)**

`lib/page/internet_settings/services/usp_internet_settings_service.dart:74–77, 132–135`

```dart
// Line 74–77 — Phase-2 fetch gate:
final connectionType = UspWanConnectionType.fromRawFields(
  addressingType: wan.addressingType, lowerLayers: pppInstance?.lowerLayers ?? '');

// Line 132–135 — _buildForm (same inputs):
final connectionType = UspWanConnectionType.fromRawFields(
  addressingType: wan.addressingType, lowerLayers: lowerLayers);
```
If `fromRawFields` ever gains a new parameter, both call sites must be updated.

**Fix:** Compute once in `fetchSettings`, pass as an explicit parameter to `_buildForm`.

</details>

<details>
<summary>Critical — none this round</summary>

No Critical findings this round. All items verified as Warning or Suggestion.

</details>

<details>

- **`_isTableExpandedPath` is complete and correct** — `path.contains('*') || path.endsWith('.')` used consistently in both `normalizeGetResponse` and `isTableQueryOnlyRequest`. Hank H-2 is not valid against the current head.
- **`isWildcardOnlyRequest` → `isTableQueryOnlyRequest` rename** — semantically accurate; new test group covers trailing-dot, mixed, and negative cases.
- **Fail-fast `StateError` mocks** — all PnP scenarios + `createFetchMockHandler` now throw on unexpected tunnel GETs.
- **`_greEmptyResponse`/`_l2tpEmptyResponse` correctly removed** — constants eliminated now that handlers fail-fast.
- **Round 2 W-1/W-3/W-4 all resolved** — DNS-split test duplication fixed; trailing-dot false warnings fixed; `_isTableExpandedPath` extraction DRY and correct.
- **Spec-W1 RESOLVED** — `PrivacyGUI-USP-E2E` commit `f691d369` re-enabled the S06 error-state test before this PR's head SHA.
- **Incremental diff is small and targeted** — 176 lines across 4 files, no accidental scope creep.

</details>

AustinChangLinksys (commented):
🤖 **Automated Review — Round 4** · `5c1ed78..c5f62a7` (incremental)

**Verdict: 💬 Self-review (comment only)** — Austin is the author; GitHub prohibits self-approval. Review provided as a reference to catch blind spots.

> **Context note:** All 6 Warnings from Round 3 (W-1 through W-6) are confirmed **resolved** in this round. The `connectionType` DRY refactor, diagnosable `StateError` contract assertion, DHCP-WAN regression guard, `allow_partial` invariant documentation, 7005 footnote, and `?? ''` dead-code removal are all verified against the HEAD code. Two new Warnings remain (one per axis).

---

#### Standards

|---|---|---|---|
| ⚠️ | 🟢High | `usp_internet_settings_service.dart:99–101` + `usp_error.dart:178–180` | `StateError` from `_buildForm` contract assertion gets silently downgraded to `UnexpectedError` via `mapUspErrorToServiceError`; diagnosable contract-violation message becomes opaque to callers |
| 💡 | 🟡Med | `usp_internet_settings_service_test.dart:975–990` | `verify(captureAny()).captured` may interact with existing `any()` stubs in Mocktail (issue #114); consider `verifyNever` or aligned stubs |
| 💡 | 🟡Med | `usp_internet_settings_service_test.dart:954–988` | [Duplicate Code] Only DHCP WAN explicitly asserts no tunnel GET; Static/Bridge/PPPoE rely implicitly on mock fail-fast rather than a shared helper assertion |

---

#### Spec

|---|---|---|---|
| ⚠️ | 🟢High | spec#Issue #1184 (open) | Issue #1184 still `open` with no comment recording the option (b) decision — PR `Closes #1184` auto-closes on merge but leaves the issue thread silent on rationale |
| 💡 | 🟢High | spec → `usp_internet_settings_service.dart:142–149` + test | No test verifying `StateError` is thrown when `_buildForm` receives `gre: null` on PPTP branch (contract branch untested) |
| 💡 | 🟢High | spec → `usp_client_normalize_response_test.dart:186–193` | Missing `[trailing-dot + concrete] → false` combination test for `isTableQueryOnlyRequest` |

---

<details>
<summary>Critical — none this round</summary>

No Critical findings this round. All Round 3 Warnings resolved; two new Warnings remain (one Standards, one Spec).

</details>

<details>

**W-1 — `StateError` silently downgraded to `UnexpectedError` (High) — flagged by both Reviewer A (Warning) and Reviewer B (Suggestion)**

`lib/page/internet_settings/services/usp_internet_settings_service.dart:99–101`
`lib/core/usp/errors/usp_error.dart:178–180`

```dart
// usp_internet_settings_service.dart:99–101
} catch (e) {
  throw mapUspErrorToServiceError(e);
}
```

```dart
// usp_error.dart:178–180
final parsed = parseUspError(error);
if (parsed == null) {
  final result = UnexpectedError(originalError: error);
  logger.w('[USP][ServiceError]: "$error" -> ${result.runtimeType}');
  return result;
}
```

`_buildForm` on the PPTP branch throws `StateError('PPTP WAN but GRE tunnel absent — fetchSettings phase-2 contract violated')`. This is not a USP-formatted string; `parseUspError` returns `null` -> the `StateError` gets wrapped as `UnexpectedError(originalError: stateError)`. `UnexpectedError.detail` is null here, so its `toString()` carries little diagnostic value. Callers see a generic `UnexpectedError` and cannot distinguish a firmware-contract violation from a network error.

**Practical impact note:** This path is theoretically unreachable in production (Phase 2's `GreTunnel.fetch` would throw 9998 *before* `_buildForm` runs, and `gre` can never be null if Phase 2 succeeded). But the `StateError` was explicitly added for diagnosability — this wrapping defeats that intent.

**Fix (any option):**
```dart
// Option A — let StateError propagate unchanged (preferred for contract violations):
} catch (e) {
  if (e is StateError) rethrow;
  throw mapUspErrorToServiceError(e);
}

// Option B — pass detail through UnexpectedError:
// UnexpectedError's originalError already carries the message; ensure callers surface it.
```

---

**Spec-W1 — Issue #1184 decision not recorded on issue thread (High)**

Issue #1184: `"state": "open"` — last comment is `2026-08-04T01:50:55Z` (author's investigation update ending with "The one decision that still needs a human"). No subsequent comment was added explaining the final decision (option b: trust firmware contract, keep required-leaf 9998, conditional fetch).

PR `Closes #1184` will auto-close the issue on merge, but anyone reading the issue thread will find an unanswered question. The rationale (firmware guarantees `Tunnel.1` when `LowerLayers` references it; 9998 is now a real alarm, not dead code) is documented only in the PR description.

**Fix:** Add one comment to issue #1184 summarising: "Went with option (b): the required-leaf contract is preserved. Conditional fetch in PR #1219 ensures 9998 only fires on a genuine firmware-contract violation. Closing via #1219."

</details>

<details>

- **All 6 Round 3 Warnings resolved** — W-1 through W-6 confirmed against HEAD code:
  - W-1/W-2: `?? ''` dead code removed; `StateError` contract assertion replaces bare `!`; Phase 2 is conditional on WAN type.
  - W-3: Dual-guard regression test added — mock fail-fast + explicit `captureAny()` assertion on GET paths.
  - W-4: `instant_safety_service.dart` single-service SET documented with cross-service tripwire comment.
  - W-5: `usp_error.dart` 7005 table entry updated with broker-level footnote.
  - W-6: `connectionType` derived once in `fetchSettings`, passed into `_buildForm` as explicit parameter.
- **Diagnosable contract assertion** — `gre ?? throw StateError(...)` is a clear upgrade from bare `!`; a violated contract now surfaces a meaningful message rather than "Null check operator used on a null value".
- **`_buildForm` parameter refactor** — DRY improvement also eliminates the implicit coupling between `fetchSettings`' derived value and `_buildForm`'s re-derivation.
- **DHCP-WAN regression guard** — test at `usp_internet_settings_service_test.dart:954–988` uses both mock fail-fast and explicit path assertion. `setUp()` recreates mock per test; no cross-test contamination.
- **7005 footnote** — correctly distinguishes broker-level cross-service rejection from standard "invalid parameter name" — useful for future debugging.
- **Spec requirements 1, 2, 3 all met** — conditional tunnel fetch (DHCP/Static/PPPoE never trigger 9998), IPv6 `allowPartial: true` (verified with test), table/wildcard warning classification (`isTableQueryOnlyRequest`) all working correctly.

</details>

AustinChangLinksys (commented):
🤖 **Automated Review — Round 4** · `5c1ed78..cc280e0` (incremental)

**Verdict: 💬 Self-review (comment only)** — Austin is the author; GitHub prohibits self-approval. Review provided as a reference to catch blind spots.

> **Context note:** Prior Round 3 W-1 (Standards: `StateError` silently downgraded to `UnexpectedError`) is **confirmed resolved** in this increment — the `throw StateError(...)` is fully removed; replaced with `assert`s (debug-only contract documentation) + `?? ''` (release graceful fallback). PPTP-absent-GRE-row test chain verified (`{} → GreTunnel._fromResponse → 9998 string → mapUspErrorToServiceError → InvalidInputError(ServiceError)`). Tests 38/38 green (confirmed by Reviewer A via `flutter test`).

---

#### Standards

|---|---|---|---|
| 💡 | 🟢High | `usp_internet_settings_service_test.dart` (L2TP absent-row) | Missing symmetric L2TP absent-row test — PPTP-absent-row guard exists; L2TP branch has the same `?? ''` path but no contract test |
| 💡 | 🟡Med | `usp_internet_settings_service.dart:117–124` | [Data Clumps] `_buildForm` takes 7 positional parameters; consider a `_FetchPhaseResult` record/class or named params |
| 💡 | 🟡Med | `instant_safety_service.dart:49–52` | `allowPartial` intent comment has no machine-searchable tag — maintainer adding a cross-service param may miss the `allowPartial: true` requirement |
| 💡 | 🟡Med | `usp_internet_settings_service_test.dart` | [Duplicated Code] `when(() => mockUsp.get(any())).thenAnswer(...)` stub repeated verbatim in DHCP and PPTP absent tests; consider a shared helper |

---

#### Spec

|---|---|---|---|
| ⚠️ | 🟢High | spec#Issue #1184 (open) → no issue comment | Issue #1184 will be auto-closed on merge, but the option-(b) decision rationale is still unrecorded on the issue thread |
| ⚠️ | 🟢High | spec → `usp_client_normalize_response_test.dart` | `[trailing-dot + concrete] → false` combination missing from `isTableQueryOnlyRequest` test group — logic-correct but boundary untested (upgraded from Suggestion round 3) |

---

<details>
<summary>Critical — none this round</summary>

No Critical findings this round. Prior Round 3 W-1 (Standards: `StateError` silently downgraded to `UnexpectedError`) fully resolved — confirmed by both Reviewer A and Reviewer B via `git show cc280e0e:...` and `flutter test` execution.

</details>

<details>

**Spec-W1 — Issue #1184 decision not recorded (High confidence)**

`gh api repos/linksys/PrivacyGUI/issues/1184` confirms: `"state": "open"`, 2 comments (both AustinChangLinksys, Aug 4). Neither records the final option-(b) choice.

PR body says `Closes #1184` — GitHub auto-closes on merge. But the issue thread will remain silent on rationale: why option (b) over option (a)? What firmware contract is trusted? What regressions did removal surface?

**Fix:** Post one comment on Issue #1184 before merge, e.g.:
> *Resolved via option (b): removed `_rawGet` back-fill in #1198; regressions fixed in #1219. Rationale: firmware-contract guarantee on PPTP/L2TP tunnels makes option (b) safe; 9998 now fires as a genuine alarm on absent required leaves.*

No code change needed — issue comment only.

---

**Spec-W2 — `[trailing-dot + concrete] → false` combination test missing (High confidence)**

`usp_client_normalize_response_test.dart` `isTableQueryOnlyRequest` group has:
- `[wildcard-only] → true` ✅
- `[trailing-dot-only] → true` ✅
- `[wildcard + trailing-dot] → true` ✅
- `[wildcard + concrete] → false` ✅ (lines 186–193)
- `[concrete-only] → false` ✅

**Missing:** `[trailing-dot + concrete] → false`

Predicate: `paths.every(_isTableExpandedPath)` where `_isTableExpandedPath = path.contains('*') || path.endsWith('.')`. Logic is correct, but this boundary is untested. If `_isTableExpandedPath` is refactored, this regression path won't be caught.

**Fix:** Add one test case in the `isTableQueryOnlyRequest` group:
```dart
test('[trailing-dot + concrete] is NOT table-query-only', () {
  expect(
    UspClient.isTableQueryOnlyRequest([
      'Device.IP.Interface.1.IPv6Address.',  // trailing-dot
      'Device.GRE.Tunnel.1.RemoteEndpoints', // concrete
    ]),
    isFalse,
  );
});
```

</details>

<details>

- **Round 3 W-1 fully resolved** — `throw StateError(...)` removed; replaced with:
  - `assert(connectionType != UspWanConnectionType.pptp || gre != null, '...')` — dev-build contract documentation, diagnosable message
  - `assert(connectionType != UspWanConnectionType.l2tp || l2tp != null, '...')` — same for L2TP
  - `gre?.remoteEndpoints ?? ''` / `l2tp?.remoteEndpoints ?? ''` — release graceful fallback (empty `serverAddress` → form validation, not crash)
  - Root defense: Phase-2 `GreTunnel.fetch` / `L2tpTunnel.fetch` throws 9998 before `_buildForm` is ever called if tunnel row is absent
- **PPTP-absent-row test chain verified** — `greResponse: const {}` → `GreTunnel._fromResponse({})` → missing required leaf → 9998 → `mapUspErrorToServiceError` → `InvalidInputError(ServiceError)` → `throwsA(isA<ServiceError>())` passes; 38/38 green
- **`connectionType` single source of truth** — derived once in `fetchSettings`, passed into `_buildForm`; eliminates old double-derivation (DRY fix from round 3)
- **7005 footnote in `usp_error.dart`** — accurately documents OBUSPA broker-level cross-service atomic SET rejection, distinct from "invalid parameter name"; cross-referenced from `_saveIpv6Settings`
- **IPv6 cross-service SET `allowPartial: true`** — correct; `_handleSetResult` still surfaces per-parameter failures as `UspPartialFailureError`
- **`instant_safety_service.dart` single-service comment** — correct safety note; current single-service SET safely atomic

</details>

lib/page/internet_settings/services/usp_internet_settings_service.dart
```diff
@@ -46,31 +46,49 @@ class UspInternetSettingsService {
...
+      GreTunnel? gre;
+      L2tpTunnel? l2tp;
+      switch (connectionType) {
+        case UspWanConnectionType.pptp:
+          gre = await GreTunnel.fetch(_usp);
+        case UspWanConnectionType.l2tp:
+          l2tp = await L2tpTunnel.fetch(_usp);
+        default:
+          break;
+      }
```
PeterJhongLinksys, line 87: **⚠️ Warning (🟡 Med)** — Phase-2 tunnel fetch keys off the freshly derived `connectionType`, but `usp_internet_settings_notifier.dart:106-113` may *restore* a preserved type afterwards. If the post-save refetch sees a transiently empty `addressingType` (the #759 race), `fromRawFields` resolves `bridge`, no GRE is fetched, `serverAddress` becomes `''`, and the notifier then restores `pptp` — the UI renders an empty "PPTP Server" field. The previous unconditional fetch masked this. Consider driving the phase-2 decision from the preserved type, and adding a notifier-level test for that combination.

lib/page/internet_settings/services/usp_internet_settings_service.dart
```diff
@@ -46,31 +46,49 @@ class UspInternetSettingsService {
...
+      );
+      GreTunnel? gre;
+      L2tpTunnel? l2tp;
+      switch (connectionType) {
+        case UspWanConnectionType.pptp:
+          gre = await GreTunnel.fetch(_usp);
+        case UspWanConnectionType.l2tp:
+          l2tp = await L2tpTunnel.fetch(_usp);
+        default:
+          break;
```
PeterJhongLinksys, line 86: **💡 Suggestion (🟢 High)** — `default: break` on a closed enum disables exhaustiveness checking: adding a third tunnel-bearing connection type would silently skip phase 2, which is the mirror image of the bug this PR fixes. Listing `dhcp/staticIp/pppoe/bridge` explicitly makes it a compile error instead.

lib/page/internet_settings/services/usp_internet_settings_service.dart
```diff
@@ -111,16 +130,27 @@ class UspInternetSettingsService {
...
+    assert(
+      connectionType != UspWanConnectionType.l2tp || l2tp != null,
+      'L2TP WAN but L2TPv2 tunnel absent — fetchSettings phase-2 contract violated',
     );
-
-    // Resolve server address from the appropriate tunnel
     final serverAddress = switch (connectionType) {
-      UspWanConnectionType.pptp => gre.remoteEndpoints,
-      UspWanConnectionType.l2tp => l2tp.remoteEndpoints,
+      UspWanConnectionType.pptp => gre?.remoteEndpoints ?? '',
```
PeterJhongLinksys, line 152: **⚠️ Warning (🟡 Med)** — `assert` is stripped in release builds, so a contract violation degrades to a silent `''`. The comment argues form validation catches it, and that does hold for the interactive path (`usp_internet_settings_form_validator.dart:44-46`) — but `pnp_service.dart:472-483` (`saveIspSettings` → `fetchSettings` → `saveAll`) bypasses the validator entirely. It is unreachable today only because `_mapConnectionType` never yields pptp/l2tp, i.e. safety rests on an upstream coincidence. Prefer throwing a `ServiceError`, or at least `logger.w` on the fallback so production is diagnosable.

lib/page/internet_settings/services/usp_internet_settings_service.dart
```diff
@@ -549,6 +586,7 @@ class UspInternetSettingsService {
           _diff(original.ipv6rdIpv4MaskLength, edited.ipv6rdIpv4MaskLength),
       ipv6rdBorderRelay:
           _diff(original.ipv6rdBorderRelay, edited.ipv6rdBorderRelay),
+      allowPartial: true,
```
PeterJhongLinksys, line 589: **⚠️ Warning (🟡 Med)** — The fix is right (OBUSPA rejects cross-service atomic SETs with 7005) and the interception chain is complete: `ipv6settings.g.dart:156` → `usp_client.dart:471-496` → `usp_operation_result.dart:400-426` → `_handleSetResult:641` throws `UspPartialFailureError`. Two follow-ups: (1) on partial failure `performSave` only rethrows and `save()` refetches on success only (`usp_internet_settings_notifier.dart:174-176, 200-222`), so already-applied parameters never reach the UI — device and form diverge; `UspPartialFailureError.successPaths` is available. (2) `allowPartial` is unconditional, so a single-service IPv6 diff also loses atomicity it could have kept.

lib/core/usp/services/usp_client.dart
```diff
@@ -341,14 +367,39 @@ class UspClient {
...
+  /// paths, so the requested key itself never appears verbatim in the response.
+  ///
+  /// Two shapes qualify, and they must be treated identically everywhere:
+  /// - a wildcard search path (contains '*', e.g. Device.Firewall.DMZ.*)
+  /// - an object/table path (ends in '.', e.g. Device.IP.Interface.1.IPv6Address.)
+  ///
+  /// Both target a multi-instance set; an absent requested key is expected, not
+  /// missing, and an empty response means the set has zero rows — not a fault.
+  static bool _isTableExpandedPath(String path) =>
+      path.contains('*') || path.endsWith('.');
```
PeterJhongLinksys, line 388: **⚠️ Warning (🟢 High)** — `endsWith('.')` also matches single-object paths, not just multi-instance tables. `lib/page/test_console/views/usp_test_console_view.dart:544,548` probes `Device.DynamicDNS.` / `Device.UPnP.` expecting fault 9005 for unimplemented modules; such an empty response is now classified "no rows" and demoted from `warn` to `debug`, making "module unsupported" indistinguishable from "zero rows" in the logs. Consider limiting the exemption to known multi-instance tables (or keeping `logger.i`), and softening the "not a fault" claim in the DartDoc.

test/page/instant_setup/services/pnp_service_test.dart
```diff
@@ -93,15 +93,14 @@ void main() {
...
         return vlanResponse;
       }
+      // PnP setup never configures a PPTP/L2TP WAN, so connectionType is never
+      // PPTP/L2TP and Phase-2 fetches no tunnel. A tunnel GET here means an
+      // unconditional fetch regressed — fail fast instead of returning empty.
       if (paths.any((p) => p.contains('GRE.Tunnel'))) {
-        return <String, dynamic>{
-          'Device.GRE.Tunnel.1.RemoteEndpoints': '',
-        };
+        throw StateError('Unexpected GRE.Tunnel GET in PnP fetch');
```
PeterJhongLinksys, line 100: **⚠️ Warning (🟡 Med)** — This guard is thrown inside `_usp.get()`, so it lands in `fetchSettings`' `catch (e) { throw mapUspErrorToServiceError(e); }` (`usp_internet_settings_service.dart:100-101`). `parseUspError`'s `^(\w+) failed: (.+)$` (`usp_error.dart:58`) does not match `"Bad state: …"`, so it becomes `UnexpectedError` — itself a `ServiceError`. Any test asserting `throwsA(isA<ServiceError>())` therefore stays green even if the unconditional fetch regresses, and elsewhere your explanatory message is replaced by `UnexpectedError`. The `verify(captureAny())` + `isNot(contains(...))` pattern at `usp_internet_settings_service_test.dart:970-987` is the robust alternative (or `fail(...)` / a flag asserted at the end).

PeterJhongLinksys (approved):
🤖 **Automated Review — Round 1** · `1175273..cc280e0` (full)

**Verdict: ✅ APPROVE** — The three fixes are correct in substance (conditional tunnel fetch, cross-service `allowPartial`, log-noise classification) and the regression guards are unusually well targeted; approving with warnings, mainly around release-mode behaviour of the new `assert`-based contract and one plausible interaction with the `_preservedConnectionType` guard.

|---|---|---|---|
| ⚠️ | 🟡 | `lib/page/internet_settings/services/usp_internet_settings_service.dart:74-87` | Conditional tunnel fetch × the `_preservedConnectionType` guard in `usp_internet_settings_notifier.dart:106-113` can leave a PPTP/L2TP user with a blank Server field (previously masked by the unconditional fetch). |
| ⚠️ | 🟡 | `usp_internet_settings_service.dart:143-155` | `assert` + `?? ''` expresses a contract that is erased in release builds → silent empty `serverAddress` instead of a diagnosable failure. |
| ⚠️ | 🟢 | `usp_internet_settings_service.dart:117-125` | [Long Parameter List] `_buildForm` now takes 7 positional params (3 nullable); `connectionType` moving out created an implicit ordering contract (suspected Temporal Coupling). |
| ⚠️ | 🟡 | `usp_internet_settings_service.dart:51-103` | [Long Method] `fetchSettings` now does 6 things; `results[4] as String` index unpacking had to be renumbered by this very change. |
| ⚠️ | 🟡 | `usp_internet_settings_service.dart:568-590` | `allowPartial: true` makes the IPv6 SET non-atomic; the partial-failure path never refetches, so device state and UI can diverge. |
| ⚠️ | 🟡 | `usp_internet_settings_service.dart:263-269` | Step 7 (IPv6) now throwing *after* partially writing amplifies the existing ordering risk: Steps 2-6 stay applied and the Step 8 bridge SET never runs. |
| ⚠️ | 🟢 | `lib/core/usp/services/usp_client.dart:387-388` | `path.endsWith('.')` also matches single-object probes, so a genuinely unsupported module's empty response is now demoted from `warn` to `debug`. |
| ⚠️ | 🟡 | `test/page/instant_setup/services/pnp_service_test.dart:100,103,311,314,380,383` | `throw StateError` guards are swallowed by `fetchSettings`' `catch (e)` → become `UnexpectedError` (a `ServiceError`), so `throwsA(isA<ServiceError>())` tests would still pass green. |
| 💡 | 🟢 | `usp_internet_settings_service.dart:85-86` | `default: break` on a closed enum disables exhaustiveness checking — a future tunnel type would silently skip phase 2. |
| 💡 | 🟢 | `usp_client.dart:316-319` | `normalizeGetResponse(paths, rawMap)` in the empty branch is a no-op; `return <String, dynamic>{}` states the intent. |
| 💡 | 🟡 | `usp_client.dart:387-401` | Naming: `_isTableExpandedPath` / `isTableQueryOnlyRequest` conflate "router-expanded path" with "multi-instance table"; `_isNonLeafPath` would match the actual predicate. |
| 💡 | 🟡 | `lib/core/usp/errors/usp_error.dart:152-160`, `instant_safety_service.dart:49-52`, `usp_internet_settings_service.dart:571-578` | The "cross-service SET cannot be atomic" rule is now duplicated as prose in three places (suspected Shotgun Surgery). |
| 💡 | 🟡 | `usp_error.dart:159-160` | Core-layer DartDoc points at a page-layer *private* method (`_saveIpv6Settings`) — a documentation-level upward reference and a stale-link risk. |
| 💡 | 🟡 | `usp_internet_settings_service.dart:589` | `allowPartial: true` is unconditional, so a single-service IPv6 diff also loses atomicity it could have kept. |
| 💡 | ⚪ | `instant_safety_service.dart:49-52` | Speculative comment: `LanNetworkInfo.update` already spans three services (`lan_network_info.g.dart:22-37`); it is only single-service because this call happens to pass `dnsServers` alone. |
| 💡 | 🟡 | `test/page/internet_settings/services/usp_internet_settings_service_test.dart:357-378` | `for (i += 2)` pairing of flattened `captured` is brittle; a new named param silently shifts the pairing. |
| 💡 | 🟡 | `test/page/internet_settings/services/usp_internet_settings_service_test.dart:340-376` | Three more inline test fixtures; `test/mocks/test_data/` still has no `internet_settings` builder (Art. VIII). |
| 💡 | ⚪ | `usp_internet_settings_service.dart:80-87` | Phase 2 adds one extra round trip for PPTP/L2TP users (tunnel fetch no longer overlaps phase 1). |
| 💡 | ⚪ | `usp_client.dart:309-310, 329-330` | New log lines interpolate `$paths` / `$missing` directly instead of the file's `_prettyList` convention. |

<sub>Confidence: 🟢 verified against head code with quoted evidence · 🟡 file:line + reasoning, not fully proven · ⚪ speculative.</sub>
<sub>`[X]` = raised by 1 agent · `[XX]` = independently raised by both. No Critical was raised this round.</sub>

<details>

**1. Conditional tunnel fetch × `_preservedConnectionType` — blank Server field** (`usp_internet_settings_service.dart:74-87`)
```dart
final connectionType = UspWanConnectionType.fromRawFields(
  addressingType: wan.addressingType,
  lowerLayers: pppInstance?.lowerLayers ?? '',
);
GreTunnel? gre;
switch (connectionType) {
  case UspWanConnectionType.pptp:
    gre = await GreTunnel.fetch(_usp);
```
`usp_internet_settings_notifier.dart:106-113` restores the saved type after a save when the refetched `addressingType` comes back momentarily empty (the #759 race):
```dart
if (savedType != null && form.connectionType != savedType) {
  form = form.copyWith(connectionType: savedType);
}
```
Trigger: PPTP user edits an unrelated field and saves → refetch sees empty `addressingType` → `fromRawFields` resolves `bridge` → phase 2 fetches no GRE → `serverAddress = ''` → notifier restores `pptp` → `_buildTunnelFields` (`views/sections/usp_ipv4_section.dart:334,344`) renders an empty "PPTP Server". Before this PR the unconditional GRE fetch masked it. No data loss (`_saveTunnelRemoteEndpoints` diffs `'' vs ''` → no write), but the setting looks lost. Suggested fix: drive the phase-2 decision from the preserved type, and add a notifier-level test for "preserved type + transiently empty `addressingType`".

**2. `assert` + `?? ''` contract is release-erased** (`usp_internet_settings_service.dart:143-155`)
```dart
assert(connectionType != UspWanConnectionType.pptp || gre != null, ...);
final serverAddress = switch (connectionType) {
  UspWanConnectionType.pptp => gre?.remoteEndpoints ?? '',
```
The comment's claim that form validation catches the fallback holds for the interactive path (`usp_internet_settings_form_validator.dart:44-46` requires a non-empty `serverAddress` for pptp/l2tp), but `pnp_service.dart:472-483` (`saveIspSettings` → `fetchSettings` → `_applyIspConfigToForm` → `saveAll`) bypasses the validator entirely. It is currently unreachable only because `_mapConnectionType` (`pnp_service.dart:509-516`) never yields pptp/l2tp — i.e. safety rests on an upstream coincidence, so this is a defensive hazard rather than a user-triggerable bug. Prefer throwing a `ServiceError` (or at minimum `logger.w` on the fallback) so release and debug agree.

**3. Non-atomic IPv6 SET without reconciliation** (`usp_internet_settings_service.dart:568-590`, `usp_internet_settings_notifier.dart:174-176,200-222`)
The interception chain was verified end to end and is sound: `ipv6settings.g.dart:156` → `usp_client.dart:436-442,471-496` → `usp_client_wasm.dart:201-204` → `web/usp_client.js:442`; `usp_operation_result.dart:400-426` yields `UspPartialSuccess`; `_handleSetResult:631-651` throws `UspPartialFailureError`. Nothing is silently swallowed. What is missing is the follow-up: `performSave` only rethrows and `save()` refetches on success only, so already-applied parameters are never pulled back into the form. `UspPartialFailureError.successPaths` is available — consider a forced refetch and/or "partially applied" wording.

**4. `throw StateError` test guards are absorbed by the SUT** (`pnp_service_test.dart:100,103,311,314,380,383`, `usp_internet_settings_service_test.dart:131-138,244-247`)
The mock throws inside `_usp.get()`, so it lands in `fetchSettings`' `catch (e) { throw mapUspErrorToServiceError(e); }` (`usp_internet_settings_service.dart:100-101`). `parseUspError`'s `^(\w+) failed: (.+)$` (`usp_error.dart:58`) does not match `"Bad state: Unexpected GRE.Tunnel GET…"`, so it becomes `UnexpectedError` — a `ServiceError`. Any test asserting `throwsA(isA<ServiceError>())` therefore still passes even if the unconditional fetch regresses, and elsewhere the carefully worded guard message is replaced by `UnexpectedError`. The pattern used at `usp_internet_settings_service_test.dart:970-987` (`verify(captureAny())` + `isNot(contains(...))`) is the right shape; prefer that, or `fail(...)` / a flag asserted at the end.

**5. Empty-response demotion also covers single-object probes** (`usp_client.dart:387-388`)
```dart
static bool _isTableExpandedPath(String path) =>
    path.contains('*') || path.endsWith('.');
```
`lib/page/test_console/views/usp_test_console_view.dart:544,548` probes `Device.DynamicDNS.` / `Device.UPnP.` expecting fault 9005 for unimplemented modules, and `usp_lan_data_service.dart:71` GETs `Device.IP.Interface.1.IPv6Address.`. Under the new rule an empty response there is classified "no rows" and logged at `debug` instead of `warn`, so "module unsupported" and "table has zero rows" become indistinguishable in the logs. Consider restricting the exemption to known multi-instance tables, or keeping `logger.i`, and softening the DartDoc's "not a fault" assertion.

**6. Save-step ordering amplified** (`usp_internet_settings_service.dart:263-269`)
With partial application enabled, an IPv6 failure at Step 7 aborts before the Step 8 bridge SET while Steps 2-6 (`AddressingType`, PPP `LowerLayers`, VLAN) are already committed — leaving the WAN in a neither-old-nor-bridge state. Moving the IPv6 SET after the bridge SET, or continuing and aggregating failures at the end, would close this.

</details>

<details>
<summary>💡 Suggestions (see table for the full list)</summary>

- Enumerate all `UspWanConnectionType` cases instead of `default: break` (`:85-86`) so a new tunnel type fails at compile time — the mirror image of the bug this PR fixes.
- `return <String, dynamic>{}` directly in the empty branch (`usp_client.dart:316-319`); `normalizeGetResponse` provably does nothing there.
- Rename `_isTableExpandedPath` / `isTableQueryOnlyRequest` to reflect "non-leaf path", and document that single-object paths are included.
- Consolidate the cross-service-atomicity rule in one place (a named constant or helper next to the 7005 mapping) and have the other two sites reference it.
- Replace the `_saveIpv6Settings` reference in `usp_error.dart:159-160` with a behavioural statement ("callers must pass `allowPartial: true` for cross-service SETs").
- Consider deriving `allowPartial` from the actual parameter set so single-service diffs keep atomicity.
- Extract the `_greResponse` / `_l2tpResponse` / `_ppp*Response` fixtures into an `InternetSettingsTestData` builder (Art. VIII) and replace the `i += 2` capture pairing with separate `captureAny(named:)` verifications.

</details>

<details>

- **Root causes actually addressed, not symptoms.** The 9998 required-leaf trip is fixed by not requesting an absent table at all, and 7005 by moving the fix to the caller rather than distorting the error mapping.
- **The partial-failure interception chain is genuinely complete** — traced hop by hop from `allowPartial: true` through the WASM bridge to `_handleSetResult`; `allowPartial` does not become "silent success".
- **The strongest test in the PR** is `usp_internet_settings_service_test.dart:954-990`: it positively asserts via `verify(captureAny())` that a DHCP WAN issues *no* tunnel GET, rather than relying on mock side effects.
- **Residual-risk path covered too** (`:992-1020`): PPTP with an absent `Tunnel.1` must surface as `ServiceError`, not a crash — consistent with the Art. XIII boundary.
- **`isTableQueryOnlyRequest` edge cases are exhaustive** (all-wildcard / all-trailing-dot / mixed / contains-concrete / empty list), and empty `paths` deliberately returns `false` so "nothing requested" is never silently demoted.
- **Log-noise fix keeps the information** while cutting the lines: one aggregated warning via `onMissingPath: missing.add` instead of N warnings, so a real partial response is no longer diluted.
- **7005's second, broker-level meaning is now documented** in the core error table, preventing a future contributor from "fixing" the mapping instead of the caller.
- **All architectural hard rules hold**: layering direction, `generated/` confined to `services/`, the four USP provider rules (L1 not `autoDispose`, L2 uses `ref.read`, mutations inside `uspMutationLockProvider.withLock()`, `PreservableContract` untouched), and the `ServiceError` boundary (`catch (e)`, never `on Exception catch`, always via `mapUspErrorToServiceError`).

</details>

> Note: no Dart/Flutter toolchain was available in the review environment, so `flutter analyze` / `flutter test` were **not** executed. All findings are from static reading of the head revision.

HankYuLinksys (approved):
## Re-review at `cc280e0e` — one correction of mine, one test-diagnostics note

Re-verified everything on a fresh worktree at the new head. Baseline: `flutter analyze` clean on all four changed lib files; `test/core/usp` + `test/page/{internet_settings,instant_setup,local_network}` green at **832 tests** (was 828 — the 4 new ones). I re-checked each item against the code and with throwaway probes rather than relying on the summaries.

### Correction — my item 2 was invalid

The trailing-dot fix landed in `5c1ed781` at 08:27Z, about three hours *before* I posted at 09:51Z. I had fetched a stale head and reviewed against it. My apologies for the noise — @AustinChangLinksys was right to mark it refuted.

For the record, verified on the current head:

```
Device.IP.Interface.1.IPv6Address.        -> isTableQueryOnlyRequest = true   (correctly demoted to debug)
[trailing-dot, concrete]                  -> false                            (concrete path still warns)
```

`_isTableExpandedPath` as the single shared predicate, plus the rename to `isTableQueryOnlyRequest`, is a better shape than the `_isRouterExpandedPath` I suggested — "table query" is what the predicate actually means.

Items 1, 3, 5, 6 all check out on the current head. On item 1 in particular: trusting the firmware contract is your call to make, and it's now explicit in the PR description with `assert`s that fail loudly in dev and `?? ''` degrading gracefully in release. That's a clean landing point for what took three passes to converge.

### (a) Note — the new PPTP test can pass for the wrong reason

Not a coverage gap, and **not a merge blocker** — see the second half of this section. Purely a diagnostics observation.

The mock handlers throw `StateError` on an unexpected tunnel GET, but `fetchSettings`' `catch (e)` funnels *everything* through `mapUspErrorToServiceError`. Probed:

```
StateError -> UnexpectedError;  isA<ServiceError> = true
```

So `PPTP WAN with absent GRE Tunnel.1 row surfaces a ServiceError` asserting only `throwsA(isA<ServiceError>())` cannot distinguish:

- 9998 from the absent GRE row — the contract it means to pin, and
- a `StateError` because the conditional fetch regressed and an unexpected tunnel GET fired.

Tightening to `isA<InvalidInputError>()`, or matching `9998` in the message, separates them. Round 4's Standards table flagged the same interaction from the Mocktail side.

**Why this doesn't matter for correctness:** the revert scenario is already caught outright by `DHCP WAN fetch issues NO GRE/L2TP tunnel GET`, which asserts on the captured GET paths directly rather than relying on the handler's fail-fast. I confirmed that assertion is not vacuous — that fetch issues 7 GETs / 29 paths, so `captured` is populated and a reintroduced unconditional fetch fails it directly. Reverting the conditional fetch therefore always turns a test red. The PPTP test's weakness is only that it might go green for the wrong reason, not that anything goes unguarded.

### (b) Refuting one warning from the approving review — the blank Server field is pre-existing

> Conditional tunnel fetch × the `_preservedConnectionType` guard can leave a PPTP/L2TP user with a blank Server field (previously masked by the unconditional fetch).

The "previously masked" part doesn't hold. Probed a transient empty `LowerLayers` on an `IPCP` WAN against this head:

```
type=pppoe, serverAddress=""
```

`_buildForm`'s switch and the phase-2 fetch both key on the *same* derived type, so the old unconditional fetch produced `""` here too — it fetched GRE and then discarded it on the `pppoe` branch. Same output, one wasted round trip. If anything, item 6's DRY refactor makes that agreement structural instead of coincidental. No action needed.

### Item 4 — worth a follow-up issue, not a change here

I re-confirmed there's no live bug: of the three generated `update()`s spanning services, `Ipv6Settings` is fixed here, `WanPppoe` already passed `allowPartial: true`, and `LanNetworkInfo` is safe at both call sites. The `instant_safety_service.dart` comment documents the constraint well. What's still unencoded is the systemic guard (a debug check in `_batchSet`, or deriving the default in codegen from the service span) so the *next* cross-service model can't silently reintroduce 7005. That's a separate hardening task and shouldn't hold this PR.

### Also noted, not blocking

`default: break` in the phase-2 switch bypasses Dart's exhaustiveness checking, and the two `assert`s cover pptp/l2tp only while three types carry a `pppLowerLayers` reference (pppoe/pptp/l2tp). The enum is closed today, so it's a future trap rather than a current defect — already on both automated rounds' lists.

---

Approving. Five of my six items are resolved or converted into explicitly stated decisions, the sixth was my own stale-head error, and the regression guard for #1184 is now real. Item (a) is a two-word assertion tightening if you want it in this PR; otherwise it can ride along next time.

PR #1221, opened 2026-08-10
feat(ai): show which step of a multi-round wait is in flight (#1220)

HankYuLinksys:
Fixes #1220.

## Summary

While the assistant works, the loading bubble cycles `Thinking / Analyzing / Processing …`. Those phrases are driven by a timer — they rotate at a fixed interval whether progress is being made or not.

That is fine for a short wait, and misleading for the common one: a single question often takes **several LLM round-trips**, the assistant asking for data before it can answer. Those intermediate responses are never rendered (a message with no `TextBlock` returns `SizedBox.shrink()`, and they carry only `tool_use`). So a multi-round wait is indistinguishable from a hang.

The controller already knew which round it was on and simply never said so. It now publishes it, and the bubble reads `Thinking… (2/5)`.

## Decisions worth reviewing

**Suppressed on round one.** A `(1/5)` next to a spinner tells the user nothing they cannot already see; showing it would add noise to the case that is already fine.

**Digits only — no new l10n keys.** The alternative, naming the tool in flight ("Checking connected devices…"), reads better but costs 15 keys × 26 locales *and* goes stale the moment the tool list is reorganised (mainline 2-2). The round number stays correct through that.

**`_processConversation` is now a thin wrapper around `_runConversation`.** Only so a `finally` can clear the round. The body has **eleven exit paths**; clearing at each is the bookkeeping that gets forgotten when a twelfth is added, and a stuck counter would leave the view reporting work that has stopped. Wrapping avoids reindenting a method whose extraction is tracked separately — the body is byte-for-byte unchanged.

## Why not streaming

This came out of investigating #1217, where I measured whether streaming would address the same complaint. It would not, and the reasoning is recorded there. The short version: the prompt requires the whole component tree in one message (measured at 186–531 characters, so a few hundred milliseconds), and the rounds that make a wait long carry **no renderable content at all** — so streaming has nothing to show during exactly the slow case.

What the user lacks is not partial output. It is whether anything is still happening.

#1217 is now blocked on mainline 2-1, which will make the output incremental and give streaming something to render.

## Verification

| Check | Result |
|---|---|
| `test/ai/orchestrator/router_chat_controller_test.dart` | 35/35 |
| `flutter test test/ai/ test/page/ai_assistant/` | 115/115 |
| `./run_tests.sh` (full suite) | **3579/3579** |
| `flutter analyze` | 0 issues on changed lines |

Three tests added: idle reports 0, a two-round exchange publishes both rounds and returns to 0, and a **failed** exchange still resets. Mutation-checked: removing the reset fails two tests, not publishing the round fails one.

## Notes for reviewers

- **The view-side display logic has no test.** The round arithmetic lives in the controller where it is covered, but the "only show when `round > 1`" condition and the `(2/5)` formatting are in `_AnimatedThinkingText` — enforced by logic, not by a test. This repo has no widget-test infrastructure for this view; called out as a deliberate trade-off rather than left to be discovered.
- Two `prefer_const_literals` infos in the test file are pre-existing (`git blame` → `f75a4ff6`, #995) and not touched here.
- `maxRounds` is exposed on the controller so the view shows progress against a known bound rather than an open-ended counter. The underlying `_maxLoops` was a local `const` inside the loop; it is now a file-level constant so both the loop and the getter cite one source.

### DIFF
--- a/lib/ai/orchestrator/router_chat_controller.dart
+++ b/lib/ai/orchestrator/router_chat_controller.dart
@@ -6,6 +6,12 @@ import 'package:privacy_gui/ai/ai_logging.dart';
 import 'package:privacy_gui/ai/abstraction/_abstraction.dart';
 import 'package:privacy_gui/ai/prompts/router_system_prompt.dart';
 
+/// Most LLM round-trips one exchange may take before it is cut off.
+///
+/// A fuse against a model that keeps requesting data instead of answering, not
+/// a budget the user is meant to feel.
+const int _maxLoops = 5;
+
 /// Results accumulated for one assistant turn, owned by the exchange that
 /// opened it.
 ///
@@ -144,6 +150,22 @@ class RouterChatController extends ChangeNotifier {
   /// Whether the controller is currently loading.
   bool get isLoading => _state.isLoading;
 
+  /// Which round-trip of the current exchange is in flight, or 0 when idle.
+  ///
+  /// Exists so the view can say something true while the user waits. A single
+  /// question can take several LLM calls — each round asks for data the previous
+  /// round revealed it needed — and the intermediate responses are invisible
+  /// (they carry only `tool_use`, which is not for the user to read). Without
+  /// this the wait is indistinguishable from a hang.
+  int get currentRound => _currentRound;
+  int _currentRound = 0;
+
+  /// The most rounds one exchange may take before it is cut off.
+  ///
+  /// Exposed so the view can show progress relative to a known bound rather
+  /// than an open-ended counter.
+  static const int maxRounds = _maxLoops;
+
   /// Whether there's an error.
   bool get hasError => _state.hasError;
 
@@ -246,6 +268,21 @@ class RouterChatController extends ChangeNotifier {
   /// - Detecting confirmation requirements
   /// - Looping for multi-turn tool use
   Future<void> _processConversation() async {
+    // Wrapped rather than guarded inline: the body has eleven exit paths, and
+    // clearing the round at each one is the kind of bookkeeping that gets
+    // forgotten when a twelfth is added. A `finally` here covers every return
+    // and every throw, without reindenting the body.
+    try {
+      await _runConversation();
+    } finally {
+      if (_currentRound != 0) {
+        _currentRound = 0;
+        notifyListeners();
+      }
+    }
+  }
+
+  Future<void> _runConversation() async {
     // This exchange owns its own batch, stamped with the generation it belongs
     // to. Every await below is a point where the user can reset the
     // conversation; because results live on the batch rather than in shared
@@ -266,13 +303,17 @@ class RouterChatController extends ChangeNotifier {
     if (_isStale(batch)) return;
 
     var loopCount = 0;
-    const maxLoops = 5;
+    const maxLoops = _maxLoops;
 
     _log(
         '_processConversation: starting loop (${tools.length} tools available)');
 
     while (loopCount < maxLoops) {
       loopCount++;
+      // Published before the call so the view can describe the wait it is
+      // about to sit through, not the one that just ended.
+      _currentRound = loopCount;
+      notifyListeners();
       _log('_processConversation: loop $loopCount/$maxLoops');
 
       // Declared outside the try so the catch blocks can still surface a
--- a/lib/ai/orchestrator/router_chat_controller.dart
+++ b/lib/ai/orchestrator/router_chat_controller.dart
@@ -6,12 +6,6 @@ import 'package:privacy_gui/ai/ai_logging.dart';
 import 'package:privacy_gui/ai/abstraction/_abstraction.dart';
 import 'package:privacy_gui/ai/prompts/router_system_prompt.dart';
 
-/// Most LLM round-trips one exchange may take before it is cut off.
-///
-/// A fuse against a model that keeps requesting data instead of answering, not
-/// a budget the user is meant to feel.
-const int _maxLoops = 5;
-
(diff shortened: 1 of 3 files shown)


AustinChangLinksys (approved):
🤖 **Automated Review — Round 1** · `c12c4d1..17c088e` (full)

**Verdict: ✅ APPROVE** — Correct implementation, clean round-counter lifecycle; one minor race window in `clearConversation` worth a follow-up.

#### Standards

|---|---|---|---|
| ⚠️ | High | `router_chat_controller.dart:851` | `clearConversation()` does not reset `_currentRound`; calls `notifyListeners()` before `finally` fires, briefly showing stale round in the loading bubble |
| 💡 | High | `router_chat_controller_test.dart:218` | `clearConversation` test group missing `expect(controller.currentRound, 0)` assertion |
| 💡 | High | `router_chat_controller_test.dart:350` | `confirmPendingAction` / `cancelPendingAction` paths not covered by round-reset tests (behavior correct, test gap) |
| 💡 | Med | `router_assistant_view.dart:993` | `_AnimatedThinkingText` has a hard static dependency on `RouterChatController.maxRounds`; passing it as a constructor param would decouple the widget |
| 💡 | Med | `router_chat_controller.dart:9` | `_maxLoops` is a file-level private const used only by `RouterChatController`; moving it inside the class alongside `maxRounds` would co-locate related constants |
| 💡 | Med | `router_assistant_view.dart:918` | Consider adding a comment that `widget.round` is live-read in `build()` to guard against a future refactor that copies it into a State field |

#### Spec

|---|---|---|---|
| ⚠️ | Med | `router_chat_controller_test.dart:350` | AC-3 ("round returns to idle when it fails") — confirm/cancel continuation paths not covered by round-reset tests |
| 💡 | High | PR description | AC-5 — PR body lists test counts but no CI-run artifact attached; recommend confirming `./run_tests.sh` clean before merge |
| 💡 | Med | `router_assistant_view.dart:993` | AC-1 edge: when round reaches (5/5) the next action is a hard error; no UX affordance distinguishes "last step" from "about to fail" — out of scope per issue but worth noting |

<details>

**`clearConversation()` does not reset `_currentRound`** — High confidence — `router_chat_controller.dart:831–851`

```dart
void clearConversation() {
  _state = ConversationState.initial();
  _conversationGeneration++;
  _pendingConfirmation = null;
  _batch = null;
  _lastUserMessage = null;
  _routerTools = null;
  // ... token counters zeroed ...
  _initializeSystemPrompt();
  notifyListeners();      // <- _currentRound is still non-zero here
  // <- _currentRound = 0 is MISSING
}
```

**Why it matters:** When the user clears the conversation mid-LLM-call, `clearConversation()` increments `_conversationGeneration`, which causes `_runConversation` to detect a stale batch and return early. However, the `finally` block in `_processConversation` (line 278) only fires *after* `_runConversation` actually returns — which requires at least one more `await` to yield. Between the `notifyListeners()` on line 851 and the `finally` clearing `_currentRound`, the View is in a state where `isLoading == false` (because `_state` was reset to `initial`) but `currentRound != 0`. Any listener that reads both fields in the same frame sees an inconsistency.

**Fix:** Add `_currentRound = 0;` before `notifyListeners()` in `clearConversation()`:
```dart
_currentRound = 0;      // ensure no stale round after a clear
_initializeSystemPrompt();
notifyListeners();
```

**Spec note (AC-3):** "The round returns to idle when the exchange ends — including when it fails." A user-initiated clear is effectively a forced end; the round counter should reflect idle immediately.

</details>

<details>
<summary>Suggestions / What looks good</summary>

**What looks good:**

- **`_processConversation` / `_runConversation` wrapper pattern** — The single-exit `finally` idiom cleanly handles all 11 exit paths without per-path bookkeeping. Correct design.
- **`maxRounds` as `static const`** — Exposing the known bound lets the View show `(N/5)` without hardcoding the denominator; avoids magic numbers in the View layer.
- **`round > 1` suppression** — Not showing `(1/5)` next to a spinner is the right UX call; documented clearly in both the PR body and commit message.
- **`_controller?.currentRound ?? 0` null-safety** — The View correctly fallbacks to 0 when the controller is not yet attached.
- **Test mutation coverage** — Three tests cover idle, successful two-round exchange (round published + reset), and failed exchange (reset confirmed); commit message explicitly verifies mutations.

**Suggestions:**

1. **`clearConversation` test** (`router_chat_controller_test.dart:218`): Add `expect(controller.currentRound, 0)` to the existing `resets all state` test to lock in the behavioral contract.

2. **`confirmPendingAction` round-reset test**: A round is published during the confirmation flow (`_runConversation` sets `_currentRound` before the confirmation pause). After `confirmPendingAction()` completes (or fails), `currentRound` should be 0. This path is not currently tested.

3. **`_AnimatedThinkingText.maxRounds` parameter** (`router_assistant_view.dart:993`): Pass `RouterChatController.maxRounds` in from the call site rather than referencing the static directly inside the widget. Keeps the widget decoupled from controller internals.

4. **`_maxLoops` placement** (`router_chat_controller.dart:9`): Moving it inside the class as `static const int _maxLoops = 5` puts it next to `static const int maxRounds = _maxLoops`, making the relationship immediately visible.

5. **`_AnimatedThinkingText` live-read comment**: `widget.round` is correctly read live in `build()`. A brief comment (`// widget.round is read live in build(); do not copy to State`) would guard against a future refactor that accidentally caches it in a State field, which would introduce stale display data.

</details>

*Cross-reviewed by three independent agents (security+correctness / architecture+maintainability / requirements-compliance). Automated — please sanity-check before merge.*

Issue #1183, opened 2026-07-29
Testing: defensive RenderFlex-overflow gate for dashboard cards


AustinChangLinksys:
## Background

While reviewing #1174 (fix for #1145 — Network Health legend labels), we confirmed the Errors/Loss legend **overflows** at several locale × width combinations, yet CI never flagged it. Investigating why surfaced a structural gap in our overflow detection worth fixing on its own.

## Branching and PR structure

**All work for this epic is based on `gate/1183-overflow-gate` (PR #1248), which
is the only PR that targets `dev-2.7.0`.** Branch from it, merge back into it.
Every child ticket now carries this in its own `## Branch` section.

Not a convention — a hard constraint. The gate, its fixture, its probe and the
design of record are all introduced *by* #1248 and are absent on `dev-2.7.0`:
`test/page/dashboard/cards/dashboard_card_overflow_test.dart`,
`test/fixtures/known_overflows.json`,
`test/util/dashboard/dashboard_card_probe.dart`,
`doc/dashboard/dashboard_density_design.md`. An acceptance criterion of the form
"N coordinates are removed from `known_overflows.json` and the gate passes"
cannot be executed off `dev-2.7.0` at all.

```
dev-2.7.0
 └── #1248  gate/1183-overflow-gate                  ← the gate itself; all work lands here
      ├── #1249  gate/test-1183-dashboard-overflow   ← MERGED (04938ad9)
      │    └── #1242  gate/sweep-1231                ← MERGED (65563d6c)
      ├── #1255  gate/sweep-1252                     ← MERGED (8f74bfe2) — carried #1252 + #1227
      ├── #1257  gate/1228-ethernet-ports            ← MERGED (52973a88)
      ├── #1263  gate/1247-arrow-icons               ← MERGED (32ac1849)
      ├── #1264  gate/sweep-1258                     ← MERGED (4749def2)
      ├── #1269  fix/1234-card-overflow              ← MERGED (4a38116b) — carried #1234/#1235/#1236/#1237/#1229/#1251
      ├── #1273  fix/1183-ratchet-truth              ← MERGED (20402ab6) — bookkeeping, no `lib/` change
      ├── #1277  fix/1230-firewall-overflow          ← MERGED (c2280288)
      └── #1280  fix/1238-connected-devices-overflow ← MERGED (43520832)
```

**Both tracks are complete.** Track A took the allowlist to 0 coordinates across
[...]

PR #1242, opened 2026-08-11
fix(dashboard/_shared): size UspInfoRow label from real available width (#1231)

AustinChangLinksys:
Closes #1231. Parent #1183.

Stacked on #1249 → #1248 → `dev-2.7.0`. **Review those first.**

> **This body replaces the original.** The original's central claim ("Behavior
> is preserved at normal widths") was wrong, and its verification numbers were
> measured 11 commits behind its own declared base. Both are corrected below,
> and the extra commits that fix the regression are the reason this PR grew.
> See the measurement comment for the full table.

## The bug this PR started from is real

`UspInfoRow` — the shared label–value row used across USP dashboard cards and
several settings pages — sized its label column with
`context.colWidth(labelColumns)`, which ui_kit derives from **screen** width
(`layout_extensions.dart`), not from the width the row is actually given. Inside
a shrunken dashboard card the two are unrelated.

Because the structure is `SizedBox` + `Expanded`, a compressed value never
raises a `RenderFlex` overflow, so the text is silently clipped by the card
surface's `Clip.antiAlias`. Users see cut-off text; the overflow gate sees
nothing. That is why this needs its own assertions and contributes zero
coordinates to `known_overflows.json`.

## The first fix traded the bug for a narrower one

Replacing `colWidth(2)` with `constraints.maxWidth * labelColumns / 12` keeps a
fraction that was **calibrated against the screen** and applies it to a
card-sized row. At a 1400px screen 2/12 is ~150px — enough for one line. Inside
a 288px row it is 48px, narrower than the 74.6px `"IP Address"` needs.

Measured with real fonts at a fixed 1400px surface (line height 20px):

| row width | `colWidth(2)` (before) | `maxWidth*2/12` (first fix) | this PR |
|---|---|---|---|
| 191, short label | 150 fits / value 41 → **3 lines** / h60 | 31.8 → **3 lines** / value fits / h60 | 76.4 fits / 114.6 fits / **h20** |
| 191, long label | 150 → 2 lines / h40 | 31.8 → **8 lines** / **h160** | ellipsized 1 line / **h20** |
| **288** | 150 fits / 138 fits / **h20** | 48 → **2 lines** / **h40** | 115.2 fits / 172.8 fits / **h20** |
| 480 | 150 fits / h20 | 80 fits / h20 | 150 fits / h20 |
| 900 | 150 fits / h20 | 150 fits / h20 | 150 fits / h20 |

288px is a **regression against the code it replaced**, at a width the product
ships. So "behavior is preserved at normal widths" was the opposite of true: the
first fix helped 191px and broke 288px.

## What this PR settles on

Keep a fraction — sibling rows must share a column width so their values line
up, which intrinsic sizing would break — but recalibrate it to the row and cap
it:

- `_kLabelShare = 0.4` — the label may claim 40% of the row. At 191px that is
  76.4px, which `"IP Address"` (74.6px) fits, leaving the value 114.6px for a
  108.6px address.
- `_kLabelMaxWidth = 150.0` — so a wide row does not open a growing gutter
  before the value. Sits inside the 130–187px band `colWidth(2)` produced across
  desktop breakpoints (measured: 360→156, 480→216, 600→276, 768→164, 1024→142.7,
  1280→130, 1440→156.7, 1920→186 — note it is **not monotonic** in screen
  width), so rows that were already wide enough keep roughly the column they had.

Apply the degradation shape settled in #1226 and applied in #1233 while here:
the label takes `maxLines: 1` + ellipsis, the value takes neither. A clipped
label is recoverable from context; a half-shown statistic misinforms.

**Removed the `labelColumns` parameter.** No consumer ever passed it — verified
across all nine call sites. It was Speculative Generality, and it was also the
vehicle for the miscalibration.

No layout scope wrapper is introduced (design of record §2.2).

## The test file was rewritten, and it now actually runs

Two independent problems:

- It was tagged **`@Tags(['ui'])`**, which `run_tests.sh` excludes
  (`--exclude-tags="golden||loc||ui"`), so it **never ran in the PR gate**. Now
  `dashboard-card`, which does gate.
- Its assertions were geometric — `SizedBox.width`, "value area > half the row"
  — which cannot see whether the text *fits*. A 48px column passes "label is
  2/12 of the row" while wrapping the label onto two lines.

It now measures rendered line count (paragraph height vs. its height at
unbounded width) and loads **real fonts**: without `loadAppFonts()` every glyph
is a fixed-width block, so every "does this fit" answer is fiction. 10 tests,
and **3 of them fail against the first fix's implementation** — verified by
reverting the widget and re-running. The two that stay green are exactly the
part the first fix got right (row-derived, not screen-derived).

Also fixed a leak the round-1 review caught: `pumpRow` set `surfaceSize`,
`physicalSize` and `devicePixelRatio` but reset only the first, leaving a
1400x800 view at DPR 1.0 for every later test in the process. (The review cited
`dhcp_card_test_harness.dart` for the pattern — that file does not exist — but
the leak itself was real.)

## Acceptance criteria

- [x] Label column sized from the row's given width, not screen width
- [x] At ~191px the value keeps >half the row and renders on one line — pinned
      by test, not by golden
- [x] No layout scope wrapper introduced
- [x] All nine consumers unchanged and analyze clean; `labelColumns` confirmed
      dead before removal
- [x] `test/fixtures/known_overflows.json` **unchanged** — this component
      contributes zero baseline coordinates
- [x] #1231's AC "verified by eye or golden" is met by neither: golden baselines
      are gitignored (`.gitignore:92`) and this project runs no pixel diff, so
      golden cannot verify anything here. Substituted measurement + assertions,
      which is strictly stronger. See #1241.

## Verification

```
$ fvm flutter test test/page/_shared/components/usp_info_row_test.dart
  10/10 passed          (3 fail against the pre-fix widget — mutation-checked)

$ fvm flutter test --tags dashboard-card
  1710/1710 passed      (1700 inherited from #1249, + 10 new; fixture untouched)

$ ./run_tests.sh
  5049/5049 passed

$ fvm dart format / fvm flutter analyze <2 changed files>
  clean / No issues found
```

The original body reported `+1493 -151` for the gate. That number was measured
against a merge-base of `c3e40e0b`, **11 commits behind the declared base** —
before `9b848994` rebuilt the overflow baseline from measurement. It described a
tree that no longer exists. This branch has since been merged up to its base and
re-measured.

## Known limits

- The 191px case clears by **1.8px** (76.4px column, 74.6px label). That is
  inside the ±2px tolerance the overflow gate itself allows for mac-vs-ubuntu
  rasterizer differences, so it is conceivable this ellipsizes on CI and not
  locally. It is now pinned by a test at all three widths, so it would fail
  loudly rather than ship a silently truncated label — but reviewers should know
  the margin is thin, and widening `_kLabelShare` past ~0.43 would push the
  *value* under its own floor instead. 191px simply does not have room for both
  with comfort.
- Non-English locales with longer labels will ellipsize at 191px. That is the
  intended degradation, not an oversight — but it is asserted only for one
  synthetic long label, not swept across all 26 locales.
- No consumer-level render test (round-1 review W-4). The nine call sites are
  covered only in that the dashboard ones now run through the overflow gate.
- `labelColumnWidth()` finds the `SizedBox` by type — white-box, and it breaks
  if the sizing mechanism is refactored. Kept because the alternative measures
  nothing useful; flagged so the next person is not surprised.
- `pumpRow` is a third near-identical pump harness in this suite (round-1 review
  W-3). Not consolidated here; that is a suite-wide change and does not belong
  in a fix PR.

### DIFF
--- a/lib/page/_shared/components/usp_info_row.dart
+++ b/lib/page/_shared/components/usp_info_row.dart
@@ -3,9 +3,13 @@ import 'package:ui_kit_library/ui_kit.dart';
 
 /// A label–value row used throughout the USP Dashboard cards.
 ///
-/// Uses the UI Kit 12-column grid system for responsive label sizing.
-/// [labelColumns] controls how many grid columns the label occupies
-/// (default 2, calculated via `context.colWidth()`).
+/// The label column is sized from the width the row is actually given
+/// (read via a [LayoutBuilder]), not from screen width. [labelColumns]
+/// controls the fraction of a nominal 12-column grid the label occupies
+/// (default 2 of 12). Sizing against the real available width keeps the
+/// value legible when the row lives inside a shrunken card, where a
+/// screen-derived width would over-claim the label column and clip the
+/// value against the card surface with no overflow raised.
 class UspInfoRow extends StatelessWidget {
   final String label;
   final String value;
@@ -22,15 +26,20 @@ class UspInfoRow extends StatelessWidget {
   Widget build(BuildContext context) {
     return Padding(
       padding: const EdgeInsets.only(bottom: AppSpacing.md),
-      child: Row(
-        crossAxisAlignment: CrossAxisAlignment.start,
-        children: [
-          SizedBox(
-            width: context.colWidth(labelColumns),
-            child: AppText.labelLarge(label),
-          ),
-          Expanded(child: AppText.bodyMedium(value)),
-        ],
+      child: LayoutBuilder(
+        builder: (context, constraints) {
+          final labelWidth = constraints.maxWidth * labelColumns / 12;
+          return Row(
+            crossAxisAlignment: CrossAxisAlignment.start,
+            children: [
+              SizedBox(
+                width: labelWidth,
+                child: AppText.labelLarge(label),
+              ),
+              Expanded(child: AppText.bodyMedium(value)),
+            ],
+          );
+        },
       ),
     );
   }
--- a/lib/page/_shared/components/usp_info_row.dart
+++ b/lib/page/_shared/components/usp_info_row.dart
@@ -1,25 +1,41 @@
+import 'dart:math' as math;
+
 import 'package:flutter/material.dart';
 import 'package:ui_kit_library/ui_kit.dart';
 
+/// Fraction of the row the label column may claim.
+///
+/// At the narrowest card realization (191px) this leaves the value ~115px,
+/// enough for the longest common value on one line. The split favours the
+/// value on purpose: per density design §2.10a the label is a name and the
+/// value is the payload, so the label is what yields.
+const double _kLabelShare = 0.4;
+
+/// Ceiling for the label column, so a wide row does not open an absurd
+/// gutter before the value. Sits inside the 130–187px band that the old
+/// screen-derived `context.colWidth(2)` produced across desktop breakpoints,
+/// so rows that were already wide enough keep roughly the column they had.
+const double _kLabelMaxWidth = 150.0;
+
 /// A label–value row used throughout the USP Dashboard cards.
 ///
-/// The label column is sized from the width the row is actually given
-/// (read via a [LayoutBuilder]), not from screen width. [labelColumns]
-/// controls the fraction of a nominal 12-column grid the label occupies
-/// (default 2 of 12). Sizing against the real available width keeps the
-/// value legible when the row lives inside a shrunken card, where a
-/// screen-derived width would over-claim the label column and clip the
-/// value against the card surface with no overflow raised.
+/// The label column is sized from the width the row is actually given (read
+/// via a [LayoutBuilder]), never from screen width: inside a shrunken card
+/// the two are unrelated, and a screen-derived column over-claims and clips
+/// the value against the card surface without raising an overflow.
+///
+/// Degradation follows the shape settled in #1226 and applied in #1233 — the
+/// label yields first with a one-line ellipsis, while the value never shrinks
+/// and never ellipsizes. A clipped label is still recoverable from context; a
+/// half-shown value misinforms.
 class UspInfoRow extends StatelessWidget {
   final String label;
   final String value;
-  final int labelColumns;
 
   const UspInfoRow({
     super.key,
     required this.label,
     required this.value,
-    this.labelColumns = 2,
   });
 
   @override
@@ -28,13 +44,18 @@ class UspInfoRow extends StatelessWidget {
       padding: const EdgeInsets.only(bottom: AppSpacing.md),
       child: LayoutBuilder(
         builder: (context, constraints) {
-          final labelWidth = constraints.maxWidth * labelColumns / 12;
+          final labelWidth =
+              math.min(constraints.maxWidth * _kLabelShare, _kLabelMaxWidth);
           return Row(
             crossAxisAlignment: CrossAxisAlignment.start,
             children: [
               SizedBox(
                 width: labelWidth,
-                child: AppText.labelLarge(label),
+                child: AppText.labelLarge(
+                  label,
+                  maxLines: 1,
+                  overflow: TextOverflow.ellipsis,
+                ),
               ),
               Expanded(child: AppText.bodyMedium(value)),
             ],
--- a/test/page/_shared/components/usp_info_row_test.dart
+++ b/test/page/_shared/components/usp_info_row_test.dart
@@ -46,7 +46,11 @@ void main() {
     await tester.binding.setSurfaceSize(wideSurface);
     tester.view.physicalSize = wideSurface;
     tester.view.devicePixelRatio = 1.0;
-    addTearDown(() => tester.binding.setSurfaceSize(null));
+    addTearDown(() async {
+      tester.view.resetPhysicalSize();
+      tester.view.resetDevicePixelRatio();
+      await tester.binding.setSurfaceSize(null);
+    });
 
     await tester.pumpWidget(
       MaterialApp(
--- a/test/page/_shared/components/usp_info_row_test.dart
+++ b/test/page/_shared/components/usp_info_row_test.dart
@@ -1,0 +1,105 @@
+@Tags(['ui'])
+import 'package:flutter/material.dart';
+import 'package:flutter_test/flutter_test.dart';
+import 'package:privacy_gui/l10n/gen/app_localizations.dart';
+import 'package:privacy_gui/page/_shared/components/usp_info_row.dart';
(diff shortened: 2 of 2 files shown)


AustinChangLinksys (commented):
🤖 **Automated Review — Round 1** · `c3e40e0..9b0f6ef` (full)

**Verdict: ✅ APPROVE** — Core fix is correct and well-tested; one double-confirmed test-isolation Warning flagged for auto-repair.

#### Standards

|---|---|---|---|
| ⚠️ | 🟢 High | `test/page/_shared/components/usp_info_row_test.dart:38` | tearDown resets `setSurfaceSize` but not `physicalSize`/`devicePixelRatio`, leaking view state to subsequent tests |
| ⚠️ | 🟡 Med | `lib/page/_shared/components/usp_info_row.dart:29` | [Primitive Obsession] Magic number `12` (grid column count) hardcoded in build; no named constant |
| ⚠️ | 🟡 Med | `test/page/_shared/components/usp_info_row_test.dart:24` | [Duplicate Code] `pumpRow` is the 3rd near-identical pump-harness pattern in the test suite; no shared helper |
| 💡 | 🟡 Med | `test/page/_shared/components/usp_info_row_test.dart:57` | `labelBoxWidth` white-box finds `SizedBox` by type — breaks if label sizing mechanism is refactored |
| 💡 | ⚪ Low | `lib/page/_shared/components/usp_info_row.dart:29` | No guard for `constraints.maxWidth == infinity`; all current call sites are bounded but future use could panic |

#### Spec

|---|---|---|---|
| ✅ | 🟢 High | `usp_info_row.dart:28–32` | AC-1: label now sized from `constraints.maxWidth`, not `context.colWidth()` — PASS |
| ✅ | 🟢 High | `usp_info_row_test.dart:69–78` | AC-2: 191px test proves value retains >50% row width — PASS |
| ✅ | 🟢 High | `usp_info_row.dart:28` | AC-3: `LayoutBuilder` at the row, no layout scope wrapper introduced — PASS |
| ⚠️ | 🟡 Med | (no test file) | AC-4: No regression widget tests for existing consumers (WiFi status, ethernet dialog, IPv4/6 sections, stats); spec says "verify those still look right" but only oral claim in commit message |
| ✅ | 🟢 High | `test/fixtures/known_overflows.json` | AC-5: `known_overflows.json` diff is empty — PASS |

<details>

**W-1 — Incomplete tearDown: `physicalSize` / `devicePixelRatio` leak (🟢 High, double-confirmed by both agents)**

`test/page/_shared/components/usp_info_row_test.dart:35–38`

```dart
await tester.binding.setSurfaceSize(const Size(1400, 800));
tester.view.physicalSize = const Size(1400, 800);
tester.view.devicePixelRatio = 1.0;
addTearDown(() => tester.binding.setSurfaceSize(null));  // only resets surfaceSize
```

Three properties are set, but tearDown only restores one. `tester.view.physicalSize` and `tester.view.devicePixelRatio` persist for the duration of the test session, potentially causing subsequent widget tests in the same `flutter test` run to see `1400x800` instead of their expected viewport.

The established repo pattern (from `test/page/dhcp/views/components/dhcp_card_test_harness.dart:46–50`):
```dart
addTearDown(() async {
  tester.view.resetPhysicalSize();
  tester.view.resetDevicePixelRatio();
  await tester.binding.setSurfaceSize(null);
});
```

**Fix:** Replace the single-line tearDown in `pumpRow` with the three-line form above. (A `review-fix` card has been queued for automatic repair.)

---

**W-2 — [Primitive Obsession] Magic number `12` in `build()` (🟡 Med, single-audit)**

`lib/page/_shared/components/usp_info_row.dart:29`

```dart
final labelWidth = constraints.maxWidth * labelColumns / 12;
```

`12` is the total grid column count, a domain constant. If ui_kit_library's grid changes, or if a reader unfamiliar with the 12-column grid encounters this, the number is opaque. The `context.colWidth()` implementation that was replaced encapsulated this constant internally.

**Fix (optional):** Extract `static const int _kGridColumns = 12;` as a private constant, or check if ui_kit_library exports a `kGridColumns` constant.

---

**W-3 — [Duplicate Code] `pumpRow` is the 3rd near-identical pump helper (🟡 Med, single-audit)**

`test/page/_shared/components/usp_info_row_test.dart:24–54`

The pattern of `setSurfaceSize` + `physicalSize` + `devicePixelRatio` + `pumpWidget(MaterialApp(theme, l10n, Scaffold(...)))` + `pumpAndSettle` already exists in `dhcp_card_test_harness.dart` and `pnp_no_internet_view_test.dart`. This is the third copy.

**Fix (tech-debt backlog):** Consider a shared `test/helpers/widget_pump_harness.dart` utility. Not blocking merge.

---

**W-4 (Spec) — No regression tests for existing consumers (🟡 Med, single-audit)**

Spec AC-4 requires verifying that all existing consumers (statistics, internet settings IPv4/IPv6/optional, WiFi status, ethernet port dialog, AI info row) still render correctly. The PR relies on the oral claim in the commit message ("All consumers use the default `labelColumns:2`; none pass a custom value") with no automated verification.

At normal page widths the new formula produces values close to the old `context.colWidth()` result, so behavioral regression is unlikely. However the spec says "verify those still look right."

**Fix (low urgency):** Add at least one consumer render test (e.g. for `ethernet_port_detail_dialog` which has a bounded `SizedBox(width: kDefaultDialogWidth)` wrapper), or explicitly document in the PR that the existing integration test suite covers these consumers.

</details>

<details>

- **Core fix is correct:** `LayoutBuilder` at the row reads `constraints.maxWidth` — exactly what the spec prescribes. The formula `maxWidth * labelColumns / 12` is semantically correct for the 12-column grid.
- **No API breakage:** All call sites use the default `labelColumns: 2` (confirmed by search). The change is purely internal to `build()`.
- **Tests pin the fix well:** Three `testWidgets` cover (1) correct fraction at 191px, (2) proportional scaling as row width changes, and (3) value retains majority of row. The screen-vs-row distinction is explicitly documented in the test file.
- **`known_overflows.json` unchanged:** Confirmed via `git diff` — zero delta. Gate result is identical to baseline.
- **Doc comment updated:** The class doc comment now correctly describes the `LayoutBuilder` approach and explains the rationale (silent clip vs RenderFlex overflow).
- **Consistent with repo pattern:** `LayoutBuilder` for local constraint reading is already used in `usp_dhcp_active_leases_card.dart` and `usp_dhcp_reservations_detail_card.dart` (#1140). This PR applies the same pattern one level up at the shared component.
- **`dart format` and `flutter analyze` clean** (per PR verification block).

</details>

PR #1248, opened 2026-08-12
test(dashboard): PR-gating RenderFlex-overflow gate for dashboard cards (#1183)

AustinChangLinksys:
Closes #1183

Adds a RenderFlex-overflow gate that actually blocks PRs, plus the measurement
tooling and the design record behind it.

The harness landed first and on its own — the commits up to `edfbf5ac` contain no
`lib/` changes at all, so the gate's go-live baseline is a measurement of `main`
rather than of anything this branch did. **The fixes the gate surfaced are then
stacked on the same branch**, ticket by ticket, each re-measuring the sweep before
touching the allowlist: #1226, #1231, #1233, #1227, #1252, #1228, #1229, #1234,
#1235, #1236, #1237, #1247, #1258, #1230 and #1238 — Track A in full.
(#1231 and #1252 are the two that moved no coordinates — #1231 recalibrated the
`UspInfoRow` label column, #1252 fixed a row outside the gate's scan domain, and
in both cases the re-measurement confirming it is exactly why the allowlist may
not shrink.)
Then Track B (#1232, #1239, #1240 and the four readability batches it split into),
the three production defects the branch had only measured (#1275, #1286, #1285),
the four tickets that came out of measuring the last two cards (#1266, #1270,
#1271, #1267), the three follow-ups those measurements owed (#1296, #1297, #1298),
#1245's Article XIV migration and #1253's localization policy — so **the epic
lands whole**, on ui_kit **v2.39.0**.
Read the harness commits first if you want to review the mechanism separately from
the fixes.

## Why the existing pipeline could not catch this

#1145's Network Health legend overflow slipped through, and it wasn't bad luck.
Three separate structural reasons, all verified against the current pipeline:

1. **Goldens are excluded from the PR test command.** `run_tests.sh` runs with
   `--exclude-tags="golden||loc||ui"`, so no golden test can gate a PR.
2. **Overflow is a warning there, not a failure.** The golden pipeline records
   overflow and moves on; a screenshot with a yellow-and-black stripe is still a
   passing screenshot.
3. **Coverage holes.** Goldens shoot the *default tab*, at *fixed widths*, and
   not every card has one. #1145's overflow needed a non-default tab at a
   non-sampled width.

## What this adds

**The gate** — `test/page/dashboard/cards/dashboard_card_overflow_test.dart`,
tagged `dashboard-card`. That tag is deliberately **not** in `run_tests.sh`'s
exclude list, so the existing PR command picks it up and a failure blocks the
[...]

Issue #1245, opened 2026-08-12
[Prefactor] De-duplicate the dashboard legend dot/entry — Article XIV decision needed


AustinChangLinksys:
## Parent

#1183

Raised from #1233, which was explicitly told not to do this: *"Do not extract a shared legend component. That would be a new shared widget, which requires approval under Article XIV — and this ticket must not be blocked on that conversation. Fix the six rows in place."* Also: *"the private colour-dot widget is duplicated verbatim in four files ... De-duplicating it would be a reasonable prefactor and is not itself overflow-related — but the same Article XIV question applies, so raise it rather than doing it here."*

This is that raise. Nothing here is overflow work; #1183's Track A does not depend on it.

## Branch

**Base: `gate/1183-overflow-gate` (PR #1248). Not `dev-2.7.0`.**

This epic is consolidated onto #1248's branch — #1248 is the only PR that
targets `dev-2.7.0`. Branch from `gate/1183-overflow-gate`, and merge back into
it.

Not a preference. The gate, its fixture, its probe and the design of record are
all introduced *by* #1248 and are **absent on `dev-2.7.0`** (verified with
`git cat-file -e origin/dev-2.7.0:<path>`):

- `test/page/dashboard/cards/dashboard_card_overflow_test.dart`
- `test/fixtures/known_overflows.json`
- `test/util/dashboard/dashboard_card_probe.dart`
- `doc/dashboard/dashboard_density_design.md`

So an acceptance criterion of the form "N coordinates are removed from
`known_overflows.json` and the gate passes" cannot be executed off `dev-2.7.0`
at all. Basing there also re-measures coordinates the landed tickets already
cleared, producing a fixture that conflicts with them and reads as a regression.

## What exists today

Two shapes, duplicated across four dashboard cards.

**`_LegendDot`** — byte-for-byte identical in four files:

- `lib/page/dashboard/views/components/usp_network_health_card.dart:451`
- `lib/page/dashboard/views/components/usp_system_status_card.dart:541`
- `lib/page/dashboard/views/components/usp_traffic_analysis_card.dart:683`
- `lib/page/dashboard/views/components/usp_device_analytics_card.dart:448`

**The legend entry** (dot + gap + `Flexible` label) — three copies, and after #1233 they are deliberately *not* identical:

| Site | Name | Label degradation |
|---|---|---|
| `usp_system_status_card.dart:563` | `_StatLegendEntry` | per-entry `ellipsize` flag |
[...]

PR #1249, opened 2026-08-12
fix(dashboard): clear the legend-row overflow cluster, prove the gate's worst-case claim (#1225, #1226, #1233)

AustinChangLinksys:
Closes #1225
Closes #1226
Closes #1233

Stacked on #1248 — **review that one first**; this PR's base is its branch, so
the diff shown here is only the six commits on top. #1248 lands the overflow
gate with 560 known exceptions; this PR takes that ratchet down to **379
coordinates across 27 keys** and hardens the gate's central correctness claim.

## 1. The gate's exhaustiveness argument was an assertion (#1225)

The gate pumps one width per column span and calls that exhaustive, on the
grounds that overflow is monotonic in width. That argument only holds if the
width really is the narrowest — and it was chosen by scanning a hand-written
list of 19 screen widths.

`narrowestRealizationOf()` replaces the list by enumerating every integer screen
width from the supported floor up. Card width is piecewise-linear and increasing
within each (columns, margin) regime, so every regime's left edge is now visited
and the worst case holds by construction.

The sample wasn't merely unproven, it was **lossy under a raised floor**: it held
no width in `602..904`, so `MIN_SCREEN=602` fell through to 1241px and reported a
3-column card as 198.25px — 6.5px wide of the real 191.75px.

- Behavioural no-op at the default floor, **verified not assumed**: same 560
  coordinates, same 36 keys, and the two sorted hit sets diff byte-identically.
  `known_overflows.json` untouched by this commit.
- The guarantee is exact over integer widths but not the continuum: breakpoints
  are exclusive, so four regimes open just above an integer and their infimum is
  approached, not attained. Bounded at 0.5px (`kEnumerationSlackPx`, worst case
  span 4) — a quarter of the gate's 2px tolerance, so it cannot flip a verdict.
  Documented rather than papered over.
- `kMinSupportedScreenWidth` (320px) now names the product commitment from
  density design §2.3 that used to be an inline literal.
- New `dashboard_card_probe_test.dart` (28 tests) checks the search against an
  **independently derived continuum infimum**, not a re-enumeration of the
  implementation's own domain. Mutation-verified: starting the walk above 601px
  fails it.
- Frozen grid formulas (column count, margin, slot width, card width) untouched.

## 2. Traffic Analysis Monitor legend (#1226) — 49 coordinates

All 49 of the card's baseline coordinates came from one site: a `Row` of two
dot+label pairs, a `Spacer`, two byte totals and the interval label, with nothing
constrained. It becomes a `Wrap` with `spaceBetween` — identical rendering while
the content fits, and the totals drop to a second line instead of overflowing
when it doesn't. Card is now clean across 26 locales × 4 tabs; both its allowlist
keys are gone. **560 → 511 coordinates, 36 → 34 keys.**

Two things measurement contradicted in the ticket, both recorded in design §2.10
because #1233 inherits the shape:

- The default-layout break was **one locale**, not a general desktop break. A
  26-locale sweep at default span-6 widths found only `fr` overflowing (+92px @
  432, +44 @ 480, +28 @ 496, +12 @ 512). The conclusion stands — a shipped locale
  broken at 1024px is a normal-form bug — but the blast radius was narrower than
  #1226 implies. Caught because the first version of the regression test used
  English and passed before the fix existed.
- #1226's widths are mis-paired: a 1024px screen yields 480px (24px margin), not
  512px; 512px is the 1440px screen. The test covers 432/480/496/512, so the
  intent holds as a superset.

## 3. The same shape across six more legend rows (#1233) — 132 coordinates

**511 → 379, 34 keys → 27.** `network_health` 84 → 3, `system_status` 85 → 34.

Measured, not assumed: a temporary harness parsed the creation location out of
every incident's `fullLog` before and after, so each coordinate is attributed to
the site that actually causes it.

Three things #1226's shape did not say, each found by measuring a row:

- **`Flexible` is load-bearing, not decoration for the ellipsis.** A `Row` hands
  non-flex children *unbounded* width, so a bare label takes its full intrinsic
  width on one line and overflows however the enclosing `Wrap` arranges the
  entries. Wrapping in `Wrap` alone fixed nothing for the single-entry legends.
  #1226's row already had two `Flexible` labels, so this never surfaced there.
- **Ellipsis vs soft-wrap follows what the label *is*, not what the row is.**
  Bare series names keep #1226's one-line ellipsis (the colour identifies the
  series). Composed statistics get neither `maxLines` nor ellipsis: the ellipsis
  lands mid-number, and a half-shown statistic misinforms in a way a missing one
  does not.
- **One row deviates, deliberately.** #1226 pays for its extra run with height
  "because the chart above is `Expanded`, so it yields the height". Network
  Health's Health tab holds a fixed 120px gauge in that `Expanded`, so it yields
  nothing: a `Wrap` there traded the row's 26 right-overflows for 12 new
  *bottom*-overflows at the gauge centre (3 → 15) — a fix on paper only, and it
  would have shipped as one had the ratchet been edited to the predicted numbers
  instead of the measured ones. That row stays a one-line `Row` of `Flexible`
  lights; the reason is commented at the site and pinned by a test that fails if
  someone "restores consistency".

That last point makes **#1235 a functional dependency of #1233**, rather than the
conflict-avoidance both tickets claim: its 3 remaining gauge-centre coordinates
exist *because* the row was held to one line.

**Not done here, on purpose:** no shared legend component and no de-duplication
of the four verbatim colour-dot copies. Both need Article XIV approval and the
ticket must not block on that conversation — raised as **#1245**, carrying the
two constraints this work measured.

New `dashboard_legend_readability_test.dart` covers the two ACs the gate is
structurally blind to: the gate cannot tell a row that fits from a row that
truncated its content to nothing. Each of its three groups was verified to fail
under a mutation of the code it guards.

## 4. Direction markers drawn as icons, not Unicode characters

Not a ticketed item — found while fixing the #1226 legend row, which did both
things at once: its speed tiles used `AppIcon.font(Icons.arrow_upward)` while the
byte totals three rows below interpolated `'↑ '` into a string.

Measured the bundled font set with fontTools: **neither
`NeueHaasGrotTextRound-55Roman/75Bold` nor any of the nine fallbacks under
`assets/fonts/fallback/` maps U+2191, U+2193 or U+2192** — the union of all
eleven fonts has none of them. So whether the glyph appeared depended on a font
outside the declared set resolving it. It does resolve in the browser, which is
why this was never visible as a bug; but `pubspec.yaml` declares those fallbacks
precisely so glyph coverage doesn't depend on the environment ("Offline-first:
all locales must render without network").

Seven sites, all now `AppIcon.font` at size 12: the Traffic Analysis legend
totals, the duplicated row in `stats_traffic_monitor_section`, mesh rate chips in
`diagnostic_result_card` (`_buildChip` grew an optional icon; hardcoded `'Stale'`
became `loc().stale`, an ARB key that already existed), mesh detail rows in
`step_result_tile` (`_ResultDetail` grew a `segments` list of (icon, text)
because the rate detail carries two markers on one line), and
`topology_section` / `devices_section`, where the markers were built inside
String formatters, so the formatters became widget builders.

**Deliberately not changed — U+2192 (32 sites).** It's a "maps to" separator
(`8080 → 192.168.1.100:80`), not a direction marker, and nothing in the app draws
it as an icon, so there's no inconsistency to fix. Five of its UI sites read
through `PortForwardingRuleUIModel.portSummary` and
`PortTriggeringRuleUIModel.summary` — String getters on UI models — so icons
there would mean models returning widgets. Raised separately as **#1247**.

**Also not changed — logger and console output** (`usp_client.dart`,
`firmware_update_notifier.dart`, `pnp_service.dart`, `usp_topology_builder.dart`,
test console box-drawing, LLM-prompt markdown in `router_context_provider.dart`).
Icons are widgets; they can't go in a log string.

One caveat a reviewer should know: this reasoning was originally justified as
"terminal font coverage is unrelated to the app's", which is **wrong for
`usp_test_console_view.dart`** — that page renders log strings as Flutter `Text`
with the app's fonts. The impact is nil (zero arrow characters across its 106
`_log()` calls), so nothing is left broken, but the stated reason doesn't cover
that one file.

The #1226 regression test asserted `find` on `Text` starting with the arrow
character, which this change makes vacuous. It now matches the formatted byte
values by pattern and separately asserts both `Icon`s survive, so it still pins
the AC that the totals never shrink — verified non-vacuous by suppressing the
totals and watching it fail.

## Verification

- Overflow gate: **1710 / 1710 passed**
- Full functional suite (`./run_tests.sh`): **5049 / 5049 passed**
- `flutter analyze` on all 13 changed Dart files: clean
- `dart format`: clean

## Known limits

- `dashboard_legend_readability_test.dart` reads `Text.data` — the string the
  widget declares, not the glyphs painted. It therefore detects a legend entry
  that got *dropped*, but not one that ellipsized down to something illegible.
  Asserting on painted glyphs needs a different mechanism than the widget tree.
- Five of the eight changed `lib/` files have no test referencing them
  (`stats_traffic_monitor_section`, `diagnostic_result_card`, `step_result_tile`,
  `devices_section`, `topology_section`, +172/−49 between them). They're the
  §4 icon changes, they span four unrelated features, and none is a dashboard
  card — so the overflow gate doesn't cover them either. Flagged rather than
  silently shipped; happy to add widget tests here if reviewers want them in
  this PR rather than as follow-up.
- 379 coordinates remain allowlisted. Tracked in #1234 (`system_status` 34),
  #1235 (`network_health` 3) and the untouched cards inherited from #1248.
- Sub-1Mbps speeds in the two AI sections now read `500.0 Kbps` rather than
  `500 Kbps`. Consequence of deleting their local bps tiering in favour of the
  repo's single formatter (`UspFormatters.formatSpeed`, one precision for all
  tiers) — see `baf5a049`. Mbps and Gbps are byte-identical.
- #1231 fixed the shared `UspInfoRow`, but **five hand-rolled copies of the same
  screen-derived sizing are still live** — two of them inside the `wifi_status`
  dashboard card, in the same `Column` as the fixed rows. Found while verifying
  the round-1 review; the gate is structurally blind to this class, so it is not
  a regression this PR introduces. Filed as #1251, with #1252 (the
  `stats_traffic_monitor` twin of the #1226 row) and #1253 (`lib/ai` has zero
  `loc()` usage across 34 files).

### DIFF
--- a/.claude/skills/dashboard-overflow-gate/SKILL.md
+++ b/.claude/skills/dashboard-overflow-gate/SKILL.md
@@ -34,8 +34,11 @@ allowlist correctly, and keep the gate exhaustive when cards change.
 
 - Writing golden/screenshot tests → use `golden-test-review`.
 - Generic pre-PR checks → use `review-pr-readiness`.
-- Rewriting the probe/grid-geometry engine — that is intentionally frozen to
-  mirror production; changing it silently changes what the gate measures.
+- Rewriting the probe's grid-geometry **formulas** (column count, margin, slot
+  width, card width) — those are intentionally frozen to mirror production;
+  changing them silently changes what the gate measures. The *width-selection*
+  step (`narrowestRealizationOf`) is separate and is not frozen, but it is
+  covered by tests — see the enumeration invariant below.
 
 ## Step 0: Read the Implementation (BLOCKING PREREQUISITE)
 
@@ -47,6 +50,7 @@ below may drift:
 3. Runner — [tool/run_overflow_test.sh](../../../tool/run_overflow_test.sh)
 4. Probe + grid math + tab registry — [test/util/dashboard/dashboard_card_probe.dart](../../../test/util/dashboard/dashboard_card_probe.dart)
 5. Report generator (Status SSoT) — [test/util/dashboard/dashboard_overflow_report_generator.dart](../../../test/util/dashboard/dashboard_overflow_report_generator.dart)
+6. Width-selection tests — [test/util/dashboard/dashboard_card_probe_test.dart](../../../test/util/dashboard/dashboard_card_probe_test.dart)
 
 ## Architecture — Data Flow
 
@@ -54,6 +58,7 @@ below may drift:
 UspWidgetSpecs.all ──┐  (card registry: id + min/pref/max column span)
 UspWidgetFactory   ──┤→ dashboard_card_probe.dart
 cardTabIndexProvider ┘     • widthCasesFor(spec): narrowest real grid px per span
+                           •   (enumerated over 320px..2560px, not sampled)
                            • buildDashboardCardApp(): pumps ONE card at real
                              screen/card width, pinned tab, locale, real fonts
                            • kTabbedCardTabCounts: tabs to sweep per card
@@ -84,6 +89,14 @@ Key invariants (do not "fix" these):
   height is measured, so testing each span's narrowest grid width is exhaustive
   (~5× fewer tests, identical coverage). Grid geometry is mirrored from
   `UspSliverDashboardView` + ui_kit `AppLayoutConfig`.
+- **Narrowest is enumerated, not sampled** (#1225). `narrowestRealizationOf()`
+  walks every screen width from `kMinSupportedScreenWidth` (320px — a *product*
+  commitment, see density design §2.3) to `kMaxScannedScreenWidth`, so the
+  worst-case claim above holds by construction. Do not reintroduce a scan list:
+  the retired 19-width sample was lossy under `MIN_SCREEN` (a floor of 602px
+  reported a 3-column card 6.5px wider than reality). Lowering the 320px floor
+  **moves the baseline** and requires a deliberate re-baseline.
+  Covered by [dashboard_card_probe_test.dart](../../../test/util/dashboard/dashboard_card_probe_test.dart).
 - **Tag must stay `dashboard-card`.** `run_tests.sh` excludes `golden||loc||ui`;
   `dashboard-card` is NOT excluded, so it runs in the PR gate. Retagging it
   golden/loc/ui silently drops it from the gate.
@@ -97,7 +110,7 @@ Always run via `fvm flutter` (the script does). Output dir: `build/overflow_test
 |------|---------|---------|
 | `-l`, `--list` | List every registered card ID + column spans + tab count (from `UspWidgetSpecs.all`); runs nothing else | off |
 | `-d`, `--dump MODE` | `0`=no output (clean PR-gate mode) · `1`=Markdown · `2`=HTML + PNG · `3`=MD+HTML+PNG | `2` |
-| `-m`, `--min-screen PX` | Only test screen widths ≥ PX (fewer cases, faster) | `0` (all) |
+| `-m`, `--min-screen PX` | Raise the enumerated range's floor to PX, so each span's narrowest width is the narrowest at or above PX (narrower screens are skipped, not the widths themselves) | `0` = no filter (the 320px product floor still applies) |
 | `-c`, `--card CARD_ID` | Target one card (passed as `--name`, matches the card's test group) | all |
 | `-L`, `--locale LOCALES` | Comma-separated locale tags, e.g. `ru` or `ru,zh_TW` | all 26 |
 | `-o`, `--open` | Open the HTML report in the browser when done | off |
@@ -256,4 +269,6 @@ Markdown report (`-d 1`) is the same data as a flat bulleted list —
 | Multi-pumping to sweep widths/tabs in one test | Each case must be its own `testWidgets` (only the first overflow is reported per render object) |
 | Editing the grid constants to change results | They mirror production geometry on purpose; changing them changes what "overflow" means |
 | Removing an allowlist locale without fixing layout | The ratchet fails that exact test — fix the card first, then remove |
+| Replacing the width enumeration with a "faster" sampled list | Sampling makes the worst-case invariant an assertion again, and is lossy under `MIN_SCREEN`; enumeration is pure arithmetic and costs nothing next to the pumps |
+| Lowering `kMinSupportedScreenWidth` to "test more" | It is a product commitment (§2.3), and lowering it adds overflow coordinates — re-baseline deliberately, with the shift explained |
 ```
--- a/doc/dashboard/dashboard_density_design.md
+++ b/doc/dashboard/dashboard_density_design.md
@@ -1,6 +1,6 @@
 # Dashboard Card Density — Design Decisions
 
-**Last Updated: 2026-08-11** · Follow-up to #1183 · Status: **agreed; tickets #1225–#1240 published, implementation not started**
+**Last Updated: 2026-08-11** · Follow-up to #1183 · Status: **agreed; tickets #1225–#1240 published; #1225 implemented (not yet merged), rest not started**
 
 ## Purpose
 
@@ -184,9 +184,15 @@ wider than span 12 @ 320px = 288px.
 
 ### 1.6 The 191px "floor" was a sampling artifact
 
-`dashboard_card_probe.dart:154` scans a **hand-written list of 19 screen widths**
-(minimum 320). 191.4px is the narrowest value in that sample, not the geometric
-minimum. An exhaustive 1px sweep (240–2560px) gives:
+> **Superseded by #1225** (implemented; not yet merged). The probe now enumerates the supported range
+> (`narrowestRealizationOf`, 320–2560px) instead of scanning a list. The
+> measurements below are what justified that change and are kept as its record;
+> the no-op claim they predicted was confirmed on the full sweep — 560
+> coordinates across 36 keys, byte-identical before and after.
+
+`dashboard_card_probe.dart` used to scan a **hand-written list of 19 screen
+widths** (minimum 320). 191.4px is the narrowest value in that sample, not the
+geometric minimum. An exhaustive 1px sweep (240–2560px) gives:
 
 | Span | True narrowest | At screen |
 |---:|---:|---:|
@@ -208,8 +214,15 @@ width for **every one of the 12 spans**:
 
 The omitted widths are nobody's worst case: 375 / 390 / 430px all sit inside the
 4-column low-margin band where 320px already dominates. So replacing the sample
-(§2.7) buys a **guarantee, not a new baseline** — it is expected to be a
-behavioural no-op.
+(§2.7) bought a **guarantee, not a new baseline** — confirmed as a behavioural
+no-op when #1225 was implemented.
+
+**Where the sample *was* lossy: a raised floor.** With `MIN_SCREEN=602` the list
+held nothing in 602–904, so it fell through to 1241px and reported a 3-column
+card as 198.25px — 6.5px wider than the real 191.75px @ 602px. That path is not
+what the PR gate runs (the gate uses no floor), which is why the committed
+baseline never saw it, but it is the concrete reason "sampling happens to be
+correct here" was not a safe place to leave the invariant.
 
 Counter-intuitively, **mobile is not the narrowest case**: a 3-column card is
 212px at a 320px phone but 191.4px at a 601px tablet.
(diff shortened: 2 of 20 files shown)


AustinChangLinksys (commented):
🤖 **Automated Review — Round 1** · `4a2b3c8..65563d6` (full)

> **Note:** APPROVE event downgraded to COMMENT — GitHub's self-review restriction applies (agent runs under Austin's account). Standards and Spec findings are below for reference; no Critical issues found.


**Verdict: ✅ APPROVE** — No Critical issues; clean across all three ticket ACs. Four Warnings noted for tracking.

#### Standards (Reviewer A — Security & Correctness · Reviewer B — Architecture & Maintainability)

|---|---|---|---|
| ⚠️ | 🟡Med | `step_result_tile.dart:367` | `_ResultDetail` has two mutually exclusive modes (string `value` vs `segments`) with no type-level separation; future callers could populate both simultaneously |
| ⚠️ | 🟢High | `usp_network_health_card.dart:473` vs `usp_system_status_card.dart:563` | `_LegendEntry` (no `ellipsize`) and `_StatLegendEntry` (has `ellipsize`) represent the same visual element but diverge silently; #1245 extraction will need to reconcile this |
| ⚠️ | 🟡Med | `dashboard_card_probe.dart:252` | `widthCasesFor` early-returns `[]` on first `null` span; correctness relies on unstated invariant that `floor` is the sole null cause and all spans share the same `floor` |
| 💡 | 🟢High | `topology_section.dart:109` + `devices_section.dart:53` | [Duplicated Code] `speedPairs` list-building logic and `_formatSpeed` helper duplicated verbatim across two AI section files |
| 💡 | 🟢High | `topology_section.dart:133` | Hardcoded `'Speed'` English string; other labels in same widget also hardcoded — document intent or track for l10n |
| 💡 | 🟡Med | `stats_traffic_monitor_section.dart:91` | Icon fix applied but `Row+Spacer` layout not converted to `Wrap` (unlike Traffic Analysis); potential overflow on narrow views — confirm this page is in gate scan scope (note: pre-existing, not introduced by this PR) |
| 💡 | 🟡Med | `usp_info_row.dart:12` | `_kLabelShare` / `_kLabelMaxWidth` are module-level constants with a single class consumer; consider `static const` or add comment clarifying intended scope |
| 💡 | ⚪Low | `dashboard_card_probe_test.dart:140` | Baseline geometry literals lack a note that updating them requires independent derivation, not just "make test green" |

#### Spec (Reviewer C — Issues #1225 / #1226 / #1233)

|---|---|---|---|
| ⚠️ | 🟡Med | `usp_traffic_analysis_card_legend_test.dart:185` | T02 AC "clean across 26 locales × 4 tabs" — new tab-clean test uses only `de` locale (1 × 4); gate ratchet provides integration coverage but the dedicated unit test only exercises 1 locale |
| 💡 | 🟢High | 5 non-legend files (devices_section, topology_section, stats_monitor, diagnostic_result_card, step_result_tile) | Scope exceeds #1225/#1226/#1233 ACs — icon substitution and UspInfoRow LayoutBuilder change are sensible companion fixes but not AC-required; design §2.10a partially records them |
| 💡 | ⚪Low | `usp_info_row.dart` | #1233 AC says "no shared legend component created" — UspInfoRow is shared and was modified; not a legend widget so literal AC is met, confirm intended reading |

<details>

**W-B-2 · `_ResultDetail` dual-mode pattern** 🟡Med  
- **Location:** `step_result_tile.dart:367–381`
- **Code:**
```dart
class _ResultDetail {
  final String label;
  final String value;
  final List<({IconData icon, String text})> segments;  // NEW
  const _ResultDetail(this.label, this.value, {this.segments = const []});
}
```
- **Why it's a concern:** Two render modes co-exist in one type: `value.isNotEmpty` → render text; `segments` non-empty → render icon rows. Nothing prevents constructing `_ResultDetail('label', 'value', segments: [...])` — the render site (lines 84–93) would then emit both, which is likely non-intended. A docstring explains the design, but there is no runtime assertion enforcing mutual exclusivity.
- **Verdict:** 技債 — no card built. Not a current breakage but a maintenance trap.
- **Fix suggestion:** `assert(value.isEmpty || segments.isEmpty, '_ResultDetail: value and segments are mutually exclusive');`

---

**W-B-1 · `_LegendEntry` / `_StatLegendEntry` naming divergence** 🟢High  
- **Location:** `usp_network_health_card.dart:473` vs `usp_system_status_card.dart:563`
- `_LegendEntry` (NH card): no `ellipsize` field; fixed soft-wrap for composed stats labels
- `_StatLegendEntry` (SS card): has `ellipsize: bool` parameter with default `false`
- Both represent the same visual concept (colour dot + label) but the key design distinction — ellipsize vs soft-wrap follows the label's *type* — is not expressed by either class's name. The #1245 extraction work will need to reconcile these before a shared widget is created.
- **Verdict:** 技債 — no card. Recommend adding a comment/docstring before #1245 work begins.

---

**W-B-3 · `widthCasesFor` early-exit hidden invariant** 🟡Med  
- **Location:** `dashboard_card_probe.dart:252–255`
```dart
final narrowest = narrowestRealizationOf(entry.value, minScreen: floor);
if (narrowest == null) return [];  // null = floor past enumerated range
```
- Currently correct because `floor` is the only null trigger and all span iterations share the same `floor`. If `narrowestRealizationOf` gains additional null-return paths in future, the early return would silently drop all subsequent spans without warning.
- **Verdict:** 技債 — no card.
- **Fix:** Add comment: `// Invariant: floor is the sole null cause, and all spans share the same floor — if any is null, all are.`

---

**W-C-1 · T02 tab-coverage test uses 1 locale, spec requires 26** 🟡Med  
- **Location:** `usp_traffic_analysis_card_legend_test.dart:185–204`
- Spec #1226 AC: "renders clean at its narrowest realization across all 26 locales and all 4 tabs"
- New test group "every tab is clean at the narrowest realization" iterates 4 tabs but hardcodes `locale: const Locale('de')` → 1 locale × 4 tabs, not 26 × 4.
- The gate ratchet (`dashboard_card_overflow_test.dart`) provides 26-locale overflow coverage at integration level (traffic_analysis allowlist keys fully cleared), so there is no functional regression.
- **Verdict:** 技債/測試完整度 — no card. Consider expanding locale coverage in this test group, or add a comment that the gate is the designated 26-locale backstop.

</details>

<details>

- **`narrowestRealizationOf` independent oracle:** `continuumInfimum` derives the expected answer from breakpoint arithmetic rather than re-enumerating the same domain — the test cannot pass by self-validation. Mutation-verified (raising the floor above 601px makes it fail). Exemplary test design.
- **UspInfoRow LayoutBuilder fix:** Correctly decouples label sizing from screen width. All callers are inside bounded card/column contexts (confirmed: none in unbounded Row), so `constraints.maxWidth` is always finite; `_kLabelMaxWidth = 150.0` cap prevents over-wide label on large rows.
- **Network Health WAN/LAN row deviation fully documented:** Inline comment at the site, design §2.10a entry, and a pinned test that fails if someone "restores consistency." Correct treatment of intentional deviation.
- **known_overflows.json ratchet with attribution:** Each residual allowlist entry points to the responsible ticket (#1234, #1235). Pass/fail count is not anonymous.
- **`labelColumns` removal is clean:** No callers in the codebase use the removed parameter (confirmed via grep across all files at FETCH_HEAD).
- **All 26 locales have the `stale` l10n key:** The `loc().stale` substitution in `diagnostic_result_card.dart` is safe across all supported locales (ARB scan confirmed).
- **1700/1700 gate + 5039/5039 full suite** as reported — strong confidence baseline.

</details>

<details>
<summary>💡 Suggestion notes</summary>

- **S-B-1 Duplicated speed logic:** `_speedRow` (topology_section) and the anonymous speed-pair builder (devices_section) produce identical `List<({IconData, String})>` with identical `_formatSpeed` helpers. These are `lib/ai/components/sections/` internals, not UI Kit widgets, so Article XIV likely does not apply. Extracting a shared helper would remove two independent update points.
- **S-B-2 Hardcoded strings in topology_section:** If this is a debug overlay not intended for l10n, document that explicitly so the pattern is not mistakenly "fixed."
- **S-B-3 stats_traffic_monitor_section Row+Spacer:** Pre-existing (confirmed by git log — Row+Spacer predates this PR). The PR correctly limits its change to the icon substitution. Suggest a follow-up issue to apply the Wrap treatment here as well.
- **S-C-1 Scope documentation:** The five non-legend files modified are sensible companion cleanups (font-coverage safety, LayoutBuilder correctness). Confirm they are intentionally bundled here for future bisect clarity; design §2.10a partially covers them.

</details>

*Cross-reviewed by three independent agents (security+correctness / architecture+maintainability / spec+requirements). Automated — please sanity-check before merge.*

Issue #1237, opened 2026-08-11
[T09] Close time_settings (21 coordinates)


AustinChangLinksys:
## Parent

#1183

Design of record: `doc/dashboard/dashboard_density_design.md` §1.1, §1.3.

## Branch

**Base: `gate/1183-overflow-gate` (PR #1248). Not `dev-2.7.0`.**

This epic is consolidated onto #1248's branch — #1248 is the only PR that
targets `dev-2.7.0`. Branch from `gate/1183-overflow-gate`, and merge back into
it.

Not a preference. The gate, its fixture, its probe and the design of record are
all introduced *by* #1248 and are **absent on `dev-2.7.0`** (verified with
`git cat-file -e origin/dev-2.7.0:<path>`):

- `test/page/dashboard/cards/dashboard_card_overflow_test.dart`
- `test/fixtures/known_overflows.json`
- `test/util/dashboard/dashboard_card_probe.dart`
- `doc/dashboard/dashboard_density_design.md`

So an acceptance criterion of the form "N coordinates are removed from
`known_overflows.json` and the gate passes" cannot be executed off `dev-2.7.0`
at all. Basing there also re-measures coordinates the landed tickets already
cleared, producing a fixture that conflicts with them and reads as a regression.

## What to build

Time Settings has 21 coordinates from one card-own site plus the shared card-template footer. Both must be fixed for them to turn green, so this ticket waits on #1227 and then does the card-own half.

This card is the one genuine exception in the whole baseline: at a fit width of 288px it is the **only** card that could be fixed by widening instead — raising its minimum span from 3 to 5 columns would give it enough room (§1.3). Every other card would need more than the 12 columns that exist.

Fix the layout anyway. Widening is available here but it is not free: it constrains what users can do on their own dashboard, and 288px is close enough that constraining the content is the smaller cost. Recording *why* we did not take the widening option is part of this ticket, since the option genuinely existed only here.

## Acceptance criteria

- [ ] The card renders clean at its narrowest realization across all 26 locales
- [ ] 21 coordinates are removed from `test/fixtures/known_overflows.json` and the gate passes
- [ ] `time_settings` has no remaining allowlist entries
- [ ] `minColumns` is unchanged, with a comment recording that widening was possible here and was declined in favour of constraining content
- [ ] Timezone names — the longest strings on this card — stay identifiable

## Blocked by

- ~~#1227~~ — **landed** (`208315ed`, `c48de478`, PR #1255). The shared
[...]

PR #1262, opened 2026-08-13
fix(dashboard): wrap the wifi_performance legend rows, and localize the channel readout the Channels tab was hiding behind (#1229, #1266)

AustinChangLinksys:
Closes #1229, closes #1266. Track A of #1183. Design of record: `doc/dashboard/dashboard_density_design.md` §2.10b and §2.10c (both added here).

Base is `gate/1183-overflow-gate` (#1248), not `dev-2.7.0`, per the ticket: the gate, its fixture, its probe and the design of record are all introduced by #1248 and are absent on `dev-2.7.0`, so "N coordinates removed from `known_overflows.json`" is not executable there.

**Two tickets, because the second was found by the first.** #1229 cleared the card's 45 coordinates from the Signal and Speed legends. It also recorded two observations about the Channels tab as deliberately unfixed — and both turned out to be wrong when measured, which is #1266. They ship together because #1266 is ratchet work with coordinates of its own (the #1249 precedent), not gate-invisible hardening (the #1258 one).

---

# Part 1 — #1229: the two legend rows

## The split was measured, not inferred

Both of this card's overflow sites are legend rows of the shape #1226 settled and #1233 replicated six times: a centred `Row` of colour dot, gap, and unconstrained label. Together they are the card's **entire** 45 coordinates.

§1.1's greedy cover credits `:190` with 33 and never names the Speed tab, so the remaining 12 had to be located by re-running §1.1's attribution method scoped to this card:

| Site | Tab | Coordinates |
|---|---|---:|
| `usp_wifi_performance_card.dart:190` | Signal | 33 |
| `usp_wifi_performance_card.dart:275` | Speed | 12 |

No coordinate needs both sites; the total is exactly 45. Inferring the second site from the ticket title would have been a guess with a 12-coordinate blast radius.

## Same shape, same fix, different cause

The locale footprint says which:

- **`:190`** — four entries at a 261px card. Overflows in `en` too (`ru` missed by 149px): **geometry**-bound.
- **`:275`** — two entries. Overflows only in `es`/`es_AR`/`fr`/`fr_CA`/`pt_PT`/`ru`/`tr`: **translation-length**-bound.

That distinction is free to read off the allowlist — `["*"]` means the card is too narrow for the content in any language, a locale list means the layout is fine and one translation is long — and it predicts which sites regress when a string changes versus when spacing changes.

Both rows sit under an `Expanded` holding a `ListView`/`AppBarChart`, so §2.10a point 3's precondition holds and the second run is affordable. **Verified rather than assumed**: re-running attribution after the fix returns 0 incidents at any site, tab or locale, so no bottom overflow was traded in — the exact failure Network Health's fixed 120px gauge produced.

Inter-entry spacing on the Signal tab goes 12px -> 16px, to `AppSpacing.lg` as all six #1233 legends and #1226's use; the measurement says 16px is affordable even at 261px.

## Choosing the mutations turned up a method finding

#1233 and #1228 recorded "verified to fail under a mutation". That is necessary but not sufficient, and this ticket found out why. Every candidate mutation was run against the gate as well:

| mutation | readability test | gate |
|---|---|---|
| drop `SignalTier.fair` from the legend list | 5 fail | green |
| `Row` + `Flexible` entries, no `Wrap` | 3 fail | green |
| drop the per-radio SNR readout | 2 fail | green |
| `Wrap` back to a bare `Row` (pre-#1229) | **green** | **33 fail** |

The last row is the instructive one. Reverting to a bare `Row` does not clip anything: a `Row` hands its non-flex children unbounded width, so each entry's inner `Flexible` never binds, every label paints at full intrinsic width, and the *outer* row overflows instead. The gate owns that regression.

The regression the gate cannot see is the one that *succeeds* at fitting: keep the single `Row` and wrap each entry in `Flexible`. Every tier name then ellipsizes to a stub — in `en`, not just the long translations — while all 157 `wifi_performance` gate cases stay green. That is the shape a well-meaning "just make it fit" edit lands on, and it is what the new test is validated against. Generalised into §2.10b as a rule for the remaining Track A tickets.

## #1229 acceptance criteria

- [x] Card renders clean at its narrowest realization across all 26 locales and all 3 tabs — `wifi_performance` gate 157/157
- [x] All 45 coordinates removed from `test/fixtures/known_overflows.json` and the gate passes — 4 keys (19 + 7 + 14 + 5), gate 1644/1644
- [x] `wifi_performance` has no remaining allowlist entries — verified programmatically; allowlist **226 -> 181** coordinates (19 -> 15 keys)
- [x] Per-band metrics stay distinguishable at the narrowest clean width — asserted, and then rewritten by #1266 (below)

---

# Part 2 — #1266: an English abbreviation was hiding a geometry problem

#1229 left the Channels tab alone on the strength of two claims: that its per-radio rows were clean "by about 48px of headroom", and that the headroom was unmeasurable because the channel string is data-dependent. The 48px was real. The reasoning from it was not — it was measured with `'Ch '`, a hardcoded English abbreviation, in the row. **The abbreviation was the bug, and it was concealing the geometry problem rather than the row not having one.**

The `channel` ARB key already exists in all 26 locales (`tr` = `Channel (Kanal)`, 15 chars, the worst of the set), so localizing is a one-line change with no ARB work. It is also **3 gate coordinates on the fixture the gate ships**:

| locale | `channel` | 261px (min) | 288px (preferred) |
|---|---|---|---|
| `tr` | `Channel (Kanal)` | +4.1px, +41.0px | +14.0px |
| `th` | `ช่องสัญญาณ` | +17.0px | clean |
| other 24 | — | clean | clean |

## Why the two halves cannot ship apart

- **Localizing alone runs the ratchet backwards.** It adds 3 coordinates to a mechanism (§2.9) whose whole contract is that the count only falls, and would have to be booked as an allowlist *addition*. Nothing in Part 4 permits that.
- **Hardening alone is gate-invisible.** With `'Ch '` in place the row is clean everywhere, so the `Wrap` changes no coordinate and the gate cannot tell whether it worked. That is the #1258 shape, and it is why #1229's bullet reached for #1258.
- **Together they are self-verifying.** The localization supplies the failure the hardening has to clear, on the shipped fixture, at both widths. Net coordinate change **0**, and the honest string ships.

## The fix, and the trap inside it

The row gets #1226's `Wrap(alignment: spaceBetween)`. Neither side yields: `band` is the block's identity and the channel string is composed data whose ellipsis would land inside `320MHz` and misinform (§2.10a point 2). So both keep their intrinsic width and the row spends a second run, paid out of the donut's `Expanded`.

`WrapAlignment.spaceBetween` is a **silent no-op under loose width constraints** — `RenderWrap` sizes to its widest run, so `freeMainAxisSpace` is 0 and there is nothing to distribute, and the channel string lands one `spacing` gap after the band instead of at the right edge. `CrossAxisAlignment.stretch` on the enclosing `Column` is what prevents that. It overflows nothing, so **all 157 gate cases pass either way** — the readability test is the only thing that can see it.

**This applies retroactively:** #1226's and #1233's `spaceBetween` legends should be checked for the same precondition, since a shrink-wrapped legend and a correct one are indistinguishable in a green gate. Recorded in §2.10c point 3.

## Attribution halved the change

#1229's bullet had assumed both per-radio rows needed the same treatment. Regexing `usp_wifi_performance_card.dart:(\d+)` over `OverflowIncident.fullLog` — §1.1's method at per-line resolution — put **every** incident, both fixtures, all locales, both widths, on the band/channel row and **none** on the clients/SNR/loader row below it, whose `Expanded` loader already binds. That row is untouched.

## What #1266 does to #1229's test

#1229's AC4 group pinned "band label and channel share a line", which the `Wrap` now permits to be false. The invariant weakens from *same line* to **attributable**: each channel readout must be nearer its own band than any other, and never below the next band label. It also had to stop matching on `'Ch '` — a test that still recognised the removed literal would silently match nothing and pass. Three new mutations, each also run against the gate:

| mutation | readability test | gate |
|---|---|---|
| Channels `Column`: `stretch` -> `start` | **1 fail** | **green** |
| `loc(context).channel` -> `'Ch '` | **4 fail** | **green** |
| Channels `Wrap` -> `Row` + `Spacer`, localized | 1 fail | 3 fail |

The first two are the pair that matters: both are real regressions that the gate's 157 cases cannot distinguish from correct code.

## Method finding: two `testWifiData` fixtures exist, and the gate reads one

The dashboard gate reads `test/golden_test/page/dashboard/cards/fixtures/cards_test_data.dart` (via `kitchenSinkOverrides()`); the Statistics page reads `test/golden_test/page/statistics/fixtures/statistics_test_data.dart`. I edited the latter first and got a confident, meaningless "all clean" from the gate. Only dumping the rendered `Text` list revealed the added radio was absent. **Rule, recorded in §2.10c point 2: when a measurement depends on fixture data, verify the render contains the data, not just that the verdict is green.**

## #1266 acceptance criteria

- [x] `'Ch '` replaced with the localized `channel` key; no ARB edit needed (all 26 locales already have it)
- [x] The band/channel row hardened to #1226's shape; the clients/SNR row left alone, per attribution
- [x] `spaceBetween` verified effective, not silently shrink-wrapped — pinned by a test the gate cannot replace
- [x] Gate net coordinate change **0**; no allowlist addition
- [x] #1229's AC4 group rewritten rather than deleted, mutations re-verified, ledger updated
- [x] 26 locales x both widths clean on the shipped fixture
- [ ] **No bottom overflow introduced** — holds on the shipped fixture; **fails on a tri-band profile the gate cannot express.** See below.

---

## Verification

- overflow gate **1644/1644**; `wifi_performance` **157/157**
- `wifi_performance` readability **13/13** (was 11; +1 locale case, +1 alignment case)
- all three readability files **62/62**
- `usp_analytics_cards_test.dart` (the only test that imports this card) **13/13**
- full functional suite `./run_tests.sh` **5360/5360** (was 5358)
- `fvm flutter analyze` clean on both changed files; `fvm dart format` clean

## Left deliberately unfixed, and one honest cost

**The tri-band bottom overflow.** The gate's data domain is a single hardcoded profile — `testRadios` is 2 radios, 2-digit channels, `160MHz` widest — and `buildDashboardCardApp()` takes no overrides, so no card can be measured against different data without editing the fixture. Measured by temporarily adding a third tri-band radio (`Channel 233 (Auto) · 320MHz`) and reverting:

| | before (`'Ch '`) | localized, old `Row` | localized + `Wrap` (this PR) |
|---|---|---|---|
| 2 radios (the gate's fixture) | clean | 3 coordinates | **clean, 26 locales x both widths** |
| 3 radios, tri-band | clean | `en` +8.3px, plus `fi`, `ja` and the two above | one **+9.0px bottom** (`tr` @261 only) |

That remaining incident is **not** the band/channel row: it is the donut's centre label at `:575`, squeezed once three two-run blocks have taken enough of the column that its `Expanded` no longer fits the label's two lines. So this PR does trade a right overflow for a bottom overflow on that unshipped profile — exactly the trade §2.10a point 3 warns about, stated rather than buried.

It is left unfixed on purpose. At that height the donut is visually useless whether or not it reports an overflow, so "shrink the donut to fit" would silence the gate without fixing anything. Deciding what the tab drops at that density — the donut, or the whole tab becoming scrollable like the Signal tab's `ListView` — is a density decision, and it is not verifiable at all until the gate can express a second data profile.

**Filed as #1267**, which pairs that decision with the gate change that makes it measurable (an `overrides` parameter on `buildDashboardCardApp()` / `kitchenSinkOverrides()`, plus a tri-band fixture). Not folded into #1235 (`network_health`'s fixed 120px gauge centre): same family, but different axis (vertical vs horizontal), different trigger (data vs translation length) and different visibility (invisible until #1267 lands vs 3 live coordinates) — and #1235's ACs are executable ratchet claims that an unverifiable one would make unclosable. Recorded in §2.10c.

**An abbreviation is the one defect this epic's instrument is structurally blind to.** A hardcoded English string makes every width measurement at that site optimistic by however much the translation is longer, and the gate reports the optimistic number in all 26 locales — a locale sweep cannot find a string that never varies. Worth a grep pass over the remaining Track A sites. §2.10c point 1.

## Follow-up

`_LegendDot`/`_LegendEntry` here is the **fifth** copy of the shape. Extracting a shared widget needs an Article XIV decision, which #1245 exists to obtain; #1233 was explicitly told not to block on that conversation and this PR follows it. #1245's inventory is written against four files and four cards (including its AC "all four cards render identically") and needs `wifi_performance` added once this merges — recorded in §1.1 and in the widget's own doc comment. Note this file lives under `lib/page/wifi_settings/cards/`, not `lib/page/dashboard/views/components/`, which is an argument against #1245's proposed location for the shared widget.

---

**Note on the last commit** (`e3d3f4bb`, docs-only): it appends a §1.1 subsection
recording the per-line attribution for the *next* batch (#1234–#1237), measured
while this branch was still the newest edit to that document. It changes nothing
in this PR's scope; it is here rather than on the follow-up branch to avoid two
branches editing the same document section concurrently.

### DIFF
--- a/doc/dashboard/dashboard_density_design.md
+++ b/doc/dashboard/dashboard_density_design.md
@@ -129,6 +129,55 @@ added to it.**
 
 **Axis split**: 464 right-only, 63 bottom-only, 33 both.
 
+#### Track A's four remaining card-own tickets reduce to four shapes across six sites (#1234–#1237, measured before implementation)
+
+**Method.** §1.1's `fullLog` attribution, re-run at per-line resolution over the
+four cards of #1234 / #1236 / #1237 (`system_status`, `lan_info`, `device_info`,
+`time_settings`). 112 incidents → the 84 allowlisted coordinates those three
+tickets own; adding #1235's 3 makes 87.
+
+| Shape | Sites | Worst | Tickets |
+|---|---|---:|---|
+| "View details" footer — `AppDivider` + `Row(mainAxisAlignment: end)` + `Semantics`/`InkWell` | `usp_system_status_card.dart:118`, `usp_device_info_card.dart:140` | +5.2px | **#1234 + #1236** |
+| Hero inner row — `Row` inside `Expanded(Column(start, [titleLarge, AppGap.xxs, Row]))` | `usp_lan_info_card.dart:56`, `usp_time_settings_card.dart:108` | +102 / +67px | **#1236 + #1237** |
+| Twin gauges — `Expanded(Row(spaceEvenly, [AppGauge(size: 100) ×2]))` | `usp_system_status_card.dart:196` | +43px | #1234 |
+| Legend row — `_LegendDot` ×2 + `Spacer` + refresh chrome | `usp_system_status_card.dart:218` | +107px | #1234 |
+
+**Two of the four shapes straddle ticket boundaries.** The footer is
+near-identical code in two files (`_buildStatisticsFooter` / `_buildNodeDetailFooter`),
+and the hero inner row is the same idiom in two more — and that idiom alone is
+**48 of the 84 coordinates**. Split across separate branches, each shape gets its
+fix invented twice and #1245's de-duplication inventory grows by two more entries.
+This is the #1249 situation again (§2.10a), so the four are best done as one
+branch with a commit per shape.
+
+**`system_status`'s 26 widest coordinates have two causes each, so neither site
+clears them alone.** The exact decomposition of #1234's 34:
+
+- `min|0` tab 0, all 26 locales — caused by **both** `:196` and `:218`; `el`/`ru`
+  additionally by `:118`
+- `min|0` tabs 1–3, `el`+`ru` — 6, `:118`
+- `preferred|0`, `fr`+`fr_CA` — 2, `:218`
+
+So `:196` alone clears 0 and `:218` alone clears 0; the pair clears 26; adding
+`:118` clears all 34. Expect zero allowlist movement until both gauge-row and
+legend-row fixes are in — the same trap §2.6a's shared-plus-card-own coordinates
+set, one file down.
+
+**`AppGauge`'s centre is fixable at the call site — no ui_kit change, so no
+Article XIV question.** `AppGauge` renders
+`SizedBox(width: size, height: size, child: Stack(alignment: center, children: [CustomPaint, …, centerBuilder(…)]))`.
+The centre is a *non-positioned* `Stack` child, so it receives loose `size × size`
+constraints and the overflowing `Column(mainAxisSize: min, …)` is the app's own
+closure. This holds for both #1235's `network_health` gauge (`size: 120`) and
+`system_status._buildGauge` (`size: 100`), which is why #1235's 3 coordinates are
+a cheap add-on to #1234 rather than a separate ui_kit conversation.
+
+> Sequencing note: these four tickets, #1229 and #1266 all edit
+> `test/fixtures/known_overflows.json` and this file. Until #1229/#1266's PR
+> merges, branch the four-ticket work from *its* head rather than from the epic
+> gate branch, or the fixture conflicts.
+
 ### 1.2 Fit width — the narrowest width at which each card is clean
 
 **Method.** Card width was driven directly through a 14-rung ladder
--- a/doc/dashboard/dashboard_density_design.md
+++ b/doc/dashboard/dashboard_density_design.md
@@ -1,6 +1,6 @@
 # Dashboard Card Density — Design Decisions
 
-**Last Updated: 2026-08-12** · Follow-up to #1183 · Status: **agreed; tickets #1225–#1240 published; #1225 + #1226 + #1233 implemented (not yet merged), rest not started**
+**Last Updated: 2026-08-13** · Follow-up to #1183 · Status: **agreed; tickets #1225–#1240 published; #1225 + #1226 + #1233 + #1227 + #1228 + #1229 implemented (not yet merged), rest not started. Allowlist 560 → 181.**
 
 ## Purpose
 
@@ -102,15 +102,27 @@ batching leverage actually is.
 > `:118` (8) + `:218` where those are the coordinate's only remaining cause. So
 > the pattern is really eight rows, and #1234 finishes it.
 
-The private colour-dot widget is additionally duplicated **verbatim in four
-files** (the three above plus `device_analytics`). De-duplicating it, or
-extracting a shared legend entry, would be a **new shared widget** and therefore
-needs approval under Article XIV — so the fix is applied in place, and the
-extraction raised separately rather than blocking on that conversation. That
-raise is **#1245**, filed after #1233; it carries the constraint #1233 measured,
-namely that any shared entry must express the ellipsize-vs-soft-wrap distinction
-per label kind (§2.10a point 2) and must not absorb the WAN/LAN row, which
-deviates for a reason (§2.10a point 3).
+> Re-measured again during #1229: `wifi_performance`'s two sites (`:190` Signal,
+> `:275` Speed) are this same shape, and they are that card's **entire** 45
+> coordinates. So the pattern is **ten rows across four files, 181 + 45 = 226
+> coordinates — 40% of the whole baseline** (not to be confused with the
+> allowlist standing at 226 just before #1229; the two numbers coincide by
+> accident), and it is the only structure in this epic that
+> accounts for a plurality of it. §1.1's greedy table hides `:275` as well as the
+> other nine: it credits `:190` with 33 and never names the Speed tab at all
+> (see §2.10b for why a "×2 sites" ticket still needs its own attribution run).
+
+The private colour-dot widget is additionally duplicated **verbatim in five
+files** (the three above plus `device_analytics` and, per #1229,
+`wifi_performance`). De-duplicating it, or extracting a shared legend entry,
+would be a **new shared widget** and therefore needs approval under Article XIV —
+so the fix is applied in place, and the extraction raised separately rather than
+blocking on that conversation. That raise is **#1245**, filed after #1233; it
+carries the constraint #1233 measured, namely that any shared entry must express
+the ellipsize-vs-soft-wrap distinction per label kind (§2.10a point 2) and must
+not absorb the WAN/LAN row, which deviates for a reason (§2.10a point 3).
+**#1245's inventory is written against four files and needs `wifi_performance`
+added to it.**
 
 **Blocked on a dependency we do not own — 45 coordinates (8%)**: fl_chart 19
 (`firewall_overview`), ui_kit `AppListTile` 26 (`connected_devices`).
@@ -634,6 +646,75 @@ from a row that truncated its content to nothing. They are covered by
 tagged `dashboard-card` so it gates; each of its three groups was verified to
 fail under a mutation of the code it guards.
 
+### 2.10b What the eighth and ninth replications taught us (#1229 — implemented)
+
+All 45 coordinates cleared as predicted (226 → 181). By this point the shape is
+routine — the interesting findings are about *method*, not about legends.
+
+1. **A "×N sites" scope still needs its own attribution run.** §1.1's greedy
+   cover is greedy: it credits each coordinate to one site, so it named
+   `:190` (33) and never mentioned the Speed tab's legend at all. The remaining
+   12 were only located by re-running §1.1's method scoped to this card, which
+   returned an unambiguous split — `:190` → 33, `:275` → 12, no coordinate
+   needing both, total exactly 45. Inferring the second site from the ticket
+   title would have been a guess with a 12-coordinate blast radius.
+2. **Same shape, same fix, different cause — and the locale footprint says
+   which.** `:190` (four entries, 261px) overflows in `en` too: it is
+   *geometry*-bound, and `ru` missed by 149px. `:275` (two entries) overflows
+   only in the long-translation locales (`es`, `fr`, `pt_PT`, `ru`, `tr`): it is
+   *translation-length*-bound. The distinction is free to read off the allowlist
+   — an `["*"]` entry means the card is too narrow for the content in any
+   language, a locale list means the layout is fine and one translation is long
+   — and it predicts which sites will regress when a string changes versus when
+   spacing changes.
+3. **§2.10a point 3's precondition was checked here, not assumed.** The
+   `Expanded` above both rows holds a `ListView`/`AppBarChart`, so it yields the
+   height a second run costs. Confirmed by re-running attribution after the fix:
+   **0 incidents at any site, any tab, any locale** — no bottom-overflow was
+   traded in, which is the exact failure Network Health's fixed 120px gauge
+   produced. Measured, because the shape's cost is paid in a currency the gate
+   only sometimes charges for.
+4. **The mutation that validates a readability test must itself be run against
+   the gate.** #1233 and #1228 recorded "verified to fail under a mutation"; that
+   is necessary but not sufficient. Here the *obvious* mutation — revert `Wrap` to
+   a bare `Row` — turns out to be **invisible to the readability test and caught
+   by the gate** (33 failures): a `Row` hands non-flex children unbounded width,
+   so the inner `Flexible` never binds, every label paints full-width, and the
+   *outer* row overflows. The genuinely gate-invisible regression is the one that
+   *succeeds* at fitting: keep the single `Row` and wrap each entry in
+   `Flexible`. Every tier name then ellipsizes to a stub — in `en`, not just the
+   long translations — and all **157** `wifi_performance` gate cases stay green.
+   That is the shape a well-meaning "just make it fit" edit lands on. **Rule for
+   the remaining Track A tickets: run each candidate mutation against the gate as
+   well: if the gate already fails it, the mutation proves nothing about the
+   readability test and a different one is needed.**
+
+The gate-invisible ACs are covered by
+`test/page/dashboard/views/components/wifi_performance_readability_test.dart`,
+tagged `dashboard-card` so it gates; its three groups were verified to fail under
+the mutations tabulated in the file, each with its gate result beside it.
+
+**AC4 was already true before this ticket, and is now pinned rather than earned.**
+"Per-band metrics stay distinguishable at the narrowest clean width" is about the
+Channels tab, which contributed none of the 45 and which #1229 does not modify. At
+the 261px narrowest realization both bands show band, channel, bandwidth, client
+count and SNR with nothing clipped. It is asserted anyway, because "already clean"
+is not "checked" and nothing else in the suite would notice a later fix collapsing
+those rows.
+
+Two observations left deliberately unfixed:
+
+- `_ChannelsTab`'s two per-radio rows use the same unconstrained shape this epic
+  keeps fixing (`Row` + `Spacer`, non-flex `AppText`), and measure clean only by
+  about **48px** of headroom. That is the #1258 situation exactly — hardening a
+  site that is not currently failing — so it belongs there, not in a ratchet
+  ticket whose contract is a coordinate count.
+- The channel readout is **data**-dependent, not translation-dependent:
+  `Ch 197 (Auto) · 320MHz` on a tri-band router is materially wider than the
+  mock's two-radio output. No amount of locale sweeping reaches it, so that
+  headroom is unmeasured rather than measured-safe — a limit of the gate's fixed
+  mock, worth stating where the 48px figure is quoted.
+
 ### 2.11 fl_chart's 19 coordinates get a primary plan and a documented fallback
 
 `firewall_overview`'s 19 coordinates originate inside fl_chart
@@ -752,7 +833,7 @@ addition — it changes no card's rendering until a threshold is declared.
 | #1233 | The other six legend rows (§1.1, §2.10a) — **implemented** | 132 |
 | #1227 | Shared blocks made overflow-safe (§2.6, §2.6a) — **implemented** | 101 |
 | #1228 | `ethernet_ports` ×2 sites (§2.12) — **implemented** | 52 |
-| #1229 | `wifi_performance` ×2 sites | 45 |
+| #1229 | `wifi_performance` ×2 sites (§2.10b) — **implemented** | 45 |
 | #1230 | `firewall_overview` own sites (§2.11) | 48 |
 | #1234 | `system_status` remaining ×3 sites | 34 |
 | #1236 | `lan_info` + `device_info` card-own | 29 |
--- a/doc/dashboard/dashboard_density_design.md
+++ b/doc/dashboard/dashboard_density_design.md
@@ -807,10 +807,24 @@ that height the donut is visually useless whether or not it reports an overflow,
 "shrink the donut to fit" would silence the gate without fixing anything. Deciding
 what the tab drops at that density (the donut, or the whole tab becoming
 scrollable like the Signal tab's `ListView`) is a density decision, and it is not
-verifiable at all until the gate can express a second data profile. **Filed
-against the #1235 shape rather than fixed here; recorded because the fix does trade
-a right-overflow for a bottom-overflow on that unshipped profile, which is exactly
-the trade §2.10a point 3 warns about.**
+verifiable at all until the gate can express a second data profile.
+
+**Filed as #1267**, which pairs the density decision with the gate change that
+makes it measurable: an `overrides` parameter on `buildDashboardCardApp()` /
+`kitchenSinkOverrides()` plus a tri-band fixture. It is deliberately *not* folded
+into #1235 — same family (fixed-size gauge in an `Expanded` that cannot pay), but
+different axis (vertical vs horizontal), different trigger (data vs translation
+length) and different visibility (invisible until #1267's Part 1 lands vs 3 live
+coordinates), and #1235's acceptance criteria are executable ratchet claims that an
+unverifiable AC would make unclosable. Recorded here because the #1266 fix does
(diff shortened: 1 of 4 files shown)


AustinChangLinksys (commented):
🤖 **Automated Review — Round 1** · `29450e0..0ceba68` (full)

**Verdict: 💬 Comment** — Self-review (comment only; GitHub prohibits self-approve)

#### Standards (Reviewer A — Security & Correctness · Reviewer B — Architecture & Maintainability)

|---|---|---|---|
| ⚠️ | 🟢 High | `usp_wifi_performance_card.dart:137,254,399` | Hardcoded English empty-state strings not using l10n keys (pre-existing, not introduced by this PR) |
| ⚠️ | 🟢 High | `usp_wifi_performance_card.dart:320–331` | [Duplicated Code] `_LegendDot` is the 5th verbatim copy across cards; #1245 inventory lists only 4 files and needs updating |
| ⚠️ | 🟡 Med | `wifi_performance_readability_test.dart:304` | `_snrMarker = 'SNR'` assumes ASCII prefix survives all translations; silently fails if future locale omits it |
| ⚠️ | 🟡 Med | `usp_wifi_performance_card.dart:452–458` | `_ChannelsTab` per-radio `Row` (same unconstrained shape, ~48 px headroom) has no machine-readable tracking key in fixture |
| 💡 | 🟢 High | `wifi_performance_readability_test.dart:306–312` | `_localeFor` uses `firstWhere` without `orElse`; obscure `StateError` on unknown tag |
| 💡 | 🟡 Med | `usp_wifi_performance_card.dart:445` | `avgSnrPerRadio[i] ?? 0` renders "SNR: 0 dB" indistinguishably from "no data" (pre-existing) |
| 💡 | 🟢 High | `usp_wifi_performance_card.dart:320–331` | `_LegendDot` in this file has only one caller (`_LegendEntry`) — could be inlined |
| 💡 | 🟢 High | `doc/dashboard/dashboard_density_design.md` | §1.1 still describes #1245 inventory as "four files"; now five after this PR |

#### Spec (Reviewer C — Requirements vs. Issue #1229)

|---|---|---|---|
| ⚠️ | 🟡 Med | `wifi_performance_readability_test.dart:173,190` | AC1: "clip-free across all 26 locales" — readability test asserts label intactness for only 3 locales (`en`/`ru`/`th`); other 23 rely on gate which is blind to ellipsis truncation |
| ✅ | 🟢 High | `test/fixtures/known_overflows.json` | AC2: 45 coordinates removed (19+7+14+5 = 45 ✓); gate keys 15 matches commit message |
| ✅ | 🟢 High | `test/fixtures/known_overflows.json` | AC3: zero `wifi_performance` allowlist entries in HEAD confirmed |
| ✅ | 🟢 High | `wifi_performance_readability_test.dart:209–270` | AC4: per-band metrics (band/channel/SNR) asserted intact at narrowest width in `en`/`ru` |
| 💡 | 🟢 High | `doc/dashboard/dashboard_density_design.md` §2.10b | Channels tab 48 px headroom recorded as out-of-scope; spec (#1229) does not cover it — reasonable, suggest cross-ref from #1258 |

<details>

**W-1 · Hardcoded empty-state strings** 🟢 High confidence
`usp_wifi_performance_card.dart:137, 254, 399`
Three `Center(child: AppText.bodyMedium('No WiFi clients connected'))` and `'No WiFi radios available'` literals exist outside l10n. These are **pre-existing** — this PR's new `_LegendEntry`/`_LegendDot` code correctly receives labels from `tier.resolveLabel(context)`. Not introduced here; flag for a follow-up ticket.
**Fix:** Use `loc(context).noWifiClientsConnected` (or equivalent); add l10n key if absent.

**W-2 · [Duplicated Code] `_LegendDot` fifth copy + #1245 inventory drift** 🟢 High confidence
`usp_wifi_performance_card.dart:320–331`
```dart
class _LegendDot extends StatelessWidget {
  final Color color;
  const _LegendDot({required this.color});
  @override
  Widget build(BuildContext context) {
    return Container(
      width: 8, height: 8,
      decoration: BoxDecoration(color: color, shape: BoxShape.circle),
    );
  }
}
```
Identical copies exist in: `usp_network_health_card.dart:451`, `usp_device_analytics_card.dart:448`, `usp_system_status_card.dart:541`, `usp_traffic_analysis_card.dart:683`. This PR is the 5th. PR doc itself notes "#1245's inventory is written against four files and needs `wifi_performance` added." Action: update #1245 issue body before or at merge to list all five files.
**Fix:** Update #1245 inventory. Extraction blocked pending Article XIV decision — follow that process.

**W-3 · `_snrMarker` locale assumption** 🟡 Med confidence
`wifi_performance_readability_test.dart:304`
```dart
const _snrMarker = 'SNR';
final snrs = texts.where((t) => t.data!.contains(_snrMarker)).toList();
```
Current test locales (`en`, `ru`) both have `"SNR: {value} dB"` — `contains('SNR')` works. Chinese ARBs use `"SNR：{value} dB"` (full-width colon) but ASCII 'SNR' prefix still matches. Risk: a future locale that doesn't preserve the ASCII prefix silently produces `snrs.isEmpty` → assertion fails with no explanation about the marker.
**Fix:** Add a comment: `// Assumes all tested locales retain the ASCII 'SNR' prefix.` Or derive from `AppLocalizations.delegate.load(locale).snrValue(0)`.

**W-4 · `_ChannelsTab` 48 px headroom untracked** 🟡 Med confidence
`usp_wifi_performance_card.dart:452–458`
```dart
Row(
  children: [
    AppText.labelLarge(radio.band),   // no Flexible/Expanded
    const Spacer(),
    AppText.bodySmall('Ch ${radio.channelDisplay}  ·  ${radio.channelBandwidth}', ...),
```
PR doc notes this shape has only ~48 px headroom, is data-dependent (tri-band `Ch 197 (Auto) · 320MHz` wider than mock), and is #1258's scope. No `known_overflows.json` key exists for this site — gate would catch it if it breaks, but with no context indicating the risk was known.
**Fix:** Add a `// TODO(#1258): monitor ~48px headroom` comment or a tracking note in the fixture.

**Spec-W1 · AC1 readability coverage gap** 🟡 Med confidence
`wifi_performance_readability_test.dart:173, 190`
```dart
for (final tag in ['en', 'ru', 'th']) {
  testWidgets('Signal tier names render in full in $tag', ...
```
AC1 requires "all 26 locales." Gate covers overflows but is blind to ellipsis truncation (a `Wrap` that fits by dropping content). The readability test is explicitly designed to catch that failure mode — but only for 3 locales. Speed tab original overflow locales include `es`, `fr`, `fr_CA`, `pt_PT`, `tr` — none of which are in the 3 tested. If a future translation grows wider than the available run, neither gate nor this test catches it.
**Fix:** Expand Speed-tab clip check to `['en', 'ru', 'tr', 'fr', 'pt_PT']`; Signal tab consider adding `'el'` (geometry-bound boundary locale).

</details>

<details>

- **Core fix is correct:** `Wrap` + `_LegendEntry(Flexible(child: AppText))` prevents unbounded label width; all entries are preserved (no dropping), matching the shape settled in #1226 and replicated by #1233.
- **Mutation ledger is exemplary:** Four mutations tabulated with both readability-test and gate results — this is the most complete ledger in the epic and correctly identifies that the gate-invisible regression (single `Row` + `Flexible`) is what the readability test is validated against.
- **Two-site attribution is thorough:** PR correctly identifies `:190` (Signal, geometry-bound, 33 coords) and `:275` (Speed, translation-bound, 12 coords) via a scoped re-run of §1.1's method; total 45 confirmed.
- **Spacing bump 12px→16px to `AppSpacing.lg` is consistent** with all six #1233 legends and #1226; measurement confirms it's affordable at 261px.
- **AC2/AC3 fully satisfied:** 45 coordinates removed (19+7+14+5 verified), 0 remaining `wifi_performance` allowlist entries confirmed.
- **AC4 pinned correctly:** Channels tab not modified but asserted anyway — "already clean" ≠ "checked" is sound reasoning.
- **New `_LegendEntry`/`_LegendDot` private widgets are clean:** single responsibility, honest names, use `AppText`/`AppGap`/`AppSpacing` from ui_kit_library throughout.
- **No security concerns:** no network calls, tokens, injections, or permission issues introduced.

</details>

*Cross-reviewed by three independent agents (security+correctness / architecture+maintainability / requirements-compliance). Automated — please sanity-check before merge.*

AustinChangLinksys (commented):
🤖 **Automated Review — Round 2** · `0ceba68..e3d3f4b` (incremental)

**Verdict: 💬 Self-review (comment only)** — GitHub prohibits self-approve; full review included as reference.

#### Standards

|---|---|---|---|
| ⚠️ | 🟢High | `usp_wifi_performance_card.dart:320–331` | [Duplicated Code] `_LegendDot` still the 5th verbatim copy; #1245 inventory still lists only 4 files — doc acknowledges but not yet updated |
| ⚠️ | 🟡Med | `wifi_performance_readability_test.dart:304` | `_snrMarker = 'SNR'` cross-locale assumption — currently safe (all 26 ARBs verified to use ASCII `SNR` prefix), but no guard against future translation drift |
| ⚠️ | 🟡Med | `wifi_performance_readability_test.dart:430–438` | `find.ancestor(of: band, matching: Column).first` — correctness of `.first` depends on `LayoutBlock` having no nested `Column`; recommend using a keyed finder |
| 💡 | 🟢High | `wifi_performance_readability_test.dart:453–459` | `_localeFor` still uses `firstWhere` without `orElse`; this round adds `'tr'` callsite — unclear `StateError` if tag not found |
| 💡 | 🟢High | `doc/dashboard/dashboard_density_design.md:3` | Header Status line lists up to `#1229`; `#1266` (implemented in this PR) not reflected in the top-level status summary |
| 💡 | 🟡Med | `wifi_performance_readability_test.dart:352–380` | `channels.reduce(...)` has no guard; `StateError` if `channels` is empty (guarded by prior `hasLength` assertion, but no early-return stops execution) |
| 💡 | ⚪Low | `doc/dashboard/dashboard_density_design.md` §2.10 | §2.10 now 3 sub-sections (~200+ lines); consider adding a brief "current state" summary at top of §2.10 pointing to §2.10c as the canonical conclusion |

#### Spec

|---|---|---|---|
| ⚠️ | 🟢High | `wifi_performance_readability_test.dart:219,300` | #1229 AC1: readability test still samples 3 locales (`en`/`ru`/`th` for legend; `en`/`ru`/`tr` for per-band) — not the full 26 locales required by spec |
| ⚠️ | 🟡Med | `usp_wifi_performance_card.dart:504` + test:335 | #1266 AC4: `clientsCount` has no `AppText` overflow param ✅ but `isClipped` loop covers only `bands`/`channels`/`snrs` — `clientsCount` not directly asserted clip-free |
| ✅ | 🟢High | `usp_wifi_performance_card.dart:494–496` | #1266 AC1: `'Ch '` replaced with `loc(context).channel`; all 26 locales have this ARB key |
| ✅ | 🟢High | `doc/dashboard/dashboard_density_design.md` §2.10c | #1266 AC2: gate 1644 green, 0 new allowlist entries; net coordinate delta = 0 (3 added by localization, 3 cleared by Wrap hardening) |
| ✅ | 🟢High | `doc/dashboard/dashboard_density_design.md` §2.10c | #1266 AC3: tri-band 3-radio measurement table recorded; fixture change not committed |
| ✅ | 🟢High | `usp_wifi_performance_card.dart:483–486` + doc §2.10c | #1266 AC5: no bottom overflow on 2-radio fixture (26 locales × both widths); tri-band +9.0px bottom filed as #1267 |
| ✅ | 🟢High | `wifi_performance_readability_test.dart:47–90,352–380` | #1266 AC6: AC4 group rewritten from "same line" to "attributable"; 3 new mutations added and verified against gate |
| ✅ | 🟢High | `doc/dashboard/dashboard_density_design.md` §2.10c | #1266 AC7: both measurement tables and the abbreviation-hiding-geometry finding documented |
| ✅ | 🟢High | `wifi_performance_readability_test.dart:283–380` | #1229 AC4: per-band metrics attributable invariant upheld at narrowest width in `en`/`ru`/`tr` |

<details>

**W-1 · [Duplicated Code] `_LegendDot` 5th copy + #1245 inventory drift** 🟢High (pre-existing, dual-agent)
`usp_wifi_performance_card.dart:320–331`
```dart
class _LegendDot extends StatelessWidget {
  final Color color;
  const _LegendDot({required this.color});
  @override
  Widget build(BuildContext context) {
    return Container(
      width: 8, height: 8,
      decoration: BoxDecoration(color: color, shape: BoxShape.circle),
    );
  }
}
```
Five identical copies confirmed: `usp_network_health_card.dart:499`, `usp_device_analytics_card.dart:448`, `usp_system_status_card.dart:511`, `usp_traffic_analysis_card.dart:621`, and this file. The design doc (§1.1 line ~124) acknowledges "#1245's inventory is written against four files and needs `wifi_performance` added". Pre-existing; not introduced by this PR. #1245 is the tracking ticket.
**Fix:** Update #1245 issue body to list all 5 files; extraction decision pending Article XIV.

---

**W-2 · `_snrMarker = 'SNR'` locale assumption** 🟡Med (pre-existing, dual-agent)
`wifi_performance_readability_test.dart:304`
```dart
const _snrMarker = 'SNR';
final snrs = texts.where((t) => t.data!.contains(_snrMarker)).toList();
```
Verified across all 26 ARB files: every locale currently uses the ASCII `SNR` prefix (e.g. `app_tr.arb`: `"SNR: {value} dB"`, `app_th.arb`: `"SNR: {value} dB"`). Currently safe. Risk is maintenance-only: a future translator could drop the ASCII prefix and this test would silently collect zero SNR widgets, passing vacuously.
**Fix:** Add a comment: `// All 26 locales verified to retain ASCII 'SNR' prefix (checked 2026-08-13).`

---

**W-B2 · `find.ancestor(Column).first` may grab wrong Column** 🟡Med (Reviewer B, single-agent)
`wifi_performance_readability_test.dart:430–438` (approx.)
```dart
final columnRect = tester.getRect(find
    .ancestor(of: find.byWidget(band), matching: find.byType(Column))
    .first);
```
`.first` is described in comments as "the innermost — the per-radio block's own Column". This is correct only if `LayoutBlock` contains no nested `Column` between `band` and the target. If `LayoutBlock` is a transparent wrapper (e.g. `Padding`), the assumption holds; if it wraps in `Column`, `.first` returns a narrower rect and the `wrapRect.width >= columnRect.width - 1.0` assertion may produce false passes or false failures.
**Fix:** Key the per-radio `Column` (e.g. `ValueKey('radio_col_$i')`) and use `find.byKey(...)` instead of topology-dependent `.first`.

---

**Spec-W1 · #1229 AC1 locale coverage: 3 of 26 locales in readability test** 🟢High (Reviewer C)
`wifi_performance_readability_test.dart:219, 300`
```dart
for (final tag in ['en', 'ru', 'th']) {  // legend group
for (final tag in ['en', 'ru', 'tr']) {  // per-band group (tr added this round)
```
Spec AC1 requires the card renders clean "across all 26 locales." Gate covers the full 26 for overflow (1644 green), but readability test (soft-clip / alignment) still samples only 3. Adding `'tr'` to the per-band group is a meaningful improvement (it's the worst-case locale for `#1266`). The gap is a known structural limitation — not introduced by this PR.
**Fix:** Expand legend group to at minimum add `'es'`, `'fr'`, `'tr'`, `'pt_PT'` (worst-case Translation-bound locales per the overflow epic's dataset).

---

**Spec-W2 · #1266 AC4 `clientsCount` not directly asserted clip-free** 🟡Med (Reviewer C)
`usp_wifi_performance_card.dart:504` + `wifi_performance_readability_test.dart:335`
```dart
// card: no maxLines/overflow on clientsCount — correct
AppText.bodySmall(loc(context).clientsCount(clientCount)),
// test: isClipped loop
for (final t in [...bands, ...channels, ...snrs]) {
  expect(isClipped(tester, t), isFalse, ...);  // clientsCount not in this list
}
```
`snrValue` is covered via `snrs`; `clientsCount` is not explicitly included. Attribution log confirms all overflow incidents were on the band/channel row (zero on the clients/SNR row), so the omission is unlikely to be a real problem in practice. But the AC specifies it "never ellipsizes."
**Fix:** Add `clientsCount` widgets to the `isClipped` assertion loop.

</details>

<details>

- **#1266 core fix is correct and elegant:** `'Ch '` → `loc(context).channel` eliminates the measurement-optimism caused by the abbreviation; `Row+Spacer` → `Wrap(spaceBetween)` + `CrossAxisAlignment.stretch` correctly handles bounded-width constraint. The PR correctly identifies that `stretch` is load-bearing (without it, `freeMainAxisSpace` = 0 and `spaceBetween` is a silent no-op).
- **Self-verifying structure:** Localization supplies the failure (3 gate coordinates on shipped fixture), hardening clears them; net delta = 0. This is the canonical self-verifying pattern from the overflow epic.
- **Attribution was rigorous:** Per-line `fullLog` regression scoped the fix to only the band/channel row; the clients/SNR row is correctly left untouched, confirmed by zero incidents on that row.
- **AC6 rewrite is sound:** The "same line → attributable" invariant change is correct — `Wrap` legitimately allows channel to move to a second run, so the prior invariant was an over-constraint. New invariant (each channel nearest its own band, not below the next band label) faithfully preserves the original intent.
- **3 new mutations + gate cross-check:** The mutation table now covers the two regressions the gate cannot detect (`stretch→start` and `loc→'Ch '` both fail readability, stay green on gate) and one it can. This is exemplary test documentation.
- **W-3 (`_snrMarker`) and W-4 (48px headroom) from Round 1 fully resolved:** `_snrMarker` verified safe across all 26 ARBs; #1266 eliminates the headroom measurement error and correctly documents the trade-off in §2.10c.
- **§2.10c is high-quality engineering documentation:** Records the abbreviation-as-measurement-hazard insight, both measurement tables, tri-band limitation, and #1267 tracking — complete audit trail.
- **No security issues:** No network calls, tokens, permissions, or injection vectors introduced.

</details>

*Cross-reviewed by three independent agents (security+correctness / architecture+maintainability / requirements-compliance). Automated — please sanity-check before merge.*

PR #1263, opened 2026-08-13
refactor(ui): draw the "maps to" arrow as an icon, not U+2192 (#1247)

AustinChangLinksys:
## Summary

U+2192 (`→`) has **no glyph in any of the eleven fonts the app declares**. Browsers render it anyway by resolving a host font outside that set — which is exactly the dependency the bundled fallbacks exist to remove. This replaces the eight rendered occurrences with `AppIcon.font(Icons.arrow_forward)`, extending what `31d71e0f` already did for U+2191/U+2193/U+23F1/U+26A0.

Closes #1247.

## Why not the cheaper options the ticket proposed

The ticket recommended option 1 (bundle a covering fallback) then option 2 (ASCII/en-dash). Measuring first showed **option 1 as written cannot work**:

- The Noto fallbacks are **locale-gated**. `FallbackFontResolver.bareFamilyForLocale` returns `null` for en/fr/de/es/pt, so for those locales the chain is **NeueHaas → Roboto only** (429 and 896 codepoints). All eight sites are hardcoded English strings — no ARB carries U+2192 — so adding arrows to a CJK subset would fix nothing at any of them.
- The full Noto CJK OTFs under `tools/font_subset/full_fonts/` **do** map U+2192; only the generated subsets drop it. That is what makes option 1 look viable when it isn't.
- The one viable form of option 1 is re-subsetting **Roboto** (the sole fallback declared for every locale). Rejected: `regenerate.sh` regenerates only the five CJK subsets and treats the non-CJK fallbacks as fixed committed artifacts. Still the right move if arrow coverage is ever wanted in bulk.

Measured with fontTools against the pinned `ui_kit_library` (`v2.34.5`, `bcee4067`):

| Codepoint | Union of the 11 text fonts |
|---|---|
| U+2192 `→`, U+2190, U+2191, U+2193 | **none** |
| U+2013 `–`, U+2014 `—` | all 11 |
| U+00BB `»`, U+203A `›` | 6 of 11 |

Full rationale, rules and a runnable re-measurement snippet: `doc/theme/unicode_glyph_coverage_decision.md`.

## Approach — AC #3 is preserved, not overridden

The ticket assumed option 3 forces `portSummary`/`summary` from `String` to `Widget`. It doesn't. `MapsToRow` takes `source` and `target` as **Strings** and composes `AppText + AppIcon.font + AppText` in the widget layer, so **the UI models keep returning Strings** — AC #3 holds, and no widgets leak into models.

- `MapsToRow` (`row_blocks.dart`) is the single definition of the arrow.
- Each model gained getters for the two halves; the composed `…summary` getters remain for diagnostics and now use ASCII `->`.
- `ToggleRow` gained `subtitleContent` (`Widget?`) alongside its String `subtitle`, since two sites are `ToggleRow` subtitles. Existing String callers unaffected.
- Per Article XIV: UI Kit has **no** inline-icon primitive (no `WidgetSpan` anywhere in it), so this composes existing atoms rather than adding a component.

Out of scope by design: logger, console and LLM-prompt strings never pass through Flutter's font stack. `lib/ai/prompts/` and `[USP]`-prefixed logs keep their arrows.

## ⚠️ One deliberate visual change — please eyeball this

`usp_firewall_overview_card.dart` read `'${rule.portSummary} → ${rule.internalClient}'`, which rendered:

```
8080 → 192.168.1.100:80 → 192.168.1.100
```

— the internal client twice, once bare and once already inside `portSummary`. It is now a single `8080 → 192.168.1.100:80`. Preserving both arrows would have drawn two icons for duplicate information. This is the only site whose output changes beyond glyph→icon.

`pnp_demo_launcher.dart` uses an en dash instead: it's a three-step sequence, not a mapping.

## Verification

- **New guard test** `test/page/_shared/components/layout_blocks/maps_to_row_test.dart` — asserts no rendered `Text` carries U+2192, that the arrow is an `Icon`, and that long operands ellipsize without overflow at 120/140px. Tagged `dashboard-card` (not `ui`) so `run_tests.sh` includes it. **Mutation-checked**: reverting `MapsToRow` to a character arrow fails 3 of its 5 tests.
- `./run_tests.sh` — **5352 passing**.
- `dashboard_card_overflow_test.dart` — **1644 passing, `known_overflows.json` untouched**. Both `port_forwarding` and `firewall_overview` are gate-probed, and the icon is narrower than the glyph box it replaced.
- Affected golden tests (port_forwarding, dashboard cards, statistics, firewall) regenerated and re-run green.
- `dart format` clean; `flutter analyze` reports no error/warning on any changed file.

## Base branch

Targets **`gate/1183-overflow-gate`** (#1248), not `dev-2.7.0` — the overflow gate, its fixture and its probe only exist there, and the arrow sites sit inside two gate-probed cards. Rebased onto the current base tip, including the stricter probe from `30c2b8f7`.

### DIFF
--- a/doc/theme/unicode_glyph_coverage_decision.md
+++ b/doc/theme/unicode_glyph_coverage_decision.md
@@ -1,0 +1,171 @@
+# Unicode Symbols in UI Strings — Glyph Coverage Decision
+
+**Last Updated: 2026-08-12** · Resolves #1247 · Follow-up to #1249 (`31d71e0f`) · Status: **decided and implemented**
+
+## Purpose
+
+The app bundles eleven fonts so that every locale renders offline, without
+network and without relying on host fonts. A UI string that uses a codepoint
+none of those eleven map is therefore a latent defect even when it looks
+correct: the browser is resolving a font the app never declared.
+
+This document records which codepoints are safe to put in UI strings, the
+measurement behind that, and the decision taken for U+2192.
+
+## The declared font set
+
+Eleven fonts, all eager-loaded and registered before the first frame:
+
+| Font | Role |
+|---|---|
+| `NeueHaasGrotTextRound-55Roman` / `-75Bold` | Primary (from `ui_kit_library`) |
+| `NotoSans-Latin` | Greek, Cyrillic, Vietnamese (`el`/`ru`/`vi`) |
+| `NotoSansCJK{sc,tc,hk,jp,kr}.subset` | CJK, per locale |
+| `NotoSansThai`, `NotoSansArabic` | `th`, `ar` |
+| `Roboto` | Engine's global fallback, declared bare so it resolves locally |
+
+Two things about the chain matter for this decision:
+
+1. **The Noto fallbacks are locale-gated.** `FallbackFontResolver`
+   (`lib/localization/fallback_font_resolver.dart`) returns `null` for
+   en/fr/de/es/pt and the other Latin locales. For those locales the chain is
+   **NeueHaas → Roboto only** — 429 and 896 codepoints respectively. A glyph
+   present only in a CJK subset does *not* rescue an English string.
+2. **The CJK subsets are generated from ARB text.** `tools/font_subset/`
+   subsets against the characters found in the `zh`/`zh_TW`/`ja`/`ko` ARB files
+   plus ASCII and the CJK punctuation blocks. Symbols that appear only in
+   hardcoded Dart literals are not in that charset even though the full source
+   fonts have them.
+
+## Measurement
+
+Measured with fontTools against the pinned `ui_kit_library` (`v2.34.5`,
+`bcee4067`) and `assets/fonts/fallback/`. `YES` = the font's `cmap` maps it.
+
+| Codepoint | NeueHaas | Roboto | NotoSans-Latin | Thai/Arabic | CJK subsets | **Union** |
+|---|---|---|---|---|---|---|
+| U+2192 `→` | no | no | no | no | no | **none** |
+| U+2190 `←` | no | no | no | no | no | **none** |
+| U+2191 `↑` | no | no | no | no | no | **none** |
+| U+2193 `↓` | no | no | no | no | no | **none** |
+| U+2013 `–` | YES | YES | YES | YES | YES | **all 11** |
+| U+2014 `—` | YES | YES | YES | YES | YES | **all 11** |
+| U+00BB `»` | YES | YES | YES | YES | no | 6 of 11 |
+| U+203A `›` | YES | YES | YES | YES | no | 6 of 11 |
+
+The arrows resolve in every browser tested, which is why this was never visible
+as a bug. That resolution comes from a system font outside the declared set —
+precisely the dependency the fallbacks exist to remove.
+
+The full Noto Sans CJK OTFs under `tools/font_subset/full_fonts/` **do** map
+U+2192. Only the generated subsets drop it. This is what makes "add arrows to
+the CJK subset" look viable when it is not: it would fix nothing for the Latin
+locales, which is where all eight sites render.
+
+## Decision
+
+**U+2192 in a rendered UI string is replaced by `AppIcon.font(Icons.arrow_forward)`.**
+Icon fonts ship with the app, so coverage is guaranteed offline. This extends
+`31d71e0f`, which did the same for U+2191/U+2193/U+23F1/U+26A0, to the
+"maps to" separator.
+
+Chosen over two cheaper alternatives:
+
+- **Re-subsetting `Roboto` to add the Arrows block** would fix all sites with no
+  code change, since Roboto is the one fallback declared for every locale. It
+  was rejected because `tools/font_subset/regenerate.sh` regenerates only the
+  five CJK subsets and treats the non-CJK fallbacks as fixed committed
+  artifacts; adding Roboto to that pipeline makes a vendored binary a
+  maintenance surface for a separator. Still the right move if arrow coverage
+  is ever wanted in bulk.
+- **An ASCII/en-dash separator** is zero-risk (U+2013 is in all eleven) but
+  loses the direction the arrow carries, and leaves the app inconsistent with
+  the icon treatment `31d71e0f` established.
+
+### Rules
+
+1. A rendered UI string **must not** depend on a codepoint outside the union
+   above. Direction markers are drawn with `AppIcon.font`; U+2013/U+2014 are
+   safe as separators.
+2. **Logger, console and LLM-prompt strings are out of scope.** They never pass
+   through Flutter's font stack. `lib/ai/prompts/` and `[USP]`-prefixed log
+   strings keep their arrows deliberately.
+3. Anything added to `lib/util/languages.dart` or the CJK ARB files needs
+   `tools/font_subset/regenerate.sh` re-run, per that script's header.
+
+## Implementation
+
+`MapsToRow` (`lib/page/_shared/components/layout_blocks/row_blocks.dart`) is the
+single definition of the arrow: it takes `source` and `target` as **Strings** and
+composes `AppText + AppIcon.font + AppText`.
+
+This keeps `PortForwardingRuleUIModel.portSummary` and
+`PortTriggeringRuleUIModel.summary` returning Strings — #1247 AC #3 — rather
+than pushing widgets into UI models. Each model gained getters for the two
+halves (`portRangeDisplay`/`internalTargetDisplay`,
+`triggerSummaryPart`/`forwardSummaryPart`); the composed `…summary` getters
+remain for diagnostics and now use ASCII `->`.
+
+`ToggleRow` gained `subtitleContent` (a `Widget?`) alongside its String
+`subtitle`, because two of the sites are `ToggleRow` subtitles. Existing String
+callers are unaffected.
+
+**One deliberate visual change.** `usp_firewall_overview_card.dart:255` read
+`'${rule.portSummary} → ${rule.internalClient}'`, rendering
+`8080 → 192.168.1.100:80 → 192.168.1.100` — the internal client twice, once
+bare and once inside `portSummary`. It is now a single
+`8080 → 192.168.1.100:80`. Preserving both arrows would have drawn two icons
+for duplicate information.
+
+### Sites converted
+
+| File | Note |
+|---|---|
+| `lib/page/_shared/models/port_forwarding_rule_ui_model.dart` | getter split; `portSummary` → ASCII |
+| `lib/page/port_forwarding/models/port_triggering_rule_ui_model.dart` | getter split; `summary` → ASCII |
+| `lib/page/statistics/views/sections/stats_port_mapping_section.dart` | `MapsToRow` |
+| `lib/page/firewall/cards/usp_firewall_overview_card.dart` | `MapsToRow`; duplicate arrow collapsed |
+| `lib/page/port_forwarding/cards/usp_port_forwarding_card.dart` | two `ToggleRow` subtitles |
+| `lib/page/port_forwarding/views/components/usp_{single_port,port_range,port_triggering}_tab.dart` | `MapsToRow` |
+| `lib/ai/registry/router_component_registry.dart` | three `AppListTile` subtitles |
+| `lib/demo/pages/pnp_demo_launcher.dart` | en dash — a three-step sequence, not a mapping |
+
+## Verification
+
+- `test/page/_shared/components/layout_blocks/maps_to_row_test.dart` asserts no
+  rendered `Text` contains U+2192, that the arrow is an `Icon`, and that long
+  operands ellipsize without overflow. Tagged `dashboard-card` (not `ui`) so
+  `run_tests.sh` includes it. Mutation-checked: reverting `MapsToRow` to a
+  character arrow fails 3 of its 5 tests.
+- `./run_tests.sh` — 5352 passing.
+- `dashboard_card_overflow_test.dart` — 1644 passing, allowlist unchanged.
+  Both `port_forwarding` and `firewall_overview` are gate-probed, and the icon
+  is narrower than the glyph box it replaced, so no coordinate moved.
+
+## Reproducing the measurement
+
+Run from the repo root. `LinksysIcons.otf` is excluded on purpose: it is the
+icon font `AppIcon.font` draws from, not part of the *text* fallback chain, so
+counting it would mask exactly the gap being measured.
+
+```bash
+python3 -m venv /tmp/fc && /tmp/fc/bin/pip install -q fonttools brotli
+/tmp/fc/bin/python - <<'PY'
+from fontTools.ttLib import TTFont
+import glob, os
+paths = glob.glob('assets/fonts/fallback/*.woff2') + [
+    p for p in glob.glob(os.path.expanduser(
+        '~/.pub-cache/git/privacyGUI-UI-kit-bcee4067*/assets/fonts/*.otf'))
+    if 'LinksysIcons' not in p]
+assert len(paths) == 11, f'expected 11 text fonts, found {len(paths)}'
+union = set()
+for p in paths:
+    f = TTFont(p, fontNumber=0, lazy=True)
+    union |= set(f.getBestCmap()); f.close()
+for cp in (0x2192, 0x2013):
+    print(hex(cp), 'covered' if cp in union else 'MISSING')
+PY
+```
+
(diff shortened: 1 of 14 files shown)


AustinChangLinksys (commented):
🤖 **Automated Review — Round 1** · `29450e0..5705670` (full)

**Verdict: ✅ APPROVE** — Clean refactor; U+2192 fully replaced across all 11 UI sites. Two architectural warnings to address.

#### Standards (Reviewer A · Security+Correctness / Reviewer B · Architecture+Maintainability)

|---|---|---|---|
| ⚠️ | 🟢High | `lib/ai/registry/router_component_registry.dart:3` | [Feature Envy] New `lib/page/_shared` import introduces first-ever cross-layer dependency from AI registry into page layer |
| ⚠️ | 🟢High | `lib/page/_shared/components/layout_blocks/row_blocks.dart:251` | `subtitleContent ?? subtitle` silent-priority logic has zero `ToggleRow` test coverage |
| 💡 | 🟢High | `lib/page/_shared/components/layout_blocks/row_blocks.dart` + 8 call sites | `MapsToRow` icon color uses `IconTheme` while `AppText` uses `DefaultTextStyle`; color may diverge when `color:` is omitted |
| 💡 | 🟡Med | `lib/page/_shared/components/layout_blocks/row_blocks.dart:251` | No `assert(subtitle == null \|\| subtitleContent == null, …)` guard for both-provided misuse |
| 💡 | 🟡Med | `lib/page/port_forwarding/models/port_triggering_rule_ui_model.dart:109` | `*SummaryPart` suffix inconsistent with `*Display` suffix in sibling model's new getters |
| 💡 | 🟡Med | `test/page/_shared/components/layout_blocks/maps_to_row_test.dart` | No test for source-side ellipsis with icon still visible under narrow container |
| 💡 | ⚪Low | `lib/ai/registry/router_component_registry.dart:856` | `MapsToRow(source: '' , target: '')` when LLM data has null mac/ip — renders orphan arrow icon |

#### Spec (Reviewer C · Issue #1247 compliance)

|---|---|---|---|
| ⚠️ | 🟢High | `doc/theme/unicode_glyph_coverage_decision.md` (option 1 section) | Decision doc only cites maintenance cost for option 1; omits the more fundamental "CJK subset is invalid for Latin locales" argument (present in commit message, absent in doc) |
| 💡 | 🟡Med | `lib/page/firewall/cards/usp_firewall_overview_card.dart:254` | Dual-arrow folding (8080→192.168.1.100:80, `internalClient` dropped) is undocumented as a spec-level visual decision; spec listed both arrows as separate sites |

<details>

**W-1 — [Feature Envy] New cross-layer import in AI registry** · 🟢High
`lib/ai/registry/router_component_registry.dart:3`

Base version had no `lib/page` imports. Head adds:
```dart
import 'package:privacy_gui/page/_shared/components/layout_blocks.dart';
```
This is the first time `lib/ai/registry/` depends on `lib/page` — a direction that may be unintended given `lib/ai/` is typically parallel to (or above) page-layer code. `MapsToRow` is a general-purpose "source → target" widget; housing it under `page/_shared` forces any non-page consumer to import into the page tree.

**Fix:** Either (a) move `MapsToRow` to a more neutral location (e.g. a shared components barrel reachable without importing `page/_shared`), or (b) confirm in AGENTS.md that `lib/ai/registry/` is explicitly permitted to import from `lib/page/_shared`. As-is the import works but sets a cross-layer precedent.

---

**W-2 — ToggleRow subtitleContent: no test coverage** · 🟢High
`lib/page/_shared/components/layout_blocks/row_blocks.dart:251`

```dart
subtitle: subtitleContent ??
    (subtitle != null
        ? AppText.bodySmall(subtitle!, ...)
        : null),
```

When a caller passes both `subtitle` and `subtitleContent`, `subtitle` is silently discarded. The doc comment says "Takes precedence over [subtitle]" — that is correct behaviour — but `test/` has zero `ToggleRow` widget tests (confirmed: `find test -name "*.dart" | xargs grep -l "ToggleRow"` → no results). If the priority logic is later reversed by mistake, no test would catch it.

**Fix:** Add at minimum one `ToggleRow` widget test covering (a) only `subtitle` shown, (b) only `subtitleContent` shown, (c) `subtitleContent` takes priority when both provided. Optionally add `assert(subtitle == null || subtitleContent == null, 'ToggleRow: provide subtitle or subtitleContent, not both')`.

---

**Spec W-1 — Decision doc missing technical infeasibility of option 1** · 🟢High
`doc/theme/unicode_glyph_coverage_decision.md` (option 1 section)

The commit message correctly explains: "FallbackFontResolver returns null for en/fr/de/es/pt, so the chain for Latin locales is NeueHaas → Roboto only — adding arrows to a CJK subset would fix nothing at any of the eight sites (all hardcoded English, no U+2192 in any ARB)." The decision doc's option 1 section only cites the Roboto pipeline maintenance cost, omitting this more fundamental point.

**Fix:** Add one sentence to the decision doc's option 1 paragraph: "More fundamentally, the CJK subsets are only loaded for CJK locales; for en/fr/de/es/pt the font chain is NeueHaas → Roboto only, and all eight U+2192 sites are hardcoded English strings (no U+2192 appears in any ARB file), so patching a CJK subset would not affect any of them."

</details>

<details>
<summary>💡 Suggestion / What looks good</summary>

**Suggestions:**

**S-1 — Icon color inheritance diverges from text color when `color:` omitted** · 🟢High
`row_blocks.dart` (MapsToRow.build)

`AppText.bodySmall` resolves color via `DefaultTextStyle.of(context)`. `AppIcon.font` resolves via `IconTheme.of(context)`. In `AppListTile.subtitle`, only `DefaultTextStyle` is set — no `IconTheme` wrapper — so the icon color may diverge from the adjacent text color in some themes when `color:` is not explicitly passed. Most affected: `router_component_registry.dart` three `MapsToRow` calls and `usp_firewall_overview_card.dart`, `stats_port_mapping_section.dart`.

Fix: In `MapsToRow.build()`, compute `final effectiveColor = color ?? DefaultTextStyle.of(context).style.color;` and pass it to both `AppText` and `AppIcon.font`.

**S-2 — Naming inconsistency: `*SummaryPart` vs `*Display`** · 🟡Med
`port_triggering_rule_ui_model.dart:109–120`

`PortForwardingRuleUIModel` uses `portRangeDisplay`/`internalTargetDisplay`; `PortTriggeringRuleUIModel` uses `triggerSummaryPart`/`forwardSummaryPart`. Both pairs serve the same role (MapsToRow source/target). Consider aligning to one convention.

**S-3 — Missing source-ellipsis test** · 🟡Med
`maps_to_row_test.dart`

Existing overflow tests cover long `target` → ellipsis. `source` is also `Flexible` + ellipsis but has no dedicated regression test for the case where source overflows while icon remains visible.

**S-4 — Orphan icon for empty-string LLM data** · ⚪Low
`router_component_registry.dart:856`

`reservation['mac'] as String? ?? ''` → `MapsToRow(source: '', ...)` renders as a bare icon. Pre-existing behaviour (old code used `→` in a similar string), but `MapsToRow`'s semantics make the empty-operand case more visible.

---

**✅ What looks good:**

- **All 11 U+2192 UI sites removed** — full grep of `lib/` confirms zero rendered U+2192 remaining; only `debugPrint`/log strings retain it (spec-excluded).
- **Model layer stays String-typed** — both `portSummary` and `summary` remain `String` getters (demoted to diagnostics); new `*Display`/`*SummaryPart` getters also `String`. No Widget types leak into models. AC3 ✅.
- **Deep-module design** — `MapsToRow(source:, target:)` hides Row + 2 Flexible + AppIcon + AppGap complexity. Deletion test passes: removing it would force 8 call sites to repeat the layout.
- **`MapsToRow` overflow protection** — both operands are `Flexible` + `maxLines: 1` + `ellipsis`; the icon is non-flexible and never squeezed out. Verified by test "icon is not squeezed out by long operands".
- **`pnp_demo_launcher.dart` uses U+2013** — 3-step sequence is correctly not a mapping; U+2013 is covered by all 11 declared fonts. Decision documented.
- **`ToggleRow.subtitleContent` is backward-compatible** — all existing `subtitle: String` callers unaffected.
- **Thorough decision documentation** — `doc/theme/unicode_glyph_coverage_decision.md` with fontTools coverage table, rejected options with evidence, re-runnable measurement script.
- **Test mutation-checked** — reverting `MapsToRow` to a character arrow fails 3 of 5 tests; test is not vacuously green.
- **AC1–AC4 all satisfied** (AC4 at medium confidence pending local test run).
</details>

---

**B2 verdict note:** Both Warnings are single-reviewer findings (not double-verified). Per dispatch-chain rules they are flagged here for awareness but no auto-fix cards are queued; they are candidates for a follow-up if confirmed.

AustinChangLinksys (commented):
🤖 **Automated Review — Round 2** · `5705670..d155ace` (incremental)

**Verdict: ✅ APPROVE** — Follow-up commit correctly addresses all Round 1 warnings; no new Criticals.

#### Standards (Reviewer A · Security+Correctness / Reviewer B · Architecture+Maintainability)

|---|---|---|---|
| ⚠️ | 🟡Med | `lib/page/_shared/components/layout_blocks/row_blocks.dart:170` | `effectiveColor` has no third-level null guard; if ancestor has no `DefaultTextStyle.color`, AppText and AppIcon fall back via different chains (diverge silently in bare contexts) |
| ⚠️ | 🟡Med | `lib/ai/registry/router_component_registry.dart:3` | [Feature Envy] W-1 rebuttal partially valid — `lib/ai` does import `lib/page` broadly, but those are Provider/Utils imports; the `layout_blocks.dart` import is the only Widget-layer cross-dependency and remains an architectural concern (pre-existing, not introduced by this incremental commit) |
| 💡 | 🟡Med | `lib/page/_shared/components/layout_blocks/row_blocks.dart:215,233` | `assert` docstring "in release builds this one wins" may imply passing both is semi-legal; recommend clarifying that this is a programming error whose release behavior is an implementation detail, not a guarantee |
| 💡 | ⚪Low | `test/page/_shared/components/layout_blocks/toggle_row_test.dart:128-131` | Color assertion fallback queries `DefaultTextStyle` at `Text` widget level; slightly different context than `MapsToRow.build()`; fragile if a new `DefaultTextStyle` wrapper is inserted between them |

#### Spec (Reviewer C · Issue #1247 compliance)

|---|---|---|---|
| 💡 | 🟢High | `doc/theme/unicode_glyph_coverage_decision.md:93-96` | `Rules` section has a duplicate `1.` list item — new color-rule entry should be `2.`; raw-source ambiguity if future maintainers cite rule numbers |

<details>

**W-1 (Round 2) — `effectiveColor` missing third-level null guard** · 🟡Med
`lib/page/_shared/components/layout_blocks/row_blocks.dart:170`

```dart
final effectiveColor = color ?? DefaultTextStyle.of(context).style.color;
```

`DefaultTextStyle.of(context).style.color` is typed `Color?` and can be null if no ancestor sets a `DefaultTextStyle` with a color. In the current production call sites (all under `AppListTile`), the subtitle `DefaultTextStyle` is non-null (`effectiveContentColor.withValues(alpha: 0.7)`), so all live paths are safe. However, if a future caller places `MapsToRow` in a bare `Scaffold` or a container without a colored `DefaultTextStyle`, both `AppText` and `AppIcon` receive `null` but fall back via *different* chains:
- `AppIcon`: `null → IconTheme.color → Colors.black`
- `AppText`: `null → inherited color → theme.surfaceBase.contentColor`

This silently breaks the "pair consistent" guarantee `MapsToRow` is meant to provide.

**Fix:** Add a third-level fallback:
```dart
final effectiveColor = color
    ?? DefaultTextStyle.of(context).style.color
    ?? Theme.of(context).colorScheme.onSurface;
```

---

**W-2 (Round 2) — [Feature Envy] Cross-layer Widget import in AI registry (pre-existing)** · 🟡Med
`lib/ai/registry/router_component_registry.dart:3`

The Round 1 W-1 rebuttal is partially valid: `lib/ai` does contain 14 other `import 'package:privacy_gui/page/...'` lines in `usp_command_provider.dart` and `speed_markers.dart`. However, those 14 imports all reference `page/_shared/providers/` or `page/_shared/utils/` — the shared service/provider layer — and represent a horizontal dependency, not an upward one.

The import in `router_component_registry.dart:3` (`layout_blocks.dart`) is the only import of a **Widget component** from `lib/ai/`, which represents an upward dependency from AI-logic layer into the Widget layer. The rebuttal conflates these two different kinds of import. The concern is not new to this incremental commit; it exists since the Round 1 SHA. Recommended action: open a follow-up issue to consider moving `MapsToRow` to a more neutral barrel (or document in `AGENTS.md` that this cross-layer dependency is intentional).

</details>

<details>
<summary>✅ Round 1 warnings — resolution status</summary>

**Round 1 W-1 [Feature Envy] — cross-layer import** · Rebuttal partially valid (see W-2 above). Pre-existing, not introduced by this incremental commit. No change required for this PR.

**Round 1 W-2 — ToggleRow zero test coverage** ✅ Resolved. `toggle_row_test.dart` added with 4 test paths: String subtitle channel, Widget `subtitleContent` channel, both-null (no subtitle), and `AssertionError` on both-provided. Color invariant test added and mutation-checked.

**Round 1 S-1 — Icon color divergence from text color** ✅ Resolved. `effectiveColor = color ?? DefaultTextStyle.of(context).style.color` is computed once in `MapsToRow.build()` and passed to both `AppText.bodySmall` and `AppIcon.font`. The residual gap (no third-level null guard) is noted as W-1 above; all current call sites are safe.

**Round 1 Spec W-1 — Decision doc missing CJK locale-gating argument** ✅ Resolved. Decision doc now clearly states: *"Those subsets are only loaded for CJK locales; for en/fr/de/es/pt the chain is NeueHaas → Roboto only, and all eight U+2192 sites are hardcoded English strings with no U+2192 in any ARB file."*

**Round 1 Spec S-1 — Dual-arrow folding undocumented** ✅ Resolved. Decision doc now has a "One deliberate visual change" paragraph documenting the `8080 → 192.168.1.100:80 → 192.168.1.100` → `8080 → 192.168.1.100:80` consolidation and the reasoning.

</details>

<details>
<summary>💡 Suggestions / What looks good</summary>

**Suggestions:**

**S-1 — `assert` docstring clarification** · 🟡Med
`row_blocks.dart:215–218`

The docstring reads: "in release builds this one wins." This is technically accurate but could be misread as "passing both in production is acceptable behavior." Suggest rephrasing: "Do not pass both — this is a programming error. The assert fires in debug mode; in release mode `subtitleContent` takes priority as an implementation detail, not a guarantee."

**S-2 — Test color assertion fragility** · ⚪Low
`toggle_row_test.dart:128-131`

The `textColor` fallback queries `DefaultTextStyle.of(tester.element(find.text('8080')))` which is the `DefaultTextStyle` at the `Text` widget's element, not at `MapsToRow.build()`'s context. In the current test tree they resolve to the same value. If a future refactor inserts a new `DefaultTextStyle` between `MapsToRow` and `Text`, the assertion could silently pass while testing the wrong invariant. Minor fragility; acceptable for now.

**Spec S-2 — Duplicate `1.` in Rules section** · 🟢High
`doc/theme/unicode_glyph_coverage_decision.md:93-96`

The new color rule was added with number `1.` but the section already had a `1.` entry. Raw Markdown renders this as two items both labeled "1" (some renderers auto-correct, some don't). Update the new entry to `2.` and renumber subsequent entries accordingly.

---

**✅ What looks good (this incremental commit):**
- **S-1 color fix is correct and well-tested**: `effectiveColor` resolved once at `MapsToRow.build()` context and handed to both children eliminates the `DefaultTextStyle` vs `IconTheme` divergence at all current call sites.
- **ToggleRow test coverage is thorough**: 4 test paths covering both channels, neither channel, mutual exclusion, and the color invariant. Mutation-checked per commit message (dropping the `DefaultTextStyle` fallback fails the color test).
- **`assert` for mutual exclusion**: Using initializer-list `assert` is idiomatic Flutter; placing it in the constructor correctly makes the constraint visible at widget creation, not at render time.
- **Decision doc is now complete**: The updated doc covers the CJK locale-gating argument (the more fundamental reason option 1 cannot work), the deliberate visual change in `usp_firewall_overview_card.dart`, and the measurement methodology.
- **All 4 ACs satisfied**: AC1 (decision recorded with font coverage answer ✅), AC2 (no UI string depends on out-of-set codepoint ✅), AC3 (model getters remain `String` ✅), AC4 (tests pass, gate unchanged ✅).
- **No security concerns**: Pure UI widget composition; no network calls, tokens, permissions, or data serialization involved.

</details>

---

**B2 verdict note:** W-A1 is a single-reviewer finding (not double-verified). W-B1 is pre-existing (introduced in Round 1 commit, not this incremental commit). Per dispatch-chain rules, neither meets the "double-verified + own-range" threshold for auto-fix card creation. Flagged here for awareness.

Issue #1267, opened 2026-08-13
[T06c] Give the overflow gate a second data profile, and decide what the Channels tab drops at that density


AustinChangLinksys:
## Parent

#1183

Design of record: `doc/dashboard/dashboard_density_design.md` §2.10c (the measurement), §2.10a point 3 (the shape), §2.7 / §2.9 (why the sweep's size and the allowlist arithmetic are not free).

## Branch

**Base: `gate/1183-overflow-gate` (PR #1248). Not `dev-2.7.0`.**

The gate, its fixture, its probe and the design of record are all introduced *by* #1248 and are absent on `dev-2.7.0`, so any acceptance criterion phrased against `known_overflows.json` cannot be executed there. Branch from `gate/1183-overflow-gate` and merge back into it. Do not create a branch under `gate/` — that namespace belongs to the dispatcher; use `fix/<issue>-<slug>`.

## Why this exists

The gate's **data** domain is a single hardcoded profile, and nothing in the harness can vary it:

- `test/util/dashboard/kitchen_sink_overrides.dart:19` — `kitchenSinkOverrides()` takes no parameters and hardcodes every fixture, including `wifiData: testWifiData`.
- `test/util/dashboard/dashboard_card_probe.dart:308` — `buildDashboardCardApp()` has no `overrides` parameter and splices `kitchenSinkOverrides()` in directly, so `probeCardOverflow()` cannot pass anything through either.
- `test/golden_test/page/dashboard/cards/fixtures/cards_test_data.dart:397` — `testRadios` is **2 radios, 2-digit channels, `160MHz` widest**. There is no tri-band profile anywhere in the dashboard fixtures.

So the gate sweeps 26 locales × every width realization × every tab — 1644 cases — against exactly one router shape. Every "clean" verdict it issues is a verdict about that shape only, and the sweep's own thoroughness makes it read as broader than it is.

#1266 hit the consequence twice in one change. Measured by temporarily adding a third tri-band radio (`Channel 233 (Auto) · 320MHz`) to `testRadios` and reverting:

| `wifi_performance` Channels tab | 2 radios (what the gate ships) | 3 radios, tri-band |
|---|---|---|
| before #1266 (hardcoded `'Ch '`) | clean | clean |
| localized, old `Row` + `Spacer` | 3 coordinates (`tr`, `th`) | `en` +8.3px, plus `fi`, `ja` and the two above |
[...]

Issue #1240, opened 2026-08-11
[T16] Declare per-card normalAbove thresholds and build the compact forms


AustinChangLinksys:
## Parent

#1183

Design of record: `doc/dashboard/dashboard_density_design.md` §2.4, §2.5, §2.9, §1.2.

## Branch

**Base: `gate/1183-overflow-gate` (PR #1248). Not `dev-2.7.0`.**

This epic is consolidated onto #1248's branch — #1248 is the only PR that
targets `dev-2.7.0`. Branch from `gate/1183-overflow-gate`, and merge back into
it.

Not a preference. The gate, its fixture, its probe and the design of record are
all introduced *by* #1248 and are **absent on `dev-2.7.0`** (verified with
`git cat-file -e origin/dev-2.7.0:<path>`):

- `test/page/dashboard/cards/dashboard_card_overflow_test.dart`
- `test/fixtures/known_overflows.json`
- `test/util/dashboard/dashboard_card_probe.dart`
- `doc/dashboard/dashboard_density_design.md`

So an acceptance criterion of the form "N coordinates are removed from
`known_overflows.json` and the gate passes" cannot be executed off `dev-2.7.0`
at all. Basing there also re-measures coordinates the landed tickets already
cleared, producing a fixture that conflicts with them and reads as a regression.

## What to build

The final phase: each card that needs one declares a `normalAbove` threshold and gains a compact form. **This is an umbrella ticket and will be split into batches** once the tracks above have converged — deciding the batches now would mean guessing which cards still need a compact form after their layout fixes land, and several may not need one at all.

The measured fit widths (worst locale × worst tab) are the starting point for the thresholds:

| Card | Fit width | Card | Fit width |
|---|---:|---|---:|
| stats_panel | 760 | firewall_overview | 360 |
| traffic_analysis | 560 | system_status | 360 ⚠️ |
| device_info | 480 | connected_devices | 320 |
| lan_info | 420 | network_status | 320 |
| wifi_performance | 420 | port_forwarding | 320 |
| ethernet_ports | 420 | time_settings | 288 |
| network_health | 420 | | |

Every one of these should be **re-measured after the layout fixes land** — the point of those fixes is to lower these numbers. A card whose fit width drops from 480 to 320 may need no threshold at all, and "no threshold declared" is the correct outcome for a card that fits.

[...]

Issue #1238, opened 2026-08-11
[T12] Close connected_devices; escalate the ui_kit AppListTile dependency (26 coordinates blocked)


AustinChangLinksys:
## Parent

#1183

Design of record: `doc/dashboard/dashboard_density_design.md` §1.1, §2.6.

## Branch

**Base: `gate/1183-overflow-gate` (PR #1248). Not `dev-2.7.0`.**

This epic is consolidated onto #1248's branch — #1248 is the only PR that
targets `dev-2.7.0`. Branch from `gate/1183-overflow-gate`, and merge back into
it.

Not a preference. The gate, its fixture, its probe and the design of record are
all introduced *by* #1248 and are **absent on `dev-2.7.0`** (verified with
`git cat-file -e origin/dev-2.7.0:<path>`):

- `test/page/dashboard/cards/dashboard_card_overflow_test.dart`
- `test/fixtures/known_overflows.json`
- `test/util/dashboard/dashboard_card_probe.dart`
- `doc/dashboard/dashboard_density_design.md`

So an acceptance criterion of the form "N coordinates are removed from
`known_overflows.json` and the gate passes" cannot be executed off `dev-2.7.0`
at all. Basing there also re-measures coordinates the landed tickets already
cleared, producing a fixture that conflicts with them and reads as a regression.

## What to build

> **Re-scoped 2026-08-13** against a gate run on `gate/1183-overflow-gate`
> @ `20402ab6`. The coordinate count (27) is unchanged, but two things in this
> body went stale: the dependency-blocked total is now **32, not 45** (fl_chart's
> share in #1230 fell 19 -> 6), and the `Blocked by` is cleared. Reproduce with
> `./tool/run_overflow_test.sh -c connected_devices -d 3`.

Connected Devices has 27 coordinates, and 26 of them need **three** sites fixed together: the card's own device-count row, the shared loading skeleton's header row (#1227), and a row inside `ui_kit_library`'s list tile. Fix the first two and 26 coordinates still stay red, because a coordinate only goes green when every one of its sites is resolved.

So this ticket does two things:

1. Fix the card-own site, closing the 1 coordinate that does not depend on ui_kit.
2. Establish exactly what ui_kit needs to change, and raise it — per Article XIV, `ui_kit_library` cannot be modified from this repo. Reproduce the overflow, identify the unconstrained row in the list tile, and file it upstream with the reproduction, then leave the 26 allowlisted with a tracking note naming the upstream request.

[...]

Issue #1230, opened 2026-08-11
[T11] Fix firewall_overview own sites; document the fl_chart fallback (21 coordinates: 15 card-own + 6 fl_chart)


AustinChangLinksys:
## Parent

#1183

Design of record: `doc/dashboard/dashboard_density_design.md` §2.11, §1.8.

## Branch

**Base: `gate/1183-overflow-gate` (PR #1248). Not `dev-2.7.0`.**

This epic is consolidated onto #1248's branch — #1248 is the only PR that
targets `dev-2.7.0`. Branch from `gate/1183-overflow-gate`, and merge back into
it.

Not a preference. The gate, its fixture, its probe and the design of record are
all introduced *by* #1248 and are **absent on `dev-2.7.0`** (verified with
`git cat-file -e origin/dev-2.7.0:<path>`):

- `test/page/dashboard/cards/dashboard_card_overflow_test.dart`
- `test/fixtures/known_overflows.json`
- `test/util/dashboard/dashboard_card_probe.dart`
- `doc/dashboard/dashboard_density_design.md`

So an acceptance criterion of the form "N coordinates are removed from
`known_overflows.json` and the gate passes" cannot be executed off `dev-2.7.0`
at all. Basing there also re-measures coordinates the landed tickets already
cleared, producing a fixture that conflicts with them and reads as a regression.

## What to build

> **Re-scoped 2026-08-13** against a gate run on `gate/1183-overflow-gate`
> @ `20402ab6`. This ticket was written at baseline and its numbers no longer
> match the fixture: it said **48 of 67 coordinates over three own sites**, and
> the live count is **21 over two**. Nothing regressed — the third own site was
> cleared as a side effect by #1247, and the allowlist was never narrowed to
> match. Reproduce with `./tool/run_overflow_test.sh -c firewall_overview -d 3`.

Firewall Overview carries **21 coordinates**, from three sites in two groups.

**Group 1 — card-own, 15 coordinates, all at `firewall_overview|min|0`**
(191px card at a 601px screen, 3x3, tab 0, in `ar el es es_AR fi fr fr_CA id nb
pl pt pt_PT ru tr vi`):

| site | coords | overflow |
|---|---:|---|
| `usp_firewall_overview_card.dart:157` — the donut's `centerWidget` `Column` | 8 | bottom, 15-31px |
| `usp_firewall_overview_card.dart:136` — the card's outer `Column` | 7 | bottom, 7-26px |

[...]

PR #1269, opened 2026-08-13
fix(dashboard): close Track A's four card-own overflow tickets, by shape (#1234, #1235, #1236, #1237)

### LINKED ISSUE #1237: [T09] Close time_settings (21 coordinates)

AustinChangLinksys:
Closes #1234, closes #1235, closes #1236, closes #1237. Track A of #1183.
Design of record: `doc/dashboard/dashboard_density_design.md` §1.1 (the shape
attribution these commits execute) and §2.10d (added here).

Base is `gate/1183-overflow-gate` (#1248), not `usp`, for the same reason #1262
gave: the gate, its fixture and its probe are all introduced by #1248 and are
absent on the release branches, so "N coordinates removed from
`known_overflows.json`" is not executable there. As with #1262, the closing
keywords above will not fire on merge — GitHub only honours them when the base is
the default branch — so the four tickets close when the gate branch lands.

**87 of the ratchet's 181 coordinates → 0.** The allowlist is left holding 94: 67
`firewall_overview` (#1230) and 27 `connected_devices` (#1238). 49 of those are
still card-own and clearable — #1230's three own sites carry 48, #1238's
device-count row carries 1 — and only 45 are dependency-blocked (fl_chart's 19
axis-title coordinates, and the 26 needing a ui_kit `AppListTile` change). What
this branch completes is the *four-shape* group, not Track A.

Four tickets on one branch, because they reduce to **four shapes across six sites**
— §1.1 of `doc/dashboard/dashboard_density_design.md` measured that attribution
before any code was written. Commits are organized by shape rather than by ticket,
so each one is a single reviewable idiom and its blast radius is exactly the sites
that share it:

| Commit | Shape | Sites | Coordinates |
|---|---|---|---:|
| `9fda58d1` | "View details" footer | `system_status:118`, `device_info:140` | 34→28, 2→0 |
| `abb82185` | Hero inner row | `lan_info:56`, `time_settings:108` | 27→0, 21→0 |
| `02413c24` | Legend row | `system_status:218` | 28→26 |
| `14162d36` | Twin gauge row | `system_status:196` | 26→**0** |
| `2673a09d` | Gauge centre | `network_health:150` | 3→0 |
| `60ca0256` | Ratchet fixture | — | 87→0 |
| `b0496457` | Review pass | — | — |

The legend row clearing only 2 and the gauge row then clearing 26 is not a
mis-split: both sites live under the same `min|0` key, so neither can clear it
alone. §1.1 predicted that; the per-commit measurement confirmed it.

## How this is verified

By the #1183 ratchet, not by new tests. The failing assertions were already
committed as entries in `test/fixtures/known_overflows.json`: fix the layout,
delete the entry, the gate goes green — and deleting an entry that still overflows
fails that test, so it cannot be faked.

- Overflow gate: **1644 cases green** (13 cards × widths × tabs × 26 locales)
- Full functional suite: **5392 green**
- Affected golden suites: **23 green**
- `flutter analyze`: clean on all 8 changed files

## Two tests, and why they are not ratchet duplication

A green gate is not a green acceptance criterion. Two situations here leave the
gate green through both the right and the wrong fix, and each got one assertion —
measured with these cards' allowlist entries already deleted:

**`usp_gauge_center_readability_test.dart`** — `FittedBox(fit: BoxFit.scaleDown)`
is the right fix for #1235's centre and it removes the overflow signal at that site
*permanently*: once a subtree may shrink, no future squeeze can produce a
coordinate. Measured — shrinking the score's font unconditionally clears all 209 of
that card's gate cases. So the fix ships a 12px painted floor as the replacement
for what `scaleDown` took away.

**`usp_hero_row_readability_test.dart`** — `maxLines: 1` and a soft-wrap clear the
*same* coordinates, and at three sites one of them destroys the string. Measured —
ellipsizing `lan_info`'s router IP, its DHCP status, or the shared InfoGrid value
renderer leaves all 1644 cases green while cutting a reading #1236 AC 4 and #1237
AC 5 name explicitly. Each mutation kills exactly the group that names it, four
locales each, so a failure identifies which criterion broke.

Everything else here is a plain `Flexible` or a deletion and gets no test — the
allowlist entry it removes is already its assertion.

## Two decisions worth a reviewer's attention

**The two gauges shrink rather than stack** (`14162d36`). Stacking them in a `Wrap`
was the alternative, and it fails *silently*: `RenderWrap` has no overflow
indicator of its own and the `Expanded` pins its height, so the second circle is
simply cut — measured at 108px between gauge centres in a 181px box with all 209
gate cases green. They are bound by `min(natural, (maxWidth - AppSpacing.md) / 2,
maxHeight)` instead, an upper bound rather than an allotment, so every width that
already fitted two 100px gauges is untouched. That reserve is deliberately tight:
`spaceEvenly` splits it three ways, so what is drawn between the circles is 4px.
Reserving a full 12px *between* them would cost 12px of diameter, and #1234's AC 4
is about the reading inside the circle staying legible — air between two rings is
the cheaper thing to give up.

**`time_settings` keeps `minColumns: 3`.** It is the only card in the baseline where
widening was arithmetically available (fit width 288px, so 3 → 5 would have covered
every locale). Declined: the floor is the user's — it decides how narrow they may
make the card on their own dashboard — and 5 of 12 columns is a permanent charge on
every layout to buy headroom in a handful of locales. The card-own fix cleared all
21 coordinates by *deleting* a single-child `Row` whose only effect was handing the
badge unbounded width. Reasoning is recorded at the declaration site, per #1237.

## Deliberately not fixed here

- **The Monitor legend's interval chip loses its flush-right position.** Verified
  visually in the golden diff at 500×400, not just asserted. `WrapAlignment.spaceBetween`
  was rejected because it distributes between all three children, pulling the two
  legend entries apart — a bigger change to the wide layout than moving a 20px chip.
- **Both closed cards are green at 191px and unusable there.** `time_settings`' hero
  clock paints 5 lines / 140px in every locale including `en`; `network_health`'s
  centre scales to 0.52 in `de`. Nothing overflows; everything that could give, gave.
  Direct **#1240** input, recorded in §2.10d point 5 — with the caveat that Track A
  has moved five cards' *clean* width without moving their *readable* width, so #1240
  must re-measure rather than read §1.2's column.
- **`usp_lan_info_card.dart:144` renders a hardcoded `'Enabled'`** while line 85 of
  the same file uses `loc(context).enabled`. A real bug, but gate-invisible twice
  over: a hardcoded string cannot vary in a locale sweep, and that branch is the
  `else` of `if (info.ipv6Addresses.isNotEmpty)`, which the fixture never takes.
  Verifying it needs #1267's `overrides` parameter first.
- **`detailRoute` cannot carry query parameters**, which is why two cards hand-roll
  the "View details" footer `DashboardCardTemplate` already owns. The fix was
  replicated verbatim rather than extracted — extracting would make a third copy
  while leaving the cause in place. Both are **#1245** inventory entries.

### DIFF
--- a/doc/dashboard/dashboard_density_design.md
+++ b/doc/dashboard/dashboard_density_design.md
@@ -878,8 +878,10 @@ explicit decision in that ticket rather than an implementation detail.
 ### 2.10d What closing the four card-own tickets taught us (#1234–#1237 — implemented)
 
 87 coordinates → **0**, on one branch, one commit per shape. §1.1's decomposition
-held to the coordinate, including its uncomfortable prediction that two of the
-four commits would move nothing on their own:
+held to the coordinate, including its uncomfortable prediction that the two
+`system_status` sites sharing `min|0`'s 26 coordinates would each clear **none**
+of them alone — the legend row moved only the 2 outside that key, and the 26 fell
+in one step when the gauge row followed:
 
 | Commit (shape) | Sites | Coordinates |
 |---|---|---:|
@@ -917,7 +919,8 @@ cannot tell you when it doesn't**, so the precondition is measured *before* the
 conversion or not at all. #1233's and #1266's conversions were safe for a reason
 that has to be checked, not inherited.
 
-**3. Every fix is one of three things to the gate, and two of them owe a test.**
+**3. Every fix is one of three things to the gate — and a green gate is not a
+green criterion.**
 
 | Fix class | Example | The gate afterwards |
 |---|---|---|
@@ -931,6 +934,28 @@ ever produce a coordinate. The mutation ledger makes that concrete — shrinking
 score's font unconditionally clears **all 209** gate cases. So a self-relaxing fix
 has to ship its own floor: `usp_gauge_center_readability_test.dart` asserts a 12px
 painted minimum, which is the replacement for what `scaleDown` took away.
+
+That table is the *revert* axis, and it is not the only one. A constraining fix
+keeps the signal, so the gate does own its revert — but it never owned the
+**choice**: `maxLines: 1` and a soft-wrap clear the same coordinates, and at three
+of these sites one of them destroys the string. Measured on this branch:
+ellipsizing `lan_info`'s router IP, its DHCP status, or the shared InfoGrid value
+renderer leaves **all 1644** gate cases green. #1236 AC 4 and #1237 AC 5 name those
+readings ("truncating an IP in the middle makes it useless"; "timezone names stay
+identifiable"), so `usp_hero_row_readability_test.dart` asserts them across four
+locales each — and each mutation kills exactly its own group, so a failure says
+which criterion broke. The rule that falls out is narrower than "constraining
+fixes need tests", which would be licence to test every `Flexible` on the branch:
+**a fix owes a test when the gate stays green through both the right and the wrong
+version of it** — whether because the fix removed the signal (#1234, #1235) or
+because the signal never told the two apart (#1236, #1237). Where the gate does
+discriminate, the allowlist entry it deletes *is* the assertion.
+
+The same test covers the one place this branch made a reading *worse*: #1237's
+badge now ellipsizes (`de` paints 45.4px of an 85.1px label), which §2.10a point 2
+requires of a capsule but which no coordinate can report, since a narrower badge
+overflows *less*. That gets a floor too, not a fidelity claim — enough glyphs to
+key the state, with the colour carrying the rest.
 
 **4. A ticket's diagnosis is a report, not a measurement.** #1235 states the tier
 label is wider than the space inside the circle. The three coordinates are
@@ -990,7 +1015,8 @@ is the only card in the baseline where §1.3's arithmetic permits the other fix
 (fit width 288px, so `minColumns` 3 → 5 would have worked). It stays 3, with the
 reasoning recorded at the declaration site: the floor is the user's, it costs 5 of
 12 columns in every layout to buy headroom in a handful of locales, and the
-card-own fix cleared all 21 coordinates for one `Flexible` and a soft-wrap.
+card-own fix cleared all 21 coordinates by deleting a single-child `Row` that
+had no other effect — the cheapest fix on the branch.
 
 ### 2.11 fl_chart's 19 coordinates get a primary plan and a documented fallback
 
@@ -1122,9 +1148,11 @@ addition — it changes no card's rendering until a threshold is declared.
 Ceiling **515 / 560**. The other 45 are the dependency-blocked ones (§1.1).
 
 After #1234–#1237 the allowlist holds **94** coordinates: 67 `firewall_overview`
-(#1230) and 27 `connected_devices` (#1238). Both remaining entries are the
-dependency-blocked kind — fl_chart's internals and a ui_kit `AppListTile` change —
-so Track A's card-own work is complete.
+(#1230) and 27 `connected_devices` (#1238). Both are the dependency-blocked kind —
+fl_chart's internals and a ui_kit `AppListTile` change — so Track A's card-own
+work is complete. Their `tracking` notes are left as the epic requires: a ticket
+touches only the notes of cards it closes, so both cards keep the `baseline #1183`
+default until their own ticket names an owner.
 
 #1266 is in this track despite clearing nothing: it is the only entry that *adds*
 coordinates (3, by localizing a hardcoded string) and removes them again in the
@@ -1163,6 +1191,7 @@ the likeliest way to get this wrong.
 | Work | Method | Why |
 |---|---|---|
 | All of Track A except #1225 | **Ratchet, not TDD** | The failing assertions are *already committed* — the 560 entries in `known_overflows.json`. The red→green move is: fix the layout, delete the allowlist entry, gate passes. Deleting an entry that still overflows fails that test, so it cannot be faked. |
+| …plus two readability tests, added by #1234–#1237 | **Ratchet + a floor test** | The ratchet verifies that the overflow is gone. It cannot verify what the fix *chose*, in two situations, and both occurred on that branch (§2.10d point 3). (a) The fix makes a subtree self-relaxing, so the signal is gone for good: measured, shrinking the health score's font unconditionally clears all 209 of its gate cases. (b) Two fixes clear the same coordinates and only one preserves the reading: measured, `maxLines: 1` on the router IP, the DHCP status, or the InfoGrid value renderer leaves all 1644 cases green while cutting a string #1236 AC 4 / #1237 AC 5 require whole. The test is owed when **the gate stays green through both the right and the wrong fix** — not whenever a fix constrains. A plain `Flexible` is fully verified by the entry it deletes. |
 | #1232 | **TDD** | The gate asserts only "no overflow"; it cannot detect *wrong density*. Threshold selection, popup cut-off, and absent-`normalAbove` behaviour can all break while the gate stays green. Tests go red first. |
 | #1231 | **Neither** | Not an assertion — it fixes a failure the gate is structurally blind to (§2.8), so it is verified by eye or golden. |
 | #1225 | **Property tests + no-op diff** | Planned as "neither", since converting a sampled invariant into a guaranteed one changes no assertion. In the event it had a testable seam after all: the search is pinned by a property (no supported width is narrower than the one pumped, which fails on a coarser step) and the no-op claim by diffing the full sweep's 560 allowlisted hits before and after — they matched exactly. |
--- a/doc/dashboard/dashboard_density_design.md
+++ b/doc/dashboard/dashboard_density_design.md
@@ -1001,14 +1001,27 @@ locale sweep, *and* the branch never renders at all — it is the `else` of
 four tickets. This is §2.10c finding 2 in its purest form: the data decides what
 the instrument can see.
 
-**7. Two entries for #1245's de-duplication inventory.** The two hand-rolled
-"View details" footers exist for one reason: `detailRoute` cannot carry query
-parameters (`?tab=`, `?deviceId=`), so neither card can use
+**7. Two entries for #1245's de-duplication inventory.** There are **three**
+hand-rolled "View details" footers, not two — `usp_system_status_card.dart:118`,
+`usp_device_info_card.dart:134` and `usp_traffic_analysis_card.dart:141` — plus
+the template's own at `dashboard_card_template.dart:393`, so four copies of one
+shape. They exist for one reason: `detailRoute` cannot carry query parameters
+(`?tab=`, `?deviceId=`), so none of the three can use
 `DashboardCardTemplate._buildDetailFooter`. #1227 fixed the template's copy and
-could not fix theirs. The fix here was replicated *verbatim* rather than extracted,
-because extracting it would make a third copy of the widget while leaving the
-cause in place. The cause — `detailRoute`'s signature — is the inventory entry.
-The second entry is the footer shape itself, now identical in three places.
+could not fix theirs. The fix here was replicated *verbatim* rather than
+extracted, because extracting it would make a fifth copy of the widget while
+leaving the cause in place. The cause — `detailRoute`'s signature — is the
+inventory entry. The second entry is the footer shape itself.
+
+These four tickets hardened two of the three. `traffic_analysis`'s is left
+unflexed on purpose: its `minColumns` is **4**, not 3, so the width enumeration
+never realizes it at the 157.4px where this shape bites, and it carries no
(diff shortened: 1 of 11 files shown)


AustinChangLinksys (commented):
🤖 **Automated Review — Round 1** · `32ac1849..535272a` (full)

**Verdict: 💬 Self-review (comment only)** — Self-authored PR; posted as reference only. GitHub prohibits self-approve on the same account. 0 Critical. Three Warnings below before merge.

#### Standards

|---|---|---|---|
| ⚠️ | 🟢High | `test/page/dashboard/views/components/usp_gauge_center_readability_test.dart:433` | `_localeFor()` uses `firstWhere` without `orElse` — throws opaque `StateError` instead of friendly `ArgumentError` if an unknown locale tag is passed |
| ⚠️ | 🟢High | `test/page/dashboard/views/components/usp_gauge_center_readability_test.dart:138` + `test/page/dashboard/views/components/usp_hero_row_readability_test.dart:85` | `UspWidgetSpecs.all.firstWhere((s) => s.id == cardId)` without `orElse` in both new test files — same opaque `StateError` hazard on id rename |
| ⚠️ | 🟢High | `lib/page/dashboard/views/components/usp_traffic_analysis_card.dart:141–181` | [Duplicated Code] `_buildStatisticsFooter` third copy lacks the `Flexible` + `maxLines: 1` + `ellipsis` guard that this PR added to the other two copies (`system_status:118`, `device_info:133`) — three copies now diverge |
| ⚠️ | 🟢High | `lib/page/local_network/cards/usp_lan_info_card.dart:144` | Hardcoded `value: 'Enabled'` vs `loc(context).enabled` at line 85 in same file — design doc §2.10d notes this as a real bug; PR modifies this file without addressing it |
| 💡 | 🟡Med | `lib/page/dashboard/views/components/usp_system_status_card.dart:381` | `AppText.titleMedium` for gauge reading vs `AppText.titleLarge` in `network_health` — intentional size difference (72.7px vs 120px circle) but comment at call site doesn't explain choice |
| 💡 | 🟡Med | `test/…/usp_gauge_center_readability_test.dart` + `test/…/usp_hero_row_readability_test.dart` | [Duplicated Code] `paragraphOf`/`isClipped`/`lineCount` helpers redefined in both new test files with different signatures — consider extracting to `test/util/dashboard/render_text_probe.dart` |
| 💡 | 🟡Med | `lib/page/admin/cards/usp_time_settings_card.dart:144` | `label: 'DST'` — design doc §2.10d finding 6 marks as "arguable and left alone"; no issue link to track the decision |
| 💡 | ⚪Low | `lib/page/dashboard/views/components/usp_system_status_card.dart:246–260` | `gaugeSize` is theoretically negative if `constraints.maxWidth < 12px` — Flutter clamps to 0 silently; pre-existing, not reachable at `minColumns: 3`, but `clamp(0.0, …)` would make intent explicit |

#### Spec

|---|---|---|---|
| ✅ | 🟢High | `known_overflows.json` | #1234 AC2: 34 coordinates removed (`system_status` — all four tabs, all 26 locales) |
| ✅ | 🟢High | `known_overflows.json` | #1235 AC2: 3 coordinates removed (`network_health` — de/ru/th) |
| ✅ | 🟢High | `known_overflows.json` | #1236 AC2: 29 coordinates removed (`lan_info` 27 + `device_info` 2) |
| ✅ | 🟢High | `known_overflows.json` | #1237 AC2: 21 coordinates removed (`time_settings`) |
| ✅ | 🟢High | `usp_widget_specs.dart:190` | #1237 AC4: `minColumns: 3` unchanged; inline comment records why widening was declined |
| ✅ | 🟢High | `usp_network_health_card.dart:194` | #1235 AC4: `FittedBox(scaleDown)` — tier label scales rather than truncates |
| ✅ | 🟢High | `usp_hero_row_readability_test.dart` | #1236 AC4 + #1237 AC5: router IP / DHCP status / timezone badge mutation ledger locks non-truncation |
| ✅ | 🟢High | `usp_gauge_center_readability_test.dart` | #1234 AC4: gauge reading legibility across en/de/ru/th at narrowest |
| 💡 | 🟡Med | `usp_gauge_center_readability_test.dart:153` | #1235 AC1 says "all 3 tabs" but only `tabIndex: 0` is tested — correct (gauge only exists on Health tab), but test file lacks a comment explaining why tabs 1/2 are omitted |

<details>

**W-1 — `_localeFor()` uses `firstWhere` without `orElse` (🟢High) [Reviewer A]**
`test/page/dashboard/views/components/usp_gauge_center_readability_test.dart:433–439`

```dart
Locale _localeFor(String tag) =>
    AppLocalizations.supportedLocales.firstWhere((l) {
      final t = l.countryCode == null || l.countryCode!.isEmpty
          ? l.languageCode
          : '${l.languageCode}_${l.countryCode}';
      return t == tag;
    });
// ← no orElse
```

All tags used in this file (`'en'`, `'de'`, `'ru'`, `'th'`, `'fr'`) are present today. But the function signature accepts any `String`; a future test with a typo (e.g. `'de_DE'`) or a not-yet-added locale produces a `StateError: No element` with no indication of which tag failed — difficult to debug. Contrast with `usp_hero_row_readability_test.dart:121` which uses a `switch` (always succeeds by construction).

**Fix:**
```dart
Locale _localeFor(String tag) =>
    AppLocalizations.supportedLocales.firstWhere(
      (l) { ... },
      orElse: () => throw ArgumentError(
          'locale tag "$tag" not in AppLocalizations.supportedLocales'),
    );
```
Or adopt the `switch`-based construction from the sibling test file and share via `test/util/`.

---

**W-2 — `UspWidgetSpecs.all.firstWhere` without `orElse` in both new test files (🟢High) [Reviewer A]**
`test/page/dashboard/views/components/usp_gauge_center_readability_test.dart:138`
`test/page/dashboard/views/components/usp_hero_row_readability_test.dart:85`

```dart
// Both files:
final constraints = UspWidgetSpecs.all
    .firstWhere((s) => s.id == cardId)   // ← no orElse
    .getConstraints(DisplayMode.normal);
```

Same hazard: if a card id is renamed/removed, throws an opaque `StateError`. Currently safe because `'system_status'` and `'network_health'` are stable.

**Fix:**
```dart
UspWidgetSpecs.all.firstWhere(
  (s) => s.id == cardId,
  orElse: () => throw ArgumentError('no WidgetSpec with id="$cardId"'),
);
```

---

**W-3 — `usp_traffic_analysis_card.dart:141–181` footer third copy missing `Flexible` (🟢High) [Reviewer B]**
`lib/page/dashboard/views/components/usp_traffic_analysis_card.dart:141–181`

This PR adds `Flexible` + `maxLines: 1` + `overflow: TextOverflow.ellipsis` to `system_status._buildStatisticsFooter` (line 118) and `device_info._buildNodeDetailFooter` (line 133). But `usp_traffic_analysis_card.dart:141–181` is the same "View details" footer shape (same `queryParameters`, same `Row(end) → InkWell → Row(min) → label + arrow` structure) and still uses the old unguarded form — `AppText.labelMedium(label)` with no `maxLines` or `overflow`, wrapped directly in `Semantics` with no outer `Flexible`.

```dart
// usp_traffic_analysis_card.dart:141 (current head — OLD shape)
Row(
  mainAxisAlignment: MainAxisAlignment.end,
  children: [
    Semantics(           // ← no Flexible wrapper
      button: true,
      label: label,
      child: InkWell(
        …
        child: Row(
          mainAxisSize: MainAxisSize.min,
          children: [
            AppText.labelMedium(label,  // ← no maxLines, no overflow
                color: Theme.of(context).colorScheme.primary),
            …
          ],
        ),
      ),
    ),
  ],
),
```

Result: all three "View details" footer copies now have divergent protection. The `traffic_analysis` copy remains overflow-unsafe at the same narrow widths this PR fixed in `system_status` and `device_info`.

**Fix:** Wrap with `Flexible` and add `maxLines: 1, overflow: TextOverflow.ellipsis` to the label, matching the other two. Or open a tracking issue with a `// TODO` comment.

---

**W-4 — `usp_lan_info_card.dart:144` hardcoded `'Enabled'` (🟢High) [Both Reviewer A and Reviewer B]**
`lib/page/local_network/cards/usp_lan_info_card.dart:144`

```dart
else if (info.ipv6Enabled)
  InfoGridItem(label: 'IPv6', value: 'Enabled'),  // ← hardcoded English
```

Line 85 in the same file uses `loc(context).enabled` correctly. Design doc §2.10d finding 2 explicitly notes this as a real bug found during this PR's §2.10c grep pass, and defers it to after #1267's `overrides` parameter. The PR modifies this file (adds `Flexible` to the hero row); the hardcoded string is one line change and does not require #1267.

**Fix:** Change `value: 'Enabled'` → `value: loc(context).enabled`. If intentionally deferred, add `// TODO(#1267): localize — 'Enabled' → loc(context).enabled`.

</details>

<details>

- **All four spec ACs satisfied**: 87 coordinates removed exactly (34 + 3 + 29 + 21); all four cards' allowlist entries gone; `minColumns: 3` unchanged with reasoning; tier label scales not truncates; hero row readings locked by mutation ledger tests.
- **Production layout correctness verified**: `math.min` bounds the gauge diameter as an upper bound (not allotment); `FittedBox(scaleDown)` is semantically correct for score/tier; `Flexible` on footer links is safe because `Row(end)` strands slack at the invisible start; `AppBadge` single-child `Row` deletion is the clean fix (not just a `Flexible` patch).
- **Test architecture is excellent**: Mutation ledger format (each row names one mutation and one group it kills) provides both documentation and discrimination power — a failure identifies which criterion broke. Two new readability test files guard the only choices the overflow gate cannot distinguish.
- **gate passes**: 1644 cases green after ratchet tightened by 87 coordinates; full suite 5392; golden 23 green; `flutter analyze` clean.
- **No security concerns**: No credentials, network calls, storage writes, or permission changes. Layout-only delta.
- **Three-layer architecture respected**: All changes are View-layer widgets; no Provider/data-layer imports or cross-layer jumps.

</details>

Issue #1275, opened 2026-08-14
Add a compact/horizontal variant to InfoGrid: two dashboard cards hand-roll _InfoGridTile because the shared block has no narrow form


AustinChangLinksys:
## Context

Raised by the `/code-review` Standards axis on #1230 (Track A of #1183), as a documented-standard breach that #1230 deliberately did not widen its scope to fix.

`constitution.md` Article XIV §14.5 Rule 2: *"Prefer shared components: before creating a new block pattern, check if `layout_blocks` already provides it."* Two dashboard cards now hand-roll the same block pattern, because the shared one has no compact form.

## What is duplicated

`_InfoGridTile` (`lib/page/_shared/components/layout_blocks/list_blocks.dart:201`) is `LayoutBlock` + `AppText.labelSmall(label.toUpperCase(), color: onSurfaceVariant)` + `AppGap.xs` + `AppText.labelMedium(value)`, at the block's default 12px padding, label above value.

- `_StackedMetrics` (`lib/page/firewall/cards/usp_firewall_overview_card.dart:285`) is that exact composition, differing only in **Row instead of Column** (label left, value right) and **8px instead of 12px padding**.
- `_SummaryTile` (`lib/page/local_network/cards/usp_ethernet_ports_card.dart:148`, from #1228) is the same shape again with a leading disc, and already carries a `compact` flag doing the same 12px → 8px trade.

Both cards arrived at it the same way: when the card is too narrow for tiles side by side, stack them full-width and buy vertical budget back by tightening the padding. The threshold constant (`_kSideBySideMinWidth` 352, `_kMetricsSideBySideMinWidth` 328) is per-card and stays per-card; the *tile* is not.

## Proposal

Add the compact/horizontal variant to the shared component rather than to a third card:

- `InfoGridItem` gains whatever the two call sites need that it does not have (a `leading`, and label/value orientation).
- `InfoGrid` (or `_InfoGridTile`) gains a `compact` flag with the same meaning it has in `_SummaryTile` — `BlockConstants.paddingSm` instead of `paddingMd`.
- The two cards drop their private tiles and pass `crossAxisCount: 1` + `compact: true` below their own thresholds.

`layout_blocks` is project-owned, so this is not an Article XV stop-and-ask: no ui_kit component is being reimplemented, only a project block being given the form its consumers already need.

## What must not change

The tightened padding and the two-line label are load-bearing, not cosmetic — both were derived from measurement and both are guarded:

[...]

PR #1277, opened 2026-08-14
fix(firewall): stack the rule metrics and guard the charts, clearing all 21 of firewall_overview's coordinates (#1230)

### LINKED ISSUE #1230: [T11] Fix firewall_overview own sites; document the fl_chart fallback (21 coordinates: 15 card-own + 6 fl_chart)

AustinChangLinksys:
Closes #1230. Part of #1183 (T11). Also bumps `ui_kit_library`/`generative_ui` **v2.34.5 → v2.34.11**, which fixes the `AppPieChart` bug this PR found upstream (privacyGUI-UI-kit#22) and lets the card-side workaround be deleted.

`firewall_overview` held 21 of the epic's coordinates: 15 card-own (`min|0`, 15 of 26 locales) and 6 fl_chart (`min|1`). All 21 are now fixed at the source and **both `firewall_overview` keys are deleted from `known_overflows.json`** — nothing is re-allowlisted. `minHeightRows` stays at 3 (AC 4), fl_chart is untouched (AC 6).

## What actually overflowed

At the narrowest realization the grid ever yields — a 191.375px card, **157.4px** of tab content in a **205px** viewport — `InfoGrid(crossAxisCount: 3)` gave each rule metric a 47.1px tile with **23.1px** of text. That is one character per line: `ru` took 8 lines to render one label, `es` 5. The grid stood 108px tall in `en` and up to **173px** in `ru`, against a viewport that also owes 20px to two gaps and 36px to the legend.

Which site reported depended on how tall the grid grew, and the two are mutually exclusive because `RenderFlex` skips an empty box:

- 8 locales (`ar el fr fr_CA pt pt_PT tr vi`) reported at the donut's `centerWidget`: the caption is 40px, the slot was 20-25px, overflow exactly `40 − slot` (+15px to +21px).
- 7 locales (`es es_AR fi id nb pl ru`) starved that slot to 0px, so the outer `Column` reported instead (+7px to +26px).

15 locales, 15 coordinates. On the Ports tab, `AppBarChart`'s vertical path passes fl_chart's default `reservedSize` for the bottom axis strip — 22px asked for whatever the chart's height is — so its own `Flex` overflowed by `22 − slot`: +5px at 17px slots (`da pl pt pt_PT sv`), +10px at `ru`'s 12px. 6 coordinates.

## The fix

Three call-site constants, all in `lib/page/firewall/cards/usp_firewall_overview_card.dart`:

| constant | value | what it does |
|---|---|---|
| `_kMetricsSideBySideMinWidth` | 328 | below it the three metrics stack into full-width rows instead of sitting three across |
| `_kDonutMinRingThickness` | 10 | the donut sizes its ring from the slot it is given, and draws nothing once that slot cannot hold this much ring |
| `_kProtocolChartMinHeight` | 70 | the protocol chart is not built at all in a slot this short |

**Stacking** (the 15): the same 157px card gives each label 111.4-127.6px on one line plus a second to wrap into, which holds every locale unbroken except `fi` (PALOMUURISÄÄNNÖT, 131.0px) and `nb` (BRANNMURREGLER, 118.6px) — single-word labels that wrap mid-word rather than clip mid-glyph, and no value of this threshold buys them an unbroken line (a 1440px desktop leaves the *grid* 130.0px for `fi`'s 131.0px word). That is #1240's compact-forms problem. 328 is bound by `es` at 73.8px; the threshold decides arrangement, not survival — both forms are overflow-safe, so getting it wrong costs a wrap, never a clip.

The stacked rows use `BlockConstants.paddingSm` (8px) rather than the block default 12px: measured, 12px re-overflows 6 locales (+3px `fi` to +7px `ru`) and starves the donut out even at the shipped height.

**Charts** (the 6): #1230's plan preferred suppressing the axis *labels*, and that lever does exist — `AppBarChart.xLabels` is nullable and the vertical path passes `showTitles: xLabels != null`, so `xLabels: null` drops the 22px strip and, measured, takes all 105 of this card's gate cases green with no height guard at all. It is not taken because of what it leaves on screen: the value axis keeps drawing, so at `ru`'s 12px slot its "2" and "0" labels overlap each other above two 8px pills. The only threshold that would admit the labelless chart where it reads (`en`, 37px) and reject it where it does not (`da`, 17px) sits between 37 and 40px — a knife edge between two shipped locales, where 70px leaves every locale 33px clear. The picture is dropped; the list above it keeps the information, and the donut's legend keeps naming every target and its count.

Both guards fire only at the 3-row minimum. At the 4 rows `HeightStrategy.strict(4)` actually gives this card, both pictures are drawn in all 26 locales at all three widths.

## Why there is a second test file

The #1183 gate cannot audit either half of this fix, and both are *removals*:

- Three tiles of one character per line **fit perfectly**. Shredded text is not overflow.
- Absent content never overflows, so the gate rewards deleting a chart exactly as much as fixing it. A threshold accidentally set above the card's shipped height would leave the gate green and the card empty, and nothing in `known_overflows.json` can express "the picture is still there at 4 rows".

So `test/page/dashboard/views/components/firewall_overview_readability_test.dart` (**74 tests**, tagged `dashboard-card` so it gates PRs) asserts on the rendered tree: the card was laid out in the app's own fonts and not the engine's block font, no metric label is broken mid-word in any of the 26 locales, the arrangement is stacked below the threshold and three-across above it, both pictures are present at the shipped height and absent at the minimum, the legend carries every target while the donut is suppressed, the content clears the footer, the ring is never drawn larger than its box, and no slice label is painted inside it.

AC 5 is a claim about the whole card, so the **Ports tab** is now swept for the same two failures even though none of the 21 coordinates were there — its rows fit at 191px, and fitting is not reading. That sweep found one thing that does not read, and the cause is outside this card (below).

Every assertion was run against a mutation of the code it guards, and every one fired:

| mutation | what failed |
|---|---|
| 1. `_kMetricsSideBySideMinWidth` 328 → 0 (the pre-#1230 card) | 33 |
| 2. `_kMetricsSideBySideMinWidth` 328 → 600 | 1 |
| 3. `_kProtocolChartMinHeight` 70 → 200 | 3 |
| 4. `_kDonutMinRingThickness` 10 → 60 | 6 |
| 5. `size: diameter` → the themed design diameter, ignoring the slot | 1 (@min only — 191px is the one width whose 157.4px slot is under the 160px this theme designs for) |
| 6. `showLabels: false` → `true` | 3 |
| 7. `_StackedMetrics` padding `sm` → block default | 12 |
| 8. stacked label `maxLines` 2 → 1 | 5 |
| 9. Ports heading `maxLines: 1` | 6 (`da pl pt pt_PT ru sv` — the locales whose heading wraps) |
| 10. the mapping on a full-width line of its own | 29 (overflow @min in all 26 locales, +11px to +36px) |
| 11. 10, plus the DMZ list dropped, plus `MapsToRow`'s source given its intrinsic width | 1 — the mapping-target ratchet, and nothing else |
| 12. ui_kit's font family renamed in `loadAppFonts` | 1 — the font identification, at 1.000 em per glyph |
| 13. the app's family registered against Noto's bytes | 1 — the same test's Noto net |
| 14. the `NotoSans` reference family renamed | 1 — the reference self-check |
| 15. the ui_kit font path broken | the whole file, at `setUpAll` |

Seven of the first eleven leave the #1183 gate green while the card loses a chart (3, 4), asks for a 160px donut in a 157.4px slot (5), prints a clipped slice label into a 20px ring (6), or clips text mid-glyph (8, 9). Only 1, 7 and 10 report an overflow, and 1 is simply the pre-#1230 card. 12-15 are a different kind: they do not break the card, they break the *measuring instrument* — see below.

Mutation 6 is worth naming: the previous revision of this file claimed an "exactly 2 texts" assertion guarded `showLabels: false`, and it did not — fl_chart paints slice labels onto its canvas, not as `Text` widgets, so flipping the flag fired nothing. The suite now reads each `PieChartSectionData.title` and requires it empty, which is what actually guards it. Overflow itself stays the gate's job, with one exception carried in this file's `pumpAt`: the gate pumps `minHeightRows` only, so the shipped height — the only one where the donut's caption and the axis strip are built at all — would otherwise be measured by nothing.

## The upstream bug this PR found is now fixed

`AppPieChart` took the section radius from the call site but the centre-hole radius from the theme, so the drawn diameter was `2 × (centre + ring)` and did not follow the `size` box. The old `size: 160` with the default 40px ring drew a **200px** donut into a 160px box and the card surface clipped 20px off every side — measured, the painted extent was 200px at every `size` from 120 to 300.

Filed as **linksys/privacyGUI-UI-kit#22**, fixed in **v2.34.11**, and this PR takes the bump: both radii are now derived from the box. **The card-side workaround is deleted** — `sectionRadius:` is gone from the call, along with the arithmetic that computed it. The four other `AppPieChart` call sites in this app are fixed by the bump alone, with no change of their own.

What remains at the call site is a *themed floor*, not a workaround. `_TargetDonut` reads `ChartStyle.pieCenterRadius` inside its `LayoutBuilder` (flat 60.0, neumorphic 65.0, pixel and brutal 50.0), sizes the box to `min(box, 2 × (centreRadius + 20))`, and returns `SizedBox.shrink()` below `2 × (centreRadius + 10)` so no theme's centre radius can invert the ring-thickness floor. Measured under v2.34.11:

| card width | drawn box | hole | ring |
|---|---|---|---|
| 191px (the 3-row minimum) | 153.0px | 57.4px | 19.1px |
| 288px | 160.0px | 60.0px | 20.0px |
| 512px | 160.0px | 60.0px | 20.0px |

Note *which* radius v2.34.11 shrinks: the hole, not the ring. At 191px the caption's half-diagonal is 32.0px against a 57.4px hole — 25.4px clear — so the drawn-vs-box assertion still has something to fail on, and does (mutation 5 below).

The bump changed three unrelated dashboard goldens for the better (`card_traffic_analysis`'s y-axis labels stopped wrapping, `card_wifi_status`'s access-point names stopped wrapping and its channel value joined the value column, `card_device_analytics`'s donut lost its 20px clip). Baselines are gitignored, so nothing is committed; regenerated locally and green. The whole `--tags golden` suite is `+73 -284` both before and after the bump — the bump adds no golden failure. Those 284 are stale local baselines in the whole-suite harness, unrelated to this branch.

## The measurements are now proven to be real-font measurements

Every assertion in this file is a claim about pixels, and nothing in it checked that the pixels came from NeueHaas rather than the test engine's block font — under which every glyph is one square em, so what fits and what ellipsizes is fiction that stays internally consistent and green. Two holes, both closed and both mutation-verified:

- **`loadAppFonts` threw nothing when a font file was missing** — it skipped silently, on the rationale that "CI without the ui_kit checkout" should still run. That is not a state worth running in, and it is not a real one either: `ui_kit_library` is a pub dependency and the five Noto files are committed under `test/fonts/`. What *is* real is a ui_kit ref bump moving the pub-cache path, which is exactly the failure that must be loud. It now collects the missing paths and throws at `setUpAll`. Icon fonts stay tolerant — advances in an icon font are one em by definition, so a missing one costs a tofu box, not a wrong measurement. This helper is shared with the golden pipeline and five other readability suites; all were re-run.

- **A new first test identifies the typeface the card was actually laid out in**, on the pure-ASCII mapping target (for Cyrillic, Arabic, Thai and CJK the app falls back to Noto *by design*, and one em per ideograph is what CJK legitimately measures, so the check would be backwards there). Three assertions, because renaming the family registration produces three different lies:

  | what breaks | what it measures | which net catches it |
  |---|---|---|
  | family unresolvable, no `fontFamilyFallback` on the style | block font: `192.168.1.101:80` at **192.0px** = 12.0 × 16 chars, and 1.000-1.045 em per glyph across the whole tab | width must be under 0.9 em per glyph (the real font is at **0.43** — 83.3px) |
  | family unresolvable, a fallback list present (`lib/app.dart` supplies one for CJK) | Noto: **95.2px**, a real proportional font, plausible enough to slip past a ratio check | width must differ from what Noto measures the same string at |
  | the Noto reference itself did not load | the reference is one em per glyph, so it identifies nothing | the reference is range-checked before it is used |

  The first net is the one a naive version of this test got wrong: an *exact* one-em signature was written first, and measurement showed the substitute block font sits at 1.000-1.045 em, so half the strings on the tab slipped through. Hence a ratio.

  It pumps through the probe rather than this file's `pumpAt`, dropping the overflow incidents: a wrong typeface reflows the card, so `pumpAt` would fail on the overflow first and hide the identification — the one failure that says *why*. Overflow at that coordinate is asserted by the tests below it.

## Follow-ups left open, not silently dropped

- **#1275** — `_StackedMetrics` is `_InfoGridTile` in a `Row` at 8px padding, a shape `layout_blocks` has no variant for, and `usp_ethernet_ports_card.dart`'s `_SummaryTile` is the same shape a third time. The ticket carries the measurements a shared compact variant must preserve.
- **#1286 (`MapsToRow`'s even split)** — the Ports-tab sweep found the one thing a 191px card cannot show whole, and it is not this card's doing. `MapsToRow` (`lib/page/_shared/components/layout_blocks/row_blocks.dart:148`) wraps both halves in equal-flex `Flexible`s, so `RenderFlex` splits the pair evenly and the shorter child's unused share is **not** redistributed: the target gets exactly half — **38.5px** — for 82.3-103.7px of `IP:port`, however short the source is (measured: a 30.7px source still leaves the target 38.5px of a 77px pair). Its own docstring claims the opposite, that the target is the part that ellipsizes "since the source is short and bounded while the target is not". Showing it whole needs 227.4px in a 191px card.

  Both card-side fixes were measured and are insufficient: putting the mapping on its own full-width line costs the +20px per rule that AC 4 forbids and overflows all 26 locales by +11px to +36px (mutation 10); dropping the DMZ list on top of that clears the overflow but leaves the target at 68.7px, still short. Only fixing the split itself works (mutation 11) — and that file is shared with seven other call sites (statistics, port-forwarding cards and tabs, the AI registry), so it is not a change this single-card PR can justify. The last test in the file is a **ratchet**: it asserts the source is never ellipsized and the target *is*, and tells whoever fixes `row_blocks.dart` to delete it. Filed as **#1286**, with the three options costed. The cleanest (one `InlineSpan` run instead of two `Flexible`s) has its mechanism in ui_kit already — `AppStyledText` is a `RichText` that takes `maxLines`/`overflow` — but it emits only `TextSpan`s from a fixed tag vocabulary (so the arrow, which must come from the icon font because U+2192 has no glyph in the app's font set, cannot sit inside the run), takes no size or colour, and skips `AppText`'s per-locale fallback injection. The upstream ask was filed as **linksys/privacyGUI-UI-kit#23** (`AppText.rich(List<InlineSpan>)`, reusing `AppText`'s existing variant/colour/locale-fallback resolution) and **landed in v2.35.0 while this PR was open**, closing all three gaps — spans take a `WidgetSpan`'d arrow, the rich path carries `variant`/`color`, and the locale fallback is the same merge. #1286's recommendation is updated to that fix rather than the layout-side one.

  **This PR deliberately stays at v2.34.11.** v2.35.1 also takes privacyGUI-UI-kit#24, which makes `AppStyledText` apply the locale fallback: inert upstream (ui_kit injects no resolver) but not here, since this app installs one — so the three `AppStyledText` call sites move their non-Latin rendering to the bundled subset and the `remote_assistance` CJK baselines shift. That is the movement #1285 already owns, so the bump belongs with it in one PR. Verified inert for *this* PR's numbers first rather than assumed: at v2.35.1 this file's 74 tests and the gate's 1644 are green with no measurement moved, and the rewrite of `AppTextVariantX.resolve` behind it is a no-op in this app for the two reasons recorded on #1285.

- **#1285** — building the font guard above turned up a production defect this PR does not touch: `lib/app.dart:185` hands `ThemeData.textTheme.apply()` an already package-prefixed fallback family, and because ui_kit's text styles carry `package: 'ui_kit_library'` the `fontFamilyFallback` getter prefixes it a second time — `packages/ui_kit_library/packages/ui_kit_library/NotoSansTC`, which matches nothing registered. Measured, not inferred. So in every non-Latin locale the bundled subset never draws raw `Text`; the engine falls through to the system font, which is why it has never looked broken. Three test harnesses copy the same line, including `dashboard_card_probe.dart:326` behind this gate, so the fix will move CJK baselines and wants its own PR. Dates from `5beff305`, not from this branch.

- **#1238** — three helpers in the new test (`pumpAt`, `desktopCase`, `loadAppFonts` setup) duplicate what a sibling readability test will need; the extraction plan is noted on that issue.

## Verification

- **`./tool/run_overflow_test.sh`** (the #1183 gate): **1644 pass**, exit 0
- **`./run_tests.sh`** (full functional suite): **5525 pass**, exit 0
- this file: **74 pass**; firewall provider/service tests: **42 pass**
- every other `loadAppFonts` caller, re-run after making it throw (6 readability suites + the shared row/info-row tests + the 1899-test overflow file): all pass. Golden pipeline: no loader error, tally unchanged.
- dashboard card goldens (`test/golden_test/page/dashboard/cards/localizations/`, 33 tests): regenerated for the bump and green. Baselines are gitignored, so nothing is committed. Note the harness's 500×400 box is 128px shorter than this card's shipped 528px, so the donut is suppressed in that golden by design — the 26-locale × 3-width sweep confirms it is drawn at 153.0-160.0px at the height the dashboard actually gives the card.
- whole `--tags golden` suite: `+73 -284` **both** at v2.34.5 and at v2.34.11, same failing set — the bump introduces no golden failure; those 284 are pre-existing stale local baselines in the whole-suite harness.
- `dart format` clean, `flutter analyze` on both changed files: no issues.

## AC checklist

- [x] **AC1** both own sites clean at the narrowest realization, 26 locales, both tabs
- [x] **AC2** the 15 `firewall_overview|min|0` coordinates removed and the gate passes
- [x] **AC3** the 6 fl_chart coordinates closed by call-site constraint, not allowlisted
- [x] **AC4** resolved by constraining width-driven content; `minHeightRows` still 3
- [x] **AC5** legible at 191px, not merely overflow-free — guarded by 74 tree assertions across both tabs, the first of which proves the other 73 measured the app's own fonts. One exception, documented and ratcheted rather than hidden: the port-forwarding mapping target ellipsizes, because `MapsToRow` splits its row evenly (see follow-ups).
- [x] **AC6** fl_chart not forked, patched, or vendored

### DIFF
--- a/doc/dashboard/dashboard_density_design.md
+++ b/doc/dashboard/dashboard_density_design.md
@@ -1,6 +1,6 @@
 # Dashboard Card Density — Design Decisions
 
-**Last Updated: 2026-08-13** · Follow-up to #1183 · Status: **agreed; tickets #1225–#1240 published; #1225 + #1226 + #1233 + #1227 + #1228 + #1229 implemented (not yet merged), rest not started. Allowlist 560 → 181.**
+**Last Updated: 2026-08-14** · Follow-up to #1183 · Status: **agreed; tickets #1225–#1240 published; Track A implemented through #1230 (#1225, #1226, #1233, #1227, #1228, #1229, #1266, #1234, #1236, #1237, #1235, #1247, #1230 — not all merged), #1238 remaining; Track B not started. Allowlist 560 → 27.**
 
 ## Purpose
 
@@ -933,7 +933,8 @@ in one step when the gauge row followed:
 | Gauge centre | `network_health:150` | 3→0 |
 
 The ratchet now stands at **48 coordinates**, all owned by #1230 (21,
-`firewall_overview`) and #1238 (27, `connected_devices`).
+`firewall_overview`) and #1238 (27, `connected_devices`). #1230 has since cleared
+its 21 (§2.11a), leaving **27**.
 
 That 48 replaces the 94 these four tickets left behind. The 46 that went were not
 removed by any of them — they were dead exemptions, and the credit belongs to
@@ -1118,6 +1119,94 @@ the axis labels above. **Not yet measured** — unlike the four shapes, this one
 no double-factor decomposition behind it, and that measurement is #1230's first
 step.
 
+Measured since, and both guessed levers were wrong: the declared height stayed at
+3 rows and the caption goes only because the whole donut does. The cause is a
+width one. See **§2.11a**.
+
+### 2.11a What the two height sites taught us (#1230 — implemented)
+
+All **21** cleared, not the 15 §2.11 scoped: fl_chart's 6 went with them, from the
+call site, with the package untouched. `minHeightRows` stays 3. Five things only
+showed up in the measurement.
+
+1. **Both bottom overflows had a *width* cause, and neither of §2.11's guessed
+   levers was one of them.** The card's height was never short. What overflowed is
+   the three-across metric grid: at the narrowest realization the tab body is
+   **157.4px wide and 205px tall** (203px where the tab bar wraps), and three
+   `InfoGrid` tiles leave each label `(157.4 − 16) / 3 − 24` = **23.1px** of text —
+   one character per line, `ru` wanting **8 lines** (129px) for one label. Those
+   three tiles measured **108px** in `en` and up to **173px** in `ru`, inside a
+   205px viewport that already owed 20px to gaps and 36px to the legend, so 15 of
+   the 26 locales overflowed and the outer `Column` (`:136`) reported up to +26px.
+   Stacking the metrics into full-width rows — a *width* decision — is what closes
+   a *height* overflow: each label gets 111.4-127.6px and the whole stack
+   112-130px, which the donut's `Expanded` absorbs. §1.8's "a
+   bottom overflow is often a width symptom" holds here in its strongest form, and
+   the ticket's AC 4 (do not raise the declared height) was never a constraint to
+   work around — it was the diagnosis.
+2. **Nested sites are mutually exclusive, which is why the two counts summed to 15
+   over 15 locales rather than 26.** `RenderFlex` reports where it is asked to
+   paint: where the grid grew tall enough to starve the donut's `Expanded` to 0,
+   the inner `centerWidget` (`:157`) had no box to overflow and only `:136`
+   reported (7 locales, grids of 156-173px); where the grid left the `Expanded`
+   20-25px, `:157` overflowed by exactly `40 − slot` — +15px to +21px, the caption
+   needing 40px — and the outer `Column` fit (8 locales). One site per locale,
+   never both. So
+   §2.11's "they must be measured together" was not caution, it was necessary:
+   fixing either alone moves the report to the other instead of clearing it.
+3. **fl_chart's 6 were not dependency-blocked; they were unmeasured.** §1.1
+   counted them as third-party and §2.11 reserved a fallback allowlist entry with
+   a tracking note for them. Measured, fl_chart's bottom `SideTitles` asks a
+   constant `reservedSize: 22` whatever the chart's height is, and `AppBarChart`'s
+   vertical path does not override it, so the strip's own `Flex` overflows by
+   `22 − slot`: +5px at 17px slots, +10px in `ru`'s 12px. Two call-site levers
+   close that, and the cheaper-looking one is the worse one. `AppBarChart.xLabels`
+   is nullable and its vertical path passes `showTitles: xLabels != null`, so
+   `xLabels: null` deletes the 22px strip outright — measured, that alone takes all
+   105 of this card's gate cases green with no height guard at all. It is not what
+   shipped: the value axis keeps drawing, so in `ru`'s 12px slot its "2" and "0"
+   overlap above two 8px pills, and the only threshold that would admit the
+   labelless chart where it reads (`en`, 37px) and reject it where it does not
+   (`da`, 17px) sits between 37 and 40px — a knife edge between two shipped
+   locales. So the card declines to draw a chart with no room to be one: below
+   **70px** (22px axis + 24px always-on value labels + 24px
+   `AppBarChart.minBarArea`, leaving every locale 33px clear) it is not built,
+   which is §2.11's Primary path applied one level up. Either way the fallback note
+   was never needed and `known_overflows.json` now holds nothing for this card.
+   Generalisation: a "third-party site" names where the exception is *thrown*, not
+   where the fix has to live.
+4. **Suppression is a fix the gate actively rewards, and the ratchet has no
+   vocabulary to constrain it.** Every earlier Track A ticket made a row give or
+   rearranged one; #1230 is the first to *remove* content below a threshold, and
+   absent content never overflows. A threshold accidentally set above the height
+   the dashboard actually gives the card leaves all 105 of its gate cases green
+   (2 widths × 2 tabs × 26 locales, plus the tab-registry check) and the card blank — verified, not feared: `_kProtocolChartMinHeight` 70 → 200 and
+   `_kDonutMinRingThickness` 10 → 60 each delete a chart at the shipped height
+   with the gate still green. So the guard test asserts *presence* at the 4 rows
+   `HeightStrategy.strict(4)` gives (chart slot 148-173px) as well as absence at
+   the 3 rows the gate pumps (12-37px). Any later ticket that clears a coordinate
+   by not drawing something owes the same pair.
+5. **Where a card overflows, look for what it also clips.** ui_kit's `AppPieChart`
+   takes the ring radius from the call site but the centre-hole radius from the
+   theme (`ChartStyle.pieCenterRadius`, 60px here), so the drawn diameter is
+   `2 × (centre + ring)` and does not follow `size`. The card's `size: 160` with
+   the default 40px ring drew a **200px donut into a 160px box** — 20px clipped off
+   every side at *every* width including desktop, invisible to the gate because a
+   clip is not an overflow. Sizing the ring from the slot fixes it and makes the
+   suppression threshold exact. Second instance of the pattern §2.10d found in
+   `network_health`'s gauge centre.
+
+The claims the gate cannot make — both arrangements legible, both charts present
+at the shipped height, the donut never wider than its box — are covered by
+`test/page/dashboard/views/components/firewall_overview_readability_test.dart`,
+tagged `dashboard-card` so it gates; each of its six groups was verified to fail
+under a mutation of the code it guards (seven mutations, tabulated in the file,
+five of which leave the #1183 gate green). It also re-asserts plain overflow at
+every realization it pumps, which is not redundant with the gate: the gate pumps
+`minHeightRows` only, and the shipped 4 rows is the one height at which the donut's
+caption and fl_chart's axis strip are built at all — the two sites #1230 fixed were
+otherwise measured by nothing at the height they actually render.
+
 ### 2.12 What the first *rearrangement* taught us (#1228 — implemented)
 
 All 52 coordinates cleared as predicted (278 → 226). #1226/#1233/#1227 all made
@@ -1226,10 +1315,11 @@ addition — it changes no card's rendering until a threshold is declared.
 | #1237 | `time_settings` card-own (§2.10d) — **implemented** | 21 |
 | #1235 | `network_health` gauge centre (§2.10d) — **implemented** | 3 |
 | #1247 | `firewall_overview` `MapsToRow` — side effect, not its brief (§2.9a) — **implemented** | 46 |
-| #1230 | `firewall_overview` remaining ×2 sites (§2.11) | 15 |
+| #1230 | `firewall_overview` remaining ×2 sites (§2.11, §2.11a) — **implemented** | 21 |
 | #1238 | `connected_devices` card-own (§2.6) | 1 |
 
-Ceiling **528 / 560**. The other 32 are the dependency-blocked ones (§1.1).
+Ceiling **534 / 560**. The other 26 are the dependency-blocked ones (§2.6) —
+fl_chart's 6 left that set when #1230 closed them from the call site (§2.11a).
 
 After #1234–#1237, and after #1247's side effect was collected (§2.9a), the
 allowlist holds **48** coordinates: 21 `firewall_overview` (#1230) and 27
@@ -1248,6 +1338,11 @@ other **32** are dependency-blocked: fl_chart's 6 and the 26 that cannot go gree
 without a ui_kit change (they need the card's own two rows fixed as well, so
 "blocked" means necessary-but-not-sufficient, not untouched).
 
+Measured since: #1230 cleared all **21**, fl_chart's 6 included, by declining to
+draw a chart into a slot shorter than its own axis strip (§2.11a). So 22 of the 48
+were clearable here, not 16, and the allowlist now holds **27** — all
+`connected_devices`, all #1238's.
+
 Two figures earlier in this document are contradicted by that attribution, both
 downward:
 
@@ -1265,15 +1360,20 @@ downward:
   19 is left as written because its table is a record of what was believed at
   baseline and its rows sum to 560.
 
-So the dependency-blocked share of the baseline is **32 of 560, not 45**, and
-`firewall_overview` is a *height* problem now — all 21 of its coordinates are
-bottom overflows, a different class from the four width shapes these tickets
-settled. #1230 is still the largest single block of clearable coordinates left,
-at 15 rather than 48.
+So the dependency-blocked share of the baseline is **26 of 560, not 45** — 32
+until #1230 measured fl_chart's 6 and closed them at the call site (§2.11a) — and
+`firewall_overview` was a *height* problem, all 21 of its coordinates bottom
+overflows, a different class from the four width shapes these tickets settled.
+#1230 closed all 21; what is left is #1238's 27.
 
-Their `tracking` notes are left as the epic requires: a ticket touches only the
-notes of cards it closes, so both cards keep the `baseline #1183` default until
-their own ticket names an owner.
+Neither card kept the epic's `baseline #1183` default in the end, and the rule
+that a ticket touches only the notes of the cards it closes is why. `#1230` closed
+`firewall_overview` outright, so both its keys and its note are gone from
+`known_overflows.json` — including the fl_chart fallback entry §2.11 had reserved,
+which measurement made unnecessary (§2.11a). `connected_devices`' note is now the
+file's only one: it names the upstream blocker (linksys/privacyGUI-UI-kit#20) for
+26 of its 27 coordinates and points the two card-own sites at #1238, which is the
+ticket that will delete it.
 
 #1266 is in this track despite clearing nothing: it is the only entry that *adds*
 coordinates (3, by localizing a hardcoded string) and removes them again in the
--- a/doc/dashboard/dashboard_density_design.md
+++ b/doc/dashboard/dashboard_density_design.md
@@ -1192,7 +1192,11 @@ showed up in the measurement.
    `2 × (centre + ring)` and does not follow `size`. The card's `size: 160` with
    the default 40px ring drew a **200px donut into a 160px box** — 20px clipped off
    every side at *every* width including desktop, invisible to the gate because a
-   clip is not an overflow. Sizing the ring from the slot fixes it and makes the
+   clip is not an overflow. Measured by pixel extent, the painted diameter is 200px
+   at every `size` from 120 to 300, so `size` does not bound the drawing in either
+   direction; filed upstream as linksys/privacyGUI-UI-kit#22, and the app's other
+   four `AppPieChart` call sites (`size: 180` ×2, `size: 120`, one caller-supplied)
+   still paint 200px. Sizing the ring from the slot fixes it here and makes the
    suppression threshold exact. Second instance of the pattern §2.10d found in
    `network_health`'s gauge centre.
 
@@ -1206,6 +1210,12 @@ every realization it pumps, which is not redundant with the gate: the gate pumps
 `minHeightRows` only, and the shipped 4 rows is the one height at which the donut's
 caption and fl_chart's axis strip are built at all — the two sites #1230 fixed were
 otherwise measured by nothing at the height they actually render.
+
+Two follow-ups left open rather than folded into #1230: the narrow arrangement's
+tile is `_InfoGridTile` in a `Row`, which `layout_blocks` has no variant for and
+`ethernet_ports` hand-rolls too (**#1275**), and the helpers these readability
+tests share are now cloned across up to seven files (noted on **#1238**, the next
+card to need them).
 
 ### 2.12 What the first *rearrangement* taught us (#1228 — implemented)
 
--- a/doc/dashboard/dashboard_density_design.md
+++ b/doc/dashboard/dashboard_density_design.md
@@ -1126,7 +1126,7 @@ width one. See **§2.11a**.
 ### 2.11a What the two height sites taught us (#1230 — implemented)
 
 All **21** cleared, not the 15 §2.11 scoped: fl_chart's 6 went with them, from the
-call site, with the package untouched. `minHeightRows` stays 3. Five things only
+call site, with the package untouched. `minHeightRows` stays 3. Six things only
 showed up in the measurement.
 
 1. **Both bottom overflows had a *width* cause, and neither of §2.11's guessed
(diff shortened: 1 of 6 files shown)


AustinChangLinksys (commented):
🤖 **Automated Review — Round 1** · `20402ab..37fd503` (full)

**Verdict: 💬 Self-review (comment only)** — Austin's own PR; review provided as reference only (GitHub self-approval prohibited). No Critical found; 4 Warnings across both axes.

#### Standards
*(Reviewer A: security/correctness · Reviewer B: architecture/maintainability)*

|---|---|---|---|
| ⚠️ | 🟢High | `usp_firewall_overview_card.dart:133` | `_kDonutMaxDiameter=160` hardcoded, silently decouples from `pieCenterRadius` if theme changes |
| ⚠️ | 🟢High | `usp_firewall_overview_card.dart:435` | `portForwardingRules` raw `List` + `rule.protocol as String` cast — type error only at runtime (pre-existing, but in PR's rewritten scope) |
| ⚠️ | 🟡Med | `usp_firewall_overview_card.dart:358` | `centreRadius` read in outer `build()` before `LayoutBuilder`, breaks if `AppDesignTheme` is missing in test harness |
| 💡 | 🟢High | `usp_firewall_overview_card.dart:304` + `usp_ethernet_ports_card.dart:155` | [Duplicated Code] `_StackedMetrics`/`_SummaryTile` are the same shape a 3rd time — tracked in #1275 |
| 💡 | 🟡Med | `firewall_overview_readability_test.dart:142` | [Duplicated Code] readability helpers (`pumpAt`, `contentBottom`…) cloned across test files — tracked in #1238 |
| 💡 | 🟡Med | `usp_firewall_overview_card.dart:375` | `showLabels: false` has no test guard; slice labels returning would be invisible to CI |
| 💡 | ⚪Low | `usp_firewall_overview_card.dart:304` | `_StackedMetrics.metrics` has no empty-list assertion |
| 💡 | 🟡Med | `firewall_overview_readability_test.dart:152` | `metricRects()` finder ambiguous on desktop width (InfoGrid also renders `LayoutBlock`) |
| 💡 | 🟢High | `usp_firewall_overview_card.dart:176` | `'No firewall rules configured'` not localized — pre-existing, but `_RulesTab.build()` was fully rewritten here |

#### Spec
*(Reviewer C: spec compliance — spec source: issue #1230)*

|---|---|---|---|
| ⚠️ | 🟡Med | AC1 spec → `firewall_overview_readability_test.dart` | AC1 requires "both tabs" legible at 191px — Ports tab has no 26-locale readability sweep (only `ru` for chart presence) |
| 💡 | 🟢High | AC1 spec → `usp_firewall_overview_card.dart` | `:136`/`:157` no longer exist at those lines — fix was elimination, not in-place repair; PR description could note this |
| 💡 | 🟢High | AC3 spec → `test/fixtures/known_overflows.json` | fl_chart 6-coord disposition documented in §2.11a only; `known_overflows.json` has no residual key — audit trail depends on doc |

---

<details>

**W-1: `_kDonutMaxDiameter` silently decoupled from theme** 🟢High — *both reviewers flagged independently*
`lib/page/firewall/cards/usp_firewall_overview_card.dart:133`
```dart
const double _kDonutMinRingThickness = 10;
const double _kDonutMaxDiameter = 160;
```
Usage at line ~362:
```dart
final diameter = math.min(
  math.min(constraints.maxWidth, constraints.maxHeight),
  _kDonutMaxDiameter,  // 160 = 2×(centreRadius60 + ring20) — not derived
);
if (diameter < 2 * (centreRadius + _kDonutMinRingThickness)) { ... }
```
`_kDonutMaxDiameter=160` is a standalone constant; the threshold `2×(centreRadius+10)=140px` is computed at runtime from the theme. If `pieCenterRadius` is ever raised (e.g. to 80px), threshold becomes 180px > 160px → donut never renders. Suggestion: at minimum add `assert(_kDonutMaxDiameter >= 2 * (centreRadius + _kDonutMinRingThickness))` inside LayoutBuilder, or derive `_kDonutMaxDiameter` from the theme.
**Fix:** Add a runtime assert or derive the cap from theme values dynamically.

---

**W-2: raw `List portForwardingRules` + unsafe cast** 🟢High — *both reviewers flagged independently (pre-existing, in PR's rewritten scope)*
`lib/page/firewall/cards/usp_firewall_overview_card.dart:435,459`
```dart
final List portForwardingRules;  // raw type — List<dynamic>
...
final proto = rule.protocol as String;  // runtime TypeError if wrong type
```
`dmzSummaries: List<DmzEntrySummary>` in the same class is typed correctly. The `as String` cast is a symptom of the missing type. Pre-existing in base, but `_PortsTab` was substantively modified in this PR (new `LayoutBuilder`, `_kProtocolChartMinHeight` guard added).
**Fix:** `final List<PortForwardingRuleModel> portForwardingRules;` — remove `as String` cast.

---

**W-3: `centreRadius` read before `LayoutBuilder` callback** 🟡Med — *Reviewer A only*
`lib/page/firewall/cards/usp_firewall_overview_card.dart:358`
```dart
// In _TargetDonut.build():
final centreRadius = AppDesignTheme.of(context).chartStyle.pieCenterRadius;
return LayoutBuilder(
  builder: (context, constraints) {
    if (diameter < 2 * (centreRadius + _kDonutMinRingThickness)) { ... }
    // centreRadius is from the outer context, not the LayoutBuilder's context
  },
);
```
Functionally correct when `AppDesignTheme` is everywhere in the tree. However, if a test pumps `_TargetDonut` without wrapping `AppDesignTheme`, the exception fires in the outer `build()` before `LayoutBuilder` even runs — harder to diagnose. Moving `centreRadius` inside the callback is a one-line change that removes this ambiguity and also fixes the W-1 assert opportunity.
**Fix:** Move `centreRadius` read inside the `LayoutBuilder` builder callback.

---

**W-4 (Spec): AC1 — Ports tab missing 26-locale readability sweep** 🟡Med — *Reviewer C only*
`test/page/dashboard/views/components/firewall_overview_readability_test.dart` — "no metric text is shredded" group
AC1 spec: "all 26 locales and both tabs". The full 26-locale text shredding sweep uses `tabIndex: 0` only. Tab 1 (Ports) is tested only for `AppBarChart` presence with locale `ru`. `MapsToRow` cells and `_ProtocolBadge` on the Ports tab could experience similar text compression at 191px that the overflow gate wouldn't catch (absent content never overflows).
**Fix:** Add a 26-locale readability group for `tabIndex: 1` checking text widget line counts.

</details>

<details>

- **Core overflow fix is correct and well-reasoned.** The three constants (`_kMetricsSideBySideMinWidth=328`, `_kDonutMinRingThickness=10`, `_kProtocolChartMinHeight=70`) are derived from actual pixel measurements across 26 locales, not guessed. The dartdoc on each constant is exemplary.
- **All 21 coordinates cleared cleanly.** Both `firewall_overview|min|0` (15 card-own) and `firewall_overview|min|1` (6 fl_chart) removed from `known_overflows.json` — no allowlisting, no silent suppression.
- **AC4 confirmed.** `minHeightRows: 3` unchanged. The overflow is correctly diagnosed as width-driven, not height-driven.
- **AC6 confirmed.** fl_chart untouched; fix is entirely at the call site via `_kProtocolChartMinHeight` guard.
- **Readability test mutation coverage is thorough.** 7 mutations tested, 5 of which are invisible to the existing overflow gate — the new test file is justified and adds genuine protection.
- **`AppPieChart` clip bug correctly worked around.** `sectionRadius: diameter / 2 − centreRadius` is the right fix for the painted-extent-vs-box-size mismatch (filed upstream as privacyGUI-UI-kit#22).
- **Layer dependencies correct.** Card only imports Layer-1 providers; no cross-layer violations.
- **AC3 fl_chart strategy is sound.** Choosing suppression over `xLabels: null` is well-argued: the label-less chart's value-axis overlap at `ru`'s 12px slot would be worse than no chart at all; 70px gives 33px clearance for every locale.

</details>

---

AustinChangLinksys (commented):
🤖 **Automated Review — Round 2** · `37fd503..31f573b` (incremental)

**Verdict: 💬 Self-review (comment only)** — Austin's own PR; auto-approve suppressed per policy. All Round 1 Warnings resolved; 3 new Warnings (test-infrastructure only) across both axes.

#### Standards
*(Reviewer A: security/correctness · Reviewer B: architecture/maintainability)*

|---|---|---|---|
| ⚠️ | 🟡Med | `test/util/app_test_fonts.dart:67` | `mainFont.load()` called unconditionally even when Bold is missing — error message can't distinguish "Roman missing (all measurements Ahem)" from "Bold missing (bold measurements off)" |
| ⚠️ | 🟡Med | `firewall_overview_readability_test.dart:261–278` | `drawnDonut()` reaches through to fl_chart's internal `PieChart` class + `dynamic` cast — a ui_kit bump that renames/restructures `PieChart` silently breaks this helper |
| ⚠️ | 🟡Med | `firewall_overview_readability_test.dart:213–219` | `portsTabTexts()` walks `MapsToRow → Column` to locate tab root — fragile coupling, silently drifts if `_PortsTab` adds a wrapper widget above `Column` |
| 💡 | 🟢High | `test/util/app_test_fonts.dart:43` | [style] `///` DartDoc on a local variable; should be `//` |
| 💡 | ⚪Low | `usp_firewall_overview_card.dart:521` | `'DMZ'` hardcoded literal (pre-existing; `'WAN'`/`'LAN'` also kept — if intentional, a one-line comment noting the policy would clarify) |
| 💡 | 🟡Med | `lib/page/_shared/components/layout_blocks/row_blocks.dart:145` | `MapsToRow` docstring describes ideal behaviour ("target ellipsizes") but the equal-flex implementation contradicts it; a TODO→#1286 (or #1230) would prevent future callers from relying on the misleading claim |

#### Spec
*(Reviewer C: spec compliance — spec source: issue #1230)*

|---|---|---|---|
| ⚠️ | 🟡Med | AC3 → `test/fixtures/known_overflows.json` | fl_chart audit trail exists only in doc §2.11a; `known_overflows.json` `tracking` block has no `firewall_overview` entry — doc refactors could silently erase the audit trail |
| 💡 | ⚪Low | AC5 → `firewall_overview_readability_test.dart:599–665` | Ratchet test cites #1286 (MapsToRow fix) but no `#1286` reference found in repo files — if the issue was opened, adding the link to the ratchet reason string would make the handoff traceable |

---

<details>
<summary>✅ Round 1 Warnings resolved (3/3)</summary>

**W-1: `_kDonutMaxDiameter=160` hardcoded** — ✅ **Resolved** (both reviewers confirmed)
`_kDonutMaxDiameter` is gone. In its place: `_kDonutDesignRingThickness = 20` (`usp_firewall_overview_card.dart:155`). The diameter is now computed as `min(box, 2 × (centreRadius + _kDonutDesignRingThickness))` inside the `LayoutBuilder`, so it scales correctly with every theme's `pieCenterRadius`. The coupling the original Warning flagged is eliminated.

**W-2: `portForwardingRules` raw `List` + `as String` cast** — ✅ **Resolved** (both reviewers confirmed)
`_PortsTab.portForwardingRules` is now `List<PortForwardingRuleUIModel>` (`:461`); the `as String` cast is gone — `rule.protocol` is accessed directly as `final String protocol`. Type chain verified end-to-end from the provider (`port_forwarding_data_provider.dart:20`).

**W-3: `centreRadius` read before `LayoutBuilder`** — ✅ **Resolved** (both reviewers confirmed)
`centreRadius` read moved inside the `LayoutBuilder.builder` callback (`usp_firewall_overview_card.dart:386–387`), with an explanatory comment. The size request and the admission floor now come from the same theme read in the same context.

</details>

<details>

**Standards W-1: `mainFont.load()` diagnostic gap when Bold is missing** 🟡Med — *Reviewer A only*
`test/util/app_test_fonts.dart:53–67`
```dart
if (mainFontFile.existsSync()) {
  mainFont.addFont(Future.value(...));   // Roman added
  if (boldFile.existsSync()) {
    mainFont.addFont(Future.value(...)); // Bold added
  } else {
    missing.add(boldFile.path);          // only logged
  }
} else {
  missing.add(mainFontFile.path);        // only logged
}
await mainFont.load();                   // ← called regardless
```
When Roman exists but Bold is missing, `mainFont.load()` succeeds with Roman only. The `throw StateError` at line ~140 fires correctly either way (Roman-missing and Bold-missing are both fatal). The diagnostic gap is that the error message cannot distinguish "all measurements were Ahem" (Roman missing) from "bold-weight measurements were off" (Bold missing only). Not a correctness bug; `throw` path is correct.
**Fix (optional):** emit separate error messages for "Roman missing" vs "Bold missing"; or fast-fail inside the `else` block for Roman absence.

---

**Standards W-2: `drawnDonut()` penetrates fl_chart's internal `PieChart` API** 🟡Med — *Reviewer B only*
`test/page/dashboard/views/components/firewall_overview_readability_test.dart:263–278`
```dart
final pie = find.byWidgetPredicate(
    (w) => w.runtimeType.toString() == 'PieChart',
    description: 'fl_chart PieChart');
final dynamic data = (tester.widget(pie) as dynamic).data;
final sections = data.sections as List;
```
`PieChart`, `.data`, `.sections`, `.radius`, `.title` are fl_chart's private API — not re-exported by `AppPieChart`. A ui_kit bump that renames the class or restructures the data model breaks this helper silently at runtime (`TypeError`) or causes `findsNothing`. The author's comment at `:258–259` acknowledges this as a deliberate tradeoff ("measuring the drawing instead of the request"). Marked as known structural debt.
**Fix (deferred):** When ui_kit exposes a `AppPieChart.debugGeometry()` test hook (or similar), replace `drawnDonut()` with that stable seam.

---

**Standards W-3: `portsTabTexts()` implicit widget-tree walk** 🟡Med — *Reviewer B only*
`test/page/dashboard/views/components/firewall_overview_readability_test.dart:213–219`
```dart
Finder portsTabTexts() => find.descendant(
      of: find
          .ancestor(
              of: find.byType(MapsToRow).first,
              matching: find.byType(Column))
          .first,
      matching: find.byType(Text),
    );
```
This locates the Ports tab root by walking up from `MapsToRow` to the nearest `Column`. If `_PortsTab.build()` wraps `Column` in a `SingleChildScrollView` or `Padding`, the finder silently changes its coverage range without a test failure.
**Fix:** Add `Key('ports_tab_content')` to `_PortsTab`'s root widget and use `find.byKey` — or add an inline comment warning that this finder is sensitive to the column's ancestry.

---

**Spec Warning — AC3: fl_chart audit trail is doc-only** 🟡Med — *Reviewer C only*
`test/fixtures/known_overflows.json` (head)
```json
{
  "tracking": { "connected_devices": "..." },
  "allowlist": { ... }
}
```
The `tracking` block has no `firewall_overview` entry. The fl_chart 6-coordinate closure is documented thoroughly in `doc/dashboard/dashboard_density_design.md §2.11a` but that's the only machine-readable record. Spec AC3 is technically satisfied (primary path closes the coordinates, no allowlist needed), but if the doc is later refactored the audit trail disappears.
**Fix (optional):** Add to `known_overflows.json`:
```json
"tracking": {
  "firewall_overview": "All 21 coordinates closed in #1230 — 15 card-own via layout fix (min|0), 6 fl_chart via call-site height guard _kProtocolChartMinHeight=70 (min|1, §2.11a item 3). No allowlist entries remain."
}
```

</details>

<details>

- **All 3 Round 1 Warnings resolved cleanly.** The `_kDonutDesignRingThickness`-based diameter, typed `List<PortForwardingRuleUIModel>`, and LayoutBuilder-scoped `centreRadius` read all confirmed by both reviewers with code evidence.
- **`loadAppFonts()` now throws instead of silently skipping missing text fonts.** This is the right call — a skipped font leaves every pixel assertion measuring Ahem, not NeueHaas, and that is not a state worth running in.
- **Font identity test architecture is sound.** The double-net (block-font ratio + Noto identity) covers all three ways the family registration can go wrong (no fallback → Ahem, fallback → Noto, reference corrupt → undetectable). Each net was mutation-verified with a renamed family registration.
- **Ports tab 26-locale readability sweep added.** Covers all 26 locales at `tabIndex: 1`; mutation 9 verifies it fires. The `MapsToRow` target-ellipsis limitation is correctly ratcheted rather than hidden.
- **`pubspec.yaml` bump (v2.34.5 → v2.34.11) is safe.** No breaking API changes; the four other `AppPieChart` call sites carry no `sectionRadius` and benefit from the upstream fix without modification. Verified.
- **Layer dependencies remain clean.** Card imports only Layer-1 providers; no cross-layer violations introduced.
- **11-mutation verification table in PR body is comprehensive.** 7 of 11 mutations would leave the overflow gate green while the card silently degrades — the readability suite genuinely adds coverage the gate cannot provide.
- **AC4 (minHeightRows=3) confirmed maintained.** Widget spec file untouched this round; the test reads `constraints.minHeightRows` from the spec rather than hardcoding it.
- **All spec ACs satisfied in this round.** AC1 (both tabs, 26 locales), AC2 (15 coordinates removed), AC3 (fl_chart 6 via call-site constraint), AC4 (height unchanged), AC5 (legibility proven with font-verified measurements), AC6 (fl_chart unforked) — all green.

</details>

---

Issue #1282, opened 2026-08-14
Static pre-compression (brotli -q11, originals deleted): build_web.sh step + lighttpd config — blocked on firmware confirmation


AustinChangLinksys:
## Context

Firmware-size measurement for Architecture#145 established that the UI occupies **9.45 MB inside the rootfs squashfs**. Static pre-compression is one of the levers on that number, and unlike the others it also cuts what the browser downloads.

This originated from a firmware-team proposal (Jenkins `gzip -9` + `rm` originals + lighttpd `mod_rewrite`). On-device measurement showed the gzip form of that proposal makes flash **larger**, so this ticket carries the corrected form: **brotli, with originals deleted**.

> **Status: blocked on firmware-team confirmation.** The server half lives in `/etc/lighttpd/conf.d/` on the router, which the firmware team owns. We have verified feasibility on a real device (M60TB, FW 1.2.3.26072920) using isolated ports without touching production config, but nothing ships until they confirm the config change is acceptable and tell us how it versions. The build-side half is inert on its own — pre-compressed files with no server rule are just 404s.

**#1281 has landed** (PR #1283, merged 2026-08-17), which changes this ticket in two ways: the `chromium` variant no longer exists, so the target list is **two files, not three**; and the combined saving below is now the live baseline rather than a projection.

## Measured effect

Measured with the vendor's own `mksquashfs` parameters, read from the shipped image's superblock (`xz` / dict 262144 / `-Xbcj arm` / block 262144). All rows are real pack runs — nothing is derived by adding per-file numbers, because XZ shares a dictionary across files.

| Combination | Distribution tgz | Extracted | **Inside rootfs squashfs** |
|---|---|---|---|
| Today | 11.08 MB | 28.85 MB | **9.45 MB** |
| gzip -9 + delete originals | 11.01 MB | 14.28 MB | **10.76 MB (+13.9%)** |
| **brotli -q11 + delete originals** | **9.21 MB** | **12.46 MB** | **8.96 MB (−5.2%)** |
| Drop `chromium` variant (#1281) | 8.96 MB | 23.28 MB | **7.73 MB (−18.1%)** |
| Drop `chromium` + gzip -9 | 8.91 MB | 12.12 MB | **8.66 MB (−8.3%)** |
| **Drop `chromium` + brotli -q11** | **7.63 MB** | **10.82 MB** | **7.39 MB (−21.8%)** |

[...]

PR #1283, opened 2026-08-14
feat(web): drop the chromium CanvasKit variant (#1281)

### LINKED ISSUE #1281: Drop the chromium CanvasKit variant (1.71 MB rootfs) — needs canvasKitVariant: "full" in the same commit
## Context

Firmware-size measurement for Architecture#145 broke the UI's 9.45 MB rootfs footprint down by component. The second-largest single removable item is the **`chromium` CanvasKit variant** — `web/assets/chromium/` — at **1.71 MB inside the rootfs squashfs (18.1% of the UI's total footprint)**.

Flutter ships two CanvasKit builds and picks one at runtime by browser capability. Both are committed to this repo (they are not `flutter build` output):

| File | Raw | Cost inside rootfs squashfs |
|---|---|---|
| `web/assets/canvaskit.wasm` (full) | 6.89 MB | 2.29 MB |
| `web/assets/chromium/canvaskit.wasm` | 5.49 MB | **1.71 MB** |

Every client downloads exactly one of them, but both occupy flash. Component costs were measured by removing the component and repacking with the vendor's `mksquashfs` parameters, because XZ shares a dictionary across files so per-file sizes do not add.

## Proposal

Drop the `chromium` variant and pin every browser to the `full` build.

This is **two changes that must land in the same commit**:

```
web/assets/chromium/          ← delete the directory (canvaskit.js + canvaskit.wasm)
web/flutter_bootstrap.js      ← add canvasKitVariant: "full" to the config block
```

Deleting the directory alone is not viable. `flutter_bootstrap.js` commits to a variant from capability detection and then `import()`s its loader with **no 404 fallback path**, so a Chromium-based browser requests `assets/chromium/canvaskit.js`, gets a 404, and the app never boots. Preliminary on-device testing reproduced exactly that — a blank screen with:

```
TypeError: Failed to fetch dynamically imported module: .../assets/chromium/canvaskit.js
```

Setting `canvasKitVariant: "full"` moves the decision to build time, so capability detection never routes to the missing path.

## What changes for users

Preliminary measurement on an M60TB (parallel lighttpd instances on isolated ports, production `/www` and config untouched) recorded which variant each browser actually fetched:

| Browser | Variant fetched today | After this change |
|---|---|---|
| Chrome 151 (desktop) | `chromium` — 5.76 MB | `full` — 7.23 MB (**+1.47 MB per cold load**) |
| iPhone Safari 18.5 | `full` — 7.23 MB | unchanged |

So the transport cost lands entirely on Chromium-based browsers; WebKit-based browsers (including all iOS browsers) already download the `full` build and are unaffected.

The trade is **1.71 MB of flash, permanently, against 1.47 MB of one-time cold-load transport for Chromium users only** — and the transport side is partly recoverable: the static pre-compression proposal (separate ticket) brings the `full` build down over the wire.

Both variants are the same CanvasKit renderer; `chromium` is a size optimisation that uses Chromium-only APIs, not a different feature set. No rendering behaviour should change. **This needs to be confirmed, not assumed** — see below.

## Still to verify

Preliminary results are promising but the coverage is not yet sufficient to ship:

1. **Browser matrix.** Only Chrome 151 and iPhone Safari 18.5 have been exercised, and only as far as loading the dashboard. Needs at least: Edge, Firefox, Android Chrome, iPad Safari, and whatever the QA baseline set is.
2. **Rendering parity.** Confirm the `full` build renders identically — the golden test suite is the natural instrument, since a variant swap would show up as pixel diffs across the board.
3. **Cold-load impact on Chromium.** Quantify the real-world cost of +1.47 MB over the router's LAN, not just the byte count. Also confirm nothing in the app has a load-time budget that this breaks.
4. **PWA / service worker.** Confirm the install path and any cached-asset manifest do not still reference the removed directory.
5. **Perf on constrained clients.** The `full` build is larger to compile as well as to download; check first-paint on a low-end Android device.

## Acceptance criteria

1. `web/assets/chromium/` removed; `web/flutter_bootstrap.js` sets `canvasKitVariant: "full"` — in one commit, so neither can ship without the other.
2. No reference to `assets/chromium` remains anywhere in the repo or generated build output.
3. Verification items 1–5 above are recorded on this ticket with results before merge.
4. Golden test suite green (or every diff explained and re-baselined deliberately).
5. `./run_tests.sh` green; `flutter analyze` clean.
6. A built `web/` tree serves and boots from the router with no `assets/chromium` request in the access log.

## Notes

- The saving is **flash inside the rootfs**, which is separate from what shrinking rootfs frees as usable flash — that depends on partition layout and is out of scope here. Sizes and method are on Architecture#145.
- The `full` variant **cannot** be the one removed: it is the only build WebKit-based browsers can use, so dropping it makes the GUI unopenable on all iOS browsers.
- Cross-check against the font-reduction discussion in Architecture#177; these are independent levers on the same 9.45 MB.

AustinChangLinksys:
Closes #1281

Flutter ships two CanvasKit builds and picks one at runtime from browser capability. Both are committed under `web/assets/` and both occupy flash, but every client downloads exactly one. This drops the `chromium` variant and pins all browsers to `full`.

**Saving: 1.72 MB inside the rootfs squashfs — 18.2% of the UI's total footprint.** Measured on the real release build with the vendor's mksquashfs parameters (`xz` / dict 262144 / `-Xbcj arm` / block 262144), not estimated:

| Dimension | Before | After | Change |
|---|---|---|---|
| Distribution tgz | 11.08 MB | 8.96 MB | −2.12 MB (−19.1%) |
| Extracted | 28.85 MB | 23.19 MB | −5.66 MB (−19.6%) |
| **Rootfs squashfs** | **9.45 MB** | **7.72 MB** | **−1.72 MB (−18.2%)** |

The three dimensions don't fall by the same amount because `canvaskit.wasm` compresses to ~31% of raw under XZ — its cost in flash is far below its file size. Context and the full component breakdown are on linksys/Architecture#145.

## Why the deletion and the config change are one commit

`flutter_bootstrap.js` commits to a variant from capability detection and then `import()`s that variant's loader **with no 404 fallback path**. Shipping either half alone white-screens:

- **Directory removed, config unchanged** → Chromium-based browsers are routed to the deleted `assets/chromium/canvaskit.js` and fail on `Failed to fetch dynamically imported module`.
- **Config changed, directory kept** → 1.71 MB of unreachable payload stays in the rootfs, i.e. the whole point is lost.

`canvasKitVariant: "full"` moves the decision to build time so detection can never route to the removed path.

## Which browsers actually change behaviour

The variant predicate is `hasChromiumBreakIterators && hasImageCodecs`, which requires `Intl.v8BreakIterator` (V8-only) and `ImageDecoder` with `navigator.vendor === "Google Inc."`. So the real distribution is:

| Browser | Variant before | After | Actually changes? |
|---|---|---|---|
| Desktop Chrome | chromium | full | **Yes** — verified |
| Edge | chromium | full | Yes (same V8, same vendor string) |
| Android Chrome | chromium | full | **Yes** |
| Firefox | **full** | full | No — no `v8BreakIterator`, already on full |
| iOS / iPadOS Safari | **full** | full | No — same |

Firefox and every iOS browser fetch byte-identical payloads before and after. That is a property of the predicate, not an assumption, which is why the manual matrix could be reduced to the Chromium family.

**The reverse change is not available:** `full` cannot be dropped. It is the only variant WebKit-based browsers can use, so removing it would break the GUI on every iOS browser.

The transport cost of the larger build (+1.4 MB) falls only on Chromium browsers. CanvasKit is always served by the router, so offline availability is unaffected by this change.

## Verification

Deployed a local release build to an M60TB's production `/www` (UI entries only; the 22 non-UI entries under `/www` were preserved and re-verified afterwards).

Confirmed on device:

- `/assets/canvaskit.js` → 200, `text/javascript`; `/assets/canvaskit.wasm` → 200 (7,229,467), `application/wasm`
- `/assets/chromium/canvaskit.js` and `/assets/chromium/canvaskit.wasm` → **404**
- Served `flutter_bootstrap.js` md5 matches the file on disk (no rewriting in between)
- All non-UI endpoints still respond (`/JNAP/`, `/sysinfo.cgi`, `/devices.html`, `/test.html`, `/ui/`, `/favicon.ico`)
- `/www` total 30.6 MB → 25.0 MB

Browser-side:

- **Desktop Chrome cold load — pass.** This is the one group whose variant actually changed.
- **Offline (WAN-down) cold load — pass.** This also proves CanvasKit resolves locally: the loader's `import()` has no fallback, so if `canvasKitBaseUrl` had resolved to gstatic.com the app could not have booted at all with no WAN.

Not covered, deliberately:

- **Android Chrome first-paint** is the one open item. It is a cost measurement (`full` is 1.4 MB larger), not a functional risk — Chrome already demonstrates the `full` variant works on V8. Tracked on #1281.
- **Router access log** cannot be produced: `/etc/lighttpd/conf.d/30-accesslog.conf` loads `mod_accesslog` but sets no `accesslog.filename`. Adding it means editing a firmware-owned config, so the 404s above stand in as the equivalent evidence. `/etc/lighttpd` was not modified and lighttpd was not restarted at any point.

## Two corrections to the ticket

- **AC#4 ("golden suite green") cannot demonstrate rendering parity.** The golden suite runs under the Flutter test VM via alchemist with no `--platform chrome` anywhere, so CanvasKit is not involved and a variant swap produces zero pixel diffs. Green here means "nothing else regressed", which is worth having, but rendering parity is only established by the manual browser checks above.
- **Verification item 4 (PWA / service worker) is closable by inspection.** The generated `flutter_service_worker.js` in this Flutter version is 28 lines that self-unregister, with no `RESOURCES` manifest — so there is no cached-asset list that could reference the removed paths.

## Why there is a test for this

`web/assets/canvaskit.*` are hand-vendored copies of the engine's `build/web/canvaskit/` output — byte-identical, but committed rather than generated. A future Flutter upgrade where someone re-copies that directory would silently restore `chromium/`, and **nothing in the build would fail**. So AC#2 needs a mechanism, not a one-time grep.

`test/web/canvaskit_variant_test.dart` asserts both halves, and additionally pins `canvasKitBaseUrl: "./assets/"` — the only reason CanvasKit resolves locally, since `buildConfig` sets no `useLocalCanvasKit`. Dropping that line would fall back to gstatic.com, pass every online test, and white-screen only on a router with no WAN. Both assertions were validated adversarially (re-creating `web/assets/chromium/` and blanking `canvasKitBaseUrl` each turn the suite red).

## Note on AC#2 (no `assets/chromium` reference anywhere)

The minified Flutter loader bundle on line 3 of `web/flutter_bootstrap.js` still contains the literal `a&&(s=l(s,"chromium"))`. That is vendored engine output — the same string appears in `flutter build web`'s own generated `flutter.js` — so it cannot be removed without patching the Flutter SDK. It is unreachable with `canvasKitVariant: "full"` (`a` is always `false`).

AC#2 is met for everything under our control. Flagging it here so the string turning up in a grep does not read as an unmet criterion.

## One stale Jenkins reference, deferred to #1282

The Jenkins `Compress=true` block pre-gzips three files, one of which is `build/web/assets/chromium/canvaskit.wasm` — a path that no longer exists after this merges. That block is **not currently enabled**, so the stale path is inert and cannot break a build as things stand.

Cleaning it up belongs to #1282 rather than here, because #1282 (static pre-compression) is the only work that would turn `Compress` on — so the flag cannot be enabled without touching those lines anyway.


AustinChangLinksys (commented):
🤖 **Automated Review — Round 1** · `c675111..949dd56` (full)

**Verdict: 💬 Self-review (comment only)** — Author is the same account; auto-approve suppressed per policy. Review findings below are for reference only; please sanity-check before merge.

#### Standards

|---|---|---|---|
| ⚠️ | 🟢High | `test/web/canvaskit_variant_test.dart:27–40` | Test uses bare relative paths (`web/assets/chromium`, `web/flutter_bootstrap.js`) that rely on process CWD being the project root — may silently false-pass in non-root CI environments |
| 💡 | 🟡Med | `test/web/canvaskit_variant_test.dart:43,50` | String-exact match on `canvasKitVariant: "full"` is whitespace-sensitive; a formatter pass could cause spurious CI failures |
| 💡 | 🟡Med | `test/web/canvaskit_variant_test.dart:33–34` | Guards Flutter engine file presence (pre-existing) rather than this PR's decision; risk of noise on future engine upgrade |
| 💡 | 🟢High | `test/web/canvaskit_variant_test.dart:1–60` | Test file lacks a reverse pointer back to `web/flutter_bootstrap.js`; only the bootstrap has `// Guarded by test/...` — not the other way around |
| 💡 | 🟡Med | `web/assets/canvaskit.wasm` | Exec bit `100755` (pre-existing, not introduced here) — minor security-audit noise; `git update-index --chmod=-x` in a separate chore commit |

#### Spec

|---|---|---|---|
| ⚠️ | 🟢High | spec#AC-3 → Issue #1281 | Verification items 1–5 not yet recorded on the ticket — spec says "before merge"; browser matrix, rendering parity, cold-load measurement, PWA/SW, low-end Android perf all missing |
| ⚠️ | 🟡Med | spec#AC-2 → `web/flutter_bootstrap.js` (minified bundle) | Engine loader bundle still contains `a&&(s=l(s,"chromium"))` branch; functionally dead (never executed when `canvasKitVariant:"full"`) but technically a `chromium` path string remains in the repo |
| ⚠️ | 🟢High | spec#AC-4 → `test/web/canvaskit_variant_test.dart` (doc comment) | Golden test suite cannot detect a CanvasKit variant swap (Flutter test VM skips CanvasKit entirely); AC-4 is vacuously satisfied — rendering parity relies entirely on the manual browser checks in #1281 |
| 💡 | 🟡Med | spec#AC-6,7 (issue comments) → `web/flutter_bootstrap.js:27` | Issue comment added AC-7 (`useLocalCanvasKit: true` as double-guard); PR chose not to adopt it — worth a brief note confirming `canvasKitBaseUrl: "./assets/"` alone is sufficient |

<details>

**Standards W-1 — CWD-relative test paths (🟢High)** *(flagged independently by both sub-reviewers)*

`test/web/canvaskit_variant_test.dart:27–40`

```dart
// line 27
expect(
  Directory('web/assets/chromium').existsSync(),   // bare relative path
  isFalse, ...
);
expect(File('web/assets/canvaskit.js').existsSync(), isTrue);   // line 31
expect(File('web/assets/canvaskit.wasm').existsSync(), isTrue); // line 32

final bootstrap = File('web/flutter_bootstrap.js').readAsStringSync(); // line 38
```

`dart:io` resolves these against `Directory.current` at runtime. `flutter test` launched from the project root works correctly. In a CI setup where the runner does `cd test && dart test`, all four paths are wrong — `Directory('web/assets/chromium').existsSync()` returns `false` because the path doesn't exist relative to CWD, not because it was deleted, so the first assertion passes silently for the wrong reason. Line 38 would throw `FileSystemException` instead, giving inconsistent behaviour.

**Fix:** Add a CWD guard at the top of the group, e.g.:
```dart
expect(
  File('pubspec.yaml').existsSync(),
  isTrue,
  reason: 'Run flutter test from the project root',
);
```
Or anchor paths via `Uri.base` / `Platform.script`.

---

**Spec Warning-1 — AC-3 verification records missing (🟢High)**

Issue #1281 AC-3 states: *"Verification items 1–5 recorded on the ticket before merge."* Current issue comments contain only discussion of `canvasKitBaseUrl` offline semantics and the #1282 savings analysis — none of the five verification results (browser matrix, rendering parity, cold-load, PWA/SW, low-end Android) are present.

This is a process gate, not a code defect, but it is an explicit merge prerequisite per the spec.

---

**Spec Warning-2 — Engine bundle chromium path string (🟡Med)**

`web/flutter_bootstrap.js` (minified Flutter loader, not hand-authored):
```javascript
let a = e && t.canvasKitVariant !== "full", s = r;
a && (s = l(s, "chromium"));   // ← path string still present
```
With `canvasKitVariant: "full"` set, `a` is always `false`; this branch is dead code in practice. The string comes from the vendored Flutter engine bundle, not from this PR, so it cannot be removed without regenerating `flutter.js`. AC-2 ("no reference to `assets/chromium` anywhere") is met at the application-configuration level; if AC-2 is interpreted to include the engine bundle, this is a known technical limitation of the approach.

---

**Spec Warning-3 — AC-4 golden suite vacuously satisfied (🟢High)**

The test file itself documents this (`canvaskit_variant_test.dart` doc comment):
> *A widget test cannot cover this. `canvasKitVariant` is read by the JS bootstrap in a real browser, and the golden suite runs under the Flutter test VM with no CanvasKit involved, so a variant swap is invisible to it.*

This means "golden suite green" proves no regressions occurred in the rest of the app, but it says nothing about rendering parity for the CanvasKit variant change. The manual Chrome desktop / offline verification in the PR description is the actual evidence for AC-4; it should be formally recorded on #1281 per AC-3.

</details>

<details>

- **Core mechanism is correct.** `canvasKitVariant: "full"` makes `a = e && ("full" !== "full") = false`, so the loader never appends `chromium/` regardless of browser capability. Verified by reading the minified loader logic end-to-end.
- **Deletion is complete.** `git show 949dd560` confirms `web/assets/chromium/canvaskit.js` and `web/assets/chromium/canvaskit.wasm` are absent; full-variant blobs (`canvaskit.js`, `canvaskit.wasm`) are unchanged and present.
- **Atomicity of the change.** Both halves (deletion + config pin) land in one commit, exactly as required by the spec. The `import()` has no 404 fallback, so this is a genuine correctness requirement, not just style.
- **Guard test design is sound.** The test correctly identifies the two invariants that must hold forever, provides meaningful failure messages pointing back to #1281, and acknowledges the golden-suite limitation in its doc comment. The `canvasKitBaseUrl: "./assets/"` assertion guards offline-critical behaviour that would otherwise be invisible to any automated test.
- **Offline verification is properly evidenced.** The PR description confirms an offline WAN-down cold load was tested adversarially (CanvasKit loaded locally, not from gstatic.com), which is the real correctness proof for the local-resolution path.
- **No security concerns.** No hardcoded credentials, no injection surface, no changes to authentication or JNAP/USP handling. TrustedTypes policy is unchanged.

</details>

*Cross-reviewed by two independent agents (security+correctness / architecture+maintainability) + spec conformance agent. Automated — please sanity-check before merge.*

PeterJhongLinksys (approved):
🤖 **Automated Review — Round 1** · `c675111..949dd56` (incremental)

**Verdict: ✅ APPROVE** — the core claim (drop the `chromium/` variant + pin `full` + keep a local base URL) was independently verified as self-consistent in the head-revision loader; all remaining findings concern the robustness of the new guard test itself, none of them merge-blocking.

|---|---|---|---|
| ⚠️ | 🟢 | `test/web/canvaskit_variant_test.dart:27,34,35,40` | All filesystem checks use CWD-relative paths; under a wrong CWD the `chromium/` assertion passes silently (false negative) while the others fail — the signal is not trustworthy. |
| ⚠️ | 🟢 | `test/web/canvaskit_variant_test.dart:44,52` | [Fragile Test] `contains()` matches raw formatting of the whole file, so a quote/spacing change red-lights it, while a comment mentioning the same keys could green-light it. |
| ⚠️ | 🟡 | `web/flutter_bootstrap.js:1,6` | [Divergent Change] The file mixes a generated minified bundle + `buildConfig` with hand-written `config{}`; the test cannot detect a partial regeneration that keeps `config` but bumps `engineRevision`. |
| ⚠️ | 🟡 | `test/web/canvaskit_variant_test.dart:52` | Guard asserts the literal line exists, not the property that matters ("never fetches CanvasKit off-box") — a duplicate `canvasKitBaseUrl` later in `config` would win in JS and still pass. |
| ⚠️ | 🟡 | `test/web/canvaskit_variant_test.dart:17,29` | [Magic Number] "1.71 MB" has no stated basis; the deleted `canvaskit.wasm` is 5,760,502 bytes (~5.49 MiB) on disk, so future readers will not be able to reconcile it. |
| 💡 | 🟢 | `test/web/canvaskit_variant_test.dart:25-35` | Three assertions in one `test`; the first failure hides whether the `full` variant is still present. |
| 💡 | 🟡 | `test/web/canvaskit_variant_test.dart:24` | Group name identifies itself by ticket number (`#1281`) rather than by behaviour; the number ages out of meaning. |
| 💡 | ⚪ | `test/web/canvaskit_variant_test.dart:1` | `import 'dart:io'` makes the file uncompilable under `--platform chrome`; `@TestOn('vm')` would state that explicitly. |
| 💡 | ⚪ | `web/assets/` | No `README.md` recording where the vendored `canvaskit.*` came from; that knowledge currently lives only in a test doc comment. |

<sub>Confidence: 🟢 High · 🟡 Medium · ⚪ Low. `[X]` = raised by 1 agent, `[XX]` = raised by both (Critical rows only; there are none this round).</sub>
<sub>Note: an earlier concern that no CI entry point runs `test/web/` was **checked and dismissed** — `run_tests.sh:92` runs `flutter test --exclude-tags="golden||loc||ui"`, `dart_test.yaml` declares only those three tags, and `.github/workflows/ci.yml:84` calls `sh run_tests.sh`, so this file does run on every PR.</sub>

<details>
<summary>⚠️ Warning details (5)</summary>

**W1 — CWD-relative filesystem assertions can pass for the wrong reason** · 🟢 High
`test/web/canvaskit_variant_test.dart:27,34,35,40`
```dart
Directory('web/assets/chromium').existsSync()        // :27  isFalse
File('web/assets/canvaskit.js').existsSync()         // :34  isTrue
File('web/flutter_bootstrap.js').readAsStringSync()  // :40
```
Why it matters: all four paths assume `CWD == package root`. `flutter test` from the repo root satisfies that, but from a subdirectory or an IDE with a different working directory, `web/assets/chromium` simply does not exist — so the assertion this PR exists to protect turns **green for the wrong reason**, while `:34/:35/:40` fail with a misleading message. Half false-pass, half false-fail. This is the only guard the PR ships, so its semantics must not be rewritable by the working directory.
Fix: add a CWD sanity precondition, e.g. `expect(File('pubspec.yaml').existsSync(), isTrue, reason: 'run from repo root')`, or resolve the package root explicitly from `Platform.script`.

**W2 — `contains()` on the whole file is fragile in both directions** · 🟢 High
`test/web/canvaskit_variant_test.dart:44,52`
```dart
contains('canvasKitVariant: "full"')
contains('canvasKitBaseUrl: "./assets/"')
```
Why it matters: this matches raw layout, including spacing and double quotes. Harmless rewrites (single quotes, `canvasKitVariant:"full"`, a line break) turn an "offline-critical" guard red, which trains the next person to delete the test. The reverse is worse: `bootstrap` is the **entire file**, and `web/flutter_bootstrap.js:11-16,20-23` already carries comments discussing exactly these two keys. Today's wording happens not to match the asserted string, but that is coincidence, not design — reword a comment to `// keep canvasKitVariant: "full"` and the test passes even with the real `config` entry deleted.
Fix: strip comments before matching (`replaceAll(RegExp(r'//.*'), '')`), anchor a RegExp inside the `config:` block, and additionally assert no *other* assignment of `canvasKitVariant` exists.
</details>

<details>
<summary>⚠️ Warning details (cont.)</summary>

**W3 — `flutter_bootstrap.js` is both a generated artifact and hand-maintained config** · 🟡 Medium
`web/flutter_bootstrap.js:1` (single minified `flutter.js` bundle, ending `//# sourceMappingURL=flutter.js.map`), `:6` (`_flutter.buildConfig = {"engineRevision":"cf56914b...", ...}`), hand-written `config{}` from `:10`.
Why it matters: the test doc comment names this risk itself ("reversible by an unrelated Flutter upgrade"). If someone re-runs `flutter build web` and pastes a fresh bootstrap, the config lines vanish together with the comments and the test goes red — good. But the failure mode that stays silent is the *partial* one: regenerate the bundle only (new `engineRevision`, `config{}` preserved) and the test is fully green while the vendored `web/assets/canvaskit.*` may no longer match the new engine. `.fvmrc` pins `{"flutter":"3.44.0"}`, so that pairing is real.
Fix: move the hand-written `config{}` into a file codegen cannot overwrite (e.g. `web/flutter_bootstrap_config.js`, or pass it via `_flutter.loader.load` from `index.html`), or additionally assert the `engineRevision` ↔ `.fvmrc` relationship, so "engine upgraded" and "shipping decision changed" get distinct failure signals.

**W4 — the guard asserts the literal, not the property that matters** · 🟡 Medium
`web/flutter_bootstrap.js:6,20-27`, `test/web/canvaskit_variant_test.dart:52`
Why it matters: the real invariant is "CanvasKit is never fetched off-box". The test only asserts one literal line is present. `buildConfig` (`:6`) is generated and can be overwritten by a build, and the loader's base resolution takes `config.canvasKitBaseUrl` first — so if `config` ever ends up with a second `canvasKitBaseUrl`, JS uses the later one and this test still passes.
Fix: assert `canvasKitBaseUrl` appears exactly once in the `config` block **and** that its value does not start with `http`.

**W5 — "1.71 MB" has no stated measurement basis** · 🟡 Medium
`test/web/canvaskit_variant_test.dart:17` ("leaves 1.71 MB of unreachable payload"), `:29-31` (`reason: '...to save 1.71 MB of rootfs'`)
Why it matters: `git diff --stat c675111..949dd56` shows the deleted `web/assets/chromium/canvaskit.wasm` at **5,760,502 bytes (~5.49 MiB)** plus 193 lines of `canvaskit.js`. 1.71 MB is presumably the compressed / squashfs-rootfs delta, but nothing in the file says so, so a future reviewer weighing "was deleting this worth it" reaches for a number that cannot be reconciled with `--stat`.
Fix: state the basis (e.g. `1.71 MB compressed in the squashfs rootfs; 5.49 MiB on disk`) or drop the figure.
</details>

<details>

**Both agents independently traced the minified loader and confirmed the change is self-consistent.**
`web/flutter_bootstrap.js:1` — variant selection:
```js
let e = i.hasChromiumBreakIterators && i.hasImageCodecs;
if (!e && t.canvasKitVariant == "chromium") throw "Chromium CanvasKit variant specifically requested, but unsupported in this browser";
let a = e && t.canvasKitVariant !== "full", s = r;
a && (s = l(s, "chromium"));
```
and base-URL resolution:
```js
function T(n,t){ return n.canvasKitBaseUrl ? n.canvasKitBaseUrl
  : (t.engineRevision && !t.useLocalCanvasKit)
    ? C("https://www.gstatic.com/flutter-canvaskit", t.engineRevision) : "canvaskit" }
```
With `canvasKitBaseUrl: "./assets/"` and `canvasKitVariant: "full"`, `a` is always `false` → `chromium/` is never appended, the `throw` path is unreachable (it fires only on an explicit `"chromium"` request), and `T()` takes the first branch so the gstatic CDN fallback is never reached. Requests land on `./assets/canvaskit.js` + `./assets/canvaskit.wasm`, both of which exist in the head tree. A repo-wide search for `chromium/canvaskit` finds only comment/message text, and `web/service_worker.js` contains no canvaskit reference — so the deletion leaves no dead precache entry. **No white screen, no 404, no network regression.**

- **The `OFFLINE-CRITICAL — do not remove or change` comment (`web/flutter_bootstrap.js:11-16`) is the best comment in this round**: it explains *why* the line cannot be deleted (no `useLocalCanvasKit` in `buildConfig` → fallback to gstatic.com) and names the test that guards it. Verified accurate against the loader.
- **Choosing `"full"` over `"chromium"` is the safe direction** — the loader has a hard-throw path for `"chromium"` on unsupported browsers; `"full"` is safe on all of them.
- **The blind spot is stated honestly** (`test/web/canvaskit_variant_test.dart:53-57`): "passes every online test and white-screens only on a router with no WAN" is exactly this change's real risk, not glossed over.
- **The reasoning that both halves must ship together is correct**, and the doc comment (`:20-22`) correctly explains why widget/golden tests cannot cover it — pre-empting a misplaced coverage request.
- **Architecturally out of scope in the best way**: nothing under `lib/` is touched, so the three-layer dependency rules, USP Provider rules, `ServiceError` isolation, stateless-Service and UI-Kit-first checks are all inapplicable. No unrelated lint drive-by; scope discipline respected.
- **Security: not applicable** — no secrets/tokens, no new network calls (it removes a potential CDN dependency), no USP/JNAP error-handling changes.
</details>

---

**Suggestions** (non-blocking): split the three assertions at `test/web/canvaskit_variant_test.dart:25-35` into separate tests so a failure says which fact broke; rename the group away from `#1281` to describe the behaviour (keep the ticket in the doc comment); add `@TestOn('vm')` given `dart:io`; add a `web/assets/README.md` recording the vendored assets' origin (engine `build/web/canvaskit/`, matching `engineRevision`, and why only the full variant is kept); consider asserting that `canvasKitBaseUrl` occurs nowhere else in the repo, since `web/index.html:104` is currently the sole loader entry point but nothing enforces that.

Issue #1246, opened 2026-08-12
Add stable Semantics identifiers to the 3 Port Forwarding tabs (E2E clicks them by localized label; follow-on to #1172)


AustinChangLinksys:
## Summary

The three Port Forwarding tabs carry no `Semantics identifier`, so the E2E suite has to click them
by their **localized tab label**. Six real click sites depend on that today.

This is the same class of implementation debt as **#1172** (stable identifiers on E2E-critical
Port Forwarding controls) — but #1172 is closed with all five phases shipped, and it covered the
add/edit/delete *controls inside* the tabs, never the tab strip itself. Filing separately rather
than reopening a completed tracker.

## Current state on `origin/dev-2.7.0` (`d13b9577`)

```
lib/page/port_forwarding/views/usp_port_forwarding_detail_view.dart:77-80

        tabs: [
          Tab(text: loc(context).singlePortForwarding),
          Tab(text: loc(context).portRangeForwarding),
          Tab(text: loc(context).portTriggering),
        ],
```

`grep -c "identifier:" usp_port_forwarding_detail_view.dart` → **0** for the whole file.

## E2E impact

`tests/P04a-port-forwarding-rendering.spec.ts` — six tab locators are declared and clicked on the
next line:

| Locator declared | Clicked | Label matched |
|---|---|---|
| `:82` | `:83` | `/Port Range Forwarding/` |
| `:101` | `:102` | `/Triggering/` |
| `:155` | `:156` | `/Port Range Forwarding/` |
| `:170` | `:171` | `/Port Range Forwarding/` |
| `:189` | `:190` | `/Triggering/` |
| `:204` | `:205` | `/Triggering/` |

(Five further declarations — `:53`, `:54`, `:55`, `:74`, `:92` — are count assertions only. Those
are fine as text lookups: asserting that localized copy renders is the assertion's purpose. Only
the click sites are the problem.)

The regexes are also load-bearing in a fragile way. `/Port Range Forwarding/` cannot be shortened
to `/Port Range/` because that also matches the **Port Range Triggering** tab — a comment at
`:89-90` records this trap. So the tests are pinned to the full localized strings of two
similar-sounding tabs, and an ARB copy tweak on either becomes a count-0 timeout (or worse, a
silent wrong-tab click) rather than a compile error.

## Ask

Add an `identifier:` to each of the three tabs — e.g. `port-forwarding-tab-single`,
[...]

Issue #1254, opened 2026-08-12
Topology node-detail popup: add stable Semantics identifier to the Details button (E2E locates it by localized label; split from #1208)


AustinChangLinksys:
## Summary

The topology **node-detail popup's `Details` button** carries no `Semantics identifier`, so the E2E
suite has to locate it by its **localized label**. This is the residual of #1208 — the node-level
identifiers that issue asked for shipped in #1211 (+ privacyGUI-UI-kit#9) and were consumed by
PrivacyGUI-USP-E2E#26, but the action button *inside* the popup those nodes open was skipped.

Filing separately so #1208 can close on its original (completed) node-identifier scope, rather than
staying open to track a different widget. Same class of implementation debt as #1246 (Port
Forwarding tabs) but a different area/file/widget.

## Current state on `origin/dev-2.7.0` (`d13b9577`)

```
lib/page/topology/views/components/node_detail_popup.dart:103-105

    child: AppButton.text(
      label: loc(context).details,
      onTap: onDetailsTap,
    ),
```

`grep -c "identifier:" node_detail_popup.dart` → **0** for the whole file.

## Why it matters

The Details button is the entry point to the node-detail page, so it is on the critical path of the
D5a/D5b interaction chains (node → popup → Details → node detail). With no hook, E2E locates it by
its localized label:

```
tests/P15-topology.spec.ts:182   page.getByRole('button', { name: /Details/i }).first()
tests/P15-topology.spec.ts:221   (same)
```

That couples the test to ARB copy — an `app_en.arb` `details` rename turns into a count-0 timeout
rather than a compile error, which is exactly the failure mode #1211 removed one layer up. The
inconsistency is the awkward part: the nodes are hook-located, then the very next tap in the same
chain falls back to translatable text.

Unlike #1246 (which may need a `Tab` passthrough in `ui_kit_library` first), this is a plain
`AppButton.text` and should be able to carry an `identifier:` directly — no known UI-kit blocker.

## Ask

Add `identifier: 'topology-node-detail-button'` (or whatever fits the established naming) to that
`AppButton.text`. Framing this as implementation completeness rather than an E2E feature request:
sibling controls in the same interaction chain already carry hooks, and this one node was skipped.

Once it lands, the E2E side is a two-line swap in `P15-topology.spec.ts` (plus `gen:ids` picking up
the new constant) — tracked in linksys/PrivacyGUI-USP-E2E#44 as verdict 2 (app-hook-blocked).

## Context

[...]

PR #1287, opened 2026-08-15
feat(port-forwarding): stable Semantics identifiers on the 3 tabs (#1246)

AustinChangLinksys:
## Summary

Adds stable `Semantics identifier`s to the three Port Forwarding tabs so the E2E
suite can target them by ID instead of by localized label. Follow-on to **#1172**,
which hooked the controls *inside* the tabs but never the tab strip itself.

Refs #1246 · triaged as verdict 2 (app-hook-blocked) in `linksys/PrivacyGUI-USP-E2E#44`.

## Root cause

`lib/page/port_forwarding/views/usp_port_forwarding_detail_view.dart:77-80` built the
tabs with `Tab(text: loc(context).…)`. Material's `Tab(text:)` exposes no identifier,
and the tabs flow through `UiKitPageView` → `AppPageView` → a Material `TabBar`, which
renders the passed `Tab` widgets verbatim. So the only stable hook available was the
localized label — six click sites in `P04a-port-forwarding-rendering.spec.ts` depended
on it, pinned to the full strings of two similar tabs (`Port Range Forwarding` vs
`Port Range Triggering`).

## Change

- **`lib/page/port_forwarding/views/usp_port_forwarding_detail_view.dart`** — swap each
  `Tab(text:)` for `Tab(child: Semantics(identifier: …, child: Text(…)))`, exposing
  `port-forwarding-tab-single` / `port-forwarding-tab-range` /
  `port-forwarding-tab-triggering`.
  - No ui_kit change needed: the label flows straight into a Material `TabBar`, and
    `Tab`'s own docs say `child` is "usually a Text widget, possibly wrapped in a
    Semantics widget". The `Text` replicates what `Tab(text:)` renders internally
    (`softWrap: false, overflow: TextOverflow.fade`) → **zero visual change**; the
    Semantics node adds only an invisible `flt-semantics-identifier` hook.
- **`test/page/port_forwarding/usp_port_forwarding_tab_identifier_widget_test.dart`** —
  new `@Tags(['ui'])` white-box test (mirrors the #1218 Advanced Settings identifier
  test): each tab is locatable via `find.bySemanticsIdentifier`, the three hooks resolve
  to distinct nodes, and the localized labels still render.

## Verification (all real output, no fabrication)

```
$ flutter analyze lib/page/port_forwarding/views/usp_port_forwarding_detail_view.dart \
                  test/page/port_forwarding/usp_port_forwarding_tab_identifier_widget_test.dart
No issues found! (ran in 13.1s)

$ flutter test test/page/port_forwarding
00:07 +62: All tests passed!        # existing 59 + 3 new

$ dart format --output=none --set-exit-if-changed <both files>
Formatted 2 files (0 changed) in 0.04 seconds.   # 2.x CI format gate — clean
```

No golden baselines are committed for this view (2.x generates/compares in CI), and the
change is Semantics-only, so it cannot alter rendered pixels.

## Notes for the E2E follow-up

Once this lands, the E2E side is the mechanical swap described in the issue: replace the
six label locators with `byId()` and let `gen:ids` pick up the three new constants.

### DIFF
--- a/lib/page/port_forwarding/views/usp_port_forwarding_detail_view.dart
+++ b/lib/page/port_forwarding/views/usp_port_forwarding_detail_view.dart
@@ -75,9 +75,47 @@ class _UspPortForwardingDetailViewState
             .fetch(forceRemote: true),
         bottomBar: _buildBottomBar(context, ref, pageState),
         tabs: [
-          Tab(text: loc(context).singlePortForwarding),
-          Tab(text: loc(context).portRangeForwarding),
-          Tab(text: loc(context).portTriggering),
+          // Stable, screen-reader-silent E2E hooks (→ `flt-semantics-identifier`)
+          // for the three Port Forwarding tabs. Follow-on to #1172, which added
+          // identifiers to the controls *inside* the tabs but not the tab strip
+          // itself; the E2E suite (PrivacyGUI-USP-E2E#44) still clicks these tabs
+          // by their localized label. `Tab(text:)` exposes no identifier, so we
+          // move to `Tab(child:)` and wrap the label in a Semantics node.
+          //
+          // The Text below intentionally mirrors what `Tab(text:)` renders
+          // internally (`Text(text, softWrap: false, overflow: TextOverflow.fade)`)
+          // so this is a zero-visual-change swap — the Semantics node adds only
+          // an invisible test hook.
+          Tab(
+            child: Semantics(
+              identifier: 'port-forwarding-tab-single',
+              child: Text(
+                loc(context).singlePortForwarding,
+                softWrap: false,
+                overflow: TextOverflow.fade,
+              ),
+            ),
+          ),
+          Tab(
+            child: Semantics(
+              identifier: 'port-forwarding-tab-range',
+              child: Text(
+                loc(context).portRangeForwarding,
+                softWrap: false,
+                overflow: TextOverflow.fade,
+              ),
+            ),
+          ),
+          Tab(
+            child: Semantics(
+              identifier: 'port-forwarding-tab-triggering',
+              child: Text(
+                loc(context).portTriggering,
+                softWrap: false,
+                overflow: TextOverflow.fade,
+              ),
+            ),
+          ),
         ],
         tabContentViews: [
           _buildTabContent(
--- a/test/page/port_forwarding/usp_port_forwarding_tab_identifier_widget_test.dart
+++ b/test/page/port_forwarding/usp_port_forwarding_tab_identifier_widget_test.dart
@@ -1,0 +1,164 @@
+@Tags(['ui'])
+library;
+
+import 'package:flutter/material.dart';
+import 'package:flutter/services.dart';
+import 'package:flutter_riverpod/flutter_riverpod.dart';
+import 'package:flutter_test/flutter_test.dart';
+import 'package:go_router/go_router.dart';
+import 'package:privacy_gui/l10n/gen/app_localizations.dart';
+import 'package:privacy_gui/page/port_forwarding/views/usp_port_forwarding_detail_view.dart';
+import 'package:privacy_gui/route/route_model.dart';
+import 'package:privacy_gui/theme/theme_json_config.dart';
+
+import '../../golden_test/golden_framework/mocks/mock_common.dart';
+import '../../golden_test/golden_framework/mocks/mock_port_forwarding.dart';
+import '../../golden_test/page/port_forwarding/fixtures/port_forwarding_test_data.dart';
+
+/// Verifies that the three Port Forwarding tabs carry the stable E2E
+/// identifiers added in issue #1246 and that each is locatable via
+/// [CommonFinders.bySemanticsIdentifier].
+///
+/// This is the fast local proxy for the E2E `byId()` contract (constitution
+/// Article XVI): once the identifier reaches the Semantics tree, the CanvasKit
+/// → `flt-semantics-identifier` DOM projection is a Flutter-runtime guarantee,
+/// so no web round is needed here. It mirrors the #1218 Advanced Settings
+/// identifier widget test that covers the sibling entry cards.
+///
+/// Before this change the tab strip carried no identifier, so the E2E suite
+/// (PrivacyGUI-USP-E2E#44) had to click each tab by its localized label — a
+/// follow-on to #1172, which only hooked the controls *inside* the tabs.
+void main() {
+  // Fixed E2E hooks assigned to the three tabs, in tab-strip order.
+  const expectedIdentifiers = <String>[
+    'port-forwarding-tab-single',
+    'port-forwarding-tab-range',
+    'port-forwarding-tab-triggering',
+  ];
+
+  setUpAll(() {
+    // UspTopBar reads package_info during build; stub the platform channel.
+    TestDefaultBinaryMessengerBinding.instance.defaultBinaryMessenger
+        .setMockMethodCallHandler(
+      const MethodChannel('dev.fluttercommunity.plus/package_info'),
+      (MethodCall methodCall) async {
+        if (methodCall.method == 'getAll') {
+          return <String, dynamic>{
+            'appName': 'PrivacyGUI',
+            'packageName': 'com.linksys.privacygui',
+            'version': '0.0.0',
+            'buildNumber': '0',
+          };
+        }
+        return null;
+      },
+    );
+  });
+
+  // Mirrors the golden framework's shell: ProviderScope + MaterialApp.router so
+  // loc(context) and navigation resolve inside the view under test. The page
+  // notifier is pinned to a fixed data state so the tab strip renders.
+  Widget wrap() {
+    final themeConfig = ThemeJsonConfig.defaultConfig();
+    final router = GoRouter(
+      initialLocation: '/',
+      routes: [
+        LinksysRoute(
+          path: '/',
+          name: 'test_root',
+          builder: (context, state) => const UspPortForwardingDetailView(),
+        ),
+      ],
+    );
+    return ProviderScope(
+      overrides: [
+        ...commonOverrides(),
+        ...portForwardingOverrides(dataState()),
+      ],
+      child: MaterialApp.router(
+        locale: const Locale('en'),
+        localizationsDelegates: AppLocalizations.localizationsDelegates,
+        supportedLocales: AppLocalizations.supportedLocales,
+        theme: themeConfig.createLightTheme(),
+        routerConfig: router,
+      ),
+    );
+  }
+
+  group('Port Forwarding tab identifiers', () {
+    testWidgets('every tab is locatable by its identifier', (tester) async {
+      final handle = tester.ensureSemantics();
+      tester.view.physicalSize = const Size(1280, 900);
+      tester.view.devicePixelRatio = 1.0;
+      addTearDown(tester.view.resetPhysicalSize);
+      addTearDown(tester.view.resetDevicePixelRatio);
+
+      await tester.pumpWidget(wrap());
+      await tester.pumpAndSettle();
+
+      // A Material TabBar builds every Tab widget up front (not just the
+      // selected one), so all three Semantics identifiers are present without
+      // needing to switch tabs.
+      for (final id in expectedIdentifiers) {
+        expect(
+          find.bySemanticsIdentifier(id),
+          findsOneWidget,
+          reason: 'tab "$id" must be locatable by identifier',
+        );
+      }
+
+      handle.dispose();
+    });
+
+    testWidgets('each identifier resolves to exactly one distinct tab node',
+        (tester) async {
+      final handle = tester.ensureSemantics();
+      tester.view.physicalSize = const Size(1280, 900);
+      tester.view.devicePixelRatio = 1.0;
+      addTearDown(tester.view.resetPhysicalSize);
+      addTearDown(tester.view.resetDevicePixelRatio);
+
+      await tester.pumpWidget(wrap());
+      await tester.pumpAndSettle();
+
+      // Each hook must be unique — the whole point is that E2E can target one
+      // specific tab without the "Port Range Forwarding" / "Port Range
+      // Triggering" label ambiguity. Assert one match per id and that the
+      // three matched elements are all distinct widgets.
+      final matched = <Element>{};
+      for (final id in expectedIdentifiers) {
+        final finder = find.bySemanticsIdentifier(id);
+        expect(finder, findsOneWidget,
+            reason: 'tab "$id" must resolve to exactly one node');
+        matched.add(finder.evaluate().single);
+      }
+      expect(matched, hasLength(expectedIdentifiers.length),
+          reason: 'the three tab identifiers must target distinct widgets');
+
+      handle.dispose();
+    });
+
+    testWidgets('the localized tab label still renders alongside the hook',
+        (tester) async {
+      final handle = tester.ensureSemantics();
+      tester.view.physicalSize = const Size(1280, 900);
+      tester.view.devicePixelRatio = 1.0;
+      addTearDown(tester.view.resetPhysicalSize);
+      addTearDown(tester.view.resetDevicePixelRatio);
+
+      await tester.pumpWidget(wrap());
+      await tester.pumpAndSettle();
+
+      // Guards against the identifier swap accidentally dropping the visible
+      // label. The three English localized strings must still be on screen.
+      // NB: `portTriggering` localizes to "Port Range Triggering" (not
+      // "Port Triggering") — the near-collision with "Port Range Forwarding"
+      // is precisely the fragility the E2E identifiers are meant to retire.
+      expect(find.text('Single Port Forwarding'), findsOneWidget);
+      expect(find.text('Port Range Forwarding'), findsOneWidget);
+      expect(find.text('Port Range Triggering'), findsOneWidget);
+
+      handle.dispose();
+    });
+  });
+}


AustinChangLinksys (commented):
🤖 **Automated Review — Round 1** · `8a3aaed..7c410e1` (full)

**Verdict: ✅ APPROVE** — 0 Critical, 0 Warnings across both axes. Clean, well-scoped change with correct Semantics identifier usage and thorough widget test coverage. Safe to merge.

> ⚙️ *Posted as `COMMENT` rather than `APPROVE` — GitHub's API prevents self-approval even for dispatch-chain PRs authored under the same account. Verdict is APPROVE (0 Critical); a second reviewer or Austin can formally approve.*

#### Standards

|---|---|---|---|
| 💡 | 🟡Med | `usp_port_forwarding_detail_view.dart:90–116` | `port-forwarding-tab-` prefix differs from module-wide `pf-` shorthand (e.g. `pf-add-single-port`) — consider documenting the intentional distinction |
| 💡 | 🟢High | `usp_port_forwarding_detail_view.dart:89–118` | [Duplicated Code] Three `Tab(child: Semantics(..., child: Text(..., softWrap: false, overflow: TextOverflow.fade)))` structures are mechanically identical; low urgency at 3 sites, but a private `_labeledTab(String id, String label)` helper would be worth extracting if a 4th tab lands |
| 💡 | 🟡Med | `usp_port_forwarding_detail_view.dart:85–87` | `Tab(text:)` internal `Text` attribute equivalence noted in a comment but has no mechanical compile-time guard against SDK drift; existing golden suite provides the safety net, but noting it explicitly |
| 💡 | 🟢High | `usp_port_forwarding_tab_identifier_widget_test.dart:61–86` | [Duplicated Code] `wrap()` scaffold (ThemeJsonConfig + GoRouter + ProviderScope + MaterialApp.router) is near-identical to `usp_advanced_settings_identifier_widget_test.dart`; worth a shared `buildIdentifierTestShell()` factory if more identifier tests follow |
| 💡 | 🟡Med | `usp_port_forwarding_tab_identifier_widget_test.dart:39` | `setUpAll` mock handler has no paired `tearDownAll` clear; consistent with sibling test (`usp_advanced_settings_identifier_widget_test.dart`) so no isolation issue in current setup, but worth cleaning up the pattern |

#### Spec

|---|---|---|---|
| ✅ | 🟢High | `usp_port_forwarding_detail_view.dart:90` | spec#Ask → `port-forwarding-tab-single` present, exact match |
| ✅ | 🟢High | `usp_port_forwarding_detail_view.dart:99` | spec#Ask → `port-forwarding-tab-range` present, exact match |
| ✅ | 🟢High | `usp_port_forwarding_detail_view.dart:108` | spec#Ask → `port-forwarding-tab-triggering` present, exact match |
| ✅ | 🟢High | `usp_port_forwarding_detail_view.dart:90–118` | spec#Change → `Tab(text:)` fully replaced by `Tab(child: Semantics(...))` on all 3 tabs |
| ✅ | 🟢High | `usp_port_forwarding_tab_identifier_widget_test.dart:133–154` | spec#zero-visual-change → localized labels still render; `softWrap`/`overflow` preserved |
| ✅ | 🟢High | `lib/components/ui_kit_page_view.dart` + `app_page_view.dart:568–579` | spec#ui_kit-dependency → `Tab` objects passed through verbatim to Flutter's `TabBar`; no passthrough needed |
| ✅ | 🟢High | `usp_port_forwarding_tab_identifier_widget_test.dart:79–154` | spec#proxy-test → 3 test cases: locatability, uniqueness, label-still-renders |
| 💡 | 🟡Med | `usp_port_forwarding_tab_identifier_widget_test.dart:148–151` | spec note: `portTriggering` locale key renders "Port Range Triggering" (not "Port Triggering"); identifier string `port-forwarding-tab-triggering` is correct per spec; worth confirming E2E fixture uses the actual rendered label when verifying display, not the spec shorthand |

<details>

- **Correct Semantics API**: All three tabs use `Semantics(identifier:)` (not `label:`), per Constitution Article XVI §16.3 which specifies `identifier` as the hook visible to the E2E generator.
- **Zero visual change verified**: `Tab(child: Semantics(..., child: Text(text, softWrap: false, overflow: TextOverflow.fade)))` exactly mirrors `Tab(text:)`'s internal `_buildLabelText()` implementation (Flutter SDK 3.44 `tabs.dart:188-189`), confirmed by reading SDK source.
- **`TabController` length unchanged**: `TabController(length: 3, vsync: this)` in `initState` is untouched; no off-by-one risk.
- **No JNAP/USP paths touched**: Change is purely accessibility-tree instrumentation with zero server-side surface.
- **ui_kit passthrough confirmed**: `ui_kit_page_view.dart._convertTabs()` passes `Tab` objects verbatim → `app_page_view.dart` feeds them directly to Flutter's `TabBar` — the `Semantics` node is preserved end-to-end with no ui_kit modification needed.
- **Widget tests: 3/3 pass** (`flutter test test/page/port_forwarding --tags ui`, 6 s). Tests cover the three required properties: locatability by identifier, identifier uniqueness (distinct nodes), and label still renders.
- **`dart analyze` clean**: Both changed files: No issues found.
- **Naming follows kebab-case convention** established by other identifier PRs (#1172, #1218, #1254).
- **PR description is thorough**: Root cause, change description, verification output, and E2E follow-up plan all documented.

</details>

*Cross-reviewed by three independent agents (security+correctness / architecture+maintainability / spec compliance #1246). Automated — please sanity-check before merge.*

AustinChangLinksys (commented):
🤖 **Automated Review — Round 1** · `8a3aaed..7c410e1` (full)

**Verdict: ✅ APPROVE** — 0 Critical, 0 Warnings across both axes. Clean, well-scoped Semantics hook change with correct API usage and thorough widget test coverage.

> ⚙️ *Posted as `COMMENT` rather than `APPROVE` — GitHub's API prevents self-approval even for dispatch-chain PRs authored under the same account. Verdict is APPROVE (0 Critical, 0 Warning); a second reviewer or Austin can formally approve.*

#### Standards

|---|---|---|---|
| 💡 | 🟡Med | `usp_port_forwarding_detail_view.dart:90–116` | `port-forwarding-tab-` prefix differs from module-wide `pf-` shorthand used elsewhere — consider documenting the intentional distinction |
| 💡 | 🟢High | `usp_port_forwarding_detail_view.dart:89–118` | [Duplicated Code] Three `Tab(child: Semantics(..., child: Text(...)))` structures are mechanically identical; low urgency at 3 sites, but a private `_labeledTab(String id, String label)` helper would be worth extracting if a 4th tab lands |
| 💡 | 🟡Med | `usp_port_forwarding_detail_view.dart:85–87` | `Tab(text:)` internal `Text` attribute equivalence noted in a comment but has no mechanical compile-time guard against SDK drift; existing widget test suite provides the safety net |
| 💡 | 🟢High | `usp_port_forwarding_tab_identifier_widget_test.dart:61–86` | [Duplicated Code] `wrap()` scaffold (ThemeJsonConfig + GoRouter + ProviderScope + MaterialApp.router) is near-identical to `usp_advanced_settings_identifier_widget_test.dart`; worth a shared `buildIdentifierTestShell()` factory if more identifier tests follow |
| 💡 | 🟡Med | `usp_port_forwarding_tab_identifier_widget_test.dart:39` | `setUpAll` mock handler (`setMockMethodCallHandler`) is deprecated in Flutter 3.13+; consistent with sibling tests but worth cleaning up; no isolation issue in current setup |

#### Spec

|---|---|---|---|
| ✅ | 🟢High | `usp_port_forwarding_detail_view.dart:91` | spec#Ask → `port-forwarding-tab-single` present, exact match |
| ✅ | 🟢High | `usp_port_forwarding_detail_view.dart:101` | spec#Ask → `port-forwarding-tab-range` present, exact match |
| ✅ | 🟢High | `usp_port_forwarding_detail_view.dart:111` | spec#Ask → `port-forwarding-tab-triggering` present, exact match |
| ✅ | 🟢High | `usp_port_forwarding_detail_view.dart:89–118` | spec#Change → `Tab(text:)` fully replaced by `Tab(child: Semantics(...))` on all 3 tabs |
| ✅ | 🟢High | `usp_port_forwarding_tab_identifier_widget_test.dart:113–139` | spec#proxy-test → 3 test cases: locatability, uniqueness, label-still-renders |
| 💡 | 🟡Med | `usp_port_forwarding_tab_identifier_widget_test.dart:148–155` | spec note: `portTriggering` locale key renders "Port Range Triggering" (not "Port Triggering"); identifier string `port-forwarding-tab-triggering` is correct per spec (matches `e.g.` exactly); test comment documents this intentionally |

<details>

- **Correct Semantics API**: All three tabs use `Semantics(identifier:)` (not `label:`), the correct hook for E2E targeting.
- **Zero visual change verified**: `Tab(child: Semantics(..., child: Text(text, softWrap: false, overflow: TextOverflow.fade)))` exactly mirrors `Tab(text:)`'s internal `_buildLabelText()` implementation, confirmed by reading Flutter SDK source.
- **`TabController` length unchanged**: `TabController(length: 3, vsync: this)` in `initState` is untouched; no off-by-one risk.
- **No JNAP/USP paths touched**: Change is purely accessibility-tree instrumentation with zero server-side surface.
- **ui_kit passthrough confirmed**: `Tab` objects passed through verbatim to Flutter's `TabBar`; no ui_kit modification needed.
- **Widget tests: 3/3 cases** covering locatability by identifier, identifier uniqueness (distinct nodes), and label still renders.
- **`dart analyze` clean**: Both changed files: No issues found.
- **Naming follows kebab-case convention** established by other identifier PRs (#1172, #1218, #1254).
- **`context.mounted` guards correct**: All async `_onSave` branches guard correctly (pre-existing, not modified by this PR).
- **No scope creep**: Only tab strip definition changed; all business logic, save handler, and dialog code untouched.

</details>

*Cross-reviewed by three independent agents (security+correctness / architecture+maintainability / spec compliance #1246). Automated — please sanity-check before merge.*

PeterJhongLinksys (approved):
🤖 **Automated Review — Round 1** · `8a3aaed..7c410e1` (full)

**Verdict: ✅ APPROVE** — Correct, narrowly-scoped, well-tested E2E-hook change with no Critical findings; the two items worth deciding before/after merge are whether the identifier lands on the *tappable* tab node, and which slug convention this screen standardises on.

|---|---|---|---|
| ⚠️ | 🟡 | `lib/page/port_forwarding/views/usp_port_forwarding_detail_view.dart:89-117` | `Semantics` uses default `container: false`, so the identifier merges onto the `Text` label node, which sits *below* `TabBar`'s tap-owning `InkWell` — an E2E `byId()` click may resolve a non-interactive label overlay and never switch tabs |
| ⚠️ | 🟢 | `lib/page/port_forwarding/views/usp_port_forwarding_detail_view.dart:85-88` | [Fragile Coupling] `Tab(text:)`→`Tab(child:)` hand-copies a Flutter internal (`softWrap: false, overflow: TextOverflow.fade`) 3×, enforced only by a comment; a Flutter bump desynchronises it silently |
| ⚠️ | 🟢 | `lib/page/port_forwarding/views/usp_port_forwarding_detail_view.dart:91,101,111` | New slugs `port-forwarding-tab-*` diverge from the `pf-*` prefix shipped by #1172 on the *same screen* → two competing E2E selector conventions |
| ⚠️ | 🟢 | `test/page/port_forwarding/usp_port_forwarding_tab_identifier_widget_test.dart:99-112,128-136` | Tests assert tree *presence* only — they would stay green in exactly the failure mode above; no tap-through and no `hasTapAction` assertion |
| ⚠️ | 🟢 | `test/page/port_forwarding/usp_port_forwarding_tab_identifier_widget_test.dart:100,112,121,145,161` | `handle.dispose()` is the last statement of each body instead of `addTearDown` → a failing `expect` leaks the `SemanticsHandle` and adds a confusing secondary failure |
| ⚠️ | 🟡 | `lib/.../usp_port_forwarding_detail_view.dart:91,101,111` + `test/.../usp_port_forwarding_tab_identifier_widget_test.dart:32-35` | [Duplicated Code] Slug literals exist in view + test + the E2E repo with no shared const; the test compares two independently-typed copies rather than enforcing a contract (144 `identifier: '…'` literals repo-wide, zero constants file) |
| ⚠️ | 🟡 | `lib/page/port_forwarding/views/usp_port_forwarding_detail_view.dart:88-119` | [Shotgun Surgery] The 10-line block is repeated 3× and 9 other `Tab(text:)` sites (`lib/page/statistics/views/usp_statistics_view.dart:58-60`, `lib/page/wifi_settings/views/usp_wifi_settings_view.dart:103`) will want to copy it, framework-mirroring included |
| 💡 | 🟢 | `test/.../usp_port_forwarding_tab_identifier_widget_test.dart:99-104,120-125,147-152` | Identical 6-line preamble triplicated; hoist to a `pumpView(tester)` helper and name the load-bearing `1280x900` viewport const |
| 💡 | 🟡 | `test/.../usp_port_forwarding_tab_identifier_widget_test.dart:155-157` | Hardcoded English labels couple the test to ARB copy; resolve via `AppLocalizations` from the pumped context instead |
| 💡 | 🟢 | `lib/.../usp_port_forwarding_detail_view.dart:101,111` | Slug/label asymmetry: the third tab reads "Port Range Triggering" but is `…-triggering` while the second is `…-range`; an E2E author may reach for `…-range-triggering` |
| 💡 | ⚪ | `test/page/port_forwarding/usp_port_forwarding_tab_identifier_widget_test.dart:1` | No workflow invoking `flutter test` was found in this worktree — worth confirming `ui`-tagged tests are actually selected by CI, otherwise the guard is nominal |
| 💡 | ⚪ | `lib/.../usp_port_forwarding_detail_view.dart:78-88` | 11 lines of issue archaeology (#1172 / E2E#44) above 30 lines of widget tree; the *why-this-shape* part must stay, the rest reads better in the PR description |

<sub>Confidence: 🟢 High = verified against head code / repo precedent · 🟡 Med = structurally argued, not executed · ⚪ Low = worth a second look. `[X]` = raised by 1 agent, `[XX]` = both agents (Critical rows only; none this round).</sub>
<sub>Note: the Flutter SDK is not available in the review worktree, so no test run, semantics dump, or web build could be executed — every claim below rests on head-code reading and in-repo precedent.</sub>

<details>
<summary>⚠️ Warning details (7)</summary>

### W1 — Identifier may not be on the node E2E clicks · `usp_port_forwarding_detail_view.dart:89-117` · 🟡 Med

```dart
Tab(
  child: Semantics(
    identifier: 'port-forwarding-tab-single',
    child: Text(
      loc(context).singlePortForwarding,
      softWrap: false,
      overflow: TextOverflow.fade,
    ),
  ),
),
```

`Semantics` here uses the default `container: false`, so it does not create a node of its own — it merges into the child's node, i.e. the **`Text` label node**. In Material's `TabBar` the interactive node is created *above* `widget.tabs[index]`: the tab is wrapped in `Center` → `Padding`/`KeyedSubtree`, then in an `InkWell` whose `onTap` drives the tab change, and the "Tab N of M" label is a *sibling* inside a `Stack`. The `identifier` therefore lands on a descendant of the tappable node, not on it.

On Flutter web (CanvasKit) each semantics node is projected to its own `<flt-semantics>` element and only nodes carrying a tap action get a click handler. A Playwright/Selenium `byId()` resolving `[flt-semantics-identifier="port-forwarding-tab-single"]` would click a plain label overlay above the canvas; the click can be absorbed there and never reach the tab handler — precisely the outcome this PR exists to prevent.

Contrast the repo's own established shape, where the hook sits on the tappable container:

```dart
// lib/page/advanced_settings/views/usp_advanced_settings_view.dart:102-105
Widget _buildCard(AppSectionItemData item) {
  return LayoutBlock(
    identifier: item.identifier,   // container that owns onTap
    onTap: item.onTap,
```

**Trigger:** any E2E step that resolves one of the three new identifiers and clicks the matched element (PrivacyGUI-USP-E2E#44).

**Fix (preferred → fallback):** (1) put the hook where the tap lives — thread an identifier through `UiKitPageView` to the tab wrapper, or `Semantics(identifier: …, container: true, child: Tab(…))` / `MergeSemantics(child: Semantics(identifier: …, child: Tab(…)))` in the `tabs:` list; (2) if it must stay on the label, confirm empirically with one semantics dump or one E2E dry-run that clicking the identified element still switches tabs, and record that in the PR description.

Confidence is 🟡 because the Flutter SDK is unavailable in the review worktree (`which flutter` → empty), so the semantics tree and generated DOM could not be observed; the argument is structural (`TabBar`/`Semantics` framework behaviour), not captured output. **Note the two reviewers reached this concern independently but from opposite directions** — Reviewer B considered it and explicitly *declined* to escalate it, on the grounds that the three widget tests do assert `find.bySemanticsIdentifier(id)` resolves, and the same nesting pattern is already in production use (`test/page/topology/helpers/topology_node_identifier_widget_test.dart`, `test/page/advanced_settings/usp_advanced_settings_identifier_widget_test.dart`). That disagreement is preserved here rather than resolved: presence in the tree is not in dispute, *hit-testability of the identified node* is. One tap-through test (W4's fix) settles it either way.

### W2 — Framework internal hand-mirrored, comment-enforced · `usp_port_forwarding_detail_view.dart:85-88, 93-97, 103-107, 113-117` · 🟢 High

```dart
// The Text below intentionally mirrors what `Tab(text:)` renders
// internally (`Text(text, softWrap: false, overflow: TextOverflow.fade)`)
// so this is a zero-visual-change swap ...
```

`softWrap: false, overflow: TextOverflow.fade` is a copy of `Tab.build`'s implementation detail. It is *correct against the current Flutter pin* — `Tab` applies typography through the enclosing `_TabStyle`/`DefaultTextStyle` rather than inside the `text` branch, and `_kTabHeight` sizing is unchanged, so the "zero-visual-change" claim holds today (this is why it is not Critical). But the invariant lives in the SDK and is enforced here by nothing except a comment, replicated 3×. A Flutter upgrade that changes `Tab`'s internal text rendering desynchronises these three call sites while every other `Tab(text:)` site in the repo picks the change up for free; the failure mode is a cosmetic overflow/wrap regression that nothing in this PR would catch (the new tests assert identifier and label *presence*, never layout).

**Fix:** keep `Tab(text:)` and attach the identifier from outside (also resolves W1), or add one narrow-viewport layout/golden assertion so the mirrored invariant is test-enforced. Note the existing goldens (`test/golden_test/page/port_forwarding/localizations/usp_port_forwarding_detail_view_test.dart`) drive tabs by index via `switchToTab(tester, 1|2)`; if the tab strip is inside the captured region the claim is already covered — worth confirming and saying so in the PR.

### W3 — Slug convention split on one screen · `usp_port_forwarding_detail_view.dart:91,101,111` · 🟢 High

Every identifier already shipped in this feature folder uses a short `pf-` prefix:

```
lib/page/port_forwarding/views/components/usp_single_port_tab.dart:37       'pf-add-single-port'
lib/page/port_forwarding/views/components/usp_port_range_tab.dart:36        'pf-add-port-range'
lib/page/port_forwarding/views/components/usp_port_triggering_tab.dart:36   'pf-add-port-triggering'
...:63/92/98                                                               'pf-rule-enable-…', 'pf-edit-…', 'pf-delete-…'
```

Those came from #1172 — the PR this change describes itself as a follow-on to. The screen now exposes two prefix conventions to the E2E suite, so anyone writing or grepping selectors has to know both. Either align (`pf-tab-single` / `pf-tab-range` / `pf-tab-triggering`) or, if `port-forwarding-*` is deliberately the new house style, say so in the PR description so #1172's identifiers get migrated instead of stranded. Cost of fixing rises the moment the E2E repo starts depending on the new strings.

### W4 — Tests cannot fail in the mode being guarded against · `usp_port_forwarding_tab_identifier_widget_test.dart:99-112, 128-136` · 🟢 High

```dart
for (final id in expectedIdentifiers) {
  expect(
    find.bySemanticsIdentifier(id),
    findsOneWidget,
    reason: 'tab "$id" must be locatable by identifier',
  );
}
```

The file's doc comment claims the CanvasKit → `flt-semantics-identifier` projection is a runtime guarantee, so no web round is needed. The projection may well be guaranteed, but the contract E2E depends on is *"the element I find by id is the element whose click switches the tab."* All three cases assert only that an identified node exists, that the three are distinct, and that labels still render. If W1 is real, every assertion stays green while E2E#44 breaks.

**Fix (highest-value single addition to this PR):**

```dart
await tester.tap(find.bySemanticsIdentifier('port-forwarding-tab-range'));
await tester.pumpAndSettle();
expect(find.byType(UspPortRangeTab), findsOneWidget);
```

and/or assert the node owns the action:

```dart
expect(
  tester.getSemantics(find.bySemanticsIdentifier('port-forwarding-tab-range')),
  matchesSemantics(identifier: 'port-forwarding-tab-range', hasTapAction: true, isFocusable: true),
);
```

### W5 — `handle.dispose()` on the happy path only · `usp_port_forwarding_tab_identifier_widget_test.dart:100,112,121,145,161` · 🟢 High

```dart
final handle = tester.ensureSemantics();
tester.view.physicalSize = const Size(1280, 900);
tester.view.devicePixelRatio = 1.0;
addTearDown(tester.view.resetPhysicalSize);
addTearDown(tester.view.resetDevicePixelRatio);
...
handle.dispose();   // last statement in the body
```

The viewport resets are correctly registered with `addTearDown`, but the `SemanticsHandle` is not. If any `expect` above throws — the whole point of the test — `dispose()` is skipped and `flutter_test` reports a leaked semantics handle *in addition to* the real failure; in some framework versions the leak surfaces as the failure of the *next* test. So the first genuine regression this file catches is harder to read than it needs to be.

**Fix:** `addTearDown(handle.dispose);` right after `ensureSemantics()`, drop the trailing calls. Note `test/page/advanced_settings/usp_advanced_settings_identifier_widget_test.dart:93,110,115,129` has the same shape — inherited pattern, not invented here, but worth fixing in both.

### W6 — Slug literals duplicated with no single source of truth · `usp_port_forwarding_detail_view.dart:91,101,111` + `usp_port_forwarding_tab_identifier_widget_test.dart:32-35` · 🟡 Med

The three strings are typed in the view, re-typed as `expectedIdentifiers` in the test, and (per the PR's own comment) a third time in PrivacyGUI-USP-E2E#44. The test therefore verifies that two independently-typed copies match — it would pass unchanged if someone renamed both while the E2E repo silently broke.

```
$ grep -rn "identifier: '" lib --include=*.dart | wc -l   → 144
$ grep -rln "semantics_identifier|SemanticsIdentifiers|identifierConst" lib test → (no matches)
```

Repo-wide and clearly out of scope to fix fully, but this page is a good place to start the pattern: `abstract final class PortForwardingIds { static const tabSingle = '…'; … }` imported by both view and test, giving the E2E repo something to mirror deliberately. At minimum a follow-up issue — the duplication grows linearly with every identifier PR.

### W7 — One-off pattern invites shotgun surgery · `usp_port_forwarding_detail_view.dart:88-119` · 🟡 Med

Nine other `Tab(text:)` call sites exist in `lib/`, including equally likely E2E targets:

```
lib/page/statistics/views/usp_statistics_view.dart:58-60      Tab(text: loc(context).network/devices/system)
lib/page/wifi_settings/views/usp_wifi_settings_view.dart:103  Tab(text: loc(context).wifi)
```

This PR solves "tab has no identifier" locally with a 10-line block repeated 3× (differing only in two strings). When Statistics or Wi-Fi Settings needs the same hook, the block gets copy-pasted along with W2's mirrored framework internals — which is how a coupling detail becomes load-bearing in five files. The Art. XV-shaped fix is one shared widget (`UspIdentifiedTab(id:, label:)`) so the mirroring lives in exactly one place and each call site is a single line. 🟡 rather than 🟢 because 3× in one file is a defensible stopping point for a targeted E2E-unblock PR; this is about trajectory, and it downgrades to a Suggestion if you would rather land the abstraction with the second consumer.

</details>

<details>

- **Right primitive.** `Semantics(identifier:)` rather than `semanticLabel` keeps the accessible name as the localized string and stays silent to screen readers — exactly what the repo's own guidance prescribes (`lib/page/models/app_section_item_data.dart:17-19`: *"Prefer this over [semanticLabel] for test slugs so the accessible name stays the localized card title"*). No a11y regression.
- **Correct layer, minimal blast radius.** Entirely within `views/` + one test. No new imports, no `generated/` leakage, no provider/service/mutation path touched — so none of the Art. IV (USP Provider hard rules), Art. V (three-tier), or Art. XIII (ServiceError isolation) boundaries are in play. No security surface either: no secrets, no `flutter_secure_storage`, no injection point, no auth logic, no USP/JNAP response handling.
- **Replacing label-based targeting is the right fix**, and the documented `portRangeForwarding` / `portTriggering` → both render "Port Range …" collision (test lines 154-156) is a genuinely good justification a reviewer would otherwise have had to discover alone.
- **Test ships with the change** — 164 lines for 41 lines of production code, reusing an existing in-repo harness (`usp_advanced_settings_identifier_widget_test.dart`, `topology_node_identifier_widget_test.dart`, `mock_common.dart` / `mock_port_forwarding.dart`, the shared `port_forwarding_test_data.dart` fixture) instead of hand-rolling state.
- **Distinctness is asserted, not assumed.** Collecting matched `Element`s into a `Set` and asserting `hasLength(3)` (test lines 133-142) catches the realistic copy-paste bug — the same slug on two tabs — that three independent `findsOneWidget` checks would miss.
- **A behavioural claim is stated and then guarded.** "Zero visual change" is asserted in the code and the label-preservation case (test lines 157-159) checks it — better discipline than most identifier PRs, which change the tree and assert nothing visible.
- **Platform-channel stub scoped correctly.** The `package_info` mock (test lines 46-62) returns `null` for unknown methods rather than throwing, installed in `setUpAll` with a comment explaining *why* (UspTopBar reads it during build). No cross-test leakage.
- **Honest, high-value comments.** The production comment explains why `Tab(text:)` had to be abandoned and links #1172 / #1246 / E2E#44; the notes that `TabBar` builds all tabs up front and about the near-collision save the next reader real investigation time.
- **`@Tags(['ui'])`** matches `dart_test.yaml` and the sibling test — correctly bucketed for CI selection.

</details>

PR #1244, opened 2026-08-11
feat(build): add an English-only build flavour to cut the delivered payload

AustinChangLinksys:
## Summary

Adds an **English-only build flavour** to `build_web.sh`, selected by a `LOCALES` environment variable. ISP builds on SDK14.0 are short on firmware flash, and localization is the largest part of the delivered payload we can give up.

Measured on a real build: **29,812 KB → 25,908 KB, a 3,904 KB (3.81 MB) reduction.** That is short of the 5 MB the issue asks for; the remainder is not in i18n.

`LOCALES` unset or `all` builds byte-for-byte what it built before, so every existing caller is untouched.

Refs linksys/Architecture#177

## Why a script, not a branch or a compile flag

- **Not a branch** — `app_en.arb` changes with nearly every feature, so a branch whose diff is "delete the other 25 ARB files" pays a merge conflict on every sync.
- **Not `--dart-define`** — it cannot shrink anything, because `supportedLocales` references every generated locale class.
- **Not deferred loading** — `use-deferred-loading` makes the payload 72 KB *larger* while turning `lookupAppLocalizations` into a `Future`.

Deleting the ARB files before `gen-l10n` runs is the only thing that removes the string tables.

## Changes

| File | What |
| --- | --- |
| `tools/locale_strip.dart` | `keep` / `restore` / `verify` commands. Strips language packs plus the fallback fonts they need. |
| `build_web.sh` | Reads `LOCALES`, strips before building, restores from an `EXIT` trap, and logs enough for a Jenkins console log to be diagnostic on its own. |
| `tools/measure_payload.sh` | Prints the *delivered* payload — `build/web` minus `canvaskit/`, which CI prunes. Quoting the whole `build/web` total overstates firmware size by more than 2x. |
| `general_settings_widget.dart` | Omits the Language tile when the build ships a single language pack. The parent makes the call rather than the tile self-hiding, or the fixed-height `SizedBox` leaves a 44px hole. |

## Safety

Restoring is `git checkout --` over a **fixed three-path list**, never a `reset`, so unrelated work in progress cannot be reverted. A strip is gated on those paths being clean, runs from an `EXIT` trap so a failed or interrupted build is still covered, and **fails the build if the restore does not complete** — a build that leaves a stripped tree behind has to fail loudly even if the build itself succeeded.

[...]

PR #1292, opened 2026-08-17
build(web): prune the unserved CanvasKit output, and pre-compress behind a flag

### LINKED ISSUE #1282: Static pre-compression (brotli -q11, originals deleted): build_web.sh step + lighttpd config — blocked on firmware confirmation

AustinChangLinksys:
Moves two build steps out of the Jenkins job's shell block and into `build_web.sh`. Nothing about the delivered payload changes by default — the prune is what CI already did, and the compression is off unless asked for.

The point is where the reasoning lives. Both steps encode non-obvious decisions (why a 37 MB directory is safe to delete, why brotli and not gzip), and in a job configuration nobody diffs, those decisions are invisible to the next person who touches them.

## `pruneWebOutput()` — unconditional

Deletes `build/web/canvaskit/` (~37 MB, ~8.6 MB of it `*.symbols`). Nothing requests it: `flutter_bootstrap.js` pins `canvasKitBaseUrl: "./assets/"`, so the loader only ever fetches `assets/canvaskit.*`, which are hand-vendored copies committed under `web/assets/`. Shipping the directory would silently give back the flash #1281 reclaimed.

**The Jenkins deletion step is deliberately left alone.** 1.x still needs it, and this function is idempotent, so on 2.x it now finds nothing:

```
========== find canvaskit ============
========== find canvaskit end ============
```

Verified under `dash -xe` — what a Jenkins shell step actually runs — that this exits 0 and the build continues: `find` with no match is 0, and `xargs rm -rf` on empty input is 0 under both BSD (never invokes `rm`) and GNU xargs (invokes it once with no operands, where `-f` absorbs the missing-operand error). Confirmed end-to-end that `artifacts/` and the tarball come out complete with zero `web/canvaskit/` entries. So there is no coordinated change and no drift between the two.

It is also written as explicit paths rather than the `find . -iname 'canvaskit' | xargs rm -rf` sweep Jenkins uses. That sweep also matches the FVM SDK's own `.fvm/flutter_sdk/bin/cache/flutter_web_sdk/canvaskit` whenever `flutter_sdk` is a real directory instead of a symlink — simulated, and it does delete the SDK's web engine, breaking the *next* build with an error pointing somewhere unrelated.

## `compressWebAssets()` — off unless `Compress=true`

brotli `-q 11` over `main.dart.js` and `assets/canvaskit.wasm`, originals deleted.

**Gated off on purpose.** It is inert until the firmware team ships the matching lighttpd rewrite rules — pre-compressed files with no server rule are just 404s, i.e. a white screen. #1282 is still `blocked:fw-support`; this lands the build half so it is ready and reviewed, not so it is enabled.

`Compress` arrives as an environment variable rather than a positional argument, because Jenkins exports build parameters — the same mechanism `LOCALES` and `FlutterVersion` already use. No shell-step change needed to turn it on later.

Three things that look like over-engineering and are not:

**The whole target list is validated before any file is touched.** Compressing and deleting while walking the array means a stale target name leaves the tree half-compressed — some files `.br`, some not. Under rewrite rules that is a mix of rewritten and unrewritten URLs, which is a white screen rather than a failed build. Reproduced on the straightforward version, hence the up-front pass.

**Two compressors.** The CI agent (Ubuntu 22.04.1) has **no `brotli` binary** and no python `brotli` module; it does have `node`. Node's zlib is the same libbrotli, so this prefers the CLI and falls back to `node -e`. Output is the same size — byte-identical for `main.dart.js`; `canvaskit.wasm` differs only in the one-byte window-size header, because the CLI auto-sizes `lgwin` to 23 for a 7.2 MB input. Both decode back to the exact original bytes.

**`lgwin` is pinned to 24.** Node defaults to 22, which costs ~7.5 KB on `main.dart.js`. 24 is `BROTLI_MAX_WINDOW_BITS`, the largest window requiring no client negotiation.

**gzip is not an option here.** Its output is near-incompressible, so it defeats the rootfs squashfs XZ layer and makes flash *larger* — 9.45 → 10.76 MB measured. See #1282 for the full table.

## `tools/measure_payload.sh`

The prune and compression run *before* `measure_payload.sh`, so what it reports is what actually gets packaged. That reordering broke the report, which the second commit fixes.

A real build printed this:

```
  removing build/web/canvaskit ( 37M)
  ...
  pruned by CI:    0 KB (canvaskit/)
```

The subtraction finds nothing left to subtract, so it reports 0 — four lines after the prune said it removed 37 MB. Under `build_web.sh` the directory is always gone by then, so it now says `already pruned`; for a standalone `flutter build web` followed by running this by hand, the subtraction is still the difference between a real payload figure and a doubled one, and there the KB number is worth printing. Both branches and the optional baseline argument verified against a real build.

The stale comment claiming the directory is "excluded because CI prunes it" is updated too — that is now this script's sibling.

## Testing

Two real `flutter build web` runs through the modified script:

| Run | Result |
|---|---|
| `Compress` unset (the default) | prune removed 37 MB, payload 23.54 MB |
| `Compress=true`, clean tree | payload 12.12 MB, both targets `.br`, no originals left |

Compression ratios on real output match #1282's table exactly — `main.dart.js` 19.1%, `assets/canvaskit.wasm` → 2,248,161 bytes, the same byte count recorded there.

Against a stubbed build tree, 5 cases:

| Case | Result |
|---|---|
| Happy path (brotli CLI) | both targets `.br`, `canvaskit/` gone, round-trip md5 identical to original |
| Prune run twice | `already absent`, exit 0 — the Jenkins no-op case |
| Missing target | aborts, **nothing touched, no stray `.br`** |
| No compressor on PATH | aborts, tree intact |
| brotli CLI hidden | selects node, round-trips to exact original bytes |

Two interactions with #1244, which landed on `dev-2.7.0` after this work started:

- An unset `Compress` does not trip `[ "$Compress" == "true" ]`.
- A compression failure still propagates `exit 1` through the new `LOCALES` EXIT trap, rather than being swallowed into 0.

Both scripts pass `bash -n`. `test/tools/locale_strip_test.dart` reproduces `build_web.sh`'s EXIT-trap sequence, so it is coupled to this file — 43/43 pass. `flutter analyze` reports no new findings (this branch changes no Dart).

## Note for whoever updates the Jenkins job

The gzip block there must be **deleted**, not just left disabled — with #1281 landed, `gzip -k -9 build/web/assets/chromium/canvaskit.wasm` targets a file that no longer exists (exit 1), and the following `rm` has no `-f` (exit 1). Verified: under `-e` that aborts the build; without `-e` it silently leaves a half-compressed tree carrying both originals and `.gz` siblings.

The `Compress` **parameter definition** must stay — `build_web.sh` reads it.

Refs #1282, #1281

### DIFF
--- a/build_web.sh
+++ b/build_web.sh
@@ -20,6 +20,14 @@
 #            Jenkins exports build parameters as environment variables.
 #
 #   LOCALES=en ./build_web.sh 100 false "/" prod false true             # English-only
+#
+#   Compress Set to "true" to pre-compress the large static assets with brotli
+#            and delete the originals (#1282). Off unless explicitly set, because
+#            it is inert — worse, a white screen — until the firmware ships the
+#            matching lighttpd rewrite rules. Same Jenkins story as LOCALES: one
+#            boolean parameter named Compress, no shell-step change.
+#
+#   Compress=true ./build_web.sh 100 false "/" prod false true          # Pre-compressed
 
 # Detect fvm: use fvm flutter if available, otherwise use flutter directly
 if command -v fvm > /dev/null 2>&1 && [ -f ".fvmrc" ]; then
@@ -77,6 +85,155 @@ function restoreLocales() {
     exit 1
   fi
   exit $buildStatus
+}
+
+# Removes build output that is never served, so it cannot reach the router's
+# flash. This runs unconditionally: it is a property of what we ship, not an
+# option, and keeping it next to the build makes the reasoning visible to
+# whoever next wonders why a directory disappears.
+#
+# build/web/canvaskit/ is the engine's own CanvasKit output (~37 MB, including
+# ~8.6 MB of *.symbols). Nothing requests it: flutter_bootstrap.js pins
+# canvasKitBaseUrl: "./assets/", so the loader only ever fetches
+# assets/canvaskit.*, which are hand-vendored copies committed under web/assets/
+# (see #1281 and test/web/canvaskit_variant_test.dart). Shipping the directory
+# would silently give back the flash #1281 reclaimed.
+#
+# Deliberately targeted paths, not a `find -iname canvaskit` sweep from the
+# workspace root: such a sweep also matches the FVM SDK's own
+# .fvm/flutter_sdk/bin/cache/flutter_web_sdk/canvaskit when flutter_sdk is a real
+# directory rather than a symlink, and rm -rf on that breaks the next build with
+# an error pointing somewhere unrelated.
+#
+# Idempotent, so the Jenkins job's own deletion step — which 1.x still needs —
+# stays a harmless no-op on 2.x rather than something to keep in sync.
+function pruneWebOutput() {
+  echo "======== Pruning unserved build output ========"
+  local prune=(
+    build/web/canvaskit
+  )
+
+  local p
+  for p in "${prune[@]}"; do
+    if [ -e "$p" ]; then
+      echo "  removing $p ($(du -sh "$p" 2> /dev/null | cut -f1))"
+      if ! rm -rf "$p"; then
+        echo "ERROR: could not remove $p"
+        return 1
+      fi
+    else
+      echo "  already absent: $p"
+    fi
+  done
+}
+
+# Pre-compresses the large static assets with brotli and deletes the originals,
+# so the router serves compressed bytes without needing a server-side brotli
+# library (its lighttpd has none). See #1282 for the measurements.
+#
+# gzip is deliberately not used here. Its output is near-incompressible, which
+# defeats the rootfs squashfs XZ layer and makes flash LARGER (9.45 -> 10.76 MB
+# measured), while also transferring ~1.3 MB more per cold load than brotli.
+#
+# The target list is explicit rather than a glob: a rewrite rule that matches
+# every .js/.wasm would also rewrite files that have no .br sibling
+# (canvaskit.js, flutter.js, usp_client.js, usp_client_bg.wasm), and each of
+# those is a 404 and a white screen. The list and the disk must agree, so a
+# missing target is a hard failure, never a silent skip.
+#
+# Inert without the server half: pre-compressed files with no lighttpd rewrite
+# rule are just 404s. Keep this off until the firmware team ships the config.
+#
+# Two compressors are accepted because the CI agent may not have the brotli CLI
+# (the current one does not; it does have node). Node's zlib is built on the same
+# libbrotli and at q11/lgwin=24 produces output of identical size — byte-identical
+# for main.dart.js, while canvaskit.wasm differs only in the one-byte window-size
+# header, since the CLI auto-sizes lgwin down to 23 for a 7.2 MB input. Both
+# decode back to the exact original.
+#
+# lgwin must be set explicitly: Node defaults to 22 while the CLI auto-sizes to
+# the input, which costs ~7.5 KB on main.dart.js. 24 is BROTLI_MAX_WINDOW_BITS,
+# the largest window needing no client negotiation, so browsers still decode it.
+function compressWebAssets() {
+  echo "======== Pre-compressing web assets (brotli -q11) ========"
+  local targets=(
+    build/web/main.dart.js
+    build/web/assets/canvaskit.wasm
+  )
+
+  local compressor=""
+  if command -v brotli > /dev/null 2>&1; then
+    compressor="cli"
+  elif command -v node > /dev/null 2>&1; then
+    compressor="node"
+  else
+    echo "ERROR: no brotli compressor available (need the brotli CLI or node)"
+    return 1
+  fi
+  echo "  compressor: $compressor"
+
+  # Checks the whole list before touching anything, so a target name that goes
+  # stale cannot leave the tree half-compressed — which serves as a mix of
+  # rewritten and unrewritten files, i.e. a white screen rather than a failure.
+  local f
+  local missing=0
+  for f in "${targets[@]}"; do
+    if [ ! -f "$f" ]; then
+      echo "ERROR: expected compression target missing: $f"
+      missing=1
+    fi
+  done
+  if [ "$missing" -ne 0 ]; then
+    echo "ERROR: target list and build output disagree; nothing was compressed"
+    return 1
+  fi
+
+  for f in "${targets[@]}"; do
+    # -c > "$f.br" rather than brotli's in-place mode: no dependency on -k
+    # semantics, and the original is removed explicitly below.
+    if [ "$compressor" = "cli" ]; then
+      if ! brotli -q 11 -c "$f" > "$f.br"; then
+        echo "ERROR: brotli failed on $f"
+        rm -f "$f.br"
+        return 1
+      fi
+    else
+      if ! node -e '
+const fs = require("fs"), zlib = require("zlib");
+const buf = fs.readFileSync(process.argv[1]);
+fs.writeFileSync(process.argv[2], zlib.brotliCompressSync(buf, {
+  params: {
+    [zlib.constants.BROTLI_PARAM_QUALITY]: 11,
+    [zlib.constants.BROTLI_PARAM_LGWIN]: zlib.constants.BROTLI_MAX_WINDOW_BITS,
+    [zlib.constants.BROTLI_PARAM_SIZE_HINT]: buf.length,
+  },
+}));' "$f" "$f.br"; then
+        echo "ERROR: node brotli failed on $f"
+        rm -f "$f.br"
+        return 1
+      fi
+    fi
+    if ! rm "$f"; then
+      echo "ERROR: could not remove original: $f"
+      return 1
+    fi
+    echo "  $(basename "$f") -> $(basename "$f").br"
+  done
+
+  # Consistency gate. A surviving original means lighttpd would serve it
+  # unrewritten; a missing .br means a 404 and a white screen.
+  for f in "${targets[@]}"; do
+    if [ ! -f "$f.br" ]; then
+      echo "ERROR: missing $f.br after compression"
+      return 1
+    fi
+    if [ -f "$f" ]; then
+      echo "ERROR: original survived: $f"
+      return 1
+    fi
+  done
+
+  echo "Pre-compression OK: ${#targets[@]} files"
 }
 
 buildNumber=${1}
@@ -154,6 +311,23 @@ if ! buildWebApp "$buildNumber"; then
     exit 1
 fi
 
+if ! pruneWebOutput; then
+    echo Web App "$buildNumber" prune failed
+    exit 1
+fi
+
+echo "Compress files? ${Compress:-false}"
+if [ "$Compress" == "true" ]; then
+    if ! compressWebAssets; then
+        echo Web App "$buildNumber" pre-compression failed
+        exit 1
+    fi
+fi
+
+# Last, so that what it measures is exactly what gets packaged: the prune has
+# already removed the unserved directory and any pre-compression has already
+# replaced the originals.
+#
 # Reporting, not gating: a measurement that cannot run must not fail a build that
 # already succeeded.
 ./tools/measure_payload.sh || echo "could not measure the delivered payload"
--- a/tools/measure_payload.sh
+++ b/tools/measure_payload.sh
@@ -4,9 +4,14 @@
 # packaged into /www/ on the router, and therefore the only part that consumes
 # firmware flash.
 #
-# The SDK's build/web/canvaskit/ directory is excluded because CI prunes it; the
-# CanvasKit builds that actually ship live under build/web/assets/. Quoting the
+# The SDK's build/web/canvaskit/ directory is excluded because it is never served;
+# the CanvasKit builds that actually ship live under build/web/assets/. Quoting the
 # whole build/web total as a firmware size overstates it by more than 2x.
+#
+# build_web.sh deletes that directory before calling this script, so the pruned
+# line reads 0 KB there. The subtraction stays for the standalone case — a plain
+# `flutter build web` followed by running this by hand — where it is the whole
+# difference between a real payload figure and a doubled one.
 #
 # Usage:
 #   ./tools/measure_payload.sh                  # print the payload size
--- a/tools/measure_payload.sh
+++ b/tools/measure_payload.sh
@@ -42,7 +42,14 @@ payloadKB=$((totalKB - prunedKB))
 
 echo "Delivered payload: ${payloadKB} KB ($(toMB "$payloadKB") MB)"
 echo "  build output:    ${totalKB} KB"
-echo "  pruned by CI:    ${prunedKB} KB (canvaskit/)"
+# Says "already pruned" rather than "0 KB" when the directory is gone. Under
+# build_web.sh it always is, and a bare 0 there reads as "nothing was pruned"
+# immediately after the prune step reported removing 37 MB.
+if [ "$prunedKB" -eq 0 ]; then
+  echo "  canvaskit/:      already pruned"
+else
+  echo "  excluded:        ${prunedKB} KB (canvaskit/, pruned before packaging)"
+fi
 
 echo
 echo "Largest contributors:"


AustinChangLinksys (commented):
🤖 **Automated Review — Round 1** · `4327b51..7760da3` (full)

**Verdict: 💬 Self-review (comment only)** — No Critical found; this is a self-authored PR (`feat/usp-1282-precompress-and-prune` not a dispatch-chain branch), so auto-approve is withheld per policy. Code is solid; see Warnings below.

#### Standards

|---|---|---|---|
| ⚠️ | 🟢High | `build_web.sh:216–218` | `rm "$f"` failure path leaves stray `$f.br` without cleanup, asymmetric to brotli/node failure paths |
| ⚠️ | 🟡Med | `tools/measure_payload.sh:48` | `prunedKB -eq 0` also fires when `canvaskit/` is an empty dir, printing misleading "already pruned" |
| 💡 | 🟡Med | `build_web.sh:~199–213` | Embedded Node.js script accesses `process.argv[1]/[2]` without named vars or comment; add a note or extract to `tools/` if targets grow |
| 💡 | 🟡Med | `build_web.sh:~113–115` | `prune` array holds one element today; design intent (future entries?) worth a one-line comment |
| 💡 | ⚪Low | `tools/measure_payload.sh:45–52` | Output format change ("already pruned" vs "excluded:…KB") may affect downstream log parsers |

#### Spec

|---|---|---|---|
| ⚠️ | 🟢High | `build_web.sh:195` | brotli CLI path lacks `--lgwin=24`; CLI auto-sizes to 23 for 7.2 MB `canvaskit.wasm`, violating spec "lgwin pinned to 24" |
| ⚠️ | 🟡Med | `tools/measure_payload.sh` | AC#9 "Measured post-change sizes recorded" — no baseline KB recorded in diff or CI config |
| ℹ️ | 🟢High | — | AC#10 "Delete Jenkins gzip block" — out-of-scope for this PR (Jenkins config not in diff), expected separate step |
| 💡 | 🟢High | `build_web.sh:120–122` | `pruneWebOutput` uses `return 1` (correct), but callers should `exit 1` on failure — current call site does; document if others ever source this function |

<details>

**W1 — `rm "$f"` failure leaves stray `.br` (Standards)** 🟢High  
`build_web.sh:216–218` (head version):
```bash
    if ! rm "$f"; then
      echo "ERROR: could not remove original: $f"
      return 1
    fi
```
The brotli CLI failure path (lines 195–198) and the node failure path (lines 210–213) both do `rm -f "$f.br"` before returning — cleaning up the partial output. The `rm "$f"` failure at line 216 returns 1 without cleaning up `$f.br`. At that point the tree has both `$f` and `$f.br`. The post-loop consistency gate will catch this (`"original survived"`) and also `return 1`, so the build does fail correctly — but `$f.br` is left stray. If CI re-runs compression without rebuilding, the pre-flight check will pass (original exists) but an existing `.br` from the prior partial run may get silently overwritten.  
**Fix:** add `rm -f "$f.br"` before `return 1` at line 218, mirroring the other failure paths.

**W2 — `prunedKB -eq 0` edge case in `measure_payload.sh` (Standards)** 🟡Med  
`tools/measure_payload.sh:38–52`:
```bash
prunedKB=0
if [ -d "$buildDir/canvaskit" ]; then
  prunedKB=$(du -sk "$buildDir/canvaskit" | cut -f1)
fi
...
if [ "$prunedKB" -eq 0 ]; then
  echo "  canvaskit/:      already pruned"
```
When `canvaskit/` directory exists but is empty, `du -sk` returns `0`, and the script prints `"already pruned"` — misleading. The comment at line 11 says "build_web.sh deletes that directory before calling this script, so the pruned line reads 0 KB there" — this is correct for the normal flow. The edge case is a developer manually emptying the directory. Cosmetic only; does not affect build correctness.  
**Fix:** use directory existence as the primary check: `if [ ! -d "$buildDir/canvaskit" ]; then echo "already pruned"`.

**W3 — brotli CLI lgwin not pinned to 24 (Spec)** 🟢High  
`build_web.sh:195`:
```bash
if ! brotli -q 11 -c "$f" > "$f.br"; then
```
The Node.js path (line 204–209) correctly sets `BROTLI_PARAM_LGWIN: zlib.constants.BROTLI_MAX_WINDOW_BITS` (= 24). The CLI path has no `--lgwin` / `-w` flag. The in-code comment at line ~150 acknowledges: "canvaskit.wasm differs only in the one-byte window-size header, since the CLI auto-sizes lgwin down to 23 for a 7.2 MB input." Issue #1282 Spec states "lgwin pinned to 24 (BROTLI_MAX_WINDOW_BITS)". The two compressor paths therefore produce non-identical output when the brotli CLI is used (vs. the current CI which uses node).  
**Fix:** `brotli -q 11 --lgwin=24 -c "$f" > "$f.br"` (or use `-w 24` on older CLI versions).

**W4 — AC#9 post-change sizes not recorded in diff (Spec)** 🟡Med  
Issue #1282 AC#9: "Measured post-change sizes recorded." The PR description mentions real build measurements (23.54 MB uncompressed, 12.12 MB compressed), but these appear only in the PR body prose — not in a committed file, a CI artifact marker, or a script constant. If the intent is to record them in the issue/PR body rather than in code, this is acceptable; but if a CI-level baseline should be updated (e.g., `measure_payload.sh` baseline argument in the Jenkins job), that's not visible in this diff.  
*(Low-risk, tracking note — does not block this PR since the PR body itself has the measurements.)*

</details>

<details>

- **Pre-flight consistency gate** (`compressWebAssets`): validates entire target list before touching any file. A stale/missing target aborts with no partial state — exactly the right design.
- **Post-compression consistency gate**: double-checks `.br` exists and original is gone — belt-and-suspenders against silent compressor failures.
- **`pruneWebOutput` idempotency**: `[ -e "$p" ]` guard + `already absent` message means running twice is safe; Jenkins step and this function coexist without coordination.
- **Compressor fallback chain**: `command -v brotli` → `command -v node` → hard fail. CI agents without brotli CLI still get a correct output; agents without either fail loudly rather than shipping uncompressed files.
- **`Compress` env var gate**: inert until explicitly enabled; aligns with #1282's "blocked on firmware confirmation" intent.
- **EXIT trap interaction**: `compressWebAssets` failure propagates correctly through the `LOCALES` EXIT trap — `exit $buildStatus` in `restoreLocales` uses the trapped `$?`, so a compression failure still exits 1 after locale restore.
- **`measure_payload.sh` fix**: the "pruned by CI: 0 KB" false-report is correctly addressed; `already pruned` output is honest in the normal `build_web.sh`-driven flow.
- **Shell safety**: no command injection vectors; all target paths are hardcoded constants.
- **`bash -n` passes** (mentioned in PR description and consistent with script structure).

</details>

PeterJhongLinksys (approved):
🤖 **Automated Review — Round 1** · `4327b51..7760da3` (incremental)

**Verdict: ✅ APPROVE** — Unusually well-engineered build-script change: the prune half's central safety claim is confirmed against head code, the destructive compression half is fail-closed and off by default, and rejected alternatives are recorded with measured numbers. Approving with one item both reviewers independently ranked #1 that needs a one-command check before the `Compress` flag is ever enabled.

|---|---|---|---|
| ⚠️ | 🟡 | `build_web.sh:110-129`, `:216` | Generated `flutter_service_worker.js` still lists `canvaskit/*` and `main.dart.js` after both new steps delete them; its `cache.addAll` pre-cache is atomic, so the worker may never activate — the lighttpd rules are **not** the only missing half |
| ⚠️ | 🟢 | `build_web.sh:225-236` | Originals are `rm`'d with only an *existence* check on the `.br`; no size or decode round-trip, so a truncated/0-byte `.br` passes the gate |
| ⚠️ | 🟢 | `tools/measure_payload.sh:45-52` | `prunedKB -eq 0` prints `already pruned` also when the directory *never existed* — misinforms the exact standalone reader the change was written for |
| ⚠️ | 🟢 | `tools/measure_payload.sh:11-14` | New header comment says the pruned line "reads 0 KB"; the new code in the same commit prints `already pruned` instead — two comments from one commit disagree |
| ⚠️ | 🟢 | `build_web.sh:186-193` | Compressor backend picked by agent capability, and the PR itself states the two outputs differ (lgwin header byte) → shipped artifact checksum depends on which CI agent ran the job |
| ⚠️ | 🟢 | `build_web.sh:112-114` + `tools/measure_payload.sh:38`, `:63` | `build/web/canvaskit` hard-coded in three places across two files; the two in `measure_payload.sh` fail *silently* into a wrong number [Shotgun Surgery] |
| ⚠️ | 🟡 | `build_web.sh:319-320` | `${Compress:-false}` in the echo, bare `"$Compress"` in the test one line later; adding `set -u` later (as the sibling script in this PR already has) breaks every **default** build |
| ⚠️ | 🟡 | `build_web.sh:159-162` | The four-file "must not be rewritten" exclusion list exists only inside a bash comment, while the rule that must honour it lives in the firmware lighttpd config; nothing detects divergence, symptom is a white screen |
| ⚠️ | 🟡 | `build_web.sh:24-28`, `:151-155` | Nothing in-repo tracks the "inert until firmware ships" prerequisite — only prose prevents someone flipping the Jenkins `Compress` parameter early |
| ⚠️ | 🟡 | `build_web.sh:24-30`, `:319` | Input style split: `LOCALES`/`Compress` via env vs everything else positional, and `SCREAMING_SNAKE` vs `PascalCase`; no documented rule for flag three |
| 💡 | 🟢 | `build_web.sh:198-210` | Node fallback hard-codes `zlib.constants.*` names; an `undefined` one becomes the literal key `"undefined"` and is silently ignored → q0 output that passes every check |
| 💡 | 🟢 | `build_web.sh:110-129` | [Speculative Generality] array + `for` + `local p` scaffolding for a single path — borderline, keep it if a second entry is imminent |
| 💡 | 🟡 | `build_web.sh:201-212` | 12-line Node script embedded in single quotes inside a bash function: no syntax check, no lint, not unit-testable → `tools/brotli_compress.js` |
| 💡 | 🟡 | `build_web.sh:222` | `compressWebAssets` mutates `build/web` destructively with no undo but a full rebuild; worth one line in the header comment |
| 💡 | 🟡 | `build_web.sh:112-114` + `tools/measure_payload.sh:38` | Prune list and measurement exclusion list are two copies of one decision — a shared `tools/unserved_paths.txt` or one more assertion in `canvaskit_variant_test.dart` |
| 💡 | 🟡 | `build_web.sh:145-149`, `:171-173`, `:225-227` | The "failure mode is a white screen" rationale is restated three times in slightly different words; state once, refer back |
| 💡 | ⚪ | `build_web.sh:110-129` | Rename `prune=(...)` → `unservedPaths` — states the *criterion* for membership, which is what a future editor needs |
| 💡 | ⚪ | `build_web.sh:110-129` | Loop is the function's last statement, so its implicit return is the last `echo`'s status — only *accidentally* 0 on the already-absent path; add explicit `return 0` |
| 💡 | ⚪ | `build_web.sh:158-232` | Neither new function is covered by a test, in a repo that already tests exactly this class of build invariant (`test/web/canvaskit_variant_test.dart`) |
| 💡 | ⚪ | `build_web.sh` (comment density) | ~60 of 174 added lines are comment; content is high quality, but the stale header in `measure_payload.sh` shows the risk — consider moving deep measurements to `#1282`/`docs/` |

<sub>Confidence: 🟢High = evidence chain closed against head code · 🟡Med = mechanism established, one link inferred · ⚪Low = worth a look, may be noise.</sub>
<sub>`[X]` = raised by 1 agent · `[XX]` = raised independently by both. No Critical this round, so no agreement markers apply.</sub>

> **No Critical findings.** Both reviewers looked specifically for one and each explicitly recorded why they could not close the evidence chain: the prune's safety is *confirmed* (see ✅ below), and the compression path's `rm` is unreachable on the default path (`build_web.sh:319-320`, `Compress` unset by any in-repo config), so per the triggerability standard those findings stay Warnings.

<details>
<summary><b>⚠️ Warning details (evidence)</b></summary>

### W1 — the service worker is a second consumer of the deleted filenames · `build_web.sh:110-129`, `:216` · 🟡Med · **raised independently by both reviewers**

This is the one item both agents ranked #1, arrived at from different directions. The PR's header comment frames the missing lighttpd rewrite rules as the only reason the flag must stay off (`build_web.sh:24-28`). That framing is incomplete — this repo ships a PWA service worker, and its generated manifest is a second, independent consumer of the exact filenames being deleted.

Chain, all quoted from head:
- `web/index.html:104` loads `./flutter_bootstrap.js`
- `web/flutter_bootstrap.js:34-37` registers the worker:
  ```js
  serviceWorkerSettings: {
    serviceWorkerVersion: "3346174710",
    serviceWorkerUrl: "service_worker.js"
  }
  ```
- `web/service_worker.js:6` pulls the generated worker in verbatim and forces it live:
  ```js
  importScripts('flutter_service_worker.js');
  ...
  self.skipWaiting();
  event.waitUntil(clients.claim());
  ```
- `flutter_service_worker.js` is generated by `flutter build web` (`build_web.sh:47`) with a `RESOURCES` map keyed by path and a `CORE` list that `install` pre-caches via `cache.addAll(...)`. Both new steps run **after** that build (`build_web.sh:311-325`), so the manifest still lists paths that are now gone from disk.

Two distinct exposures:
- `pruneWebOutput` deletes `build/web/canvaskit/` **unconditionally** — not behind `Compress` — so if the manifest enumerates `canvaskit/*`, this lands on *every* build in this PR, not just opt-in ones.
- `compressWebAssets` `rm`s `main.dart.js` (`build_web.sh:216`), which is near-certain to be in `RESOURCES`.

Failure mode: `addAll` is atomic. Without a rewrite rule it 404s and rejects the whole install promise, so the worker never activates; with a rewrite rule the fetch succeeds but brotli bytes get cached under the uncompressed key without the `Content-Encoding` a later cache hit needs. Because `skipWaiting`/`clients.claim` are unconditional, a worker installed by earlier firmware can keep serving a stale `main.dart.js`. Neither branch is fixed by correct lighttpd config.

Held at Warning (not Critical) by both reviewers for the same honest reason: `flutter_service_worker.js` is a build artifact absent from the worktree, so the `RESOURCES` keys are inferred from the standard Flutter template rather than quoted. **The mechanism is fully established; only the key names are unconfirmed.**

★ **One command settles it** — after a build:
```bash
grep -oE '"(canvaskit/[^"]*|main\.dart\.js)"' build/web/flutter_service_worker.js | sort -u
```
Empty output clears this item entirely. Non-empty means the manifest needs filtering/regeneration in the same pass, or the build needs `--pwa-strategy=none`, or `serviceWorkerSettings` should come out of `web/flutter_bootstrap.js`. Either way the conclusion belongs in the header comment, because that comment currently tells the next reader lighttpd is the only blocker and it will be trusted.

### W2 — originals deleted on an existence check alone · `build_web.sh:225-236` · 🟢High

The consistency gate checks only that the file exists:
```bash
if [ ! -f "$f.br" ]; then ... fi
if [ -f "$f" ]; then ... fi
```
No non-empty check, no decode round-trip — while `rm "$f"` at `:216` has already removed the only good copy. A disk filling mid-write on a 7.2 MB `canvaskit.wasm` leaves a short `.br` that `[ -f ]` happily accepts (the `>` truncation happens before `brotli` even runs). The function is admirably careful to pre-flight the *entire* target list before touching anything (`:186-189`) — which is exactly what makes it inconsistent that the post-condition is weaker than the pre-condition.

Both branches already have the tooling for the fix, before the `rm`:
```bash
if ! cmp -s <(brotli -d -c "$f.br") "$f"; then
  echo "ERROR: $f.br does not decode back to $f"; rm -f "$f.br"; return 1
fi
```
That turns "trust the compressor" into "the original is deleted only once its replacement provably reproduces it."

### W3 + W4 — `measure_payload.sh` overstates what it knows, and its own new comment contradicts it · `tools/measure_payload.sh:11-14`, `:45-52` · 🟢High

```bash
if [ "$prunedKB" -eq 0 ]; then
  echo "  canvaskit/:      already pruned"
```
`prunedKB` is 0 in two distinguishable cases (`:36-40`): the directory was pruned, **or it never existed**. The message asserts the first unconditionally. The motivation is right — a bare `0` immediately after the prune step reported removing 37 MB does read as "nothing was pruned" — but the fix over-corrects into a claim the script cannot support, and it does so for the one reader the standalone path exists to serve (`:11-14`): someone on a future SDK that stops emitting the directory, or running against a partially-copied `build/web`, is told "already pruned" and concludes the prune worked.

The information needed is already available and then discarded — `[ -d "$buildDir/canvaskit" ]` is tested at `:38`. Keep the boolean instead of re-deriving it from the size:
```bash
if [ -d "$buildDir/canvaskit" ]; then
  prunedKB=$(du -sk "$buildDir/canvaskit" | cut -f1)
  echo "  excluded:        ${prunedKB} KB (canvaskit/, pruned before packaging)"
else
  echo "  canvaskit/:      absent (pruned, or not emitted by this SDK)"
fi
```
Two side notes: an existing-but-empty directory can make `du -sk` return non-zero on some filesystems, printing `excluded: 0 KB` — the exact string this change set out to eliminate. And the new header comment at `:11-14` says the pruned line "reads 0 KB there", which the code in the same commit makes false; it should say `already pruned`. Two comments introduced in one commit already disagree about observable output.

### W5 — build output is not reproducible across CI agents · `build_web.sh:186-193` · 🟢High

```bash
if command -v brotli > /dev/null 2>&1; then compressor="cli"
elif command -v node > /dev/null 2>&1; then compressor="node"
```
The PR's own comment concedes the paths do not agree byte-for-byte ("canvaskit.wasm differs only in the one-byte window-size header, since the CLI auto-sizes lgwin down to 23 for a 7.2 MB input"). So the shipped artifact's checksum depends on which agent picked up the job — which defeats artifact hash comparison and firmware-image diffing. It is also two implementations of one behaviour that are *already documented as divergent* [Duplicated Code / Divergent Change]. Either pick one backend and make its absence a hard failure, or pass `--lgwin 24` to the CLI so both paths converge.

### W6 — `set -u` divergence between the two scripts in this PR · `build_web.sh:319-320` · 🟡Med

`tools/measure_payload.sh:20` enables `set -uo pipefail`; `build_web.sh` has no `set` line at all while gaining 174 lines including `rm -rf` and `rm`. Locally this shows up as `${Compress:-false}` in the echo at `:319` and bare `"$Compress"` in the test at `:320`. Harmless *only because* `set -u` is absent — adding it later, a routine hardening change, kills every default build with "Compress: unbound variable", i.e. the failure lands on the path that doesn't even want the feature. One-character fix, matching how `locales` is handled at `:269`:
```bash
if [ "${Compress:-false}" == "true" ]; then
```

### W7 — `build/web/canvaskit` hard-coded in three places · `build_web.sh:112-114`, `tools/measure_payload.sh:38`, `:63` · 🟢High

One decision, three literals across two files [Shotgun Surgery]. The two in `measure_payload.sh` fail *silently* into a wrong number rather than an error. A shared constant/exported var, or one more assertion in `test/web/canvaskit_variant_test.dart` (already the enforcement point for the other half of #1281), would collapse it.

### W8 + W9 — the cross-repo contract has no in-repo guard · `build_web.sh:159-162`, `:24-28` · 🟡Med

The code is strict about one half of the contract — the target list must match what is on disk, checked twice — but the comment states the real failure mode lives elsewhere: a rewrite rule matching every `.js`/`.wasm` would also rewrite `canvaskit.js`, `flutter.js`, `usp_client.js`, `usp_client_bg.wasm`, each a 404 and a white screen. That four-file exclusion list exists only inside a bash comment here, while the rule that must honour it lives in the firmware lighttpd config. Nothing detects divergence. Similarly, nothing but prose prevents someone flipping the Jenkins `Compress` parameter before the firmware half ships, for a flag the PR itself describes as actively harmful if enabled early. `test/web/canvaskit_variant_test.dart` is the pattern this repo already uses for exactly this problem — a sibling test asserting the compressible/non-compressible partition would give the same protection.

</details>

<details>

- **The prune is genuinely safe, and this was confirmed rather than taken on trust.** `web/flutter_bootstrap.js:27` pins `canvasKitBaseUrl: "./assets/"`; the vendored copies exist (`web/assets/canvaskit.js`, `web/assets/canvaskit.wasm`, 7.2 MB); and `test/web/canvaskit_variant_test.dart:50-56` fails the suite if that line is ever dropped. `build/web/canvaskit/` really is unreachable output, so deleting it cannot regress the loader. The reasoning in the comment matches the code.
- **Unconditional prune vs flagged compression is exactly the right split** (`build_web.sh:88-91`). Removing never-served output is a property of the artifact; changing the wire encoding needs a server counterpart. Correct architectural call, stated explicitly.
- **`build_web.sh:100-108` — the explanation for using explicit paths over `find -iname canvaskit` is the best thing in this PR.** It names the actual failure (the sweep matching `.fvm/flutter_sdk/bin/cache/flutter_web_sdk/canvaskit` when `flutter_sdk` is a real directory rather than a symlink) *and* the symptom (next build fails pointing somewhere unrelated). That is a trap someone would otherwise fall into twice.
- **Idempotency was designed in, not discovered.** `pruneWebOutput` handles already-absent (`:124`) specifically so the Jenkins job's own deletion step — still needed by 1.x — stays a harmless no-op. Real migration thinking rather than a flag day.
- **Fail-closed, checked twice.** Pre-flighting the entire target list before mutating anything (`:186-189`) is the correct ordering for a destructive step, and the stated reason — a half-compressed tree serves a mix of rewritten and unrewritten files, i.e. a white screen rather than a clean failure — is precisely why "hard failure, never a silent skip" is right here.
- **Failure cleanup is correct on both compressor branches**: `rm -f "$f.br"` before `return 1` (`:200`, `:208`) means a failed run leaves no partial `.br` for a later rewrite rule to find.
- **Gating is applied where it matters and withheld where it doesn't.** Prune/compress are gating with explicit messages (`:314-325`); the measurement stays non-gating via `|| echo` (`:331`). Payload corruption should fail the build; a broken measurement should not. The ordering comment at `:327-330` ("last, so that what it measures is exactly what gets packaged") explains the sequence.
- **`Compress` defaults off and the default is echoed** (`:319`), so every build log states which mode produced the artifact.
- **Rejected alternatives are recorded with numbers, not vibes**: gzip rejected with measured flash impact (9.45 → 10.76 MB, because gzip output defeats the squashfs XZ layer), `lgwin` 22-vs-24 with its ~7.5 KB cost, brotli in-place mode rejected over `-k` semantics. Explaining why the obvious choice is wrong is more valuable than justifying the one that was made, and it makes future revisits cheap.
- **`measure_payload.sh` keeps the subtraction for the standalone case** (`:11-14`) rather than deleting it as now-dead code — correct, the script has a second caller (a human) with different preconditions.
- **Cross-references are concrete** — `#1281`, `#1282`, `test/web/canvaskit_variant_test.dart` — so the reasoning is traceable to its origin.
- **Scope discipline** (Art. I §1.3): two shell scripts, zero Dart, tightly scoped to what the change needs. No USP/JNAP, tier-dependency, `ServiceError`, Equatable or UI-Kit surface area — nothing to violate.

</details>

---

**Bottom line:** approving. Nothing here blocks the merge — the prune half is verified safe and the compression half cannot fire on the default path. Before the `Compress` flag is ever switched on, please run the one-line `flutter_service_worker.js` grep in W1 (both reviewers independently landed on it as the top item) and consider the decode round-trip in W2, since that is the difference between "the compressor is trusted" and "the original is only deleted once proven reproducible."

Issue #1299, opened 2026-08-18
Density: let edit mode choose a card's form (popup/compact/normal), with the form constraining resize


AustinChangLinksys:
## Parent

#1183 — new, raised from the #1293/#1294 work on `gate/1183-overflow-gate` rather
than split out of an existing sub-ticket.

Design of record: `doc/dashboard/dashboard_density_design.md` **§2.1** (which this
ticket **partially overrides** — see below), §2.2, §2.4, §2.5, §2.6, §D2, §D3.

## Branch

Base **`dev-2.7.0` once PR #1248 has merged**; until then base
`gate/1183-overflow-gate`. The density mechanism (#1232), the popup form (#1239),
the gate, its probe and the design of record all arrive with #1248 and are absent
on `dev-2.7.0`. Branch name `feat/<this-issue>-density-toggle` — the `gate/`
namespace belongs to the dispatcher.

## What this overrides in §2.1, and why that clause has a hole

§2.1 states the choice is not the user's:

> Selection is **automatic from the card's real pixel width**. It is not a user
> preference: the width is already the consequence of a user action (resizing the
> card, or the screen they are on), so requiring a second, separate choice to
> avoid broken layout would push the framework's responsibility onto the user.

The reasoning holds wherever the user controls the width. **On a phone they do
not.** The 4-column grid pins every card to `x: 0, w: cols` and, as of the
#1293 left-edge lock, horizontal resize there is forbidden outright — so on the
one form factor where 18 full cards is a punishing scroll, the user has no
influence over density at all. §D2 (*no mobile preset*) is still deferred, so
nothing else addresses it either.

`popup` never reaches a phone by width (§2.1: a 320px phone's 3-column card is
212px, and on the mobile grid the card is full width), and `compact` reaches it
only for the two widest thresholds. Manual selection is the only mechanism that
puts a phone user in control of card density.

## The inversion — the ticket's central design claim

The current mechanism runs **width → density**. This ticket runs
**density → the sizes that are legal**:

| the user picks | the size rule | why |
|---|---|---|
| **popup** | cannot be resized at all | an icon and one value has no use for a larger box, and a locked-but-huge popup would be unrecoverable |
| **compact** | can be enlarged, not shrunk | its floor is the width at which the reduced form still reads |
| **normal** | today's behaviour | floor derived from `normalAbove` |

This is **not** the thing §2.1 rejected. §2.1 refuses to make the user repair a
[...]

Issue #1037, opened 2026-07-01
[Dashboard] Edit mode still work if changing tab to others


JohnnyWuLinksys:
**Description**
Edit mode still enable if changing to other tab instead of finish or cancel on Home page
<img width="1378" height="567" alt="Image" src="https://github.com/user-attachments/assets/ca8f11d8-005a-4e6c-b9c8-d25b478821f9" />
(not only blue border, also can resize and drop)

**Steps**
1. Enable edit mode on dashboard
2. Click "Menu"(or "Support") tab
3. Back to dashboard

**Current result**
Although icon show edit, but Edit mode already enabled(With blue border still and can resize also)

**Expected result**
Edit mode disabled

**Information**
Pinnacle 2.0 FW: 1.2.2.26062820
Linksys Now Version: 2.5.0.700525

**Severity**
S3

JohnnyWuLinksys:
Passed on FW_Pinnacle2.0_v1.2.3.26072618_USP_release.img
gui_version='2.6.0.700535'
Close this ticket as fixed.

PR #1304, opened 2026-08-18
fix(dashboard): keep each breakpoint's layout to itself, and leave edit mode on logout (#1293, #1294)

### LINKED ISSUE #1293: [Dashboard] Custom card size saved in mobile layout breaks the desktop layout
**Steps:**
1. Login.
2. Resize the browser window to a mobile viewport
3. Enter dashboard edit mode, change 1 card size, and save
4. Log out, then refresh the page.
5. Resize the browser window to a desktop viewport and log in again.

**Actual result:**
The dashboard layout is broken

**Expected result:**
The desktop layout is not affected by the card size saved in the mobile layout.

**Device:**
Pinnacle 2.0 FW: 1.2.3.26070802
Linksys Now Version: 2.7.0

**Additional context**

Image

### LINKED ISSUE #1294: [Dashboard] Dashboard stays in edit mode after logging out and logging in again
**Steps:**
1. Log in.
2. Enter dashboard edit mode.
3. Log out from the UI (do NOT refresh the browser), then log in again.

**Actual result:**
The dashboard is still in edit mode

**Expected result:**
Edit mode is exited when logging out, so the dashboard is in normal
(read-only) mode after logging in again.

AustinChangLinksys:
Closes #1293
Closes #1294

> **Stacked PR — do not squash-merge.** This targets `gate/1183-overflow-gate` (#1248), and #1299 is stacked on top of *this* branch. Squashing rewrites the SHAs the upper branch was cut from, which turns #1299 into a duplicate diff with false conflicts. Merge bottom-up with a **merge commit** or **rebase merge**: #1248 first, then this, then #1299.

## #1293 — resizing a card on a phone wrecked the desktop dashboard, permanently

Two separate defects hid behind one report.

**The layout pref had no idea which grid wrote it.** `exportLayout()` returns coordinates in whatever slot count the controller is currently on, and the pref was a bare list, always read back as 12 columns. A save at mobile came back at desktop as third-width cards whose `minW`/`maxW` had been scaled down with them — and those caps then blocked widening them again, which is what made it permanent.

The pref becomes an envelope keyed by slot count (`UspLayoutEnvelope`; legacy bare lists decode as the desktop entry, so no one loses their dashboard on upgrade), and save/load walk every breakpoint on the live controller so geometry stays per-grid. Membership deliberately does *not* — deleting a card on a phone deletes the card. `alignMembership` covers the case the package would otherwise get wrong: a stored narrower entry written before a card existed would reconcile that card *out* of the wider grids.

Resize enforcement also moves out of the view into `UspWidgetSpecs.correctedSize`, which scales the spec's 12-column figures to the grid actually in front of the user. Read literally, `stats_panel`'s `minColumns: 6` was being enforced as six of a tablet's *eight* columns, and as a width wider than a 4-column grid has.

**On a phone, width is not supposed to be editable at all — and pinning the caps only half achieved that.** At 4 columns every card is pinned full-width (`minW == maxW == slotCount`), leaving height and order. That holds the right-hand handles, because `sliver_dashboard` clamps a resize delta to `[minW, maxW]` and a width that is already both floor and ceiling cannot move. The left-hand handles move `x` as well, and the package fits the item to the row *after* that clamp:

```
newX = x + dW;  newW = (w - dW).clamp(minW, maxW)
if (newX < 0) { newW += newX; newX = 0; }
if (newX + newW > cols) newW = cols - newX
```

so one column of drag on a card's left edge narrows it to three either way — to `x: 1, w: 3` inwards, and to `x: 0, w: 3` outwards, where the row's own edge does the trimming. No cap can express "x may not move", and the package builds all eight handles unconditionally (its one switch, `isResizable`, would take the height with it), so the correction has to happen on the live layout.

`UspSliverDashboardControllerNotifier` now subscribes to the controller's layout beacon and puts any horizontal change on the phone grid back. Subscribing rather than snapping back at gesture end is what makes the drag look *inert* instead of rubber-banding: the correction is queued alongside the rebuild the same write triggers and runs first, because it is registered at construction, before any widget starts watching. The write goes straight to the beacon — `importLayout` would compact the whole grid mid-gesture and close gaps the user is still making. Height and ordering are untouched. The lock belongs to the instance it watches, so every controller swap goes through `_swapController`, which re-arms it.

## #1294 — a new session opened mid-edit, holding the previous session's snapshot

Logging out from the dashboard's own menu neither navigates nor reloads, so the `onExit` route guard never fires and the root-scoped edit state survived. `DashboardEditModeNotifier` now watches for the logged-in → logged-out edge and reverts, matching #1037's silent-discard policy. The edge is tracked locally because `logout()` emits a value-less `AsyncLoading` before the logged-out value.

## Verification

Run on this branch's head (`eda9f28c`):

- `./run_tests.sh` — **6531 passed**
- `flutter test --tags dashboard-card` — **2744 passed**
- `flutter test test/golden_test/page/dashboard/localizations/usp_dashboard_view_test.dart --tags golden` — **6 passed**
- `flutter analyze` — no errors or warnings in changed files

New coverage: `usp_layout_envelope_test.dart`, `usp_widget_specs_test.dart`, `usp_layout_breakpoint_persistence_test.dart` (the round-trip across all three breakpoints, plus the legacy-bare-list migration), and additions to `usp_layout_controller_test.dart` and `dashboard_edit_mode_provider_test.dart`.

### DIFF
--- a/lib/page/dashboard/models/usp_layout_envelope.dart
+++ b/lib/page/dashboard/models/usp_layout_envelope.dart
@@ -1,0 +1,112 @@
+import 'dart:convert';
+
+/// The persisted form of the USP dashboard layout: one serialised grid per
+/// breakpoint, keyed by slot count.
+///
+/// ## Why the slot count has to be recorded
+///
+/// [DashboardController.exportLayout] returns coordinates in whatever slot count
+/// the controller is on at that moment. The pref used to hold a bare list of
+/// those coordinates with no note of which grid produced them, and load always
+/// read them back as 12-column values. A save made on a phone therefore came
+/// back on a laptop as a 12-column layout of 4-column cards — a third of their
+/// width, with `minW`/`maxW` scaled down to match, which made the damage
+/// permanent because those caps then blocked any attempt to widen the cards
+/// again (#1293).
+///
+/// ## Compatibility
+///
+/// A legacy bare list decodes as the [desktopSlotCount] entry — that is what it
+/// always was, since load created a 12-slot controller before importing. The
+/// other breakpoints are left absent rather than guessed, so the caller can tell
+/// "the user arranged this" from "we derived this" and re-derive the second kind
+/// whenever the scaling rules change.
+class UspLayoutEnvelope {
+  /// Slot counts the dashboard renders and persists, matching the ui_kit
+  /// breakpoint regimes (`GridLayoutContext.currentMaxColumns`).
+  static const int desktopSlotCount = 12;
+  static const int tabletSlotCount = 8;
+  static const int mobileSlotCount = 4;
+
+  /// Ordered widest-first: the desktop entry is the source the narrower grids
+  /// are derived from, so it has to be read and seeded first.
+  static const List<int> persistedSlotCounts = [
+    desktopSlotCount,
+    tabletSlotCount,
+    mobileSlotCount,
+  ];
+
+  /// Bumped whenever the payload shape changes in a way older builds would
+  /// misread. Version 1 is the implicit version of the legacy bare list.
+  static const int currentVersion = 2;
+
+  const UspLayoutEnvelope(this.layouts);
+
+  /// Serialised layouts by slot count. Keys outside [persistedSlotCounts] are
+  /// preserved on decode so a build that renders fewer breakpoints cannot
+  /// silently discard a layout a newer build wrote.
+  final Map<int, List<dynamic>> layouts;
+
+  List<dynamic>? operator [](int slotCount) => layouts[slotCount];
+
+  List<int> get slotCounts => layouts.keys.toList();
+
+  /// Returns a copy with [slotCount]'s geometry replaced.
+  UspLayoutEnvelope withLayout(int slotCount, List<dynamic> layout) =>
+      UspLayoutEnvelope({...layouts, slotCount: layout});
+
+  String encode() => jsonEncode({
+        'version': currentVersion,
+        'layouts': {
+          for (final entry in layouts.entries)
+            entry.key.toString(): entry.value,
+        },
+      });
+
+  /// Parses [raw], or returns null when it cannot be read as layouts on known
+  /// grids.
+  ///
+  /// Null means "fall back to the default layout", so the bar is deliberately
+  /// low: anything importable is imported, including ids this build does not
+  /// ship (they may be package widgets whose specs load later). What is rejected
+  /// is only what cannot be placed at all — a payload from a future version, a
+  /// slot count that is not a number, or items that are not maps. Importing
+  /// those would either throw inside the grid or render a layout the user never
+  /// arranged, and both are worse than starting from the default.
+  static UspLayoutEnvelope? tryDecode(String raw) {
+    Object? decoded;
+    try {
+      decoded = jsonDecode(raw);
+    } catch (_) {
+      return null;
+    }
+
+    // Legacy: a bare list of 12-column items.
+    if (decoded is List) {
+      if (!_isItemList(decoded)) return null;
+      return UspLayoutEnvelope({desktopSlotCount: decoded});
+    }
+
+    if (decoded is! Map) return null;
+
+    final version = decoded['version'];
+    if (version is! int || version > currentVersion) return null;
+
+    final rawLayouts = decoded['layouts'];
+    if (rawLayouts is! Map) return null;
+
+    final layouts = <int, List<dynamic>>{};
+    for (final entry in rawLayouts.entries) {
+      final slotCount = int.tryParse('${entry.key}');
+      if (slotCount == null || slotCount < 1) return null;
+      final layout = entry.value;
+      if (layout is! List || !_isItemList(layout)) return null;
+      layouts[slotCount] = layout;
+    }
+
+    return UspLayoutEnvelope(layouts);
+  }
+
+  static bool _isItemList(List<dynamic> layout) =>
+      layout.every((item) => item is Map);
+}
--- a/lib/page/dashboard/models/usp_widget_specs.dart
+++ b/lib/page/dashboard/models/usp_widget_specs.dart
@@ -1,5 +1,6 @@
 import 'package:ui_kit_library/ui_kit.dart';
 import 'package:privacy_gui/page/dashboard/models/display_mode.dart';
+import 'package:privacy_gui/page/dashboard/models/usp_layout_envelope.dart';
 import 'package:privacy_gui/page/dashboard/models/widget_spec.dart';
 import 'package:privacy_gui/page/dashboard/providers/layout_item_factory.dart';
 import 'package:sliver_dashboard/sliver_dashboard.dart';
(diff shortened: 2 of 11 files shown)


AustinChangLinksys (commented):
🤖 **Automated Review — Round 1** · `ed8a3c2..eda9f28` (full)

**Verdict: 💬 Self-review (comment only)** — `is_own_pr=true, is_dispatch_chain=false`; GitHub prohibits self-approve; automated flow does not endorse Austin's own PR. Review provided for reference/blind-spot detection.

---

#### Standards

|---|---|---|---|
| ⚠️ | 🟢High | `dashboard_edit_mode_provider.dart:76–78` | [A+B] `cancelEditMode()` on logout reverts layout from single-breakpoint snapshot; if an async `commitEditMode` is in flight when logout fires, the finally-cleared `isEditing` race means cancel can wipe a just-saved layout |
| ⚠️ | 🟡Med | `dashboard_edit_mode_provider.dart:67–79` | [B] `_exitEditMode` (async) fired-and-forgotten inside `ref.listen` callback lacks re-entrancy guard; concurrent commit + logout can both enter `_exitEditMode` simultaneously |
| ⚠️ | 🟢High | `usp_layout_controller.dart:244–263` | [A+B] `_exportAllBreakpoints()` mutates `controller.slotCount` as side-effect during a read; name does not signal this; can transiently notify layout listeners mid-walk |
| ⚠️ | 🟢High | `usp_layout_envelope.dart:93` | [A] `version > currentVersion` silently discards *all* saved layouts on rollback; no partial-decode fallback |
| 💡 | 🟢High | `usp_sliver_dashboard_view.dart:495–499` | [B] View calls `UspWidgetSpecs.correctedSize()` and makes the resize/save decision; business logic for constraint correction belongs in Provider/Notifier |
| 💡 | 🟢High | `usp_layout_envelope.dart:54–56` | [B] `withLayout()` has no production caller (only tests); [Speculative Generality] |
| 💡 | 🟡Med | `usp_widget_specs.dart:638–640` | [A] `scaleLayout` max-clamp defence for `minW>maxW` edge is undocumented; intent differs subtly from `correctedSize` |
| 💡 | 🟡Med | `dashboard_edit_mode_provider.dart:66–79` | [A] `fireImmediately` seeding guard relies on `wasLoggedIn=false`; if `authProvider` initial state ever becomes `isLoggedIn=true`, callback reads `state` during `build()` before assignment |
| 💡 | ⚪Low | `usp_widget_specs.dart:734–736` | [A] `alignMembership` id-to-item map silently takes last entry on duplicate id; harmless in practice |
| 💡 | 🟡Med | `test/.../usp_layout_breakpoint_persistence_test.dart:38` | [B] Fixed 100 ms `pumpAsync()` delay is brittle; should await notifier `initialized` future |
| 💡 | ⚪Low | `test/.../dashboard_edit_mode_provider_test.dart:350–363` | [B] `_FakeAuthNotifier` inherits full `AuthNotifier` but uses only `build()`; [Refused Bequest] — acceptable given Riverpod's `overrideWith` constraint |

---

#### Spec

|---|---|---|---|
| ⚠️ | 🟢High | `dashboard_edit_mode_provider.dart:77` → spec #1294 expected | Logout calls `cancelEditMode()` (revert), but spec only says "exit edit mode" — revert-on-logout can silently undo a user's just-saved resize in a `commitEditMode` async-gap race |
| ⚠️ | 🟢High | `dashboard_edit_mode_provider.dart:133` → spec #1293 fix | `cancelEditMode` revert uses non-breakpoint-aware `importLayout(snapshot)` then `saveLayout()`; snapshot is single-breakpoint, so the walk in `_exportAllBreakpoints` may re-store stale coords for other breakpoints, partially undermining the #1293 fix |
| ⚠️ | 🟡Med | `test/.../usp_layout_breakpoint_persistence_test.dart` → spec #1293 step 4–5 | Spec step 4 requires "log out, then **refresh the page**"; only unit-level prefs round-trip is tested; no golden/integration test verifies the desktop UI *renders* the correct layout post-refresh |
| 💡 | 🟢High | `usp_widget_specs.dart:540+` | Scope creep: `correctedSize`, `scaleSpan`, `alignMembership` are not mentioned in spec #1293/#1294; fine if intentional, but no discussion found in either issue |
| 💡 | 🟡Med | `dashboard_edit_mode_provider.dart:68` → spec #1294 step 3 | Cold-init edge: if provider is first instantiated *after* logout, `wasLoggedIn` starts false, so the cancel is not triggered; out of spec scope but untested boundary |

---

**[A+B — Standards] `cancelEditMode()` on logout — single-breakpoint snapshot race** · `dashboard_edit_mode_provider.dart:76–78` 🟢High

- **Location**: `dashboard_edit_mode_provider.dart:76–78` (head)
- **Code**:
```dart
// dashboard_edit_mode_provider.dart:100–101
final controller = ref.read(uspSliverDashboardControllerProvider);
final layoutSnapshot = controller.exportLayout();   // ← single breakpoint only
```
```dart
// dashboard_edit_mode_provider.dart:76–78
if (loggedOutJustNow && state.isEditing) {
  cancelEditMode();   // → _exitEditMode(revert: true): importLayout(snapshot) + saveLayout()
}
```
- **Why it matters**: `enterEditMode()` snapshots only the *current* breakpoint at the time of entry. If the user is at tablet-8-col when entering edit mode, moves to desktop, resizes, commits (async `saveLayout` in flight), then logout fires during that `await` — `state.isEditing` is still `true` (cleared only in `finally`). `cancelEditMode()` then imports the stale 8-col snapshot and calls `saveLayout()`, which walks all breakpoints from that stale base, silently overwriting the just-saved desktop layout.
- **Fix**: On logout, prefer `commitEditMode()` (exit flag only, no revert). If intentional revert-on-logout is wanted, capture a full `UspLayoutEnvelope` snapshot at `enterEditMode()`, and restore it breakpoint-aware via `_replaceController`.

---

**[B — Standards] `_exitEditMode` re-entrancy** · `dashboard_edit_mode_provider.dart:127` 🟡Med

- Async `cancelEditMode()` is fire-and-forgotten inside `ref.listen`. If a concurrent `commitEditMode()` is also in flight (e.g. route guard), both enter `_exitEditMode` simultaneously; the `finally` block runs twice, second run may `importLayout` already-cleared state.
- **Fix**: Add `if (!state.isEditing) return;` at the top of `_exitEditMode`, or use `unawaited()` + comment to make fire-and-forget intent explicit.

---

**[A+B — Standards] `_exportAllBreakpoints` mutating side-effect** · `usp_layout_controller.dart:251–262` 🟡Med

- Method name suggests a read, but `controller.setSlotCount(slots)` is called for each breakpoint — a ValueNotifier mutation that can synchronously notify layout watchers mid-walk.
- Pre-existing smell (exists before this PR), but this PR extends its call sites (`saveLayout`).
- **Fix**: Rename to `_walkAndExportAllBreakpoints()` or add a `/// WARNING: mutates slotCount as side-effect` javadoc.

---

**[A — Standards] `version > currentVersion` discards all saved layouts on app rollback** · `usp_layout_envelope.dart:93` 🟢High

```dart
// usp_layout_envelope.dart:93
if (version is! int || version > currentVersion) return null;
```
- Rollback from future app version → all breakpoint layouts silently reset to defaults.
- **Fix**: Attempt partial decode (read only keys in `persistedSlotCounts`), skip unknown keys; reserve full-discard for structurally incompatible payloads.

---

**[Spec C-1] `cancelEditMode()` on logout may silently undo committed layout** · `dashboard_edit_mode_provider.dart:77` 🟢High

Spec #1294 says: *"Edit mode is exited when logging out."* — says nothing about reverting. `cancelEditMode()` reverts to pre-edit snapshot. In the async-gap race described above, a just-committed save is lost.

---

**[Spec C-3] `importLayout(snapshot)` is not breakpoint-aware** · `dashboard_edit_mode_provider.dart:133` 🟢High

```dart
// dashboard_edit_mode_provider.dart:131–137
if (state.layoutSnapshot != null) {
  controller.importLayout(state.layoutSnapshot!);   // ← single-bp flat list
  await ref.read(uspSliverDashboardControllerProvider.notifier).saveLayout();
}
```
`importLayout` sets the *current* breakpoint only; subsequent `saveLayout → _exportAllBreakpoints` walks other breakpoints from whatever state the controller was left in after the import, which may not reflect pre-edit geometry for those breakpoints. This partially undermines the #1293 per-breakpoint isolation fix.

</details>

- **`UspLayoutEnvelope` migration path**: legacy bare-list prefs are transparently decoded as desktop entry — no user data lost on upgrade. Well-tested in `usp_layout_envelope_test.dart`.
- **`_widthLockGuard` dispose safety** (`usp_layout_controller.dart:232–234`): `dispose()` cancels the subscription; no listener leak.
- **`_exportAllBreakpoints` origin restoration** (`usp_layout_controller.dart:261`): `setSlotCount(origin)` called synchronously after the walk; controller never left at a wrong breakpoint.
- **`_exitEditMode` try/finally** (`dashboard_edit_mode_provider.dart:144–147`): edit mode is always cleared even if `saveLayout` throws; no stuck-in-edit-mode zombie state.
- **`lockItemsToFullWidth` returns `null` on no-change** (`usp_widget_specs.dart:703–715`): avoids re-triggering layout watcher on the same value — subtle but correct.
- **Test coverage breadth**: new `usp_layout_breakpoint_persistence_test.dart` round-trips all three breakpoints including the legacy migration path. `usp_layout_controller_test.dart` and `dashboard_edit_mode_provider_test.dart` both extended to cover new paths.
- **`authProvider` loading edge handled**: `next.valueOrNull` returns null on `AsyncLoading`/`AsyncError`; logout listener ignores transient loading emissions correctly.
- **Provider L1/L2 rules**: `dashboardEditModeProvider` uses `NotifierProvider` (not autoDispose); controller reads use `ref.read` in mutation paths — no L2 violations found.

</details>

PR #1306, opened 2026-08-19
fix(remote-assistance): send every RA call to the Guardian host

### LINKED ISSUE #1307: Remote Assistance: subscription/SSE requests leave for the web app origin instead of the Guardian API host
## Summary

In Remote Assistance (RA) mode the app talks to **two different hosts** for what is logically one session:

- USP `get` / `set` / `operate` → `https://qa.guardian.tools/v1/guardians/remote-assistances/sessions/<id>/actions/usp` (correct)
- subscription register/unregister + SSE notifications → the origin the web app is served from, e.g. `https://qa-usp.remote-support.guardian.tools/v1/guardians/remote-assistances/sessions/<id>/usp/subscriptions` (wrong)

Captured request (real QA session, agent Bearer token, RA subscription path — but the app's own origin):

```
POST https://qa-usp.remote-support.guardian.tools/v1/guardians/remote-assistances/sessions/<id>/usp/subscriptions
authorization: Bearer AGENT…
x-linksys-client-type-id: BB426FA7-16A9-5C1C-55AF-63A4167B26AD
{"action":"unregister","subscription_id":"remote-ethernet-valuechange"}
```

That host is a static S3/CloudFront deployment of the Flutter app (`<base href="/app/">`), not a Guardian proxy, so the call cannot succeed:

| Request | App origin (`qa-usp.remote-support.guardian.tools`) | Guardian API (`qa.guardian.tools`) |
|---|---|---|
| `POST /v1/guardians/…/usp/subscriptions` | `502` | `404 RESOURCE_NOT_FOUND` — real Guardian error, token + session resolved |
| `POST /v1/guardians/…/actions/usp` | — | `415` — endpoint exists, authenticated |
| `POST /api/v1/usp` | `200 text/html` (SPA fallback — silent failure) | — |

Net effect: RA sessions run without any value-change notifications; the UI never refreshes from SSE and the failure is silent (CloudFront answers `200 text/html` for the WASM client's default path).

## Root cause

Two independent sources decide the host:

| | get / set / operate | subscription / SSE |
|---|---|---|
| Transport | WASM client built by `UspClientBuilder` | Dart `UspBridgeClient` (`http` + Fetch/ReadableStream) |
| Host source | baked in at `activate()` from the RA config | `UspClient.baseUrl` (Dart-side field) |

`lib/core/usp/services/usp_bridge_client_web.dart` built every REST/SSE URL as `_usp.baseUrl + path`, and the `UspClient` it received came from `uspClientProvider` — a plain, permanently cached `Provider` that returns `getIt<UspClient>()` (`lib/core/usp/providers/usp_client_provider.dart:15`).

`authProvider.init()` reads that provider during app startup (`lib/app.dart:67` `_initAuth()` → `lib/providers/auth/auth_provider.dart:57`), long before RA activates, which pins the **boot-time** client registered with `Uri.base.origin` (`lib/di.dart:46-55`). `RemoteAssistanceNotifier.activate()` then swaps the GetIt singleton and calls `dispose()` on the old client — but nothing invalidated the provider, so:

- the bridge kept building URLs from the boot client's `baseUrl` → the app's own origin;
- the same stale instance is also what `get` consumers hold, and its WASM object was already `free()`d by `activate()`.

Verified against the deployed QA bundle (`2.6.0.600010`): the minified `activate()` frees the old client and contains no provider invalidation, and the bridge factory is compiled as `new UspBridgeClient(m, endpoints…)` where `m` is the `uspClientProvider` value — no explicit host.

Historically the same symptom had an earlier cause: `UspClient.fromBuilder` set `_baseUrl = ''` (fixed in f75a4ff6, 2026-06-26), which made the bridge issue relative URLs that the browser resolved against the page origin.

## Expected

All RA traffic — session REST, USP requests, subscriptions, SSE notifications — uses the Guardian API host from the RA config.

## Fix (PR #1306)

- `RemoteAssistanceConfig.guardianOrigin` — single accessor for the host all RA traffic must use.
- `UspBridgeClient` takes an explicit `baseUrl`; the remote branch of `uspBridgeClientProvider` passes `config.guardianOrigin`, so the host can no longer be inferred from `UspClient.baseUrl`.
- `activate()` / `deactivate()` invalidate `uspClientProvider` after the GetIt swap, so nothing keeps calling into a freed WASM client.

## Acceptance criteria

- [ ] During an RA session, DevTools shows `actions/usp` **and** `usp/subscriptions` both on the Guardian API host.
- [ ] SSE notifications stream connects and value-change events refresh the UI.
- [ ] Swapping the client on `activate()` leaves no consumer holding the boot-time client.

### LINKED ISSUE #1309: Remote builds register a boot-time UspClient that can never reach a valid host
## Summary

Follow-up from the review of #1306 (raised by @HankYuLinksys).

`dependencySetup()` registers a `UspClient` unconditionally at boot:

```dart
// lib/di.dart:46-55
if (kIsWeb && !getIt.isRegistered<UspClient>()) {
  getIt.registerSingleton<UspClient>(UspClient(Uri.base.origin));
}
```

`GlobalConfig.remote.isActive` is `BuildConfig.isRemote()` → `forceCommandType == ForceCommand.remote`, resolved from `const String.fromEnvironment('force')` — i.e. **fixed at build time**. So in a Remote build that boot-time client can never point at a usable host:

- its only observable effects are making the boot-time `restoreSession()` talk to the web app's own origin, and
- creating the stale-instance problem #1306 had to patch: `uspClientProvider` caches it on first read (`authProvider.init()`, during startup), and RA's `activate()` then swaps and disposes it.

#1306 closes the window by invalidating `uspClientProvider` inside the swap, but that fix carries an invariant: invalidation only reaches `ref.watch` consumers, so the ~40 feature services that capture the client with `ref.read(uspClientProvider)` in a non-autoDispose provider body keep their first instance for the container's lifetime. The swap therefore must happen before any feature service provider is first read. Nothing enforces that ordering today.

## Options

1. Skip the GetIt registration in `dependencySetup()` when `BuildConfig.isRemote()` — no boot client exists, so nothing can capture the wrong one.
2. Have `uspClientProvider` build the client itself (from the RA config in Remote mode, from `Uri.base.origin` otherwise) instead of pulling it out of GetIt, making the provider the single owner of client identity.

Either removes the whole class of bug and makes the invalidation in `activate()` unnecessary.

## Acceptance criteria

- [ ] In a Remote build, no `UspClient` is created against the web app's own origin.
- [ ] Swapping in the RA client cannot leave any consumer holding a pre-swap instance, regardless of when feature providers are first read.
- [ ] Local (on-router) builds are behaviourally unchanged.
- [ ] The ordering invariant documented in `remote_assistance_provider.dart` is either enforced by construction or no longer needed.

AustinChangLinksys:
## Problem

In Remote Assistance mode, subscription and SSE notification requests were sent to the origin the web app is served from (e.g. `https://qa-usp.remote-support.guardian.tools`), while USP `get`/`set` went to the Guardian API host (`https://qa.guardian.tools`). On a static-hosted deployment (S3/CloudFront) those subscription `POST`s never reach Guardian: the app origin answers `502`, and `POST /api/v1/usp` there returns `200 text/html` (SPA fallback), so the failure is silent rather than loud.

Two independent sources decided the host for what is logically one session:

| | get / set / operate | subscription / SSE |
|---|---|---|
| Transport | WASM client (`UspClientBuilder`) | Dart `UspBridgeClient` (http + Fetch/ReadableStream) |
| Host source | baked in at `activate()` from the RA config | derived from `UspClient.baseUrl` |

The instance the bridge received was the **boot-time** client registered in `di.dart` with `Uri.base.origin`: `uspClientProvider` is a plain (cached) `Provider` returning `getIt<UspClient>()`, and `authProvider.init()` reads it during app startup (`app.dart` `initState` -> `_initAuth()`), long before RA activates. `activate()` swaps and disposes the GetIt singleton, but nothing invalidated the provider — so every consumer kept the stale instance, whose WASM object `activate()` had just `free()`d.

## Fix

- `RemoteAssistanceConfig.guardianOrigin` — one accessor for the host all USP-over-Guardian traffic must use (USP requests, subscriptions, SSE).
- `UspBridgeClient` accepts an explicit `baseUrl`; the remote branch of `uspBridgeClientProvider` passes `config.guardianOrigin`, so the host can no longer be inferred from `UspClient.baseUrl`.
- `activate()` / `deactivate()` invalidate `uspClientProvider` inside the `withLock` body that swaps the GetIt singleton, so requests stop going through a disposed client and the two steps cannot drift apart.
- **No boot-time client in Remote builds** (#1309): `dependencySetup()` only registers `UspClient(Uri.base.origin)` when the app origin can actually serve USP — i.e. Local and unforced builds, where the router serves the app. In a Remote build the slot stays empty until `activate()` fills it, so `activate()` is the only thing that can ever create the client and no consumer can capture a wrong-host instance regardless of read order. The decision is a single testable predicate, `canUseAppOriginUspClient()`.

## Scope

- **Session REST is not routed through `guardianOrigin`.** `GuardianApiClient._buildUrl` still builds `https://${cloudEnvironmentConfig[kCloudBase]}…` itself. It lands on the same host by construction — the RA config's `guardianBaseUrl` is set from that same entry (`remote_assistance_confirm_view.dart:201`) — but it remains a separate code path.
- **Static assets intentionally stay on the app origin.** `package_widget_loader.dart`, `package_widget_renderer.dart`, `usp_apps_service.dart` and `usp_apps_view.dart:99` keep using `Uri.base.origin`; they fetch bundled web assets, not Guardian APIs. The title's "every RA call" means every USP/Guardian call.
- **Ordering invariant, now enforced by construction.** Invalidation only reaches `ref.watch` consumers; feature services capture the client with `ref.read(uspClientProvider)` in the body of a non-autoDispose provider (~40 sites) and keep their first instance for the container's lifetime. With no boot client in a Remote build, the only instance that can exist is the session one, so the worst a too-early read can capture is `null` — fails loudly instead of silently talking to a host that answers no USP request. All 16 `ref.read(uspClientProvider)!` sites are feature services under `lib/page/**/services/`, first read after the dashboard route, which RA activation gates.
- **Local behaviour unchanged**: same registration, same origin, same lifetime.

## Verification

Probed with a live QA session token:

- `POST https://qa.guardian.tools/v1/guardians/remote-assistances/sessions/<id>/usp/subscriptions` -> `404 {"code":"RESOURCE_NOT_FOUND", ... "of type Subscription cannot be found."}` — endpoint exists, token authenticates, session resolves.
- `POST https://qa.guardian.tools/v1/guardians/remote-assistances/sessions/<id>/actions/usp` -> `415` — exists and authenticated.
- Same paths on the app origin -> `502` / `200 text/html`.

Local checks: `dart format` clean, `flutter analyze` reports only a pre-existing info in an unrelated test file, and the affected + dependent suites (`test/core/usp`, `test/core/connection`, auth notifier, new `test/di_test.dart`) pass — 453 tests, including a build-count test that asserts `deactivate()` really rebuilds `uspClientProvider` and three cases pinning `canUseAppOriginUspClient()` per build mode.

## QA note

In DevTools during an RA session, both `actions/usp` and `usp/subscriptions` should now show `qa.guardian.tools`. In a Remote build, boot logs `[DI]: UspClient boot registration skipped` instead of `[DI]: UspClient registered (Web)`.

Closes #1307
Closes #1309

### DIFF
--- a/lib/core/usp/providers/remote_assistance_provider.dart
+++ b/lib/core/usp/providers/remote_assistance_provider.dart
@@ -127,15 +127,15 @@ class RemoteAssistanceNotifier extends Notifier<RemoteAssistanceState> {
       //
       // uspClientProvider caches whatever GetIt held when it was FIRST read,
       // and authProvider.init() reads it during app boot — long before RA
-      // activates. Without this, every USP request and every bridge call keeps
-      // using the disposed boot-time client, whose baseUrl is the web app's own
-      // origin instead of the Guardian API host.
+      // activates. In a Remote build that first read now yields null
+      // ([canUseAppOriginUspClient] keeps the boot slot empty), and this drops
+      // that cached null so watchers pick the session client up.
       //
-      // Invariant: only ref.watch consumers rebuild. Feature services capture
-      // the client via ref.read(uspClientProvider) in the body of a
-      // non-autoDispose provider, so they keep their first instance for the
-      // container's lifetime — the swap MUST happen before any feature service
-      // provider is first read.
+      // Only ref.watch consumers rebuild; a ref.read consumer keeps whatever it
+      // captured. That is no longer a correctness risk: in a Remote build the
+      // only client that can ever exist is this one, so the worst a too-early
+      // read can capture is null — which fails loudly instead of silently
+      // talking to the wrong host.
       ref.invalidate(uspClientProvider);
     });
 
--- a/lib/core/usp/providers/remote_assistance_provider.dart
+++ b/lib/core/usp/providers/remote_assistance_provider.dart
@@ -1,6 +1,7 @@
 import 'package:flutter/foundation.dart';
 import 'package:flutter_riverpod/flutter_riverpod.dart';
 import 'package:privacy_gui/constants/cloud_const.dart';
+import 'package:privacy_gui/core/usp/providers/usp_client_provider.dart';
 import 'package:privacy_gui/core/usp/providers/usp_mutation_lock.dart';
 import 'package:privacy_gui/core/usp/services/usp_client.dart';
 import 'package:privacy_gui/core/utils/logger.dart';
@@ -20,6 +21,13 @@ class RemoteAssistanceConfig {
     required this.temporaryAccessToken,
     this.clientTypeId,
   });
+
+  /// Guardian API origin — the single host for ALL Remote Assistance traffic
+  /// (session REST, USP requests, subscriptions, SSE notifications).
+  ///
+  /// Must NOT be confused with the origin the web app is served from; the
+  /// Guardian API lives on a different host (e.g. `qa.guardian.tools`).
+  String get guardianOrigin => 'https://$guardianBaseUrl';
 
   /// Constructs the Guardian USP endpoint path.
   String get uspEndpoint =>
@@ -75,11 +83,11 @@ class RemoteAssistanceNotifier extends Notifier<RemoteAssistanceState> {
     }
 
     logger.i('[RA] Activating Remote Assistance: ${config.sessionId}');
-    logger.d('[RA] Guardian URL: https://${config.guardianBaseUrl}');
+    logger.d('[RA] Guardian URL: ${config.guardianOrigin}');
     logger.d('[RA] USP Endpoint: ${config.uspEndpoint}');
 
     // Build the client using UspClientBuilder
-    var builder = UspClientBuilderJS('https://${config.guardianBaseUrl}')
+    var builder = UspClientBuilderJS(config.guardianOrigin)
         .endpoint(config.uspEndpoint)
         .authToken(config.temporaryAccessToken);
 
@@ -89,8 +97,8 @@ class RemoteAssistanceNotifier extends Notifier<RemoteAssistanceState> {
     }
 
     final jsClient = builder.build();
-    final guardianUrl = 'https://${config.guardianBaseUrl}';
-    final raClient = UspClient.fromBuilder(jsClient, baseUrl: guardianUrl);
+    final raClient =
+        UspClient.fromBuilder(jsClient, baseUrl: config.guardianOrigin);
 
     // Swap UspClient atomically with mutation lock to prevent races
     await ref.read(uspMutationLockProvider).withLock(() async {
@@ -111,6 +119,13 @@ class RemoteAssistanceNotifier extends Notifier<RemoteAssistanceState> {
       isActive: true,
       config: config,
     );
+
+    // uspClientProvider caches whatever GetIt held when it was FIRST read
+    // (authProvider.init() reads it during app boot, before RA activates).
+    // Without this invalidation, every USP request and every bridge call keeps
+    // using the disposed boot-time client — whose baseUrl is the web app's own
+    // origin, not the Guardian API host.
+    ref.invalidate(uspClientProvider);
   }
 
   /// Deactivates Remote Assistance mode.
@@ -131,6 +146,10 @@ class RemoteAssistanceNotifier extends Notifier<RemoteAssistanceState> {
     });
 
     state = const RemoteAssistanceState();
+
+    // Drop the cached (now disposed) client so downstream providers tear down
+    // instead of calling into freed WASM memory.
+    ref.invalidate(uspClientProvider);
   }
 }
 
--- a/lib/core/usp/providers/remote_assistance_provider.dart
+++ b/lib/core/usp/providers/remote_assistance_provider.dart
@@ -22,12 +22,23 @@ class RemoteAssistanceConfig {
     this.clientTypeId,
   });
 
-  /// Guardian API origin — the single host for ALL Remote Assistance traffic
-  /// (session REST, USP requests, subscriptions, SSE notifications).
+  /// Guardian API origin — the host every USP-over-Guardian call must use:
+  /// USP requests, subscriptions and SSE notifications.
   ///
   /// Must NOT be confused with the origin the web app is served from; the
   /// Guardian API lives on a different host (e.g. `qa.guardian.tools`).
-  String get guardianOrigin => 'https://$guardianBaseUrl';
+  ///
+  /// The session REST API ends up on the same host by construction — both this
+  /// config ([RemoteAssistanceConfig.guardianBaseUrl], set from
+  /// `cloudEnvironmentConfig[kCloudBase]`) and `GuardianApiClient._buildUrl`
+  /// read that same entry — but it builds its URL independently.
+  String get guardianOrigin {
+    // Kept as an assert rather than a constructor assert: the constructor is
+    // const, and const asserts only accept potentially-constant expressions.
+    assert(!guardianBaseUrl.contains('://'),
+        'guardianBaseUrl must be a bare host, without a scheme');
+    return 'https://$guardianBaseUrl';
+  }
 
   /// Constructs the Guardian USP endpoint path.
   String get uspEndpoint =>
@@ -110,6 +121,22 @@ class RemoteAssistanceNotifier extends Notifier<RemoteAssistanceState> {
       }
       getIt.registerSingleton<UspClient>(raClient);
       logger.i('[RA] UspClient replaced with Guardian-proxied client');
+
+      // Invalidated in the same critical section as the swap, so "client
+      // identity changed" and "cache dropped" can never drift apart.
+      //
+      // uspClientProvider caches whatever GetIt held when it was FIRST read,
+      // and authProvider.init() reads it during app boot — long before RA
+      // activates. Without this, every USP request and every bridge call keeps
+      // using the disposed boot-time client, whose baseUrl is the web app's own
+      // origin instead of the Guardian API host.
+      //
+      // Invariant: only ref.watch consumers rebuild. Feature services capture
+      // the client via ref.read(uspClientProvider) in the body of a
+      // non-autoDispose provider, so they keep their first instance for the
+      // container's lifetime — the swap MUST happen before any feature service
+      // provider is first read.
+      ref.invalidate(uspClientProvider);
     });
 
     // Set login type to remote so auth checks pass
@@ -119,13 +146,6 @@ class RemoteAssistanceNotifier extends Notifier<RemoteAssistanceState> {
       isActive: true,
       config: config,
     );
-
-    // uspClientProvider caches whatever GetIt held when it was FIRST read
-    // (authProvider.init() reads it during app boot, before RA activates).
-    // Without this invalidation, every USP request and every bridge call keeps
-    // using the disposed boot-time client — whose baseUrl is the web app's own
-    // origin, not the Guardian API host.
-    ref.invalidate(uspClientProvider);
   }
 
   /// Deactivates Remote Assistance mode.
@@ -143,13 +163,13 @@ class RemoteAssistanceNotifier extends Notifier<RemoteAssistanceState> {
         client.dispose();
         logger.d('[RA] UspClient unregistered and disposed');
       }
+
+      // Drop the cached instance in the same critical section, so watchers
+      // rebuild against an empty GetIt instead of holding a disposed client.
+      ref.invalidate(uspClientProvider);
     });
 
     state = const RemoteAssistanceState();
-
-    // Drop the cached (now disposed) client so downstream providers tear down
-    // instead of calling into freed WASM memory.
-    ref.invalidate(uspClientProvider);
   }
 }
 
--- a/lib/core/usp/providers/sse_providers.dart
+++ b/lib/core/usp/providers/sse_providers.dart
@@ -38,6 +38,8 @@ final uspBridgeClientProvider = Provider<UspBridgeClient?>((ref) {
     bridge = UspBridgeClient(
       usp,
       endpoints: BridgeEndpoints.remote(config.sessionId),
+      // Same host as the Guardian session REST API — NOT the app's own origin.
+      baseUrl: config.guardianOrigin,
       authToken: config.temporaryAccessToken,
       clientTypeId: config.clientTypeId,
       authBehavior: AuthBehavior.remote,
--- a/lib/core/usp/providers/usp_client_provider.dart
+++ b/lib/core/usp/providers/usp_client_provider.dart
@@ -10,6 +10,15 @@ import 'bridge_request_throttler_provider.dart';
 /// Returns null on non-Web platforms since USP is only available
 /// through the WASM transport layer.
 ///
+/// Also null in a Remote Assistance build until
+/// `RemoteAssistanceNotifier.activate()` has registered the session client —
+/// such builds get no boot-time client, because the app's own origin is not the
+/// USP host (see `canUseAppOriginUspClient` in `di.dart`).
+///
+/// Caches for the container's lifetime: a `ref.read` consumer keeps the instance
+/// it first saw. `activate()` invalidates this provider after the swap so
+/// `ref.watch` consumers follow.
+///
 /// Binds the [BridgeRequestThrottler] so all `usp.get()` calls are
 /// automatically throttled to prevent overwhelming the router.
 final uspClientProvider = Provider<UspClient?>((ref) {
--- a/lib/core/usp/services/usp_bridge_client_base.dart
+++ b/lib/core/usp/services/usp_bridge_client_base.dart
@@ -25,6 +25,7 @@ class UspBridgeClient {
   UspBridgeClient(
     UspClient usp, {
     BridgeEndpoints? endpoints,
+    String? baseUrl,
     String? authToken,
     String? clientTypeId,
     AuthBehavior authBehavior = AuthBehavior.local,
--- a/lib/core/usp/services/usp_bridge_client_web.dart
+++ b/lib/core/usp/services/usp_bridge_client_web.dart
@@ -38,6 +38,7 @@ external set _jsSseAbort(JSAny? value);
 class UspBridgeClient {
   final UspClient _usp;
   final BridgeEndpoints _endpoints;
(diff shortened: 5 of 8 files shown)


HankYuLinksys:
Reviewed the net diff (5 files, +45/-5) plus the surrounding pieces it depends on: `di.dart`, `uspClientProvider`, `sse_providers.dart`, `bridge_endpoints.dart`, `GlobalConfig.remote`, both SSE strategies, and `UspMutationLock`.

The diagnosis is right and the two fixes reinforce each other — the explicit bridge `baseUrl` stops the host from being inferred from `UspClient.baseUrl`, and the `uspClientProvider` invalidation is what actually fixes `get`/`set` going to the wrong host. A few things worth addressing before/after merge.

---

### 1. The invalidation is detached from the GetIt swap

The swap and `oldClient.dispose()` (which `free()`s the WASM object) happen inside `withLock`, but `ref.invalidate(uspClientProvider)` sits at the very end of `activate()` — with `setLoginType()` and `state = ...` in between.

This is correct **today**, but only because those three statements are synchronous and Riverpod flushes a dirty provider on read. Note that `setLoginType()` mutates `authProvider`, which drives `RouterNotifier` → the redirect in `router_provider.dart:141`. The moment anyone inserts an `await` between the swap and the invalidation, the exact window this PR closes reopens — and it will be a silent regression, because the symptom is a stale client, not an exception.

Suggest moving the invalidation to immediately after the swap (inside the `withLock` body, or the line right after it). "Client identity changed" and "cache invalidated" belong together. Same for `deactivate()`.

### 2. The fix only reaches `ref.watch` consumers — worth documenting the invariant

`uspClientProvider` has ~47 call sites, and a large share of them are `ref.read(uspClientProvider)!` in the body of **non-autoDispose** service providers (`pnp_service.dart:27`, `usp_firmware_update_service.dart:12`, `usp_admin_service.dart:12`, `usp_firewall_service.dart:11`, and about a dozen more). `ref.read` does not create a dependency, so `invalidate` will not rebuild any of them — they keep whatever `UspClient` they captured on first read, forever.

That is safe as things stand, because in a Remote build none of those feature services is instantiated before `activate()` runs. But nothing enforces that ordering. Please write the invariant down somewhere visible (a comment on `uspClientProvider`, or next to the invalidation in `activate()`): *the client may only be swapped before any feature service provider is first read.* Otherwise this becomes a real bug the day `deactivate()` gets wired up (see #4) or the RA entry point moves later in the boot sequence.

### 3. Root cause is in `di.dart` — worth a follow-up

`di.dart:47-51` registers `UspClient(Uri.base.origin)` unconditionally. But `GlobalConfig.remote.isActive` is `BuildConfig.isRemote()` — **compile-time**. So in a Remote build that boot-time client can never point at a valid host; its only effect is to create the stale-instance problem this PR is patching (and to make the boot-time `restoreSession()` talk to the app origin).

Skipping the registration in Remote builds — or having `uspClientProvider` build the client from the RA config instead of pulling it out of GetIt — removes the whole class of bug, and the invalidation becomes unnecessary. Fine to land this as-is, but the structural fix is the one that makes the invariant in #2 unnecessary.

### 4. `deactivate()` has no production call site

`git grep deactivate` only hits the definition and the tests, so the added invalidation there is currently unreachable in the app.

Also, the comment says downstream providers would otherwise be "calling into freed WASM memory" — I checked, and both `RemoteSseStrategy` and `LocalSseStrategy` tear down over bridge HTTP only and never touch `_usp`, so that specific failure can't happen. Worth toning the comment down so it doesn't assert something that isn't true.

### 5. Test coverage

Only the trivial `guardianOrigin` getter is covered. Neither behavioural change is. The web-only half is understandably hard to test, but the `deactivate()` invalidation is testable on the VM:

```dart
var builds = 0;
container = ProviderContainer(overrides: [
  uspClientProvider.overrideWith((ref) { builds++; return null; }),
  authProvider.overrideWith(() => mockAuthNotifier),
]);
container.read(uspClientProvider);
await container.read(remoteAssistanceProvider.notifier).deactivate();
container.read(uspClientProvider);
expect(builds, 2);
```

Separately, the existing `deactivate resets state to inactive` test (`remote_assistance_provider_test.dart:190`) doesn't `await` the future, so it asserts on the state before `deactivate()` has done anything — it passes vacuously. Adding `await` would make it meaningful.

### 6. Minor

- `guardianOrigin` does no normalization. Every current `kCloudBase` value is a bare host, so this is fine now, but the getter is public API — a `kCloudBase` that ever carries a scheme yields `https://https://…`. `assert(!guardianBaseUrl.contains('://'))` in the const constructor is free.
- The description says "one accessor for the single host all RA traffic must use (session REST, USP, subscriptions, SSE)", but session REST still builds the host by hand in `guardian_api_client.dart:31` (and `linksys_http_client.dart:94` does the same). Either route those through one source too, or narrow the claim.
- The title ("send every RA call to the Guardian host") is broader than the change: `package_widget_loader.dart`, `usp_apps_service.dart`, and `usp_apps_view.dart:99` still use `Uri.base.origin`. Those are static assets that *should* stay on the app origin — just worth stating the exclusion explicitly so a future reader doesn't treat the title as an invariant.

---

### Things I checked that look correct

- The non-web stub accepting and ignoring `baseUrl` is consistent with how it ignores every other parameter — no change needed.
- `_overrideBaseUrl ?? _usp.baseUrl` keeps local mode byte-identical, and every `BridgeEndpoints` path is root-relative, so no double slashes.
- Rebuilding `uspClientProvider` re-attaches the throttler to the new client (`usp_client_provider.dart:19`).
- `uspAuthCoordinatorProvider` uses `ref.watch`, so it rebuilds against the new client and re-wires its callbacks; the coordinator holds no timers, so the rebuild leaks nothing.
- `sseBootstrapProvider` already skips the health check in remote mode, so the absent Guardian health endpoint is a non-issue.
- Turbo channel is only reached from firmware upload (blocked in RA by `allowConfigChanges`) and the test console, so the remote turbo paths aren't on the critical path.

### What I verified locally

`dart format` is clean on all five files at the project's language version. I did not run `flutter analyze` or the test suite against this branch, so the reported 627 passing tests are not independently confirmed.

AustinChangLinksys:
Thanks — verified every point against the code. Five of the six are addressed in `64306778`; #3 is filed as a follow-up.

**1. Invalidation detached from the swap — fixed.** Both `ref.invalidate(uspClientProvider)` calls now live inside the `withLock` body, on the line after the `registerSingleton`. Agreed on the reasoning: correctness rested on the three statements in between being synchronous, and the regression an inserted `await` would cause is silent.

**2. Only `ref.watch` consumers rebuild — documented.** Confirmed: 41 `ref.read(uspClientProvider)` sites, most as constructor arguments in non-autoDispose service providers. The invariant is now written at the invalidation site: *the swap must happen before any feature service provider is first read.* (Minor: the path is `lib/page/instant_setup/services/pnp_service.dart:27`, not `lib/page/pnp/…` — the claim holds.)

**3. Root cause in `di.dart` — filed as #1309** (milestone 2.7.0). Confirmed `BuildConfig.isRemote()` resolves `const String.fromEnvironment('force')`, so it is build-time: in a Remote build the boot client can never reach a valid host. The issue lists both options you raised (skip the registration, or let `uspClientProvider` own client construction) and carries the AC that the invariant in #2 becomes unnecessary.

**4. `deactivate()` — comment toned down.** Confirmed no production call site (`git grep` hits only the definition and tests), and confirmed both strategies tear down over bridge HTTP; `SseManager.dispose()` only nulls Dart callbacks on the client. The comment now says watchers rebuild against an empty GetIt instead of claiming freed-WASM access. Left the invalidation in place so `deactivate()` is correct whenever it does get wired up.

**5. Tests — added.** Your build-count test is in as `deactivate invalidates uspClientProvider` (asserts `builds == 2`). Also confirmed the vacuous test: `deactivate()` resets state only after the `await withLock(...)`, so the un-awaited call asserted the initial state. Both un-awaited `deactivate()` calls now `await`. The remote-branch `baseUrl` assertion in `sse_providers_test.dart` is still missing — it needs `GlobalConfig.remote.isActive` to be mockable, which it isn't today (build-time flag), so I left it out rather than restructure that for a test.

**6. Minor — done.**
- Scheme guard added, but as an assert in the `guardianOrigin` getter rather than the constructor: the constructor is `const`, and const asserts only accept potentially-constant expressions, so `!guardianBaseUrl.contains('://')` is rejected there (`invalid_constant`).
- Narrowed the claim. `GuardianApiClient._buildUrl` still builds the session-REST host itself; it agrees by construction because `guardianBaseUrl` is set from the same `cloudEnvironmentConfig[kCloudBase]` (`remote_assistance_confirm_view.dart:201`). Both the getter doc and the PR description now say that instead of implying one accessor already feeds session REST.
- Static-asset exclusions (`package_widget_loader`, `package_widget_renderer`, `usp_apps_service`, `usp_apps_view.dart:99`) are now called out explicitly in the PR description as intentional `Uri.base.origin` users.

On the last note: `flutter analyze lib/core/usp test/core/usp` reports only the pre-existing `prefer_const_declarations` info in `invalidation_domain_test.dart:16`, and `flutter test test/core/usp test/core/connection test/providers/auth/auth_notifier_test.dart` is green at 490 tests on `64306778`.

AustinChangLinksys:
Follow-up #1309 is now in this PR (milestone moved to 2.6.1), so the ordering invariant this PR had to document is gone — enforced by construction instead.

`dependencySetup()` now registers `UspClient(Uri.base.origin)` only when the app origin can actually serve USP:

```dart
if (kIsWeb && !getIt.isRegistered<UspClient>()) {
  if (canUseAppOriginUspClient()) { ...register... } else { ...log skip... }
}

@visibleForTesting
bool canUseAppOriginUspClient() => !BuildConfig.isRemote();
```

Consequences:

- In a Remote build no client exists until `activate()` builds one from the session config, so there is no wrong-host instance to capture — whatever the read order.
- The worst a too-early `ref.read` can capture is `null`, which fails loudly at the `!` instead of silently POSTing to a host that answers `200 text/html`. All 16 `ref.read(uspClientProvider)!` sites are feature services under `lib/page/**/services/`, first read on the dashboard route, which RA activation gates.
- The `activate()` invalidation stays: it drops the cached `null` from the boot read so `ref.watch` consumers pick the session client up. Its comment now describes that, not an invariant callers have to uphold.
- Local / unforced builds are untouched — same registration, same origin, same lifetime.
- `BuildConfig.isRemote()` is build-time, so the branch itself is not reachable from VM tests; the decision was extracted into `canUseAppOriginUspClient()` and pinned by three cases in `test/di_test.dart` (local / none / remote).

`dart format` clean, `flutter analyze` clean on the changed files, `test/di_test.dart` + `test/core/usp` + auth notifier → 453 passed.

HankYuLinksys (approved):
Approving. Went through all six points against the code — everything is addressed, and #3 actually landed here (`6d1aef0a`) rather than being deferred, so the thread above reads as more open than it is. The description already carries `Closes #1309`, so nothing to do there.

**Verified as described:** both invalidations now sit in the `withLock` body next to the GetIt mutation; the invariant is written at the invalidation site and on `uspClientProvider`; the `deactivate()` comment no longer claims freed-WASM access; the scheme guard as a getter assert with the const-assert reasoning is right (`String.contains` isn't potentially-constant, so it can't live in the const constructor).

**Tests:** I ran them rather than taking the number on faith — `flutter test test/di_test.dart test/core/usp/providers/remote_assistance_provider_test.dart` on `6d1aef0a` is green, 23 tests, including `deactivate invalidates uspClientProvider` (`builds == 2`) and the three `canUseAppOriginUspClient` cases.

**The `di.dart` change wasn't in what I originally reviewed, so I traced it independently.** It holds up:

- `main.dart` runs `await BuildConfig.load()` before `dependencySetup()`, which is the order this needs.
- The pre-activation path in a Remote build now sees `uspClientProvider == null` throughout, and every consumer on that path tolerates it: `UspAuthCoordinator`'s constructor uses `_usp?.`, and `restoreSession()` has an explicit `if (_usp == null) return;` ahead of `_restoreSessionImpl`, which would otherwise deref `_usp!`. `authProvider.init()` reads it as `?.isAuthenticated ?? false`, `SessionService` takes `UspClient?`, `autoConfigurationLogic` returns on `BuildConfig.isRemote()` before `authCheck`/`_prepare`, the RA confirm view touches no USP provider, and every `getIt<UspClient>()` read is behind `isRegistered`.
- Local and unforced builds are byte-identical.

Nice side effect worth noting in the description: this also stops the boot-time `restoreSession()` from talking to the app origin in a Remote build, which was a silent no-op-ish failure before.

Two notes, neither blocking:

**1. `forceCommandType` is a mutable static, not a compile-time constant.** It's *initialized* from `const String.fromEnvironment('force')`, but `BuildConfig.load()` reassigns it on non-Web. That's harmless today — `load()`'s only write is inside `if (!kIsWeb)`, and the registration is inside `if (kIsWeb)` — but it means the `load()`-before-`dependencySetup()` ordering in `main.dart` is load-bearing rather than incidental. Worth a word in `canUseAppOriginUspClient`'s doc, since the surrounding comments read as if the value were const.

**2. The gate claim in the Scope section is slightly stronger than the router.** "first read after the dashboard route, which RA activation gates" — `/pnp` and `/pnpNoInternetConnection` bypass that gate unconditionally in every build (`router_provider.dart:131-134` returns the URI with no auth check), so in a Remote build a direct hit on `/pnp` reads `pnpServiceProvider` → `ref.read(uspClientProvider)!` → null-check throw. Practically zero risk: there's no link to `/pnp` in a Remote deployment, and that path was already broken before this PR, just at the network layer instead. But since the sentence is stated as an invariant someone may later rely on, the exception is worth naming.

Related nuance for whoever reads that invariant next: those service providers are non-autoDispose, and `Provider` caches a thrown error as well as a value — so a too-early read isn't just "captures null", it's "this service provider stays broken for the container's lifetime". Loud and permanent is a fine fail-fast, just not obvious from the comment.

Issue #1319, opened 2026-08-19
Test conventions: the PR gate imports a fixture from test/golden_test/


AustinChangLinksys:
## Parent

#1183 — raised by the standards review of PR #1248. Deferred out of that PR
deliberately: this is not a defect the gate introduced, and it touches files the gate
does not own.

**Re-scoped.** This issue originally carried two parts. Part 2 (constitution §3.6.2
group naming) was withdrawn and closed as a duplicate of #1313, which owns that
decision along with §3.2 and §3.3.4; its three measurements were transplanted there
([comment](https://github.com/linksys/PrivacyGUI/issues/1313#issuecomment-5350062272)).
What remains is one concrete move.

## The PR gate transitively depends on a fixture inside `test/golden_test/`

`test/golden_test/page/dashboard/cards/fixtures/cards_test_data.dart` is 29KB and
holds **55 top-level fixtures** (`testSystemInfo`, `testWanOnlineData`,
`testLanDhcpEnabled`, `testEthernetPorts`, `testRadiosTriBand`,
`testTrafficWithDiscards`, …). It has **10 importers**, and only 4 of them are
golden tests:

| importer | tree |
|---|---|
| `test/util/dashboard/kitchen_sink_overrides.dart` | shared util |
| `test/util/dashboard/card_data_profiles.dart` | shared util |
| `test/page/dashboard/views/components/usp_hero_row_density_test.dart` | functional |
| `test/page/dashboard/views/components/usp_connected_devices_density_test.dart` | functional |
| `test/page/dashboard/views/components/usp_network_health_card_legend_test.dart` | functional |
| `test/page/statistics/views/sections/stats_traffic_monitor_legend_test.dart` | functional |
| 4 × `test/golden_test/page/dashboard/cards/localizations/usp_*_cards_test.dart` | golden |

The first row is the load-bearing one: `test/util/dashboard/dashboard_card_probe.dart:23`
imports `kitchen_sink_overrides.dart`, so **every suite that calls
`probeCardOverflow` — the whole #1183 gate — depends on a file under
`test/golden_test/`**. `run_tests.sh` runs with `--exclude-tags="golden||loc||ui"`,
which excludes the golden *tests*; the fixture carries no tag and is just a library,
so nothing is broken today. What is wrong is the ownership signal: a change made for
a golden baseline lands in the PR gate's data, and the directory says the opposite.

The obvious destination is `test/mocks/test_data/` per constitution §1.6.2 /
[...]

PR #1325, opened 2026-08-20
fix(dashboard): stack the DHCP Active Leases row and declare its threshold (#1321)

### LINKED ISSUE #1321: DHCP card's Active Leases row overflows by 50px at the gate's own width — the fixture's lease expired in 2024, so #1183 reports it clean
## Parent

#1183 — the RenderFlex-overflow gate for dashboard cards. This is a **live overflow at a
width that gate sweeps on every PR**, reported clean because the fixture it pumps renders
less content than production can.

Found by hand on a real router, not by the gate. That is the point of the ticket.

## What ships broken

`dhcp_reservations`, Active Leases section, on a 4-column card (~261px) — the narrowest
realization the grid gives this card and one of the two widths the gate sweeps:

Observed in a debug build: all three lease rows carry a RenderFlex overflow stripe —
`OVERFLOWED BY 50`, `OVERFLOWED BY 44`, `OVERFLOWED BY 44` — with the lease duration
clipped at the right edge (`192.168.1.130` visible, `10h…` cut). Reservations section is
empty (`0`), Active Leases `3`.

Reproduced under the gate's own probe with an **unexpired** lease, `en`, tab 0:

| card width | source | incidents |
|---|---|---|
| **261px** | `widthCasesFor(dhcpReservations)` → `min`, span 4 @601px screen | **+50.0px right, +37.0px right** |
| **288px** | `widthCasesFor(dhcpReservations)` → `preferred`, span 6 @320px screen | +29.0px right, +16.0px right |
| 320px | probe | +5.3px right |
| 360px / 400px / 440px / 480px / 520px | probe | clean |

Both of the gate's widths are dirty. `de` at 261px measures identically (+50.0/+37.0), so
this is geometry, not translation length.

The overflowing box is the row's trailing slot:

```dart
// lib/page/local_network/cards/usp_dhcp_reservations_card.dart:121-129
trailing: Row(
  mainAxisSize: MainAxisSize.min,
  children: [
    AppText.bodySmall(client.ip, ...),
    if (lease.isNotEmpty) ...[
      AppGap.md(),
      AppText.bodySmall(lease, ...),   // <- the ~50px that does not fit
    ],
  ],
),
```

`DeviceRow` passes `trailing` straight through to `AppListTile`
(`lib/page/_shared/components/layout_blocks/row_blocks.dart:76`), so nothing constrains
it: title and subtitle ellipsize as designed, and the trailing `Row` overflows instead.

## Why the gate reports it clean

`cards_test_data.dart:356-381` pins the lease expiry to an absolute date in the past:

```dart
// test/golden_test/page/dashboard/cards/fixtures/cards_test_data.dart:363
leaseExpiry: DateTime(2024, 6, 16, 14, 30),
```

and `DhcpClientUIModel.leaseTimeFormatted` returns the **empty string** for a lease that
is both expired and active:

```dart
// lib/page/_shared/models/dhcp_client_ui_model.dart:44-46
final remaining = leaseExpiry!.difference(DateTime.now());
if (remaining.isNegative) {
  return leaseActive ? '' : 'Expired';
}
```

So `if (lease.isNotEmpty)` is false for all three fixture clients — printed from the
fixture at test time: `lease strings: "", "", ""`. The gate's trailing slot is IP-only,
~50px narrower than any live router's, and it fits.

Two consequences worth stating separately:

- **The fixture went dead on 2024-06-16, on the wall clock, with nobody touching it.**
  Before that date it rendered a lease string. #1183's gate was built after it, so the
  gate has never once rendered one.
- **The repo already knows the right idiom.** The sibling fixture for the DHCP detail page
  uses `DateTime.now().add(...)`:
  `test/golden_test/page/dhcp/fixtures/dhcp_test_data.dart:27`. The dashboard fixture is
  the outlier.

## Why no other path in the gate catches it either

- `dashboard_card_popup_overflow_test.dart`'s dialog groups render the full card form at
  `kCardPresentationWidth = 400.0`
  (`lib/page/_shared/components/card_popup_form.dart:30`) — measured clean at 360px and
  above, so the row fits there by 40px+.
- `dhcp_reservations` declares `minColumns: 4`, whose narrowest realization is 261px, well
  above `kPopupBelow = 200` — the card never selects the popup form on the grid, so the
  forced-form and popup-band sweeps do not reach this row at a dirty width.
- The data axis has exactly one non-default profile in the whole gate
  (`wifi_performance` tab 2, tri-band). `card_data_profiles.dart:39-42` already states the
  limit in its own words: *"a card is only covered on the second profile once someone adds
  it here. This list is not a claim about the other 17 cards."*

This case is one step worse than "the data axis is unswept": the **default** fixture
renders *less* than production, so "clean at the widths we sweep" was not true either.

## The class, not just the instance

`device_analytics` has the same shape — `_baseTime = DateTime(2024, 6, 15, 14, 0, 0)`
(`cards_test_data.dart:794`) feeding `testDeviceAnalyticsWithData.hourlyHistory`, while the
card builds a 24-slot axis ending at the *current* hour
(`usp_device_analytics_card.dart:350-357`, `:425-429`) and looks each slot up by exact
`DateTime` equality. None of the 12 history entries match any slot, so every bar renders
as zero.

Measured: 4 tabs × {stale fixture, now-relative fixture} at 261px — **all clean, no
overflow**. So this one is a coverage weakness, not a second bug. It is in scope only as
the second known member of the class the guard in AC4 has to cover.

## Scope

In scope: the DHCP lease row's overflow, the fixture staleness that hid it, a guard so a
fixture cannot silently stop rendering conditional content again, and the one doc
correction this finding forces (see AC7 — the correction currently lives only in a comment
on PR #1248, which is not where the next reader will look).

Not in scope: sweeping the data axis across all 18 cards (that is the growth
`card_data_profiles.dart` describes as option 2, and it belongs to a ticket that measures
each card). This ticket does not claim the other 16 cards are clean on production-shaped
data — it claims nobody has looked.

## Acceptance criteria

- [ ] The Active Leases row is clean at **both** widths the gate sweeps for
      `dhcp_reservations` (261px span 4, 288px span 6) in all 26 locales, **with a
      non-empty lease string rendered** — currently +50.0px and +29.0px
- [ ] The narrow form still identifies each lease: dropping the lease duration to make
      room is acceptable, dropping or ellipsizing the IP is not (the IP is what the row
      exists to show)
- [ ] `cards_test_data.dart`'s DHCP clients use `DateTime.now()`-relative expiries, so
      `leaseTimeFormatted` is non-empty by construction
- [ ] A marker-style assertion proves the lease string reached the rendered tree, on the
      same argument `CardDataProfile.markers` makes (`card_data_profiles.dart:60-69`): a
      fixture whose content silently disappears must fail, not pass
- [ ] `_baseTime`-derived fixtures (`testSystemMonitorWithHistory`,
      `testDeviceAnalyticsWithData`) are now-relative too, or carry a comment stating why
      an absolute date is correct for them
- [ ] No fixture in `cards_test_data.dart` pins an absolute date that a renderer compares
      against `DateTime.now()` — enumerated rather than asserted is acceptable if a lint
      is not practical
- [ ] `known_overflows.json` ends this ticket at `{}` — the ratchet's contract is that the
      allowlist only falls, so this is a fix, not an entry
- [ ] **AC7 — the doc correction.** `card_data_profiles.dart:39-42` currently bounds the
      gate's data coverage as *"a card is only covered on the second profile once someone
      adds it here"*. This ticket proves that understates it: the **default** profile can
      also render less than production, so "clean on the default profile" is not the same
      claim as "clean at the widths swept". State that where the gate's coverage is
      documented — beside the code, not only in a PR comment — and say plainly that
      `known_overflows.json == {}` is a statement about what the gate measures, not about
      the product

## Repro

Pump the card through the gate's own probe with a lease that has not expired:

```dart
final liveClients = [
  DhcpClientUIModel(
    mac: '52:D7:1A:44:3C:9E', ip: '192.168.1.130',
    leaseActive: true, isOnline: true, hostName: 'Android-e8f2c1',
    leaseExpiry: DateTime.now().add(const Duration(hours: 10, minutes: 30)),
  ),
  // ...
];

await probeCardOverflow(
  tester,
  cardId: 'dhcp_reservations',
  widthCase: widthCasesFor(
    UspWidgetSpecs.all.firstWhere((s) => s.id == 'dhcp_reservations')).first,
  cardHeightRows: 3,
  tabIndex: 0,
  locale: const Locale('en'),
  extraOverrides: cardOverrides(
    dhcpData: DhcpData(reservationModels: const [], clientModels: liveClients),
  ),
);
// -> [+50.0px right, +37.0px right]
```

## Ordering

**No longer blocked.** PR #1248 merged into `dev-2.7.0` as `b0fbfc58`, so
`cards_test_data.dart`, `dashboard_card_probe.dart`, `card_data_profiles.dart` and
`known_overflows.json` all exist there now. The earlier note here said the fixture half
collided with #1248's unmerged diff and offered to split the widget fix from the fixture
fix — that choice is gone. Branch off `dev-2.7.0` and do both halves in one change.

**Do this before #1319.** #1319 moves `cards_test_data.dart` out of `test/golden_test/`.
This issue rewrites fixtures inside it. Moving first would turn this into a conflict
resolution across a renamed path for no gain.

AustinChangLinksys:
Closes #1321.

## The defect

The Active Leases row put an IP and a lease duration side by side in an unconstrained `trailing` slot, and overflowed by **51.0px at 260.5px** and **31.0px at 288.0px** — both widths #1183's gate sweeps. The gate reported clean because `testDhcpClients` pinned `leaseExpiry` to 2024-06-16: `leaseTimeFormatted` returns the empty string for a lease that is both expired and active, so every swept case measured an IP-only slot production does not have. The fixture went dead on the wall clock, with nobody touching it, before the gate was ever built.

## The fix

The compact form **stacks** the pair rather than dropping either operand (AC 2: the IP is what the row exists to show, and an ellipsized `10h…` is the reported defect, not a fix for it). Stacking is the half that matters at the narrow end — `DeviceRow.compact`'s 60px saving alone clears only 252px, not the 200px the band starts at.

`normalAbove: 369`, measured against the product's own bounds instead of the fixture — a 15-character IPv4 quad, and `364d 23h`, the widest string `leaseTimeFormatted` can render for a lease `validateLeaseTime` accepts (525600-minute cap):

| content | normal form needs |
|---|---|
| fixture IP + `23h 59m` | 329 |
| fixture IP + `364d 23h` | 330–336 |
| quad + `23h 59m` | 361–369 |
| **quad + `364d 23h`** | **369** (368 clips 0.7px) |

25 of the 40px difference is the two extra address characters, not the exotic lease — so the number is driven by the IPv4 format, the same way `connected_devices`' is. Compact is clean with worst-case rows across the entire band, including all 26 locales at both edges (367 compact, 369 normal). The band's cost is one always-green 8px dot between 329 and 368.

## Fixture and gate

- `testDhcpClients` and `_baseTime` are now-relative (AC 3, AC 5). `_baseTime` is truncated to the hour because `usp_device_analytics_card` looks its 24 slots up by exact `DateTime` equality — the stale constant was rendering an all-zero chart that the gate swept four tabs of.
- `kDefaultFixtureMarkers` asserts the lease string reached the rendered tree (AC 4).
- AC 6 is a guard rather than an enumeration: no `DateTime(<int>` literal anywhere in the shared fixture, comment lines exempt. A list of "the absolute dates as of today" is a comment that goes stale the next time someone adds a fixture, which is the failure mode this ticket is about.
- `usp_dhcp_reservations_density_test.dart` — 14 cases over the compact band, asserting no-overflow **and** painted-inside-the-card, since a `ClipRect` would silence the first while cutting the glyphs (#1240 AC 1). Mutation ledger in the file header, re-measured on this revision.

## Fallout worth reading

- The seventh threshold moves several pinned inventories from six to seven.
- It also **retires `card_form_choice_test.dart`'s one surviving mutation**. `minColumns: 4` is the first spec floor that differs from the compact floor once scaled, so reading it as raw twelfths is now observable on the 8-column grid — the note said this would become observable the first time such a card existed, and the test it was owed is in this PR.
- **Known consequence of AC 3:** the `card_dhcp_reservations` golden gains a cell that churns with the wall clock. Baselines are gitignored, so CI cannot fail on it, and the sibling fixture `test/golden_test/page/dhcp/fixtures/dhcp_test_data.dart:27` has shipped `DateTime.now()` behind an absolute `yyyy-MM-dd HH:mm` stamp that churns every minute. The systemic fix is a clock seam (`clock.now()` in the model, `withClock` in the golden harness) covering both files; it is deliberately not done here because it is a production model change outside what this ticket measured.

## Verification

| command | result |
|---|---|
| `./run_tests.sh` | 7159 passed, 0 failed |
| `flutter test --tags dashboard-card` | 3284 passed, no KNOWN OVERFLOW |
| `test/fixtures/known_overflows.json` | still `{}` |
| `dart format` / `dart analyze` | clean; no new issues in touched files |

### DIFF
--- a/lib/page/dashboard/models/usp_widget_specs.dart
+++ b/lib/page/dashboard/models/usp_widget_specs.dart
@@ -337,6 +337,79 @@ abstract class UspWidgetSpecs {
   static const dhcpReservations = WidgetSpec(
     id: 'dhcp_reservations',
     displayName: 'DHCP Reservations',
+    // Measured floor, #1321: the narrowest width at which the **normal** Active
+    // Leases row seats an IP *and* a lease duration side by side — measured
+    // against the widest row the product can produce, not against the fixture.
+    //
+    // ## What "widest" means here, and why it is not the fixture
+    //
+    // Both operands are bounded, so both are measured at their bound. This is
+    // the criterion `usp_dhcp_reservations_density_test.dart` inherits from
+    // #1289: a bounded token must never be clipped, an unbounded one may.
+    //
+    //  - the IP is a 15-character IPv4 quad. The pool is user-editable
+    //    (`AppIpv4TextField`, Local Network page), so `192.168.100.200` is an
+    //    ordinary address rather than a corner case, and the fixture's
+    //    `192.168.1.102` is two characters short of the format.
+    //  - the lease is `leaseTimeFormatted`'s widest bucket. `validateLeaseTime`
+    //    caps a pool's lease at 525600 minutes
+    //    (`usp_local_network_service.dart:197`), so `364d 23h` is the widest
+    //    string the getter can render for a lease this product accepts.
+    //
+    // 1px sweeps of the pinned normal form, `en`, three lease rows per shape:
+    //
+    // | row content                   | first width with no clipping |
+    // |-------------------------------|------------------------------|
+    // | fixture IP + `23h 59m`        | 329                          |
+    // | fixture IP + `364d 23h`       | 330–336                      |
+    // | 15-char quad + `23h 59m`      | 361–369                      |
+    // | **15-char quad + `364d 23h`** | **369** (368 clips 0.7px)    |
+    //
+    // Row 1 is what this threshold was set to first, and the table is why it
+    // moved: of the 40px between, 25 are the two extra address characters and 5
+    // are the exotic lease. The number is driven by the *address format*, which
+    // is what the row exists to show — a lease nobody configured is not what
+    // makes 369 the floor.
+    //
+    // ## Why 369 and not the 368 that "widest failing + 1" gives
+    //
+    // Every other threshold in this file is the widest *failing* width plus one,
+    // where failing means "over `kOverflowTolerancePx`". That derivation puts
+    // this one at 368 — and 368 still clips, by 0.7px, which only the 2.0px
+    // tolerance absorbs. The tolerance exists for rasterizer drift between the
+    // mac and ubuntu font stacks (`overflow_probe.dart:4-15`), so a threshold
+    // set inside it spends the very margin the tolerance is there to provide.
+    // 369 is the narrowest width where the row does not clip at all, which is
+    // the claim a threshold should be making, and it costs 1px.
+    //
+    // ## What the bands buy
+    //
+    // `widthCasesFor` realizes exactly two widths here — 260.500px (the
+    // 4-column floor, @601px screen) and 288.000px (6- and 8-column spans both
+    // clamped to the 4-column mobile grid, @320px). **Both were shipping
+    // broken**: the normal row overflowed by 51.0px and 31.0px against an
+    // unexpired lease, with the duration clipped off the right edge on a real
+    // router. This threshold puts both in compact, and it stays below
+    // `desktopCaseFor` (512.000px) so the side-by-side row is intact wherever
+    // there is room for it.
+    //
+    // The compact form holds **both** operands across the whole band, and it
+    // does so by stacking them rather than dropping either half: at the worst
+    // case above it is clean at 200, 260.5, 288, 330, 350, 367 and 368px in
+    // `en`, and at 367px — the band's top edge — in all 26 locales. The
+    // side-by-side row needs ~126px of trailing slot at the fixture's content
+    // and ~166px at the bound; stacked it needs `max(IP, lease)` ≈ 93px, which
+    // is what makes [200, 369) safe to *declare* rather than a band that happens
+    // to work at the two widths the grid realizes. `DeviceRow.compact`'s 60px is
+    // spent on the same row and is not enough on its own — it clears 252px, not
+    // 200px.
+    //
+    // The cost of the band is one 8px dot per row. `build` filters to
+    // `isOnline == true` before building any lease row, so that dot is always
+    // the success colour: it is the only thing on this row that can be spent
+    // without losing information, which is what makes 369 cheap to declare and
+    // would not be true of a card whose leading slot carried a state.
+    normalAbove: 369,
     constraints: {
       DisplayMode.normal: WidgetGridConstraints(
         minColumns: 4,
@@ -746,7 +819,7 @@ abstract class UspWidgetSpecs {
   /// and a content line — and it is deliberately not a *measured* raise: the
   /// compact form is shorter than normal, so a number above what each card
   /// already declares could only be invented, and §2.4 is explicit that an
-  /// unmeasured constant must not be frozen into the code. Every one of the six
+  /// unmeasured constant must not be frozen into the code. Every one of the seven
   /// compact consumers already declares `minHeightRows` of 2 or 3, so today this
   /// floor is the mechanism without the raise; the raise it would apply is real
   /// the moment a card declares less.
@@ -757,10 +830,10 @@ abstract class UspWidgetSpecs {
   ///
   /// [CardDensity.normal] appears only alongside something to return *from* — on
   /// its own it is not a choice, just the status quo with a control attached.
-  /// [CardDensity.compact] appears only for the six cards that read the density
+  /// [CardDensity.compact] appears only for the seven cards that read the density
   /// (`normalAbove != null` is the existing predicate); offered anywhere else it
   /// would render exactly the normal form, which is a control that visibly does
-  /// nothing. Building compact forms for the other twelve is card-own design work
+  /// nothing. Building compact forms for the other eleven is card-own design work
   /// at #1288-#1291's scale and is out of this ticket's scope.
   static List<CardDensity> selectableForms(String id) {
     final spec = getById(id);
--- a/lib/page/local_network/cards/usp_dhcp_reservations_card.dart
+++ b/lib/page/local_network/cards/usp_dhcp_reservations_card.dart
@@ -2,6 +2,8 @@ import 'package:flutter/material.dart';
 import 'package:flutter_riverpod/flutter_riverpod.dart';
 import 'package:go_router/go_router.dart';
 import 'package:privacy_gui/localization/localization_hook.dart';
+import 'package:privacy_gui/page/_shared/components/card_density_scope.dart';
+import 'package:privacy_gui/page/_shared/models/card_density.dart';
 import 'package:privacy_gui/page/_shared/models/dhcp_client_ui_model.dart';
 import 'package:privacy_gui/page/_shared/models/dhcp_reservation_ui_model.dart';
 import 'package:privacy_gui/components/shortcuts/dialogs.dart';
@@ -27,6 +29,7 @@ class UspDhcpReservationsCard extends ConsumerWidget {
     // Dashboard shows only online clients (based on Hosts.Active, not DHCP lease).
     final onlineClients = clients.where((c) => c.isOnline == true).toList();
     final isLoading = ref.watch(uspMutationLoadingProvider) == 'dhcp';
+    final compact = CardDensityScope.of(context) == CardDensity.compact;
 
     return DashboardCardTemplate.multiSection(
       title: 'DHCP',
@@ -64,7 +67,7 @@ class UspDhcpReservationsCard extends ConsumerWidget {
           content: Column(
             children: [
               for (var i = 0; i < onlineClients.length; i++) ...[
-                _buildClientRow(context, onlineClients[i]),
+                _buildClientRow(context, onlineClients[i], compact: compact),
                 if (i < onlineClients.length - 1) AppGap.sm(),
               ],
             ],
@@ -100,12 +103,44 @@ class UspDhcpReservationsCard extends ConsumerWidget {
     );
   }
 
-  Widget _buildClientRow(BuildContext context, DhcpClientUIModel client) {
+  /// The Active Leases row.
+  ///
+  /// **Why this row has a compact form (#1321).** The trailing slot carries an IP
+  /// *and* a lease duration, and `AppListTile` hands `trailing` through
+  /// unconstrained, so the row overflowed by 50.0px at 260.5px and 29.0px at
+  /// 288.0px — both widths the #1183 gate sweeps — with the lease clipped at the
+  /// right edge on a real router. Neither operand can give: the IP is what the row
+  /// exists to show, and an ellipsized `10h…` is the defect that was reported, not
+  /// a fix for it.
+  ///
+  /// So the compact form **stacks** them instead of dropping either. That is the
+  /// half of the fix that matters at the narrow end: [DeviceRow.compact] returns
+  /// 60px (the 44px icon block plus the 16px gap ui_kit adds per occupied slot),
+  /// which alone clears 252px and not the 200px the band starts at. Stacked, the
+  /// slot's demand falls from IP + gap + lease to `max(IP, lease)` — the address
+  /// at every reachable content, since the widest lease a pool can hand out
+  /// (`364d 23h`, capped by `validateLeaseTime`) is narrower than a 15-character
+  /// quad. The band is therefore bounded by the address alone, which is why
+  /// `normalAbove: 369` covers it from 200px up.
+  ///
+  /// The icon the compact form drops is the only part of this row that can go
+  /// without loss: [build] filters to `isOnline == true` before building any row,
+  /// so the dot is **always** the success colour here and the ternary below has
+  /// one reachable branch. It is kept for the normal form because the row is a
+  /// live lease and the dot says so at a glance — it carries no information the
+  /// caller cannot already infer, which is exactly what makes it the right thing
+  /// to spend.
+  Widget _buildClientRow(
+    BuildContext context,
+    DhcpClientUIModel client, {
+    required bool compact,
+  }) {
     final colorScheme = Theme.of(context).colorScheme;
     final appColors = Theme.of(context).extension<AppColorScheme>();
     final lease = client.leaseTimeFormatted;
 
     return DeviceRow(
+      compact: compact,
       icon: Container(
         width: 8,
         height: 8,
@@ -118,16 +153,28 @@ class UspDhcpReservationsCard extends ConsumerWidget {
       ),
       title: client.displayName,
       subtitle: client.hostName.isNotEmpty ? client.mac : null,
-      trailing: Row(
-        mainAxisSize: MainAxisSize.min,
-        children: [
-          AppText.bodySmall(client.ip, color: colorScheme.onSurfaceVariant),
-          if (lease.isNotEmpty) ...[
-            AppGap.md(),
-            AppText.bodySmall(lease, color: colorScheme.onSurfaceVariant),
-          ],
-        ],
-      ),
+      trailing: compact
+          ? Column(
+              mainAxisSize: MainAxisSize.min,
+              crossAxisAlignment: CrossAxisAlignment.end,
+              children: [
+                AppText.bodySmall(client.ip,
+                    color: colorScheme.onSurfaceVariant),
+                if (lease.isNotEmpty)
+                  AppText.bodySmall(lease, color: colorScheme.onSurfaceVariant),
+              ],
+            )
+          : Row(
+              mainAxisSize: MainAxisSize.min,
+              children: [
+                AppText.bodySmall(client.ip,
+                    color: colorScheme.onSurfaceVariant),
+                if (lease.isNotEmpty) ...[
+                  AppGap.md(),
+                  AppText.bodySmall(lease, color: colorScheme.onSurfaceVariant),
+                ],
+              ],
+            ),
     );
   }
 
--- a/test/golden_test/page/dashboard/cards/fixtures/cards_test_data.dart
+++ b/test/golden_test/page/dashboard/cards/fixtures/cards_test_data.dart
@@ -353,6 +353,40 @@ final testDhcpReservations = [
   ),
 ];
 
+/// Active DHCP leases, with **now-relative** expiries (#1321).
+///
+/// These were `DateTime(2024, 6, 16, ...)`, and that is how #1183's gate came to
+/// sweep this card at two widths it was overflowing at. `leaseTimeFormatted`
+/// returns the empty string for a lease that is both expired and still active
+/// (`dhcp_client_ui_model.dart:44-46`), so every one of these rendered an
+/// IP-only trailing slot — about 50px narrower than any live router's — and the
+/// gate measured a row production does not have. The fixture went dead on
+/// 2024-06-16 on the wall clock, with nobody touching it, and the gate was built
+/// after that date, so it had never once rendered a lease string.
+///
+/// The offsets are width-maximal, not arbitrary. `leaseTimeFormatted` buckets
+/// into `Nd Nh` / `Nh Nm` / `Nm`, and measured as an overflow delta in the
+/// normal form at 300px, `23h 59m` is the widest string the getter can produce —
+/// 5.0px wider than the `days` bucket's own widest (`10d 23h`) and 22.0px wider
+/// than a bare `59m`. The first client carries it so the gate sweeps the worst
+/// case; the other two are shorter on purpose, so a row that only breaks at the
+/// maximum is still distinguishable from one that breaks everywhere.
+///
+/// The trailing `seconds: 59` is load-bearing: without it the few milliseconds
+/// between construction and layout drop `23h 59m` to `23h 58m`, which is a
+/// narrower string, and the fixture would quietly stop measuring the maximum it
+/// was written to measure.
+///
+/// **Known consequence.** A now-relative expiry makes the rendered string change
+/// with the wall clock, so the `card_dhcp_reservations` golden gains a cell that
+/// churns between baseline generation and verification. This is the sibling
+/// fixture's existing behaviour rather than a new class of problem —
+/// `test/golden_test/page/dhcp/fixtures/dhcp_test_data.dart:27` has shipped
+/// `DateTime.now().add(...)` behind a golden that renders `leaseExpiryFormatted`,
+/// an absolute `yyyy-MM-dd HH:mm` stamp, which churns every minute. The systemic
+/// fix is a clock seam (`clock.now()` in the model, `withClock` in the golden
+/// harness) covering both files; it is deliberately not done here, because a
+/// production model change is outside what this ticket measured.
 final testDhcpClients = [
   DhcpClientUIModel(
     mac: 'AA:BB:CC:DD:EE:02',
@@ -360,7 +394,9 @@ final testDhcpClients = [
     leaseActive: true,
     isOnline: true,
     hostName: 'iPhone-15',
-    leaseExpiry: DateTime(2024, 6, 16, 14, 30),
+    // `23h 59m` — the widest string leaseTimeFormatted can render.
+    leaseExpiry:
+        DateTime.now().add(const Duration(hours: 23, minutes: 59, seconds: 59)),
   ),
   DhcpClientUIModel(
     mac: 'AA:BB:CC:DD:EE:03',
@@ -368,7 +404,9 @@ final testDhcpClients = [
     leaseActive: true,
     isOnline: true,
     hostName: 'MacBook-Air',
-    leaseExpiry: DateTime(2024, 6, 16, 10, 00),
+    // `10h 30m`
+    leaseExpiry:
+        DateTime.now().add(const Duration(hours: 10, minutes: 30, seconds: 59)),
   ),
   DhcpClientUIModel(
     mac: 'AA:BB:CC:DD:EE:04',
@@ -376,7 +414,9 @@ final testDhcpClients = [
     leaseActive: true,
     isOnline: true,
     hostName: 'Smart-Speaker',
-    leaseExpiry: DateTime(2024, 6, 16, 8, 00),
+    // `6h 15m`
+    leaseExpiry:
+        DateTime.now().add(const Duration(hours: 6, minutes: 15, seconds: 59)),
   ),
 ];
 
@@ -791,7 +831,38 @@ final testFirewallEmptyData = FirewallData(
 // System Monitor
 // ---------------------------------------------------------------------------
 
-final _baseTime = DateTime(2024, 6, 15, 14, 0, 0);
+/// The clock every history series here hangs off — **now-relative, truncated to
+/// the hour** (#1321).
+///
+/// This was `DateTime(2024, 6, 15, 14, 0, 0)`, the second member of the same
+/// class as [testDhcpClients]: a fixture whose renderer compares it against
+/// `DateTime.now()`. `usp_device_analytics_card.dart:350-357` and `:425-431` build
+/// a 24-slot axis ending at the *current* hour and look each slot up by exact
+/// `DateTime` equality, so none of the 12 `hourlyHistory` entries matched any
+/// slot and both the trend chart and the heatmap rendered as all-zero. The gate
+/// swept four tabs of that card against an empty chart.
+///
+/// Unlike the lease row, this one was never a *bug* — 4 tabs × {stale fixture,
+/// now-relative fixture} at 260.5px measure clean either way, because a zero bar
+/// and a full bar occupy the same box. It is a coverage weakness, and it is fixed
+/// here because a fixture that renders nothing is the thing #1321 is about.
+///
+/// **Truncating to the hour is what makes it work**, not a tidiness choice: the
+/// lookup is `==` against `DateTime(now.year, now.month, now.day, now.hour)`, so
+/// a `_baseTime` carrying minutes or seconds would still miss every slot and the
+/// charts would stay empty while looking fixed.
+///
+/// This adds no churn class to the `device_analytics` golden that the card does
+/// not already have. Its heatmap axis renders wall-clock hour labels — `'00'`,
+/// `'06'`, `'12'`, `'18'` at slot positions derived from `DateTime.now()`
+/// (`:448`) — so that image already moves once an hour on its own, regardless of
+/// what this fixture holds.
+final _baseTime = _currentHour();
+
+DateTime _currentHour() {
+  final now = DateTime.now();
+  return DateTime(now.year, now.month, now.day, now.hour);
+}
 
 final testSystemMonitorWithHistory = SystemMonitorState(
   history: List.generate(
--- a/test/page/_shared/models/card_form_choice_test.dart
+++ b/test/page/_shared/models/card_form_choice_test.dart
@@ -43,23 +43,27 @@ import 'package:privacy_gui/page/dashboard/models/usp_widget_specs.dart';
 /// | 13 | normal restores the minima but not the maxima | normal restores the handles and the spec bounds |
 /// | 14 | normal's `maxW` not scaled to the grid | a card parked outside its spec bounds is pulled inside them |
 /// | 15 | `_applyFloors` drops the `maxW < minW` repair | a raised floor never ends up above its own ceiling |
-/// | 16 | `selectableForms` drops the `spec.normalAbove != null` guard | 4, incl. offers compact only to the six cards that read it |
+/// | 16 | `selectableForms` drops the `spec.normalAbove != null` guard | 4, incl. offers compact only to the seven cards that read it |
 /// | 17 | `cardsWithoutPopupForm` emptied | 3, incl. offers popup to every card built through the template |
 /// | 18 | `selectableForms` returns `[normal]` instead of `const []` when nothing else applies | stats_panel offers no form at all |
 /// | 19 | `CardForms.props` → `[]` (equality on type alone) | a pick that differs anywhere is a different value |
 /// | 20 | `CardForms` drops `Equatable` (back to identity) | a rebuilt CardForms equals the one it was rebuilt from |
 /// | 21 | `CardFormChoice.props` → `[density]` | the restore size is part of what makes a choice equal |
+/// | 22 | compact's `specMinW` reads the raw 12-column figure instead of `_scaleFromTwelfths(constraints.minColumns, …)` | so is the spec's own floor, which is also written in twelfths |
 ///
-/// ### One survivor, and why it is left alive
+/// ### The survivor that stopped surviving (#1321)
 ///
-/// Replacing compact's `specMinW = _scaleFromTwelfths(constraints.minColumns, …)`
-/// with the raw 12-column figure survives. It is an *equivalent* mutation on today's
-/// data, not a gap: every one of the six compact consumers declares `minColumns: 3`,
-/// and `minW` is the max of the scaled spec floor and the scaled compact floor, which
-/// the compact floor wins on all three grids. The scaling there is defensive —
-/// it becomes observable the first time a card declares a `minColumns` above 4 —
-/// and writing a test that only passes because of a card that does not exist would
-/// be worse than recording this.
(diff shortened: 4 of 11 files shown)


AustinChangLinksys (commented):
🤖 **Automated Review — Round 1** · `71f76e6..ef7c19f` (full)

**Verdict: 💬 Self-review (comment only)** — GitHub does not permit self-approval; review posted for reference. Core fix is solid; 5 warnings noted below for your consideration.

#### Standards

|---|---|---|---|
| ⚠️ | 🟡Med | `test/golden_test/page/dashboard/cards/fixtures/cards_test_data.dart:398` | `testDhcpClients` is a top-level `final`; now-relative `DateTime.now()` evaluated once at import time gives a 24h TTL before lease strings silently go empty again |
| ⚠️ | 🟡Med | `lib/page/local_network/cards/usp_dhcp_reservations_card.dart:78` | Reservations section (`_buildReservationRow`) has no compact handling; density test does not cover `ToggleRow` at compact band (200–369px) |
| ⚠️ | 🟡Med | `lib/page/dashboard/models/usp_widget_specs.dart:365` | Comment claims `364d 23h` is the widest `leaseTimeFormatted` output but doesn't explain why `365d 0h` (maximum legal lease, 1 char shorter in hours) is not wider |
| 💡 | 🟢High | `lib/page/local_network/cards/usp_dhcp_reservations_card.dart:32,70` | [Feature Envy] `compact` computed in `build()` and passed into `_buildClientRow` — matches established codebase pattern; OK as-is |
| 💡 | 🟡Med | `test/page/dashboard/cards/dashboard_card_overflow_test.dart:770` | Test reads source file via `dart:io` at test runtime; pragmatic given no linter infrastructure, but path-coupled to repo root CWD |

#### Spec

|---|---|---|---|
| ⚠️ | 🟡Med | `test/page/dashboard/cards/dashboard_card_overflow_test.dart` (marker group) | AC1: marker assertion runs at `en` + 1440px only — does not explicitly verify non-empty lease string in the 26-locale × {261px, 288px} sweep |
| ⚠️ | 🟡Med | `lib/page/local_network/cards/usp_dhcp_reservations_card.dart:156` | AC2: `AppText.bodySmall(client.ip, ...)` has no explicit `overflow:` — non-ellipsis guarantee relies on `AppText.bodySmall` default behaviour (runtime density test covers this) |
| ✅ | 🟢High | `cards_test_data.dart` | AC3: all three DHCP clients now use `DateTime.now().add(...)` — `leaseTimeFormatted` non-empty by construction |
| ✅ | 🟢High | `card_data_profiles.dart` + `dashboard_card_overflow_test.dart` | AC4: `kDefaultFixtureMarkers` + `findsNWidgets(3)` assert lease string reached rendered tree |
| ✅ | 🟢High | `cards_test_data.dart:860` | AC5: `_baseTime` now uses `_currentHour()` → now-relative; truncation to hour matches `usp_device_analytics_card`'s `==` lookup |
| ✅ | 🟢High | `dashboard_card_overflow_test.dart:770` | AC6: static guard (`DateTime\(\s*\d` regex) enforces no absolute DateTime literal in shared fixture |
| ✅ | 🟢High | `test/util/dashboard/card_data_profiles.dart:50` | AC7: doc correction added — "clean on the default profile ≠ clean at the widths swept" now stated in-code |
| ✅ | 🟢High | `test/fixtures/known_overflows.json` | AC8: `known_overflows.json` remains `{}` |

*No linked spec items missed: AC1–AC8 all addressed above.*

<details>

**W1 — Fixture top-level `DateTime.now()` TTL race** 🟡Med
`test/golden_test/page/dashboard/cards/fixtures/cards_test_data.dart:398`
```dart
leaseExpiry:
    DateTime.now().add(const Duration(hours: 23, minutes: 59, seconds: 59)),
```
`testDhcpClients` is a top-level `final`, evaluated once when the file is first imported. On a test host where the full suite runs longer than ~24 h (extremely unlikely), `leaseTimeFormatted` would return `''` and `kDefaultFixtureMarkers` would catch it in the same run. The real concern is the design pattern: a factory function `testDhcpClients()` would guarantee a fresh timestamp on every pump. Author acknowledges this in the PR description and defers the systemic fix (clock seam) deliberately. **Non-blocking.**
Fix: convert to factory function, or introduce a `clock.now()` seam alongside #1319's work.

**W2 — Reservations section has no compact form** 🟡Med
`lib/page/local_network/cards/usp_dhcp_reservations_card.dart:78`
```dart
Widget _buildReservationRow(BuildContext context, WidgetRef ref,
    DhcpReservationUIModel reservation, bool isLoading) {
  return ToggleRow(
    title: reservation.mac,   // 17 chars max
    subtitle: reservation.ip, // 15 chars max
    trailing: AppIconButton(  // fixed-size icon — not unbounded
      ...
    ),
  );
}
```
The trailing is a fixed-size `AppIconButton` (not unbounded text like Active Leases was), so the overflow risk is lower. However, `usp_dhcp_reservations_density_test.dart` only asserts the `_buildClientRow` path; the Reservations section's `ToggleRow` at 200–369px is unverified by this PR's new density tests. The mutation ledger (mutations C/D) covers client rows only.
Fix: either add a no-overflow assertion for `_buildReservationRow` in `usp_dhcp_reservations_density_test.dart`, or add a doc comment on `_buildReservationRow` explaining why `ToggleRow` with fixed-icon trailing is safe at the compact band floor (200px).

**W3 — `364d 23h` widest-string justification incomplete** 🟡Med
`lib/page/dashboard/models/usp_widget_specs.dart:365`
The comment correctly identifies `364d 23h` as the widest `leaseTimeFormatted` output for a legal lease. However, it doesn't explain why `365d 0h` (exactly 525,600 minutes — the `validateLeaseTime` cap) is not wider. In a proportional font, `'23h'` (two digits) is pixel-wider than `'0h'` (one digit), making `364d 23h` (8 chars with two double-digit fields) correct. A one-sentence note would prevent future readers from questioning the choice.
Fix: add `// (365d 0h has one fewer digit in the hours field → narrower than 364d 23h in a proportional font)`.

**C-W1 (Spec) — AC1 marker assertion scope** 🟡Med
`test/page/dashboard/cards/dashboard_card_overflow_test.dart` — marker group
The marker assertion (`findsNWidgets(3)`) runs only at `en` + `desktopCaseFor(spec)` (1440px). The 26-locale overflow sweep does verify no-overflow at 261px/288px, but it does not simultaneously assert that lease strings are non-empty in each locale. Given AC3 fixes the fixture at the source (now-relative, not locale-sensitive), and the PR description argues that lease-string shape does not change with locale (it's a number + unit abbreviation), the practical gap is narrow. Still, the spec says "in all 26 locales, *with a non-empty lease string rendered*" — the test architecture splits these two assertions across different test configurations.
Fix: either add a note in the density test explaining why locale does not affect lease-string presence (numbers are locale-invariant), or loop the marker assertion over the same locale set the sweep uses.

**C-W2 (Spec) — AC2 IP non-ellipsis implicit dependency** 🟡Med
`lib/page/local_network/cards/usp_dhcp_reservations_card.dart:156`
```dart
AppText.bodySmall(client.ip, color: colorScheme.onSurfaceVariant),
```
No `overflow:` or `maxLines:` is passed. The guarantee that IP is never ellipsized rests on `AppText.bodySmall`'s default overflow behavior. The runtime density tests (`expectBothOperandsWhole`) confirm IP is fully visible at all tested widths with worst-case content. Acceptable as-is since the tests verify the runtime guarantee, but an explicit `overflow: TextOverflow.clip` or a doc comment would make the intent explicit.

</details>

<details>

- **Core overflow fix**: compact stacking (`Column` instead of `Row`) correctly eliminates the 50px overflow at 261px and 29px at 288px; `normalAbove: 369` is precisely derived from worst-case content (`255.255.255.255` + `364d 23h`) and confirmed by the mutation ledger.
- **Mutation ledger** in `usp_dhcp_reservations_density_test.dart`: mutations A–D give quantitative evidence for both the threshold value and the compact branch, making future threshold changes reviewable without re-measuring.
- **`_baseTime` fix**: `_currentHour()` correctly truncates to the hour to match `usp_device_analytics_card`'s `DateTime ==` slot lookup — a subtle interaction explained clearly in the PR description.
- **DateTime literal guard** (AC6): choosing a static guard over enumeration is strictly stronger; the implementation correctly exempts comment lines and `DateTime(now.year, ...)` patterns.
- **`kDefaultFixtureMarkers`** (AC4): `RegExp(r'\b\d{1,2}[dh] \d{1,2}[hm]\b')` matches format shape rather than literal value — appropriate for a now-relative fixture.
- **AC7 doc correction**: the new paragraph in `card_data_profiles.dart` directly addresses the gap #1321 found and states both consequences plainly.
- **Density pattern consistency**: `CardDensityScope.of(context)` in `build()` matches the established pattern across `usp_ethernet_ports_card`, `usp_connected_devices_card`, and `usp_network_health_card`.
- **Test suite**: PR description reports 7159 passed, 0 failed; `known_overflows.json` remains `{}`.

</details>

*Cross-reviewed by two independent agents (security+correctness / architecture+maintainability) + spec-compliance agent (Reviewer C vs. Issue #1321). Automated — please sanity-check before merge.*

PeterJhongLinksys (approved):
🤖 **Automated Review — Round 1** · `71f76e6..ef7c19f` (full)

**Verdict: ✅ APPROVE** — the root cause is correctly identified (a fixture whose renderer compares against `DateTime.now()` went stale, so the #1183 gate swept an IP-only trailing slot ~50px narrower than production), the widget fix is loss-free, and all 8 acceptance criteria of #1321 have a locatable implementation. No Critical on either axis. The residual risk is concentrated in the *new guard rails*: two of them do not enforce what their own failure messages promise.

#### Standards

|---|---|---|---|
| ⚠️ | 🟢 | `test/golden_test/page/dashboard/cards/fixtures/cards_test_data.dart:390-419` | Golden-consumed shared fixture becomes wall-clock dependent, pinned to a **59-second-wide** `23h 59m` boundary; the doc comment claims `seconds: 59` solves it, which holds for one fast pump, not a 234-case isolate |
| ⚠️ | 🟢 | `test/page/dashboard/cards/dashboard_card_overflow_test.dart:770-802` | The new date guard's failure message advertises an escape hatch (*"say why in a comment on the line above and this check will not see it"*) that the loop never implements — following it still fails, with no clue why |
| ⚠️ | 🟢 | `test/golden_test/.../cards_test_data.dart:860-865` | `_baseTime` is a lazily-initialised top-level `final`, so a suite that crosses an hour boundary puts the fixture's hour and the card's own `DateTime.now()` slot axis one apart — the all-zero chart returns, silently |
| ⚠️ | 🟢 | `test/util/dashboard/card_data_profiles.dart:213` | Marker regex `\b\d{1,2}[dh] \d{1,2}[hm]\b` cannot match the bare-`Nm` bucket its own comment names, nor the 3-digit `364d 23h` bound the rest of the PR argues is width-defining |
| ⚠️ | 🟢 | `lib/page/dashboard/models/usp_widget_specs.dart:339-412` | **[Comments]** 74 comment lines inserted *between* `displayName:` and `constraints:` split one `const` record's fields; the 369 derivation now exists in 3 prose copies + 1 executable copy |
| ⚠️ | 🟢 | `lib/page/dashboard/models/usp_widget_specs.dart:819,833,836` (+6 sites) | **[Shotgun Surgery]** Adding one compact card required editing an English cardinal ("six"→"seven") in 7 places; the count is derivable from `UspWidgetSpecs.all.where((s) => s.normalAbove != null).length` |
| ⚠️ | 🟡 | `lib/page/local_network/cards/usp_dhcp_reservations_card.dart:156-166` | Compact `Column` reaches `AppListTile.trailing` **unconstrained** with no `maxLines`/`overflow`; the 15-char bound is a property of what the router *should* send, enforced nowhere in this row |
| ⚠️ | 🟡 | `test/util/dashboard/card_data_profiles.dart:217` | `expected: 3` is a hand-copied literal for a property the comment states derivably (`testDhcpClients.where((c) => c.isOnline == true).length`); a 4th online client fails with the wrong diagnosis |
| ⚠️ | 🟡 | `lib/page/local_network/cards/usp_dhcp_reservations_card.dart:124-129` | The "dot has one reachable branch" invariant depends on the `isOnline == true` filter at `:29-30` but is asserted in a comment two functions away; the dead ternary is left in place |
| ⚠️ | 🟡 | `lib/page/dashboard/models/usp_widget_specs.dart:412` | `369` deviates from the file's own "widest failing width + 1" rule (368 clips 0.7px, inside `kOverflowTolerancePx`); the repo now has two derivation rules distinguished only by prose |
| 💡 | 🟢 | `lib/page/local_network/cards/usp_dhcp_reservations_card.dart:156-176` | Both ternary arms repeat the two `AppText.bodySmall` children and the `if (lease.isNotEmpty)` condition; only the axis, alignment and `AppGap.md()` differ |
| 💡 | 🟢 | `test/golden_test/.../cards_test_data.dart:387-389` | The deferred clock-seam analysis (the most valuable artifact in the diff) exists only as a fixture paragraph — a ticket ref would make it greppable by symptom |
| 💡 | 🟡 | `test/page/dashboard/views/components/usp_dhcp_reservations_density_test.dart:135` | `spec.normalAbove ?? 369.0` re-hardcodes the constant the file exists to test; the fallback is unreachable today (`:269` asserts not-null) — use `!` |
| 💡 | 🟡 | `test/page/dashboard/views/components/usp_dhcp_reservations_density_test.dart` | Density suite sits under `views/components/` while the card is in `lib/page/local_network/cards/`; consistent with 4 sibling suites, so flagged for awareness only, not for change here |
| 💡 | ⚪ | `test/golden_test/.../cards_test_data.dart:371-375` | The load-bearing `seconds: 59` pad deserves one mechanical assertion (`expect(testDhcpClients.first.leaseTimeFormatted, '23h 59m')`) rather than a comment |
| 💡 | ⚪ | `test/golden_test/.../cards_test_data.dart:862-865` | `_currentHour()` is private, but the file-wide no-absolute-date guard means every future time-series fixture needs the same hour truncation |

#### Spec

Spec source: #1321

|---|---|---|---|
| ⚠️ | 🟡 | "The Active Leases row is clean at **both** widths…" (AC1) | **[Wrong]** Satisfied by moving both widths into compact via `normalAbove: 369`, not by repairing the row — the PR's own comment records the normal form still overflows there by 51.0/31.0px, i.e. the coordinate is fixed and the defect is now unreachable rather than gone |
| ⚠️ | 🟢 | "A marker-style assertion proves the lease string reached the rendered tree" (AC4) | **[Omission]** The guard runs at `desktopCaseFor(spec)` in `en`, so it proves the lease renders in the **normal** form; the compact `if (lease.isNotEmpty)` branch the fix introduced has no default-fixture marker behind it |
| ⚠️ | 🟢 | "No fixture … pins an absolute date that a renderer compares against `DateTime.now()`" (AC6) | **[Wrong]** Implemented as a source-text regex over one literal path: `DateTime.parse`, `DateTime.utc`, `fromMillisecondsSinceEpoch` and hoisted constants all pass, while the failure message overstates what was checked |
| ⚠️ | 🟢 | "`_baseTime`-derived fixtures … now-relative, or carry a comment" (AC5) | **[Omission]** Met via the now-relative branch, but `testSystemMonitorWithHistory` (10s spacing, no hour-equality lookup) silently inherits an hour truncation justified entirely by `device_analytics`' `==` slot lookup |
| ⚠️ | 🟡 | "clean at both widths … in all 26 locales" (AC1) | **[Omission]** The two named widths *are* swept in 26 locales by the gate; the wider "the whole 200–368px band is clean in every locale" claim rests on a comment — the new suite pins `locale: const Locale('en')` |
| ⚠️ | 🟡 | Issue scope list (overflow / fixture / guard / doc) | **[Scope creep]** `card_form_choice_test.dart`'s retired mutation-table survivor is the one change with no AC behind it; defensible as fallout from being the first card to pair `minColumns: 4` with a compact form, but worth an explicit confirmation |
| ⚠️ | 🟡 | "State that where the gate's coverage is documented" (AC7) | **[Omission]** Delivered near-verbatim at `card_data_profiles.dart:47-80`, but a reader arriving at `known_overflows.json == {}` from the gate entry point or the JSON itself never passes through that file |
| 💡 | 🟢 | "the second known member of the class the guard in AC4 has to cover" | **[Omission]** `kDefaultFixtureMarkers` has one entry; nothing asserts the analytics chart renders non-zero bars, so loosening `_currentHour()` returns it to all-zero with no failure |
| 💡 | 🟢 | AC6 escape-hatch wording | **[Wrong]** Same defect as Standards row 2, seen from the spec side: the stated contract and the implemented skip differ |
| 💡 | 🟡 | AC1 threshold basis | **[Scope creep]** 369 is derived from hypothetical worst-case content (15-char quad + `364d 23h`); the fixture clears at 329-336px, so widths 330-368 now render compact for content that would have fit side by side — the PR's largest discretionary call |
| 💡 | ⚪ | `normalAbove` derivation convention | **[Scope creep]** If "no clipping at all" is the right rule, the six pre-existing thresholds are each potentially 1px low by the same argument — follow-up ticket, not a change here |

<sub>Confidence: 🟢 High = evidenced end-to-end · 🟡 Med = mechanism shown, one link inferred · ⚪ Low = worth a second look. `[X]` = raised by 1 agent, `[XX]` = by 2 independently — **Standards table only** (the Spec axis has a single reviewer, so agreement markers carry no meaning there). Standards rows 1-6 were each raised independently by both Standards reviewers.</sub>

<details>
<summary><b>Standards — Warning detail (evidence)</b></summary>

**W1 · Golden-consumed fixture becomes wall-clock dependent** — `cards_test_data.dart:390-400`

```dart
final testDhcpClients = [
  DhcpClientUIModel(
    // `23h 59m` — the widest string leaseTimeFormatted can render.
    leaseExpiry:
        DateTime.now().add(const Duration(hours: 23, minutes: 59, seconds: 59)),
```

and the renderer (`lib/page/_shared/models/dhcp_client_ui_model.dart:44-53`):

```dart
final remaining = leaseExpiry!.difference(DateTime.now());
if (hours > 0) return '${hours}h ${minutes}m';
```

`testDhcpClients` is a top-level `final`, lazily initialised **once per isolate**, while `leaseTimeFormatted` calls `DateTime.now()` again at every render. So the rendered bucket is `23h 59m` only while `render_time − init_time < 59s`; after that it renders `23h 58m` — a different glyph run. The fixture feeds a now-234-case sweep plus 26-locale groups in one isolate, and it is consumed by `testDhcpData` (`:423-426`) which `usp_list_cards_test.dart:37,43` pumps as `viewName: 'card_dhcp_reservations'`. The fixture's own comment names the hazard and says `seconds: 59` is load-bearing — true for one pump, not for the suite.

This was **not** promoted to Critical: golden baselines are gitignored (`.gitignore:92` `test/**/goldens/*`), golden verification is owned by a separate repo (`ci.yml:11,93` → `linksys/PrivacyGUI-golden-ci`), and `run_tests.sh:80,92,96` all pass `--exclude-tags="golden||loc||ui"`. So in-repo CI cannot go red on it. It is still a real trap for whoever regenerates baselines. Cheapest fix inside this ticket's scope: keep the density/overflow tests now-relative (they build their own clients at `usp_dhcp_reservations_density_test.dart:167,178`) and give the **golden-consumed** fixture an offset in the middle of a bucket — `Duration(hours: 23, minutes: 30)` renders `23h 30m` stably for ~30 minutes instead of 59 seconds, for a couple of px of measured width.

---

**W2 · The date guard's advertised escape hatch does not exist** — `dashboard_card_overflow_test.dart:783-802`

```dart
final trimmed = lines[i].trimLeft();
// Prose about the dates this ticket removed is not a date.
if (trimmed.startsWith('//')) continue;
if (RegExp(r'DateTime\(\s*\d').hasMatch(lines[i])) {
  offenders.add('  ${i + 1}: ${lines[i].trim()}');
}
```

The failure `reason` printed to whoever trips it ends:

> …and if a constant genuinely is correct for a fixture, say why in a comment **on the line above** and this check will not see it.

Nothing in the loop looks at `lines[i-1]`. The only thing that suppresses a hit is the offending literal *itself* being on a comment line — i.e. the documented hatch is really "put the `DateTime(2024, …)` inside a comment", which no reader would infer. A contributor who follows the instruction adds the comment, re-runs, and still fails with no explanation. This is worse than having no hatch, because a failure message is the one place a reader has no reason to distrust. Either implement the lookbehind (accept a marker like `// fixture-date-ok:` on the preceding non-blank line, or trailing on the offending line) or delete the sentence.

Secondary: the check is line-based, so `DateTime(\n  2024,\n  6,\n)` — a shape `dart format` can produce for long argument lists — is not matched, since the regex needs a digit after `DateTime(` on the same line.

---

**W3 · `_baseTime` can straddle an hour boundary** — `cards_test_data.dart:860-865`

```dart
final _baseTime = _currentHour();

DateTime _currentHour() {
  final now = DateTime.now();
  return DateTime(now.year, now.month, now.day, now.hour);
}
```

Same lazy-`final` mechanics as W1. Per the PR's own comment, `usp_device_analytics_card` builds its 24-slot axis from `DateTime.now()` at **build** time and looks slots up by exact `DateTime` equality. If the suite's first touch of `_baseTime` lands at 13:59:58 and the card is pumped at 14:00:01, the axis ends at hour 14 while the fixture sits at hour 13 — every entry is off by one slot and the newest bar drops out of the window. That is the same failure class the change was written to eliminate (all-zero chart swept as clean), with its period reduced from two years to one hour. Fix: make the series a function evaluated inside the test (`testSystemMonitorWithHistory({DateTime? now})`) rather than a lazily-initialised global, so the fixture clock and the render clock cannot disagree.

---

**W4 · Marker regex cannot match the buckets it documents** — `card_data_profiles.dart:213-217`

```dart
// `leaseTimeFormatted`'s three buckets: `Nd Nh`, `Nh Nm`, `Nm`.
pattern: RegExp(r'\b\d{1,2}[dh] \d{1,2}[hm]\b'),
expected: 3,
```

The pattern requires two space-separated tokens, so it matches `Nd Nh` and `Nh Nm` but **cannot** match the third bucket it names (bare `Nm`), and `\d{1,2}` caps the day count at two digits — so `364d 23h`, the very bound `usp_widget_specs.dart` argues the threshold was measured at, is invisible to the guard meant to prove the widest content renders. Net effect: the guard silently encodes "all three fixture clients must be in the hours bucket". Shorten one offset below an hour — an edit the fixture's own comment invites ("the other two are shorter on purpose") — and it fails with `'the fixture has gone stale again (absolute expiry dates walk into the past)'`, pointing the reader at the wrong file. Suggested: `RegExp(r'\b\d+[dhm]( \d+[hm])?\b')`, or keep the strict pattern and correct the comment and `why` string to say what is actually asserted.

---

**W7 · Compact `Column` is unconstrained** — `usp_dhcp_reservations_card.dart:156-166`

```dart
trailing: compact
    ? Column(
        mainAxisSize: MainAxisSize.min,
        crossAxisAlignment: CrossAxisAlignment.end,
        children: [
          AppText.bodySmall(client.ip, color: colorScheme.onSurfaceVariant),
          if (lease.isNotEmpty)
            AppText.bodySmall(lease, color: colorScheme.onSurfaceVariant),
        ],
      )
```

Stacking reduces the demand to `max(IP, lease)` as claimed, but there is no upper bound: `AppListTile` hands `trailing` through unconstrained (per the PR's own comment, `row_blocks.dart:76`), and `ip` is a plain `String`. An IPv6 address or a malformed router value overflows the compact form exactly as the normal one did. `maxLines: 1, overflow: TextOverflow.ellipsis` on both `AppText.bodySmall` calls costs nothing and matches what `DeviceRow` already does for `title`/`subtitle` (`row_blocks.dart:63-74`) — and by this epic's own criterion an ellipsized IP is graceful degradation whereas a `RenderFlex` overflow is not.

</details>

<details>
<summary><b>Spec — findings vs #1321, and requirements met (10/10)</b></summary>

**W1 (AC1) — the coordinate is fixed; the row is not.** Spec: *"The Active Leases row is clean at **both** widths the gate sweeps for `dhcp_reservations` (261px span 4, 288px span 6) in all 26 locales, with a non-empty lease string rendered — currently +50.0px and +29.0px."* The PR satisfies this by making those widths render a different form — `usp_widget_specs.dart:412` declares `normalAbove: 369`, and `dashboard_card_overflow_test.dart:625-631` records the consequence:

```dart
// `minColumns: 4` keeps it above `kPopupBelow`, so it has no popup form
// to reach by width and *both* its realizations go to compact (#1321).
'dhcp_reservations': ['261=compact', '288=compact'],
```

The normal side-by-side `Row` that the issue measured at +50.0/+29.0px is never rendered at those widths again, and `usp_widget_specs.dart:391-398` states plainly that it *"overflowed by 51.0px and 31.0px"* — i.e. the defect is documented as still present, only unreachable by width. Warning, not Critical: AC2 explicitly licenses degradation (*"dropping the lease duration to make room is acceptable, dropping or ellipsizing the IP is not"*) and stacking is strictly stronger than what AC2 permits. Worth an author sign-off that "fix the coordinate" was the agreed reading of AC1, because an unreachable-but-latent overflow is the kind of thing a later ticket rediscovers.

**W2 (AC4) — the marker guard cannot fail where the bug was.** Implemented faithfully (`card_data_profiles.dart:168-227` + the *"default fixture conditional content"* group), but the group pumps `widthCase: desktopCaseFor(spec)` and is explicitly *"not part of the width sweep"*. So it proves the lease string reaches the tree in the **normal** form at ~512px. If the compact `if (lease.isNotEmpty)` branch (`usp_dhcp_reservations_card.dart:162-163`) ever stops rendering the lease, the marker group still passes. `usp_dhcp_reservations_density_test.dart:298-325` does cover that path via `expectBothOperandsWhole`, but against its own `worstClients` override rather than the default fixture, so the two halves do not compose into the guarantee AC4 describes.

**W3 (AC6) — a text scan, not a property about renderers.** `DateTime.parse('2024-06-16')`, `DateTime.utc(2024, …)`, `fromMillisecondsSinceEpoch(...)` and a constant hoisted into another file all pass the regex while pinning an absolute date, and the scope is one literal path (`_sharedFixturePath`), so the class the issue describes is guarded only inside `cards_test_data.dart`. That matches AC6's wording (which offers "enumerated rather than asserted is acceptable"), so it is a fair trade — but the failure message *"These fixtures pin an absolute date"* overstates what was checked, and the sibling the issue itself cites (`test/golden_test/page/dhcp/fixtures/dhcp_test_data.dart:27`) stays unguarded.

**W4 (AC5) — one clock, two consumers, one justification.** `_baseTime = _currentHour()` is consumed by `testSystemMonitorWithHistory` (`:871`) and `testDeviceAnalyticsWithData` (`:971`). The hour truncation is documented as required for `device_analytics`' `==` slot lookup; the system-monitor series is 10 samples at 10-second spacing and does no hour-equality lookup, so it inherits a truncation for a reason that is about the other consumer. Two fixtures sharing one clock whose shape is justified by one of them is the coupling that made the original `_baseTime` a class-wide problem. AC5's own *"or carry a comment"* clause is the precedent for a one-line note closing this.

**Requirements met**

1. **AC1 — clean at 261px/288px with a non-empty lease string** ✅ `normalAbove: 369` (`usp_widget_specs.dart:412`); `dashboard_card_overflow_test.dart:625-631` pins `['261=compact','288=compact']`; the gate sweeps both in 26 locales; the compact row renders a non-empty lease fed by now-relative expiries. Caveats: W1, W5 above.
2. **AC2 — narrow form still identifies each lease; IP not dropped or ellipsized** ✅ and exceeded — the compact form *stacks* both operands rather than taking AC2's licence to drop the duration. `usp_dhcp_reservations_density_test.dart:298-325` asserts both are whole and stacked across the band via painted rects, so a `ClipRect`-style silent truncation also fails (#1240 AC1).
3. **AC3 — DHCP clients use `DateTime.now()`-relative expiries** ✅ all three clients at `cards_test_data.dart:396-421`, so `leaseTimeFormatted` is non-empty by construction.
4. **AC4 — marker assertion proves the lease string reached the tree** ✅ `DefaultFixtureMarker`/`kDefaultFixtureMarkers` with a shape-matching `RegExp`, `expected` pinned rather than `findsWidgets`, `why` used as the failure message — the same argument `CardDataProfile.markers` makes, as the AC required.
5. **AC5 — `_baseTime`-derived fixtures now-relative or commented** ✅ via the now-relative branch (`:860-865`, consumed at `:871`/`:971`).
6. **AC6 — no fixture pins an absolute date a renderer compares against `DateTime.now()`** ✅ and by the stronger of the two options offered (asserted, not enumerated); `grep` over the fixture confirms no remaining `DateTime(<int>` literal outside comments.
7. **AC7 — the doc correction beside the code** ✅ `card_data_profiles.dart:47-80` carries both required sentences near-verbatim, including that `known_overflows.json == {}` is a statement about what the gate measures, not about the product.
8. **AC8 — `known_overflows.json` ends at `{}`** ✅ reads `{"tracking": {}, "allowlist": {}}` at head and is not in the changed-file list, so no allowlist entry was added and the ratchet's monotonic contract holds.
9. **Ordering — branch off `dev-2.7.0`, both halves in one change** ✅ base is `dev-2.7.0`; widget half and fixture half ship together.
10. **Ordering — done before #1319** ✅ and anticipated: `_sharedFixturePath` is a single named constant *"so #1319 — which moves it out of `test/golden_test/` — has one line to change"*, with the existence check's failure reason pointing at it.

**Coverage: 10 / 10** (8 acceptance criteria + 2 ordering constraints), 7 with residual Warning-level caveats noted above.

</details>

<details>

- **Root-cause framing, not symptom-patching.** The diagnosis chain is exact and independently checkable: `leaseExpiry` pinned to `DateTime(2024, 6, 16)` → `leaseTimeFormatted` returns `''` for a lease both expired and active (`dhcp_client_ui_model.dart:44-46`) → the trailing slot lost an operand → the gate swept 260.5/288.0px against a row ~50px narrower than production and reported clean. Fixing the *fixture* instead of adding an allowlist entry is the right call, and the PR then adds two independent anti-recurrence mechanisms rather than one.
- **The compact form is loss-free.** Stacking beats ellipsizing either operand, and the justification for what *is* dropped (the 8px dot) holds: `build` filters `isOnline == true` at `:29-30`, so the dot carries no information the section title does not. Both reviewers verified the filter and the ternary in head code.
- **The threshold is defended executably, not only in prose.** The density suite pins that the normal form *must* overflow at the widest realization, *must* still clip 2px below the threshold (using `pixels > 0` rather than `> tolerancePx`, deliberately), and that the fixture alone cannot distinguish 369 from 329 — that last one proves the old measurement basis was wrong by measuring it.
- **`normalAbove: 369` is bounded on both sides** (`> kPopupBelow`, `> widestRealization`, `< desktopCase.cardWidth`), so it cannot silently drift into degrading a card that has room.
- **Architecturally conventional.** `CardDensityScope.of(context) == CardDensity.compact` is the same pattern the six existing compact cards use; `DeviceRow(compact:)` is reused as designed rather than hand-rolling a variant (UI-Kit-first satisfied). `DeviceRow.compact` is additive and defaults false, so no existing caller changes behaviour, and `CardDensityScope.of` returns `CardDensity.normal` when no scope is present, so the new read cannot crash the card outside a dashboard cell.
- **The gate's own blind spot is written down where it will be read.** `card_data_profiles.dart:47-79` states that `known_overflows.json == {}` is a claim about what the gate measures — the kind of caveat that normally only lives in a reviewer's head.
- **The mutation survivor was retired honestly.** `card_form_choice_test.dart` had recorded row 22 as an *equivalent* mutation conditional on "the first card declaring `minColumns` above 3"; this PR is that card and writes the test that was owed instead of leaving the ledger stale.
- **Clean on the axes that most often break here**: no generated-model import outside `services/`, no provider/notifier/mutation/`PreservableContract` surface touched, no `catch` added, no USP/JNAP error path in the diff, and no credential/storage/input-sanitisation surface anywhere.

</details>

Issue #1008, opened 2026-06-30
Clicking the "Logout" button should return to the login page


KarenChen-Linksys:
**Preconditions:**
1. Dashboard style is Professional.
2. Screen size: 936.2 x 752.6.

**Steps:**
1. Login.
2. Click the settings icon then click "Logout" button.

**Actual result:**
It goes to the "Internet Settings" page.

**Expected result:**
Logout successfully and return to the login page.

**Device:**
Pinnacle 2.0 FW: 1.2.2.26062820
Linksys Now Version: 2.5.0.700525

**Additional context**
[log-logout.txt](https://github.com/user-attachments/files/29486773/log-logout.txt)
https://github.com/user-attachments/assets/704020d0-dfc5-448c-a2d5-df1ee216149b

KarenChen-Linksys:
Verified in the env below, the issue can still be reproduced.
Linksys Now Version: 2.6.0.700535
Pinnacle 2.0 FW: 1.2.3.26072618
[log-2026-08-17_14_24_34.txt](https://github.com/user-attachments/files/31131576/log-2026-08-17_14_24_34.txt)

AustinChangLinksys:
Root cause identified and documented in #1301. The failure is not in the logout handler (which is correct at `general_settings_widget.dart:130-139`) but in semantics absorption: in release web builds the dashboard card's semantics node absorbs the detail footer's tap action across the whole card, and `flutter_portal` does not hoist popup followers in the semantics tree — so the click lands on the card underneath instead of the Logout button. This explains why the issue reappeared after being marked fixed on 2026-07-02. Fix proposal in #1301.

AustinChangLinksys:
**Correction to my earlier comment above.** I stated the cause was the card absorbing the footer's tap action, implying #1301's fix would close this. That was wrong on the second half.

There are two independent defects. #1301 covers the card absorption (cards tappable across their whole surface). **This issue is a separate defect** and #1301's fix does not close it.

The engine assigns `pointer-events` in three tiers (`web_ui/.../semantics/semantics.dart:2018-2033`). Today the card is tier `all` — it actively intercepts the click. After #1301's fix it drops to tier `auto`, which merely defers to browser z-index — and the measured z-index still puts the card (z=3) **above** the Logout node (z=2). So the click still lands on the card.

[...]

Issue #1035, opened 2026-07-01
Click system information page item but lower item response


JohnnyWuLinksys:
**Description**
Click item on upper level on system information page item but actually lower level item works
<img width="1595" height="442" alt="Image" src="https://github.com/user-attachments/assets/785e1ead-5022-40a4-b8a0-080f4dd0ad11" />
Click yellow area, however it responses as red area

**Steps**
1. Login
2. Go to "Menu" tab
3. Open user information on the top-right corner
4. Click "Dark Mode" which overlap "Devices" (item might different base on resolution)

**Current result**
Go to Devices page

**Expected result**
Show Mode select popup
(top level item "in visual" should response)

**Information**
Pinnacle 2.0 FW: 1.2.2.26062820
Linksys Now Version: 2.5.0.700525

**Severity**
S3

AustinChangLinksys:
Likely the same root cause as #1008 — see #1301. In release web builds, dashboard cards absorb their detail footer's tap action across the entire card surface, so a click can activate a different item than the one aimed at. Worth verifying against #1301's fix before treating this as separate.

AustinChangLinksys:
**Correction to my earlier comment above.** I said this was "likely the same root cause as #1008". That framing was based on a since-retracted conclusion — see the correction on #1301.

There are in fact **two** independent defects, and this issue has not been reproduced against either yet:

1. **#1301** — dashboard cards absorb their detail footer's tap action across the whole card surface (semantics absorption). If the misrouting here happens *within the dashboard grid*, this is the likely candidate.
2. **#1008** — `flutter_portal` does not hoist popup followers in the semantics tree, so overlay clicks land on whatever is beneath them. If the misrouting here involves a popup/dropdown/overlay, this is the likely candidate.

The title mentions the system information page rather than the dashboard, so it is worth reproducing before assigning it to either. Both are release-web-only (semantics tree is off in debug), so reproduction needs a release build.

JohnnyWuLinksys:
Failed on FW_Pinnacle2.0_v1.2.3.26072618_USP_release.img
gui_version='2.6.0.700535'
Please check the UI log as below : 
[log-2026-09-02_16_30_00_issue_1035_verified_failed.txt](https://github.com/user-attachments/files/31729048/log-2026-09-02_16_30_00_issue_1035_verified_failed.txt)

Issue #1285, opened 2026-08-14
Theme's CJK fontFamilyFallback is double-prefixed and resolves to nothing: the bundled subset never draws raw Text


AustinChangLinksys:
## Context

Found while building the font-reality guard for #1230 (Track A of #1183): the new test had to establish which typeface the test harness actually measures in, and the harness copies `lib/app.dart`'s theme setup. Both are wrong in the same way.

Not a regression from #1230 — `lib/app.dart` is untouched on that branch. The line dates from `5beff305` (2026-07-16, the offline font-subset work).

## The defect

[`lib/app.dart:183-190`](../blob/main/lib/app.dart#L183-L190):

```dart
final cjkFallback = FallbackFontResolver.prefixedFallbackFor(effectiveLocale);
if (cjkFallback != null) {
  appLightTheme = appLightTheme.copyWith(
    textTheme: appLightTheme.textTheme.apply(fontFamilyFallback: cjkFallback),
  );
```

`prefixedFallbackFor` returns an **already package-prefixed** family (`fallback_font_resolver.dart:90`, `_prefix = 'packages/ui_kit_library'`). That is correct for a `TextStyle` with no `package` — and the prefix itself is deliberate, because [`pubspec.yaml:142-165`](../blob/main/pubspec.yaml#L142-L165) registers the app's own subset assets under family names *literally* called `packages/ui_kit_library/NotoSansSC`, so that ui_kit's `AppText` (which composes `packages/ui_kit_library/<bare>`) resolves to the app's file.

But ui_kit's `appTextTheme` styles are built with `package: 'ui_kit_library'` (`app_typography.dart:13-20`), and the Flutter SDK then prefixes a second time:

- `TextStyle.apply` stores the list raw and keeps the package: `fontFamilyFallback: fontFamilyFallback ?? _fontFamilyFallback` and `package: package ?? _package` (`painting/text_style.dart:1018`, `:1049`)
- the `fontFamilyFallback` **getter** composes the prefix onto every entry: `_fontFamilyFallback?.map((str) => 'packages/$_package/$str')` (`painting/text_style.dart:601-603`)

Measured against ui_kit v2.34.11's real `appTextTheme` (Flutter 3.44.0):

```
prefixedFallbackFor(zh_TW)      = [packages/ui_kit_library/NotoSansTC]
app.dart:189 result             = [packages/ui_kit_library/packages/ui_kit_library/NotoSansTC]   <- unresolvable
[...]

PR #1327, opened 2026-08-20
fix(dashboard): let each card footer keep its own tap target (#1301)

### LINKED ISSUE #1301: Dashboard cards are tappable across their whole surface in release web builds (semantics absorption)
## Summary

In **release web builds only**, every dashboard card that renders a detail footer becomes tappable across its *entire* surface, and its semantics node is exported to the DOM as `role=button`. Tapping any "dead zone" inside such a card navigates to the card's detail route. The same defect also swallows taps aimed at widgets that visually overlay a card — which is the actual cause of #1008 ("Logout" navigating to a settings page instead of logging out) and, we believe, #1035 (clicking an item activates a different item).

All three reported symptoms share **one** root cause. This issue exists to record that cause and the fix, because #1008 was previously closed as fixed (2026-07-02) yet remained reproducible in 2.6.0 on current firmware — the earlier fix addressed the logout handler, which was never the problem.

## Why release-only

`lib/main.dart:47-51` force-enables the semantics tree on web, gated on `!kDebugMode`:

```dart
// TODO Revisit again until Flutter SDK 3.27.x
if (!kDebugMode &&
    defaultTargetPlatform != TargetPlatform.iOS &&
    defaultTargetPlatform != TargetPlatform.android) {
  SemanticsBinding.instance.ensureSemantics();
}
```

This exists so E2E tests can locate elements by `flt-semantics-identifier` (constitution §16.1). With semantics on, Flutter Web renders an invisible `<flt-semantics>` DOM overlay, and clicks are routed by **DOM hit testing** against that overlay rather than by Flutter's own hit test. In debug the overlay is absent, so the bug is invisible — matching the field report exactly.

Note the TODO's horizon: the project is now on **Flutter 3.44.0**, well past 3.27.x.

## Root cause

`SemanticsConfiguration.absorbAll` (`rendering/object.dart:5843`) merges a child fragment's semantics **upward** into the nearest enclosing semantics boundary. `sliver_dashboard` 0.9.0 wraps every grid item in a boundary:

```dart
// ~/.pub-cache/hosted/pub.dev/sliver_dashboard-0.9.0/lib/src/view/dashboard_item_widget.dart:227
child: Semantics(
  container: true,
  label: semanticLabel,   // 'Item ${id}, Row ${y}, Column ${x}'
  ...
```

The detail footer in `DashboardCardTemplate._buildDetailFooter` (`lib/page/_shared/components/dashboard_card_template.dart:495`) declares `button: true` but creates **no boundary of its own**:

```dart
Flexible(
  child: Semantics(
    button: true,
    label: label,
    child: InkWell(
      onTap: () => context.pushNamed(detailRoute!),
```

So the footer's `tap` action and `button` flag are absorbed up into the grid item's boundary node — **whose rect is the whole card**. The card node ends up carrying the footer's tap handler over a 549×392 area instead of the footer's ~90×20.

This is directly observable in DevTools on a release build: the card's `flt-semantics` node reports `Role: button`, `Keyboard-focusable ✓`, and a merged accessible name of `"Item network_status, Row 1, Column … / View details"` — the `View details` suffix is the leaked footer.

### Why this breaks Logout (#1008)

`ClickDebouncer` (`web_ui/lib/src/engine/pointer_binding.dart:203`) withholds pointer events on `pointerdown` over a `flt-tappable`, then on the browser `click` dispatches `SemanticsAction.tap` and **discards the original pointer coordinates**. The action is delivered to whichever semantics node the DOM resolved — the card, not the popup button on top of it.

`flutter_portal`'s `RenderPortalTheater` (`portal_theater.dart:67-85`) overrides `paint` and `hitTestChildren` but **not** semantics, so popup followers are never hoisted in the semantics tree. A z-index dump confirms the card is ordered *above* the popup (card z=3 vs Logout z=2). This is not web-engine-specific: the framework's own `SemanticsOwner.performActionAt` reproduces the same misrouting in a widget test.

## Evidence

Probed with the real `SliverDashboard` + real `DashboardCardTemplate`, inspecting `hasAction(SemanticsAction.tap)` and the `isButton` flag on the grid item's node:

| Case | Card node rect | `hasTap` | `isButton` | Verdict |
|---|---|---|---|---|
| A `detailRoute` only | 396×194 (whole card) | **true** | **true** | BROKEN |
| B `detailRoute` + trailing button | 396×194 | **true** | **true** | BROKEN |
| C no footer, no tappable | 396×194 | false | false | OK |
| D tappable only inside content | 396×194 | false | false | OK |

Two findings worth recording, both of which corrected earlier hypotheses:

- **Tappable count is irrelevant; position is what matters.** An earlier probe suggested the bug required exactly one tappable (via the action-bit merge conflict at `semantics.dart:6697`). That was an artifact of hand-built geometry. Case B disproves it.
- **An `identifier:` alone accidentally masks the bug**, because `object.dart:4929` makes `identifier != null` create a boundary: `config.isSemanticBoundary = container || (_properties.identifier != null);`. Cards that already set an identifier on the footer will not reproduce. This is a latent trap: adding or removing an E2E identifier silently toggles the bug.

## Blast radius

Cards rendering a detail footer, i.e. `detailRoute != null` **or** a hand-written `footer:` containing the replicated `Semantics(button: true)` shape:

**Via `DashboardCardTemplate._buildDetailFooter` (fixed centrally, 11 cards):**
`usp_time_settings_card`, `usp_connected_devices_card`, `usp_firewall_overview_card`, `usp_network_status_card`, `usp_dhcp_reservations_card`, `usp_lan_info_card`, `usp_port_forwarding_card`, `usp_network_topology_card`, `usp_speed_test_card`, `usp_wifi_networks_card`, `usp_wifi_status_card`

**Hand-written footers replicating the same shape (need the same edit):**
- `lib/page/admin/cards/usp_device_info_card.dart:162`
- `lib/page/dashboard/views/components/usp_system_status_card.dart:145`
- `lib/page/dashboard/views/components/usp_traffic_analysis_card.dart:159`

Unaffected (no footer tappable): `usp_ethernet_ports_card`, `usp_device_analytics_card`, `usp_network_health_card`, `usp_wifi_performance_card`.

## Proposed fix

**Option A (recommended) — add `container: true` to the footer's `Semantics`.** This makes the footer its own boundary so its tap action stops being absorbed into the card. Four sites: `dashboard_card_template.dart:496` plus the three hand-written footers above.

```diff
 Flexible(
   child: Semantics(
+    container: true,
     button: true,
     label: label,
     child: InkWell(
```

Verified against the real widget: all four probe cases flip to `hasTap=false`, `isButton=false`, and `identifier` values are preserved (E2E-safe).

**Option B — `explicitChildNodes: true` on the card's outer `Semantics`.** Also correct, but broader: it fragments the card's accessibility reading into separate nodes, degrading the screen-reader experience for no extra benefit.

**Option C — remove the `ensureSemantics()` call in `main.dart`.** Would fix all symptoms, but breaks the `identifier`-based E2E anchoring mandated by constitution §16.1 and its `identifiers.generated.ts` selector map. Not recommended.

### Related: menu cards are *not* affected

`AppMenuCard` (`lib/components/styled/menus/widgets/app_menu_card.dart:34-35`) passes `AppCard(onTap: onTap)` and declares `Semantics(button: true)` at line 92. There the whole card is *intentionally* tappable, so `role=button` is correct and should stay. Only dashboard cards — which pass no `onTap` to `AppCard` — are leaking.

## Suggested regression test

A widget test asserting that the `sliver_dashboard` grid item node does **not** carry `SemanticsAction.tap`, for a card configured with `detailRoute`. This is cheap, runs headless, and would have caught the original #1008 regression. Note the harness needs `AppLocalizations.localizationsDelegates` (the footer calls `loc(context).viewDetails`) and a non-`const` `breakpoints` map.

## Caveats

- Verification was done via `SemanticsOwner.performActionAt` in widget tests plus DevTools inspection of a release build. A full release-build click-through at the reported 936×752 viewport has **not** yet been run, and should be part of the fix's acceptance.
- Relates to #1008, #1035.

AustinChangLinksys:
Closes #1301

## Problem

In **release web builds only**, every dashboard card that renders a detail footer is tappable
across its *entire* surface, and its semantics node is exported to the DOM as `role=button`.
Tapping any dead zone inside such a card navigates to the card's detail route.

`sliver_dashboard` wraps every grid item in a semantics boundary spanning the whole card
(`dashboard_item_widget.dart:227`). A footer button that declares no boundary of its own has its
tap action and `button` flag absorbed *up* into that node by `SemanticsConfiguration.absorbAll`
— so the card's node ends up carrying a tap handler over the whole card rect instead of the
footer's ~90x20.

It is release-only because `main.dart:47-51` keeps the semantics tree alive on web for E2E
(`!kDebugMode`). With semantics on, clicks are routed by DOM hit testing against the
`<flt-semantics>` overlay rather than by Flutter's own hit test. The absorption itself is
platform-independent and is what the new test asserts on.

## Fix

`container: true` on the footer's `Semantics`, which gives it a boundary so the action stays on
the link. One edit in `DashboardCardTemplate._buildDetailFooter` covers 11 cards; the other
three cards have hand-written footers and are edited individually:

| File | Footer |
|---|---|
| `dashboard_card_template.dart` | `_buildDetailFooter` (11 cards) |
| `usp_device_info_card.dart` | `_buildNodeDetailFooter` |
| `usp_system_status_card.dart` | `_buildStatisticsFooter` |
| `usp_traffic_analysis_card.dart` | `_buildStatisticsFooter` |

Purely additive: 245 insertions, 0 deletions. No layout or paint change — `Semantics` only
annotates, so golden baselines are unaffected by construction.

## What this does NOT fix

**#1008's popup pass-through.** That is a separate defect in `ui_kit_library`'s `AppPopupButton`
(no semantics on the portal follower), tracked as `linksys/privacyGUI-UI-kit#45`. The fix here
only moves the card from the engine's `pointer-events: all` tier to `auto`; the card is still
above the popup in z-order. #1301's body carries the full correction.

## Test

`test/page/dashboard/views/components/dashboard_card_semantics_absorption_test.dart` (3 tests,
`ui` tag).

It replicates the grid item's boundary rather than pumping `SliverDashboard`, because the grid's
measurement pass lays the item builder out under unbounded height — which makes
`DashboardCardTemplate`'s `Expanded` overflow and drops the footer from the tree entirely,
leaving nothing to assert on. `expectFooterOwnsItsTapAction` guards that substitution by proving
the footer is really rendered and really owns a link-sized tappable node.

Two traps recorded in the test's own docs:

- **Do not use `SemanticsOwner.performActionAt`** as the oracle. It resolves a point by searching
  for a node that both contains it *and* carries the requested action, silently skipping others,
  whereas browsers dispatch to the topmost DOM element regardless. It reports this bug as fixed
  when it is not — that is how #1008 came to be misdiagnosed.
- The `AppTheme` in the harness is required, not cosmetic. Without it the kit's spacing resolves
  to degenerate values, the footer lays out ~100000px outside the card, and every assertion goes
  vacuous.

**Non-vacuity, re-measured after the rebase:** removing `container: true` from the template fails
2 of the 3 tests, with the tappable node measuring 260.0px of height (card-sized) instead of
link-sized. The third is a control case (no footer) that is clean either way and exists to catch
a harness reporting green for the wrong reason.

## Verification

| check | result |
|---|---|
| `dart format` | 5 files, 0 changed |
| `flutter analyze` | no issues |
| `--tags ui` | 67 passed |
| `--tags dashboard-card` (#1183 gate) | 3239 passed, `known_overflows.json` still `{}` |
| `./run_tests.sh` | 7113 passed, 0 failed |

## Note for E2E

The semantics tree shape changes: the card node no longer exports `role=button`, and the footer
becomes its own node. No `identifier` value is renamed and no hook is migrated, so the generated
selector map is unchanged — but any spec that navigated by clicking a card's dead zone was
relying on this bug and must target the footer link instead.

## Rebase note

Written on 2026-08-18 and rebased onto `dev-2.7.0` for this PR. The three hand-written footers
conflicted with #1227, which wrapped each of them in `Flexible` after the branch point; #1227's
shape is kept verbatim and `container: true` is inserted into the `Semantics` inside it.

### DIFF
--- a/lib/page/_shared/components/dashboard_card_template.dart
+++ b/lib/page/_shared/components/dashboard_card_template.dart
@@ -494,6 +494,15 @@ class DashboardCardTemplate extends StatelessWidget {
               // row still fits (#1227).
               Flexible(
                 child: Semantics(
+                  // The grid item that wraps every card is a semantics boundary
+                  // (sliver_dashboard's `Semantics(container: true)`), so
+                  // without a boundary of our own this button's tap action and
+                  // `button` flag are absorbed *up* into it — and that node's
+                  // rect is the whole card. Release web builds keep the
+                  // semantics tree alive for E2E, and clicks there route through
+                  // the DOM overlay, so the absorbed action makes the entire
+                  // card navigate (#1301).
+                  container: true,
                   button: true,
                   label: label,
                   child: InkWell(
--- a/lib/page/admin/cards/usp_device_info_card.dart
+++ b/lib/page/admin/cards/usp_device_info_card.dart
@@ -161,6 +161,10 @@ class UspDeviceInfoCard extends ConsumerWidget {
             // out exactly as before.
             Flexible(
               child: Semantics(
+                // Own boundary, or the grid item absorbs this tap action across
+                // the whole card — see DashboardCardTemplate._buildDetailFooter
+                // for the full reasoning (#1301).
+                container: true,
                 button: true,
                 label: label,
                 child: InkWell(
--- a/test/page/dashboard/views/components/dashboard_card_semantics_absorption_test.dart
+++ b/test/page/dashboard/views/components/dashboard_card_semantics_absorption_test.dart
@@ -1,0 +1,224 @@
+@Tags(['ui'])
+library;
+
+import 'package:flutter/material.dart';
+import 'package:flutter/semantics.dart';
+import 'package:flutter_test/flutter_test.dart';
+import 'package:privacy_gui/l10n/gen/app_localizations.dart';
+import 'package:privacy_gui/page/_shared/components/dashboard_card_template.dart';
+import 'package:ui_kit_library/ui_kit.dart';
+
+/// Widget-layer regression coverage for PrivacyGUI#1301.
+///
+/// `sliver_dashboard` wraps every grid item in a semantics boundary spanning the
+/// whole card (`dashboard_item_widget.dart:227`):
+///
+/// ```dart
+/// child: Semantics(
+///   container: true,
+///   label: semanticLabel,      // 'Item ${id}, Row ${y}, Column ${x}'
+///   hint: ...,
+///   selected: isSelected,
+/// ```
+///
+/// A footer button that declares no boundary of its own has its tap action and
+/// `button` flag absorbed *up* into that node by
+/// `SemanticsConfiguration.absorbAll` — so the card's semantics node ends up
+/// carrying a tap handler over the entire card rect instead of the footer's
+/// ~90x20, and the DOM exports the card as `role=button`.
+///
+/// This only *manifests* in release web builds, because `main.dart` keeps the
+/// semantics tree alive there for E2E (`!kDebugMode`) and clicks are then routed
+/// by DOM hit testing against the `<flt-semantics>` overlay. The absorption
+/// itself is platform-independent and observable here.
+///
+/// The fix is `container: true` on the footer's `Semantics`.
+///
+/// ## Why this test reproduces the boundary instead of pumping `SliverDashboard`
+///
+/// Driving the real grid needs the whole dashboard provider stack, and its
+/// drag/measurement layer lays the item builder out under unbounded height —
+/// which makes `DashboardCardTemplate`'s `Expanded` overflow and drops the
+/// footer from the tree entirely, leaving nothing to assert on. What matters
+/// for absorption is only that an ancestor boundary exists, so the wrapper
+/// above is replicated verbatim. `expectFooterOwnsItsTapAction` guards the
+/// substitution by proving the footer is really rendered and really tappable.
+///
+/// ## Do not use `performActionAt` here
+///
+/// `SemanticsOwner.performActionAt` resolves a point by searching for a node
+/// that both contains it *and* carries the requested action, silently skipping
+/// others — whereas browsers dispatch to the topmost DOM element regardless. It
+/// reports this bug as fixed when it is not, which is exactly how #1008 came to
+/// be misdiagnosed. Assert on node ownership, not on delivered actions.
+void main() {
+  const cardLabel = 'Item network_status, Row 0, Column 0';
+
+  // Required, not cosmetic: without an `AppTheme` the kit's spacing resolves to
+  // degenerate values and the footer is laid out ~100000px outside the card, so
+  // it never reaches the semantics tree and every assertion below goes vacuous.
+  final testTheme = AppTheme.create(
+    brightness: Brightness.light,
+    seedColor: Colors.blue,
+    designThemeBuilder: (c) => CustomDesignTheme.fromJson({'style': 'flat'}),
+  );
+
+  /// Mirrors `sliver_dashboard`'s per-item wrapper so the card sits beneath a
+  /// card-sized semantics boundary, exactly as it does in the real grid.
+  Future<SemanticsNode> pumpCardUnderGridBoundary(
+    WidgetTester tester, {
+    required Widget content,
+    String? detailRoute,
+    Widget? footer,
+    int? itemCount,
+  }) async {
+    await tester.pumpWidget(MaterialApp(
+      theme: testTheme,
+      localizationsDelegates: AppLocalizations.localizationsDelegates,
+      supportedLocales: AppLocalizations.supportedLocales,
+      home: Scaffold(
+        body: Center(
+          // A representative desktop card footprint (4 of 12 slots).
+          child: SizedBox(
+            width: 420,
+            height: 260,
+            child: Semantics(
+              container: true,
+              label: cardLabel,
+              selected: false,
+              child: DashboardCardTemplate(
+                title: 'Network Status',
+                detailRoute: detailRoute,
+                footer: footer,
+                itemCount: itemCount,
+                content: content,
+              ),
+            ),
+          ),
+        ),
+      ),
+    ));
+    await tester.pumpAndSettle();
+
+    SemanticsNode? cardNode;
+    void walk(SemanticsNode n) {
+      if (n.label.contains('Item network_status')) cardNode ??= n;
+      n.visitChildren((c) {
+        walk(c);
+        return true;
+      });
+    }
+
+    // `rootPipelineOwner` (the suggested replacement) resolves to a different
+    // tree than the one `pumpWidget` builds, so the node lookup finds nothing
+    // there. This is the owner that actually holds the test's tree.
+    // ignore: deprecated_member_use
+    walk(tester.binding.pipelineOwner.semanticsOwner!.rootSemanticsNode!);
+    expect(cardNode, isNotNull,
+        reason: 'The card-sized semantics boundary was not found, so this test '
+            'is not exercising the absorption path at all.');
+    return cardNode!;
+  }
+
+  /// Proves the footer link is on screen AND owns a tappable semantics node of
+  /// its own. Without this, "the card has no tap action" passes trivially
+  /// whenever the footer failed to lay out — a false green that a previous
+  /// iteration of this probe actually hit.
+  void expectFooterOwnsItsTapAction(WidgetTester tester, String label) {
+    expect(find.text(label), findsOneWidget,
+        reason: 'The footer link was not laid out, so any assertion about the '
+            'card not absorbing its tap action would be vacuous.');
+
+    // The footer node's own label plus the `AppText` beneath it both mention
+    // the link, so the node's merged label reads "View details / View details" —
+    // hence `contains` rather than equality.
+    SemanticsNode? footerNode;
+    void walk(SemanticsNode n) {
+      if (n.label.contains(label) &&
+          n.getSemanticsData().hasAction(SemanticsAction.tap)) {
+        footerNode ??= n;
+      }
+      n.visitChildren((c) {
+        walk(c);
+        return true;
+      });
+    }
+
+    // ignore: deprecated_member_use — see the note in pumpCardUnderGridBoundary.
+    walk(tester.binding.pipelineOwner.semanticsOwner!.rootSemanticsNode!);
+    expect(footerNode, isNotNull,
+        reason: 'Expected a dedicated semantics node labelled "$label" owning '
+            'the tap action. Its absence means the action was absorbed into an '
+            'ancestor (the #1301 regression).');
+
+    // Pin the scope of that node: it must cover the link, not the card. Without
+    // this the test would still pass if the action were merged into a
+    // card-sized node that happened to carry the label too.
+    expect(footerNode!.rect.height, lessThan(60),
+        reason: 'The tappable node spans ${footerNode!.rect.height}px of '
+            'height — that is card-sized, not link-sized, so the tap target is '
+            'still far larger than the footer link (#1301).');
+  }
+
+  void expectCardIsNotTappable(SemanticsNode cardNode) {
+    final data = cardNode.getSemanticsData();
+    expect(data.hasAction(SemanticsAction.tap), isFalse,
+        reason: 'The card-wide node absorbed the footer button, so tapping '
+            'anywhere on the card navigates (#1301).');
+    expect(data.flagsCollection.isButton, isFalse,
+        reason: 'The card is not a button. The flag leaked up from the footer '
+            'and is exported to the DOM as role=button (#1301).');
+  }
+
+  group('#1301 the grid item must not absorb the card footer button', () {
+    testWidgets('detailRoute footer keeps its tap action to itself',
+        (tester) async {
+      final handle = tester.ensureSemantics();
+
+      final cardNode = await pumpCardUnderGridBoundary(
+        tester,
+        content: const Center(child: Text('card body')),
+        detailRoute: 'uspInternetSettings',
+      );
+
+      expectFooterOwnsItsTapAction(tester, 'View details');
+      expectCardIsNotTappable(cardNode);
+
+      handle.dispose();
+    });
+
+    testWidgets('itemCount variant ("View all") behaves the same',
+        (tester) async {
+      final handle = tester.ensureSemantics();
+
+      final cardNode = await pumpCardUnderGridBoundary(
+        tester,
+        content: const Center(child: Text('card body')),
+        detailRoute: 'uspInternetSettings',
+        itemCount: 3,
+      );
+
+      expectFooterOwnsItsTapAction(tester, 'View all');
+      expectCardIsNotTappable(cardNode);
+
+      handle.dispose();
+    });
+
+    testWidgets('a card with no tappable footer stays non-interactive',
+        (tester) async {
+      final handle = tester.ensureSemantics();
+
+      // Control case: no detailRoute and no footer. Clean both before and after
+      // the fix — it is here to catch a harness that reports "clean" for the
+      // wrong reason.
+      final cardNode = await pumpCardUnderGridBoundary(
+        tester,
+        content: const Center(child: Text('card body')),
+      );
+
+      expectCardIsNotTappable(cardNode);
+
+      handle.dispose();
+    });
+  });
+}
--- a/test/page/dashboard/views/components/dashboard_card_semantics_absorption_test.dart
+++ b/test/page/dashboard/views/components/dashboard_card_semantics_absorption_test.dart
@@ -1,4 +1,3 @@
-@Tags(['ui'])
 library;
 
 import 'package:flutter/material.dart';
--- a/lib/page/dashboard/views/components/usp_system_status_card.dart
+++ b/lib/page/dashboard/views/components/usp_system_status_card.dart
@@ -148,6 +148,10 @@ class _UspSystemStatusCardState extends ConsumerState<UspSystemStatusCard> {
             // out exactly as before.
             Flexible(
               child: Semantics(
+                // Own boundary, or the grid item absorbs this tap action across
+                // the whole card — see DashboardCardTemplate._buildDetailFooter
+                // for the full reasoning (#1301).
+                container: true,
                 button: true,
                 label: label,
                 child: InkWell(
--- a/lib/page/dashboard/views/components/usp_traffic_analysis_card.dart
+++ b/lib/page/dashboard/views/components/usp_traffic_analysis_card.dart
@@ -167,6 +167,10 @@ class _UspTrafficAnalysisCardState
           mainAxisAlignment: MainAxisAlignment.end,
           children: [
             Semantics(
+              // Own boundary, or the grid item absorbs this tap action across
+              // the whole card — see DashboardCardTemplate._buildDetailFooter
+              // for the full reasoning (#1301).
+              container: true,
               button: true,
               label: label,
               child: InkWell(


AustinChangLinksys (commented):
🤖 **Automated Review — Round 1** · `71f76e6..ad30c2c` (full)

**Verdict: 💬 Self-review (comment only)** — Austin's own PR; automated approval withheld per self-review policy. Review provided as a reference to catch blind spots.

#### Standards

|---|---|---|---|
| ⚠️ | 🟢High | `dashboard_card_semantics_absorption_test.dart:72` | [Duplicated Code] Custom-`footer` parameter path in `pumpCardUnderGridBoundary` is never exercised by any test case |
| ⚠️ | 🟢High | `dashboard_card_semantics_absorption_test.dart:116,148` | Deprecated `pipelineOwner` API used with force-unwrap `!` — silent failure risk on Flutter SDK upgrade |
| ⚠️ | 🟢High | `usp_device_info_card.dart:156–198`, `usp_system_status_card.dart:143–189`, `dashboard_card_template.dart:488–533` | [Duplicated Code] Footer `InkWell` widget tree verbatim-copied × 3 cards; shotgun-surgery already proven (this PR's 4-site edit) |
| ⚠️ | 🟢High | `usp_traffic_analysis_card.dart:169` | Missing outer `Flexible` wrapper — diverges from other 3 copies; label may overflow on narrow viewports |
| ⚠️ | 🟢High | `usp_device_info_card.dart`, `usp_system_status_card.dart`, `usp_system_status_card.dart`, `usp_traffic_analysis_card.dart` | Three hand-written footer fixes have zero regression test coverage; removing their `container: true` would not be caught |
| ⚠️ | 🟢High | `_buildDetailFooter` / `_buildNodeDetailFooter` / `_buildStatisticsFooter` | Method names inconsistent across 4 footer sites — reduces searchability |
| 💡 | 🟡Med | `dashboard_card_semantics_absorption_test.dart:136–144` | `walk()` uses `contains` substring match + first-found — ambiguous if multiple nodes share `label` text |
| 💡 | 🟢High | `usp_device_info_card.dart:147,160` | Footer label hardcoded `viewDetails` (cannot use `detailLabel` override); pre-existing, but `container: true` edit widens copy surface |
| 💡 | 🟡Med | `dashboard_card_semantics_absorption_test.dart:66–97` | `pumpCardUnderGridBoundary` is good infrastructure; should be extracted to `test/helpers/` for reuse |
| 💡 | 🟡Med | `usp_device_info_card.dart:164–166`, `usp_system_status_card.dart:151–153`, `usp_traffic_analysis_card.dart:170–172` | Comment explaining `container: true` verbatim-duplicated × 3 — will drift if reasoning evolves |

#### Spec

|---|---|---|---|
| ⚠️ | 🟡Med | `dashboard_card_semantics_absorption_test.dart` (whole file) | spec#REQ-4 → test: `non-const breakpoints map` not found anywhere in the test harness |
| ⚠️ | 🟢High | `dashboard_card_semantics_absorption_test.dart` (whole file) | spec#REQ-6 → test: non-vacuity only guaranteed for central `DashboardCardTemplate` path; three hand-written footer fixes have no test — removing their `container: true` would not fail any test |

<details>

**W-1 — Custom `footer` parameter branch untested** 🟢High  
`dashboard_card_semantics_absorption_test.dart:72` — `pumpCardUnderGridBoundary` accepts a `Widget? footer` parameter, but none of the three `testWidgets` pass a value to it. `DashboardCardTemplate._buildFooter` has three branches: (a) `footer != null` (raw Padding wrapping — no `container:true`), (b) `detailRoute != null` (fixed by this PR), (c) no-op. Branch (a) is entirely untested; the same semantics absorption would still occur if a caller passes a custom footer widget containing a tap action.  
Fix: add a `testWidgets` case passing a `Semantics(button:true, child:InkWell(...))` via the `footer` parameter, then assert the grid item node is still non-tappable.

**W-2 — Deprecated `pipelineOwner` API + force-unwrap** 🟢High  
`dashboard_card_semantics_absorption_test.dart:116,148` — Two occurrences of:
```dart
// ignore: deprecated_member_use
walk(tester.binding.pipelineOwner.semanticsOwner!.rootSemanticsNode!);
```
The author's own inline comment notes that `rootPipelineOwner` (the suggested replacement) resolves to a different tree. The `!` operator on `semanticsOwner` will throw `Null check operator used on a null value` if a future Flutter version changes the relationship. The repo's pubspec upper bound is `<4.0.0` — broad enough for this to be a real risk.  
Fix: either pin a `// ignore: deprecated_member_use` with a tracked TODO, or file a follow-up to migrate to the stable API once the semantics tree owner model is stable.

**W-3 — [Duplicated Code] Footer `InkWell` tree verbatim-copied** 🟢High  
Author's own comment in `usp_device_info_card.dart:156-158`:
```dart
// Both `Flexible`s are #1227's detail-footer shape, replicated
// verbatim from `DashboardCardTemplate._buildDetailFooter`.
```
This PR required a 4-site edit to add `container: true`. Next time (e.g. changing the icon, the border radius, or the tap handler shape) will require the same 4-site edit — classic Shotgun Surgery. Fix: extract a `static Widget buildSemanticFooterLink({required String label, required VoidCallback onTap})` factory on `DashboardCardTemplate` and replace the 4 sites.

**W-4 — `usp_traffic_analysis_card` missing outer `Flexible`** 🟢High  
`usp_traffic_analysis_card.dart:169` — diff context shows:
```dart
Row(
  mainAxisAlignment: MainAxisAlignment.end,
  children: [
    Semantics(           // ← no Flexible wrapper
      container: true,
      button: true,
```
The other three copies (`dashboard_card_template.dart:495`, `usp_device_info_card.dart:162`, `usp_system_status_card.dart:149`) all have `Flexible(child: Semantics(...))` plus an inner `Flexible(child: AppText(..., overflow: TextOverflow.ellipsis))`. Without the outer `Flexible`, the footer label text may overflow on narrow viewports.  
Fix: wrap the `Semantics` in `Flexible(child: ...)`.

**W-5 — Three hand-written footer fixes have no test** 🟢High  
`usp_device_info_card.dart`, `usp_system_status_card.dart`, `usp_traffic_analysis_card.dart` — the new test file only pumps `DashboardCardTemplate`. Removing `container: true` from any of the three hand-written footers would not cause any test to fail (the PR description's "removing it fails 2 of 3 tests" applies only to the central path). The regression net has holes for 3 of the 4 fix sites.  
Fix: add three `testWidgets` cases (one per hand-written card) using the same `pumpCardUnderGridBoundary` / `expectFooterOwnsItsTapAction` infrastructure, adapted for each card's footer method.

**W-6 — Footer method names inconsistent** 🟢High  
Four methods doing functionally identical things: `_buildDetailFooter` (template), `_buildNodeDetailFooter` (device-info), `_buildStatisticsFooter` (system-status, traffic-analysis). Inconsistent naming makes grep-based audits harder (proven by this PR, which required manually tracking 4 sites). Fix: align on a single naming convention such as `_buildSemanticFooterLink`.

**Spec W-1 — Non-const breakpoints map not found in test harness** 🟡Med  
Issue #1301 lists "non-const breakpoints map" as a test-harness requirement. `grep breakpoints dashboard_card_semantics_absorption_test.dart` returns empty. If `DashboardCardTemplate` reads breakpoints from an `InheritedWidget` / Provider at layout time and the harness falls back silently, the test may be measuring a degenerate layout rather than the real one. The PR description does not explain the omission. Low-risk if `DashboardCardTemplate` does not depend on breakpoints at the footer layout step, but that should be confirmed.

**Spec W-2 — Test non-vacuity for hand-written footers (see W-5 above)** 🟢High  
Spec REQ-6 requires that removing `container: true` fails tests. True for the central path; not true for the three hand-written footers. Already captured in W-5 — included here for completeness in the Spec axis.

</details>

<details>

- **Core fix is correct and well-reasoned.** `container: true` is exactly the right API — it creates a semantics boundary preventing `SemanticsConfiguration.absorbAll` from hoisting the footer's tap action into the enclosing grid item node. The PR description and issue #1301 provide excellent root-cause analysis.
- **All 4 required fix sites addressed.** Central `_buildDetailFooter` + all 3 hand-written footers received the change.
- **Scope discipline.** PR correctly limits itself to semantics boundary changes; no layout/paint changes, no golden failures. The "what this does NOT fix" section for #1008 is clear.
- **Test non-vacuity for central path.** PR description confirms that removing `container: true` fails 2 of 3 tests with the wrong node size (260px vs link-sized). The `expectFooterOwnsItsTapAction` `lessThan(60)` height guard is a well-designed oracle.
- **`AppLocalizations.localizationsDelegates` included.** Spec REQ-4 explicitly required this; it's present at test line 77.
- **`AppTheme` harness documented.** The test comment explains why `AppTheme` is required (degenerate spacing without it) — good anti-regression documentation.
- **Release-only behavior documented.** The `!kDebugMode` semantics guard and the `<flt-semantics>` overlay click-routing mechanism are clearly explained.
- **`identifier` trap documented.** The note that `identifier != null` accidentally creates a boundary (making the bug latent for identified cards) is a valuable find, properly called out in both the issue and PR description.
- **Dart format / flutter analyze / full test suite passed.** CI hygiene confirmed per PR verification table.

</details>

*Cross-reviewed by two independent agents (security+correctness / architecture+maintainability) plus a spec compliance check. Automated — please sanity-check before merge.*

PeterJhongLinksys (commented):
🤖 **Automated Review — Round 1** · `71f76e6..ad30c2c` (full)

**Verdict: 💬 Comment** — the four-line product fix is correct, minimal and unusually well-explained; the blocker is that the 224-line regression guard it ships is tagged out of the CI test command, so nothing enforces the fix after merge.

#### Standards

|---|---|---|---|
| 🔴 `[XX]` | 🟢 | `test/.../dashboard_card_semantics_absorption_test.dart:1` | `@Tags(['ui'])` is excluded by all three `run_tests.sh` invocations (`--exclude-tags="golden||loc||ui"`), so the #1301 guard never runs in CI — the repo's purpose-built `dashboard-card` tag exists for exactly this |
| ⚠️ | 🟢 | `dashboard_card_template.dart:505` | `container: true` promotes the pre-existing duplicate label into a real double announcement — the node's merged name becomes "View details / View details", and the test relaxes equality to `contains` to accommodate it |
| ⚠️ | 🟢 | `test/.../dashboard_card_semantics_absorption_test.dart:8` | The guard covers 1 of the 4 patched sites — the three hand-rolled footers are never instantiated, and they are the copies most likely to drift |
| ⚠️ | 🟢 | `usp_device_info_card.dart:167` | [Shotgun Surgery] Fourth verbatim copy of the detail footer; #1227 and #1301 have each now cost an identical four-site edit with nothing enforcing sync |
| ⚠️ | 🟢 | `test/.../dashboard_card_semantics_absorption_test.dart:113` | Unguarded `!` on `semanticsOwner` / `rootSemanticsNode` in both walkers — a missing `ensureSemantics()` surfaces as a bare null-check crash instead of a diagnosable failure |
| ⚠️ | 🟡 | `test/.../dashboard_card_semantics_absorption_test.dart:157` | The absorption probe accepts any label-matching tappable node bounded only by `height < 60`; width is never checked, yet #1301 is about horizontal span |
| ⚠️ | 🟡 | `test/.../dashboard_card_semantics_absorption_test.dart:57` | Hand-rolled `AppTheme.create(seedColor: Colors.blue, style: flat)` diverges from the production theme path that `dashboard_card_probe.dart:610` uses — and documents why a local copy is a trap (#1285) |
| ⚠️ | 🟡 | `test/page/dashboard/views/components/` | Test lives under `dashboard/views/components/` but exercises a `_shared/components/` widget; sibling card gates live in `test/page/dashboard/cards/` |
| 💡 | 🟢 | `test/.../dashboard_card_semantics_absorption_test.dart:188` | The `itemCount` case re-enters the same `_buildDetailFooter` branch as test 1; that slot would cover more ground aimed at a hand-rolled card |
| 💡 | 🟢 | `test/.../dashboard_card_semantics_absorption_test.dart:101,136` | [Duplicated Code] Two near-identical `walk` closures, each with its own deprecation suppression and owner dereference |
| 💡 | 🟢 | `test/.../dashboard_card_semantics_absorption_test.dart:82` | Magic dimensions `420`/`260` and the `60` bound have no stated derivation; named consts would survive a card-geometry change |
| 💡 | 🟢 | `usp_system_status_card.dart:151` | The three back-reference comments point at `_buildDetailFooter` by private method name, which a rename silently invalidates |
| 💡 | 🟡 | `dashboard_card_template.dart:505` | Now that the footer owns a DOM node, adding a stable `identifier:` is cheap and spares the E2E suite from selecting on localised label text |
| 💡 | ⚪ | `setting_blocks.dart:175`, `flow_card.dart:28` | Sibling `Semantics(button: true)` sites without a boundary; confirmed **not** reachable from a dashboard grid item today, so latent only — worth a follow-up audit, out of scope here |

#### Spec

Spec source: #1301

|---|---|---|---|
| ⚠️ | 🟡 | "Acceptance … must be a real release build clicked through at 936×752" | [Omission] Verification is a headless debug widget test only; the issue's own analysis says the symptom is invisible in debug, so no release-build or DevTools `pointer-events`/`z-index` evidence exists yet — closable by posting the result, no code change implied |
| ⚠️ | 🟢 | "assert that the `sliver_dashboard` grid item node does not carry…" | [Omission] The boundary is hand-replicated rather than instantiated (no `sliver_dashboard` import), so the guard is blind if the dependency's own wrapper changes; the spec's minimum assertion is nonetheless satisfied |
| 💡 | 🟢 | "Case B `detailRoute` + trailing button … BROKEN" | [Omission] Case B is untested although `pumpCardUnderGridBoundary` already accepts `footer:` and no call site passes it |
| 💡 | 🟡 | "adding or removing an E2E identifier silently toggles the bug" | The fix makes the boundary unconditional (correct), but the four comments never carry the `identifier:` trap forward for the next author |
| 💡 | 🟡 | "`identifier` values are preserved (E2E-safe)" | Asserted nowhere — no test constructs a footer with an identifier to confirm it survives the new boundary |
| 💡 | ⚪ | "It does **not** close #1008" | Handled correctly in the commit body; only nit is that the source comments mention `#1301` alone |

<sub>Confidence: 🟢 High / 🟡 Medium / ⚪ Low. `[XX]` = both Standards reviewers independently raised it, `[X]` = one — this marking applies to the Standards table only (the Spec axis has a single reviewer, so there is no second opinion to corroborate against).</sub>

<details>
<summary><b>🔴 Critical — full evidence chain (1)</b></summary>

### C1. The #1301 regression guard is tagged out of the PR test command — it never runs

**`test/page/dashboard/views/components/dashboard_card_semantics_absorption_test.dart:1`** · 🟢 High · raised independently by both Standards reviewers

**(1) The new test declares only the `ui` tag:**

```dart
// dashboard_card_semantics_absorption_test.dart:1-2
@Tags(['ui'])
library;
```

**(2) Every `flutter test` invocation in `run_tests.sh` excludes that tag:**

```sh
# run_tests.sh:80
# run_tests.sh:92
$FLUTTER test --exclude-tags="golden||loc||ui"
# run_tests.sh:96
$FLUTTER test --file-reporter json:$reportPath/tests.json --exclude-tags="golden||loc||ui"
```

**(3) CI runs exactly that script** — `.github/workflows/ci.yml:84`:

```yaml
        run: sh run_tests.sh
```

**(4) The repo already created a tag for precisely this case, and says so in `dart_test.yaml`:**

```yaml
tags:
  golden:
  loc:
  ui:
  # failure here blocks the PR.
  dashboard-card:
```

Three sibling guards for this same widget already follow that convention: `test/page/dashboard/cards/dashboard_card_overflow_test.dart:1`, `dashboard_card_forced_form_overflow_test.dart:1`, `dashboard_card_popup_overflow_test.dart:1` — all `@Tags(['dashboard-card'])`.

**Why it is a bug, not a style nit.** The stated deliverable of this PR is a regression guard, and nothing else in the repo enforces the four `container: true` lines — no lint, no golden, no analyzer rule can notice their removal. So the failure mode is concrete: delete `container: true` from `dashboard_card_template.dart:505` and push, and CI goes green. #1301 returns silently in release-web E2E, and the reviewer of that PR sees a file named `dashboard_card_semantics_absorption_test.dart` in the tree and reasonably assumes the path is guarded. That is worse than having no test. The `dart_test.yaml` comment shows the project already hit this exact failure mode once (#1183) and built `dashboard-card` as the remedy.

**Trigger condition.** Any run through `run_tests.sh`, i.e. all CI runs and the documented local PR command. The test executes only if someone runs `flutter test <path>` by hand or passes `--tags=ui`.

**Suggested fix** — use the tag designed for a gate that must block the PR, matching the siblings:

```dart
@Tags(['dashboard-card'])
library;
```

Note Dart's `--exclude-tags` drops a test if *any* of its tags match, so `['ui', 'dashboard-card']` would still be excluded — the `ui` tag has to go. Please include a `./run_tests.sh` excerpt showing the three new tests executing rather than skipped; "the gate actually runs" is the one claim here that cannot be established by reading the diff.

</details>

<details>
<summary><b>⚠️ Standards Warnings — detail (7)</b></summary>

### W1. `container: true` makes the duplicate label a real double announcement — 🟢 High

`dashboard_card_template.dart:505-521` (same shape at `usp_device_info_card.dart:167`, `usp_system_status_card.dart:154`, `usp_traffic_analysis_card.dart:173`)

The label is supplied twice — once on the `Semantics` wrapper, once by the `AppText` inside the `InkWell`:

```dart
container: true,
button: true,
label: label,          // (1)
child: InkWell(
  onTap: () => context.pushNamed(detailRoute!),
  child: Row(
    children: [
      Flexible(
        child: AppText.labelMedium(
          label,         // (2) same string
```

Before this PR both copies were absorbed into the card-sized node, where they were two contributions among many. Now that the footer owns a boundary, the descendant text config merges into *this* node and its accessible name becomes the two strings concatenated. This is not inference — the PR's own test states it and codes around it:

```dart
// dashboard_card_semantics_absorption_test.dart:132-134
// The footer node's own label plus the `AppText` beneath it both mention
// the link, so the node's merged label reads "View details / View details" —
// hence `contains` rather than equality.
```

The relaxation from equality to `contains` is the part worth pushing back on: `contains` also passes for any superset, removing the one assertion that would catch a regressed accessible name. Note this codebase already argued this out on a neighbouring card and landed the other way — `usp_time_settings_card.dart:87-92` records "the name goes on the button, not on a `Semantics` around it … `semanticLabel` … puts both on one node."

**Fix:** drop `label: label` from the wrapper (the `AppText` already names the node), or keep it and add `excludeSemantics: true`; then tighten to `expect(footerNode!.label, label)`.

### W2. The guard covers 1 of the 4 patched sites — 🟢 High (both reviewers)

`dashboard_card_semantics_absorption_test.dart:8` — the file's only production import is `dashboard_card_template.dart`, and all three `testWidgets` bodies pump `DashboardCardTemplate` via `pumpCardUnderGridBoundary`. `UspDeviceInfoCard`, `UspSystemStatusCard` and `UspTrafficAnalysisCard` — 3 of the 4 changed product files — are never instantiated. Those three carry hand-rolled footer copies, which is exactly why each needed its own `container: true`, and being copies is exactly why they will drift. The one site *least* likely to lose the flag (the shared template, which carries an 8-line explanatory comment) is the only one under test. Combined with C1, effective coverage today is zero; after C1 it is 1 of 4.

**Fix:** parameterise `pumpCardUnderGridBoundary` over the widget under test and run the same two assertions against each concrete card. `kitchenSinkOverrides()` in `test/util/dashboard/dashboard_card_probe.dart` already stands these cards up with provider data.

### W3. Fourth verbatim copy of the detail footer, with nothing enforcing sync — 🟢 High (both reviewers)

`dashboard_card_template.dart:495-509` · `usp_device_info_card.dart:162-170` · `usp_system_status_card.dart:149-157` · `usp_traffic_analysis_card.dart:168-176`

The code documents its own duplication — `usp_device_info_card.dart:156-158` says the shape is "replicated verbatim from `DashboardCardTemplate._buildDetailFooter`", and this PR adds a *second* cross-reference to the same origin in all three copies. Each copy is `AppDivider` + `AppGap.md()` + end-aligned `Row` + `Flexible(Semantics(button, label, InkWell(...)))`. Two separate defects (#1227 layout, #1301 semantics) have now each required an identical four-site edit. There is no lint, wrapper widget or shared helper backing the flag.

**Fix:** extract `CardDetailFooterLink({required String label, required VoidCallback onTap, int? itemCount})` into `lib/page/_shared/components/`, have the template and the three cards call it, and move both explanatory comments into it. That collapses W2's coverage gap to a single target. Deferring to a follow-up is defensible if this is being treated as a hotfix — please file it and reference the ticket in the comments so the fourth copy is recorded debt rather than an accepted pattern.

### W4. Unguarded `!` on the semantics owner yields a misleading failure — 🟢 High

`dashboard_card_semantics_absorption_test.dart:113`, `:148`

```dart
// ignore: deprecated_member_use
walk(tester.binding.pipelineOwner.semanticsOwner!.rootSemanticsNode!);
```

Both helpers dereference unconditionally. If `ensureSemantics()` is ever dropped, or a future Flutter defers root-node creation past `pumpAndSettle`, this throws a bare `Null check operator used on a null value` from inside a walker — reading as "the harness is broken" rather than "semantics were never enabled". Every other failure path in this file has a considered `reason:`; this one has none, and it is the most likely to fire after a Flutter upgrade (both sites are already flagged as on a deprecation path).

### W5. The absorption probe can pass on a node that is not the footer — 🟡 Med

`dashboard_card_semantics_absorption_test.dart:136-160` — `expectFooterOwnsItsTapAction` accepts any node whose label contains the link text and which has a tap action, bounded only by `expect(footerNode!.rect.height, lessThan(60))`. Two gaps: (1) `60` is unanchored — the link is ~20px and the card 260px, so a derived bound (`cardNode.rect.height / 2`, or `find.text(label)`'s rendered size) would neither break on a density change nor pass on a 50px wrapper node; (2) **width is never checked**, yet #1301 is about the tap target spanning the card *horizontally* — a full-width, 40px-tall node is exactly the shape of a wrong answer this assertion admits. Med rather than High because no concrete wrong-but-passing node was demonstrated in the current tree; the objection is that the probe's guarantee is weaker than the property it is named after.

### W6. Hand-rolled theme diverges from the production theme path — 🟡 Med

`dashboard_card_semantics_absorption_test.dart:57-64` builds `AppTheme.create(seedColor: Colors.blue, designThemeBuilder: … {'style': 'flat'})` — values production never uses. The shared harness at `dashboard_card_probe.dart:610-613` goes through the production path *and explains why a local copy is a trap*: "the copy that used to live here reproduced #1285's double prefix, so the harness measured the defect and could not have seen it fixed." Since this test's assertions are geometric (`rect.height < 60`), a theme whose spacing differs from production can make the guard pass or fail for reasons unrelated to absorption — and the file's own comment at `:56-58` concedes the theme is load-bearing. Med because the two themes' spacing was not measured; the objection is to the divergent duplicate harness.

### W7. Test path does not mirror the widget under test — 🟡 Med

The widget lives in `lib/page/_shared/components/`, the test in `test/page/dashboard/views/components/`. The repo is already inconsistent here — three other `DashboardCardTemplate` tests sit in `test/page/dashboard/cards/` alongside the other card gates — so this is a "please pick the mirror path" request rather than a claim of no precedent. If the file grows to cover the three USP cards (W2), `test/page/dashboard/cards/` becomes the better home.

</details>

<details>
<summary><b>📋 Spec axis — detail against issue #1301</b></summary>

### SW1. Release-build acceptance evidence is absent — ⚠️ 🟡 Med

Issue #1301 states the requirement twice:

> **this fix still needs release-build verification before being accepted.**

> A full release-build click-through at the reported 936×752 viewport has **not** yet been run, and should be part of the fix's acceptance. Acceptance for both defects must be a real release build clicked through at 936×752, plus DevTools inspection of the node's computed `pointer-events` and `z-index`.

The PR's entire verification surface is one headless widget test asserting on `SemanticsNode` data (`:174`, `:190`, `:207`), and the commit body's claim is likewise debug-only ("removing `container: true` fails 2 of the 3 tests"). Nothing in the diff or commit message records a release web build clicked through at 936×752, nor a DevTools reading of the card node's computed `pointer-events`/`z-index`. This matters because the issue is emphatic that the bug is invisible in debug ("In debug the overlay is absent, so the bug is invisible") — so by the issue's own analysis a passing debug test is not sufficient evidence the release symptom is gone. Warning, not Critical: the code change is the prescribed one, and this is discharged by the author posting the release-build result. No code change implied.

### SW2. The guard replicates the `sliver_dashboard` boundary rather than instantiating it — ⚠️ 🟢 High

Issue #1301, "Suggested regression test":

> A widget test asserting that **the `sliver_dashboard` grid item node** does **not** carry `SemanticsAction.tap`, for a card configured with `detailRoute`.

The test hand-builds the wrapper instead (`:85-88`, with `cardLabel = 'Item network_status, Row 0, Column 0'` at `:55` imitating `sliver_dashboard`'s `'Item ${id}, Row ${y}, Column ${x}'`); confirmed there is no `sliver_dashboard` import in the file. The substitution is documented and defended at `:37-45` and the reasoning is sound — the issue itself notes the real grid drops the footer under unbounded height, and the added `expectFooterOwnsItsTapAction` closes the vacuous-pass hole the issue warned about. So this is a defensible deviation, not a defect. What it costs: the guard is blind to changes in the dependency it guards against — if `sliver_dashboard` moves off `container: true`, the test keeps passing against a stale hand-copy while production regresses. The issue records `sliver_dashboard 0.9.0` as the specific version whose wrapper causes absorption; pinning that assumption (even a comment naming the version) would keep the test honest.

### Suggestions

- **Case B untested** (🟢 High) — the issue's Evidence table records `detailRoute` + trailing button at 396×194 as BROKEN, and draws the finding "Tappable count is irrelevant; position is what matters … Case B disproves it." The three tests cover Case A, an `itemCount` variant of A, and Case C. `pumpCardUnderGridBoundary` already accepts `footer:` (`:72`) and forwards it (`:92`) but no call site passes it — ~10 lines would lock in the finding that corrected an earlier wrong hypothesis.
- **`identifier:` trap not carried into the code** (🟡 Med) — the issue warns `identifier != null` creates a boundary as a side effect, "a latent trap: adding or removing an E2E identifier silently toggles the bug." `container: true` is the right structural answer, but the four fix-site comments never mention the interaction, so the next author copying the pattern has no warning. One sentence in the `dashboard_card_template.dart:497-504` block would carry it forward.
- **`identifier` preservation unasserted** (🟡 Med) — the issue claims "`identifier` values are preserved (E2E-safe)"; the PR verifies the card is no longer tappable but no test constructs a footer with an identifier. Since release semantics exist precisely for `flt-semantics-identifier` E2E anchoring (constitution §16.1), this is the unasserted claim that would hurt most if wrong. Practical risk is low — `container: true` widens rather than removes boundaries.
- **#1008 boundary is code-invisible** (⚪ Low) — correctly disclaimed in the commit body and redirected to `linksys/privacyGUI-UI-kit#45`; the only nit is that source comments mention `#1301` alone.

### Requirements met

Derived from issue #1301, one line each:

1. ✅ `container: true` on `DashboardCardTemplate._buildDetailFooter`'s `Semantics` (Option A, site 1) — `dashboard_card_template.dart:505`, covering the 11 template-driven cards in the blast radius.
2. ✅ Same edit at hand-written footer — `usp_device_info_card.dart:167`.
3. ✅ Same edit at hand-written footer — `usp_system_status_card.dart:154`.
4. ✅ Same edit at hand-written footer — `usp_traffic_analysis_card.dart:173`.
5. ✅ Exactly four sites, no more — diff touches only those four `lib/` files; `grep` for `button: true` confirms no other dashboard-grid footer was missed and no unrelated site changed.
6. ✅ Option A chosen over Option B — no `explicitChildNodes` in the diff, avoiding the screen-reader fragmentation the issue warns about.
7. ✅ Option C rejected — `lib/main.dart` untouched, so `identifier`-based E2E anchoring (constitution §16.1) is preserved.
8. ✅ `AppMenuCard` deliberately left alone — `app_menu_card.dart:95` untouched, per "the whole card is *intentionally* tappable".
9. ✅ Unaffected cards not touched — `usp_ethernet_ports_card`, `usp_device_analytics_card`, `usp_network_health_card`, `usp_wifi_performance_card` absent from the diff.
10. ✅ Regression test asserting the grid item node lacks `SemanticsAction.tap` for a `detailRoute` card — `:174` + `expectCardIsNotTappable:165`; also asserts `isButton` false. (See SW2 on the hand-replicated boundary.)
11. ✅ Harness supplies `AppLocalizations.localizationsDelegates` (the footer calls `loc(context).viewDetails`) — `:77-78`.
12. ✅ Non-`const` breakpoints map hazard avoided — harness uses a fixed `SizedBox(420×260)` (`:82-84`) rather than driving the grid.
13. ✅ `performActionAt` not used as the oracle — asserts node ownership instead, and documents the prohibition at `:47-53`.
14. ✅ Does not claim to close #1008 — commit body disclaims it and points to `linksys/privacyGUI-UI-kit#45`.
15. ✅ Does not claim to close #1035 — no mention anywhere in the diff or commit message.
16. ✅ Vacuous-pass hazard guarded — `expectFooterOwnsItsTapAction` (`:127`) requires `find.text(label)` to match and pins `rect.height < 60` (`:157`).
17. ⚠️ Release-build acceptance at 936×752 + DevTools inspection — not evidenced. See SW1.
18. ⚠️ Evidence-table Case B coverage — not tested. See Suggestions.

**Coverage: 16 of 18 requirements satisfied.**

**Scope creep: none found.** Every changed line is the prescribed `container: true` edit, an explanatory comment for it, or the regression test the issue itself asks for. No new dependencies, no public API change, no unrelated module touched.

</details>

<details>

- **The root cause is identified, not worked around.** `container: true` on the descendant is the right lever: `SemanticsConfiguration.absorbAll` merges a child's config into the nearest ancestor boundary, and the only way to stop that is to *be* a boundary. Both alternatives are worse — `explicitChildNodes` on the ancestor means patching third-party code, and `excludeSemantics` would delete the button rather than scope it. Four lines for a whole-card mis-click is an excellent ratio.
- **The comment at `dashboard_card_template.dart:497-504` is genuinely excellent.** It names the absorbing widget *and its source location*, explains why the defect only manifests in release web (`main.dart` keeps the tree alive for E2E; DOM overlay hit testing), and cites #1301. This is the kind of comment that stops the next person from "cleaning up" the flag.
- **Minimal, surgical diff.** 21 production lines, of which 17 are comment and 4 are the fix. No refactor smuggled in, no drive-by lint fixes, no reformatting, no visual-tree change — `git diff --stat` matches the PR description exactly.
- **The test records its own negative results.** Three hard-won facts are preserved rather than lost: why `SliverDashboard` is not pumped (unbounded height makes `Expanded` overflow and drops the footer, leaving nothing to assert on); why `performActionAt` must not be used (it searches for a node carrying the action and silently skips others, which is how #1008 was misdiagnosed); why `AppTheme` is mandatory (without it the footer lands ~100000px outside the card and every assertion goes vacuous). Each is a trap a future author would otherwise re-enter.
- **It guards against its own false green.** `expectFooterOwnsItsTapAction` first proves the footer rendered and owns a tappable node before asserting the card does not — the obvious version of this test ("card node has no tap action") passes trivially whenever the footer failed to lay out, and the file says an earlier iteration of the probe actually hit that. Rect-scoping the node (`:157`) is a further step most authors skip.
- **Test 3 is a real control**, not filler: a card with no `detailRoute` and no `footer` is clean before and after the fix, which catches a harness reporting "clean" for the wrong reason.
- **No security findings and no architecture violations.** Checked and negative: no secrets or `flutter_secure_storage` surface; no network, JNAP/USP transport or service layer touched, so `mapUspErrorToServiceError()` / `on Exception catch` hazards and the four USP provider hard rules are out of scope; no `generated/` imports introduced in views; no `catch` blocks added (ServiceError isolation unaffected); no hand-rolled UI component — `AppDivider`/`AppGap`/`AppText` all come from ui_kit; no naming violations; no new models. The only dynamic value in the vicinity, `queryParameters: {'tab': tabIndex.toString()}` at `usp_traffic_analysis_card.dart:180`, is pre-existing and unchanged.

</details>

test/page/dashboard/views/components/dashboard_card_semantics_absorption_test.dart
```diff
@@ -0,0 +1,224 @@
+@Tags(['ui'])
```
PeterJhongLinksys: 🔴 **Critical** — `@Tags(['ui'])` means this guard never runs in CI.

All three `flutter test` invocations in `run_tests.sh` (`:80`, `:92`, `:96`) pass `--exclude-tags="golden||loc||ui"`, and `.github/workflows/ci.yml:84` runs that script. So the 224 lines here are skipped on every CI run, and deleting `container: true` from `dashboard_card_template.dart:505` would go green.

`dart_test.yaml` already documents the tag built for exactly this case:

```yaml
  # failure here blocks the PR.
  dashboard-card:
```

The three sibling guards for this widget in `test/page/dashboard/cards/` are all `@Tags(['dashboard-card'])`. Suggested:

```dart
@Tags(['dashboard-card'])
library;
```

Note `--exclude-tags` drops a test if *any* tag matches, so `['ui', 'dashboard-card']` would still be excluded — `ui` has to go. Please attach a `./run_tests.sh` excerpt showing the three tests executing.

lib/page/_shared/components/dashboard_card_template.dart
```diff
@@ -494,6 +494,15 @@ class DashboardCardTemplate extends StatelessWidget {
...
                 child: Semantics(
+                  // The grid item that wraps every card is a semantics boundary
+                  // (sliver_dashboard's `Semantics(container: true)`), so
+                  // without a boundary of our own this button's tap action and
+                  // `button` flag are absorbed *up* into it — and that node's
+                  // rect is the whole card. Release web builds keep the
+                  // semantics tree alive for E2E, and clicks there route through
+                  // the DOM overlay, so the absorbed action makes the entire
+                  // card navigate (#1301).
+                  container: true,
```
PeterJhongLinksys, line 505: ⚠️ **Warning** — with the boundary now in place, the label is contributed twice: once here (`label: label`) and once by the `AppText.labelMedium(label)` inside the `InkWell`. The footer node's merged accessible name becomes "View details / View details". The PR's own test acknowledges this at `dashboard_card_semantics_absorption_test.dart:132-134` and relaxes equality to `contains` to accommodate it — which also makes the assertion pass for any superset, removing the check that would catch a regressed name.

This codebase already settled the same question the other way on a neighbouring card (`usp_time_settings_card.dart:87-92`: "the name goes on the button, not on a `Semantics` around it … it puts both on one node").

Suggested: drop `label: label` here (the `AppText` already names the node), or keep it and add `excludeSemantics: true`; then tighten the test back to `expect(footerNode!.label, label)`.

lib/page/admin/cards/usp_device_info_card.dart
```diff
@@ -161,6 +161,10 @@ class UspDeviceInfoCard extends ConsumerWidget {
             // out exactly as before.
             Flexible(
               child: Semantics(
+                // Own boundary, or the grid item absorbs this tap action across
+                // the whole card — see DashboardCardTemplate._buildDetailFooter
+                // for the full reasoning (#1301).
+                container: true,
```
PeterJhongLinksys, line 167: ⚠️ **Warning** (Shotgun Surgery) — this is the fourth verbatim copy of the detail footer, and the second defect in a row (#1227 layout, now #1301 semantics) to require an identical four-site edit. The code documents its own duplication at `:156-158` ("replicated verbatim from `DashboardCardTemplate._buildDetailFooter`"), and nothing — no lint, no shared helper, no test — enforces that the four stay in sync. The new guard covers only the template copy, i.e. the one *least* likely to drift.

Suggested: extract `CardDetailFooterLink({required String label, required VoidCallback onTap, int? itemCount})` into `lib/page/_shared/components/`, have the template and the three cards call it, and move both explanatory comments there — that collapses the coverage gap to one target. Deferring to a follow-up is defensible for a hotfix; please file it and reference the ticket here so the fourth copy is recorded debt rather than an accepted pattern.

AustinChangLinksys:
C1 fixed in `c9ee43ab`.

The finding holds up — I measured it with CI's own exclude expression rather than reading the tag, before and after:

```
before:  No tests ran.
         No tests match the requested tag selectors:
           exclude: "golden || loc || ui"

after:   00:03 +3: All tests passed!
```

So the guard really would have merged unexecuted, exactly as described.

**One deviation from the suggested fix.** You proposed `@Tags(['dashboard-card'])`, matching the three sibling gates. That would work, and the reasoning was right, but I removed the tag outright instead: the taxonomy is under separate review, and `dashboard-card` names dashboard card gates while this file exercises a `_shared/components/` widget (which is W7's point too). An untagged widget test already lands in the default `flutter test` set, so C1's property is obtained without pre-empting the taxonomy decision. Happy to switch to `dashboard-card` if you'd rather keep the tag explicit.

Also confirming your note that `--exclude-tags` drops a test if **any** tag matches — `['ui', 'dashboard-card']` would still have been skipped, so `ui` had to come off either way.

**Your finding generalises further than this PR.** I audited every `ui`-tagged file against the merged CI log for run `32441744841`: **12 files, 66 tests, zero executions between them.** The sanity check that the lookup itself works — `dashboard-card` files hit 15/39/8/5 times in the same log — so those zeros are real.

Two of them matter beyond hygiene:

- `static_route_dialog_test.dart` merged in #1347 this morning carrying the identical defect. #1353 corrects it.
- The five identifier guards cover the semantics identifiers `PrivacyGUI-USP-E2E` locates elements by, referenced in 24 places across its specs. Delete an identifier app-side, this repo's CI passes, and the breakage appears the next day in another repo's scheduled run.

All twelve pass when run directly, so the tag is the only thing holding them out. `dart_test.yaml` documents the remedy in a comment (`#1183`) but nothing enforces it, which is why two PRs landed the same mistake on the same day. A lint that fails when a new widget test tagged `ui` looks like a regression guard would stop the third.

**Not addressed, deliberately:**

- **W1** (`container: true` making the duplicate label a real double announcement, with the test relaxed to `contains` to accommodate it) — this is a test accommodating a defect rather than guarding against it, so it deserves its own decision, not a quiet fix bundled into a tag change.
- **W2** (guard covers 1 of 4 patched sites), **W3** (fourth verbatim footer copy) — both accurate. W3 is the one I'd file rather than fix here, per your own note about recording it as debt.
- **SW1** (no release-build evidence, and the issue says the symptom is invisible in debug) — still outstanding; it needs a release build clicked through at 936×752, which is not something I can produce from the diff.

I have not re-verified W1–W7 independently — this reply only covers C1, which I did verify end to end.

PeterJhongLinksys (approved):
🤖 **Automated Review — Round 2** · `ad30c2c..c9ee43a` (incremental)

**Verdict: ✅ APPROVE** — Round 1's only Critical (the #1301 guard was tagged out of every CI invocation) is verified fixed by measurement, and no Critical remains on either axis; the open Warnings are decisions and follow-ups, not blockers.

#### Standards

|---|---|---|---|
| ⚠️ | 🟡 | `test/.../dashboard_card_semantics_absorption_test.dart:156` | NEW — un-tagging makes the hard `height < 60` pixel bound CI-live; a ui_kit spacing drift now reds an unrelated PR with a misleading "#1301" message. |
| ⚠️ | 🟢 | `test/.../dashboard_card_semantics_absorption_test.dart:1` | NEW — C1's property is now obtained by an *implicit* untagged state, where `dart_test.yaml` and `.claude/skills/dashboard-overflow-gate/SKILL.md:112` both treat `dashboard-card` as the explicit "must not be excluded" marker. |
| ⚠️ | 🟢 | repo-wide (`test/page/**`) | NEW — 11 more files still carry `@Tags(['ui'])` (66 tests, 0 CI executions); nothing mechanised prevents the twelfth. Scope note, not a blocker on this PR. |
| ⚠️ | 🟢 | `dashboard_card_template.dart:505`, test `:136` | OPEN (R1 W1) — `container: true` promotes the duplicated footer label to a real double announcement ("View details View details"); the test was relaxed to `contains` to accept it. Now *enforced*, since the test runs. |
| ⚠️ | 🟢 | `test/.../dashboard_card_semantics_absorption_test.dart:7` | OPEN (R1 W2) — guard covers 1 of the 4 patched sites; the 3 uncovered ones are the hand-rolled copies most likely to drift. |
| ⚠️ | 🟢 | test `:115`, `:147` | OPEN (R1 W4) — unguarded `!` on `semanticsOwner`/`rootSemanticsNode`; a Flutter upgrade surfaces as a bare null-check throw reading as "harness broken". Two `expect(..., isNotNull, reason:)` lines fix it. |
| ⚠️ | 🟡 | test `:156` | OPEN (R1 W5) — width is never asserted, yet #1301 is about horizontal span: a full-width 40px-tall node still passes. |
| ⚠️ | 🟡 | test `:56-63` | OPEN (R1 W6) — hand-rolled `AppTheme.create` diverges from the production theme path while the file's own comment concedes the theme is load-bearing for its geometry. |
| ⚠️ | 🟡 | `test/page/dashboard/views/components/` vs `lib/page/_shared/components/` | OPEN (R1 W7) — test path does not mirror the widget under test; now entangled with the tagging choice (moving the file makes `dashboard-card` obviously correct). |
| ⚠️ | ⚪ | `usp_traffic_analysis_card.dart` (+2 siblings) | OPEN (R1 W3) — fourth verbatim footer copy, accepted as debt; only ask is that the ticket number lands in a comment at the copy sites. |
| 💡 | 🟢 | test `:1` | Vestigial bare `library;` — it existed only to host the annotation; this is now the only `library;` in `test/` whose purpose was removed. |
| 💡 | 🟢 | test `:1` | Nothing records that `ui` must never be re-added; two comment lines convert a tacit invariant into a review-enforced one. |
| 💡 | 🟢 | `dart_test.yaml` / CI | File the umbrella lint: fail when a widget test under `test/page/**` declares `@Tags(['ui'])`. The `#1183` remedy is a comment with nothing enforcing it — which is why two PRs landed this defect the same day. |
| 💡 | 🟡 | PR description | The before/after `--exclude-tags` measurement lives only in a thread reply; two lines in the body carry it into the merge commit. |
| 💡 | ⚪ | test `:101`/`:136`, `:420`/`:260`/`:60`, footer `identifier:` | R1 suggestions carried, unchanged: twin `walk` closures, magic dimensions, missing stable `identifier:`, Case B untested. |

#### Spec

Spec source: #1301

|---|---|---|---|
| ⚠️ | 🟡 | "this fix still needs release-build verification before being accepted" | [Omission] No release build clicked through at 936×752, no DevTools `pointer-events`/`z-index` reading; all evidence is headless debug, which the issue says cannot see the symptom. |
| ⚠️ | 🟢 | "asserting that **the `sliver_dashboard` grid item node** does not carry…" | [Wrong] The wrapper is hand-replicated (no `sliver_dashboard` import at head); documented and defensible, but the guard is blind to the dependency it guards. |
| 💡 | 🟢 | "Tappable count is irrelevant; position is what matters" (Case B) | [Omission] Case B (`detailRoute` + trailing button) untested although `pumpCardUnderGridBoundary:91` already forwards `footer:` — ~10 lines. |
| 💡 | 🟡 | "An `identifier:` alone accidentally masks the bug… latent trap" | [Omission] None of the four fix-site comments carries the trap forward for the next author. |
| 💡 | 🟡 | "`identifier` values are preserved (E2E-safe)" | [Omission] No test constructs a footer with an identifier and confirms it survives the new boundary. |
| 💡 | ⚪ | "It does **not** close #1008." | [Omission] Handled correctly in the commit bodies; in-source comments cite `#1301` alone. |

*Confidence: 🟢 High / 🟡 Med / ⚪ Low — low-confidence items are kept and marked, not dropped.*
*`[X]` = raised by 1 agent, `[XX]` = independently raised by 2 (Standards table only; the Spec axis has a single reviewer). No Critical rows this round, so no agreement markers appear.*

<details>
<summary>🔴 Standards — Critical: none (Round 1's C1 verified fixed)</summary>

**C1 — RESOLVED ✅ 🟢 High.** Confirmed independently by both Standards reviewers.

The whole Round-2 diff is the removal:

```diff
--- a/test/page/dashboard/views/components/dashboard_card_semantics_absorption_test.dart
+++ b/test/page/dashboard/views/components/dashboard_card_semantics_absorption_test.dart
@@ -1,4 +1,3 @@
-@Tags(['ui'])
 library;
```

Head carries no annotation, and all three CI invocations filter on tags only:

```sh
# run_tests.sh:80 / :92 / :96
$FLUTTER test --exclude-tags="golden||loc||ui"
```

`--exclude-tags` drops a test only when one of *its* tags matches, so an untagged file lands in the default set that `.github/workflows/ci.yml:84` (`sh run_tests.sh`) executes. Grepping every workflow file for tag selectors returns exactly that one invocation — there is no `--tags=ui` job, so `ui` was a delete, not a deferral.

**On the deviation from the suggested fix.** Round 1 proposed `@Tags(['dashboard-card'])`; the author removed the tag outright. Reviewer A judges untagged the *stronger* option (a tag-based exclusion filter has no handle on a file with no tags, whereas `dashboard-card` holds only while it stays out of the exclude expression — the same contingency that caused C1). Reviewer B judges the explicit tag more durable *in this repo*, since `dart_test.yaml` and `.claude/skills/dashboard-overflow-gate/SKILL.md:112` both make `dashboard-card` the machine-readable "must not be excluded" marker, and all three sibling gates carry it despite also exercising `_shared/components/` widgets. Both readings are recorded; the choice is yours. Either way the residual ask is the same one line — take the tag, or leave a comment saying why not.

</details>

<details>
<summary>⚠️ Standards — Warning evidence</summary>

**NW1 (A, 🟡 Med) — the geometric bound is now CI-live.** `test/.../dashboard_card_semantics_absorption_test.dart:156`:

```dart
    expect(footerNode!.rect.height, lessThan(60),
        reason: 'The tappable node spans ${footerNode!.rect.height}px of '
            'height — that is card-sized, not link-sized, so the tap target is '
            'still far larger than the footer link (#1301).');
```

Trigger: a ui_kit spacing/typography change pushes the footer past 60px under the hand-rolled theme at `:56-63` (`seedColor: Colors.blue`, `style: 'flat'`) that production never uses → red CI on an unrelated PR, with failure text blaming a semantics regression that did not happen. Before this commit the file never ran, so W5/W6 were dormant; that is the delta. Capped at Med because `flutter` was unavailable in the review environment, so the *margin* to 60px was not measured — only the coupling. Cheap close: assert a derived bound (`lessThan(cardNode.rect.height / 2)`, or against `find.text(label)`'s rendered size).

**W8 (B, 🟢 High) — implicit convention where the repo has an explicit one.** `dart_test.yaml`:

```yaml
  # failure here blocks the PR.
  dashboard-card:
```

and `.claude/skills/dashboard-overflow-gate/SKILL.md:112-114`: *"Tag must stay `dashboard-card`… Retagging it golden/loc/ui silently drops it from the gate."* Of the 57 files under `test/page/dashboard/`, the untagged ones are overwhelmingly plain unit tests, so untagged carries no signal and this file is invisible to `grep dashboard-card`. One of the author's two stated reasons does not survive the tree: the three sibling gates (`dashboard_card_forced_form_overflow_test.dart:11`, `dashboard_card_popup_overflow_test.dart:8`) also import `_shared/components/dashboard_card_template.dart` yet all carry `dashboard-card`.

**W4 (A+B, 🟢 High) — unguarded `!`, two lines to fix.** `:115` and `:147`, both already on a deprecation path:

```dart
    walk(tester.binding.pipelineOwner.semanticsOwner!.rootSemanticsNode!);
```

Every other failure path in this file carries a considered `reason:`; these two do not, so a deferred root node reads as "harness broken" rather than "semantics were never enabled".

**W1 (A+B, 🟢 High) — the one item both Standards reviewers push hardest.** `container: true` gives the footer its own boundary while the wrapper still supplies `label:` alongside the `AppText` repeating it, and test `:136` matches `n.label.contains(label)` rather than equality. This is a behaviour change shipped by *this PR's production diff*, not inherited debt: a screen-reader user hears the name twice, and `contains` passes for any superset so it cannot catch a future accessible-name regression either. The author's framing — "a test accommodating a defect rather than guarding against it deserves its own decision, not a quiet fix bundled into a tag change" — is a correct read and good scope hygiene, and it is why this is not a blocker. It does need a *recorded* decision (fix here, or a filed ticket referenced at the four fix sites) rather than a silent carry-forward, which is why `need-decide` is applied.

**W9 (B, 🟢 High) — scope note, deliberately not a request to fix here.** Sampled at head: `usp_advanced_settings_identifier_widget_test.dart:1`, `node_detail_popup_identifier_widget_test.dart:1`, `topology_node_identifier_widget_test.dart:1`, `usp_network_health_card_test.dart:1` all `@Tags(['ui'])`, beside directory siblings tagged `dashboard-card`. The identifier guards in that set are the sharpest, since another repo's E2E specs select on those semantics identifiers in 24 places — an app-side deletion passes here and surfaces a day later elsewhere.

</details>

<details>
<summary>⚠️ Spec — findings + Requirements met (17/19)</summary>

**SW1 — release-build acceptance evidence absent (🟡 Med).** Issue #1301, twice: *"this fix still needs release-build verification before being accepted"* and *"Acceptance for both defects must be a real release build clicked through at 936×752, plus DevTools inspection of the node's computed `pointer-events` and `z-index`."* Round 2 does add measured evidence, but about *test execution*, not the symptom:

```
before:  No tests ran.
         No tests match the requested tag selectors:
           exclude: "golden || loc || ui"
after:   00:03 +3: All tests passed!
```

That is `flutter test` in debug. The product-fix verification surface is still three headless widget tests (`:173`, `:189`, `:206`); grep for `936` / `release` / `pointer-events` / `z-index` in the test file returns only the explanatory comment at `:29`. The issue's own analysis is that a debug pass proves nothing here ("In debug the overlay is absent, so the bug is invisible"), so the bar it sets sits deliberately outside the widget-test layer. Discharged by posting a result — no code change implied.

**SW2 — the guard mirrors the `sliver_dashboard` boundary instead of instantiating it (🟢 High).** Head imports are `flutter/material.dart`, `flutter/semantics.dart`, `flutter_test/flutter_test.dart`, `app_localizations.dart`, `dashboard_card_template.dart`, `ui_kit_library/ui_kit.dart` — no `sliver_dashboard`; the only references are prose at `:12` and `:65`. Defensible (the real grid drops the footer under unbounded height, as the issue itself records) and documented, but now more consequential because this is the test CI actually runs: if `sliver_dashboard` changes its wrapper, the hand-copy keeps passing while production regresses. Cheapest close: name the version the mirror was copied from — the issue pins it ("`sliver_dashboard` 0.9.0 wraps every grid item in a boundary"), and `0.9.0` appears nowhere in the test file.

**Requirements met — 17 of 19** (Round 1: 16 of 18; #17 is new because the suggested test's "runs headless / would have caught the regression" property only became verifiable once the tag came off).

1. ✅ `container: true` on `DashboardCardTemplate._buildDetailFooter`'s `Semantics` (Option A, site 1 — covers the 11 template-driven cards).
2. ✅ Same edit at `lib/page/admin/cards/usp_device_info_card.dart`.
3. ✅ Same edit at `lib/page/dashboard/views/components/usp_system_status_card.dart`.
4. ✅ Same edit at `lib/page/dashboard/views/components/usp_traffic_analysis_card.dart`.
5. ✅ Exactly four sites, no more, no fewer — re-confirmed against the full-PR diff over `lib/`.
6. ✅ Option A chosen over Option B — no `explicitChildNodes`, avoiding the screen-reader fragmentation the issue warns about.
7. ✅ Option C rejected — `lib/main.dart` untouched, so `ensureSemantics()` and identifier-based E2E anchoring are preserved.
8. ✅ `AppMenuCard` deliberately untouched (whole-card tap is intentional there).
9. ✅ The four "unaffected, no footer tappable" cards untouched.
10. ✅ Regression test asserts the grid item node carries no `SemanticsAction.tap` for a `detailRoute` card (`:173`), plus `isButton`. Caveat: SW2.
11. ✅ Harness supplies `AppLocalizations` delegates, per the issue's note about `loc(context).viewDetails`.
12. ✅ Non-`const` `breakpoints` map hazard avoided (fixed-size box rather than the real grid).
13. ✅ `performActionAt` not used as the oracle, per the issue's explicit prohibition; asserts node ownership and documents why.
14. ✅ Does not claim to close #1008 — commit body disclaims it and redirects to the `ui_kit_library` ticket.
15. ✅ Does not claim to close #1035.
16. ✅ Vacuous-pass hazard guarded — `expectFooterOwnsItsTapAction` proves the footer rendered and owns a tappable node before asserting the card does not, answering the issue's `[!WARNING]` about a probe that partly measured an absent widget.
17. ✅ **NEW this round** — the suggested regression guard is now actually executable in CI (the entire Round-2 diff), with the before/after measurement recorded in the commit body.
18. ⚠️ Release-build acceptance at 936×752 + DevTools reading — not evidenced (SW1).
19. ⚠️ `sliver_dashboard` grid item node instantiated rather than mirrored — deviated, documented (SW2).

**Scope creep: none.** Every changed line across the PR is either the prescribed `container: true` edit, its explanatory comment, or the regression test the issue requests. Declining to fix the 11 sibling tag-defect files here was the correct call.

</details>

<details>

- **The Critical was retired by measurement, not assertion.** The author ran CI's own exclude expression before and after and posted both outputs (`No tests match the requested tag selectors` → `00:03 +3: All tests passed!`) — the one claim in this PR that could not be established by reading the diff, closed empirically without being asked twice. The sanity check on top (confirming `dashboard-card` files hit 15/39/8/5 in the same log, so the zeros are real rather than a broken lookup) is what makes the number trustworthy.
- **The deviation from the suggested fix was declared, reasoned, and offered back** rather than silently substituted, closing with "happy to switch to `dashboard-card` if you'd rather keep the tag explicit". Reviewer A considers the deviation an improvement on the suggestion; Reviewer B prefers the explicit tag. That the two independent reviewers split on it is itself a sign it was a real judgement call, handled well.
- **The finding was generalised instead of locally patched** — auditing all 12 `ui`-tagged files against a real merged CI log, quantifying it (66 tests, zero executions), spotting that #1347 landed the identical defect the same morning, tracing the cross-repo consequence for the E2E identifier guards, and proposing a lint so the class closes. Materially more than the review asked for.
- **Scope discipline is exact.** One deleted line: no product code, no test-logic edits, no opportunistic reformatting, no drive-by fixes to the 11 siblings (correctly routed to #1353 and a follow-up).
- **Honest, precisely-scoped disclosure** — "I have not re-verified W1–W7 independently; this reply only covers C1, which I did verify end to end", and W1 left open *with a stated reason* rather than quietly skipped. Directly checkable against the diff, and it makes the round cheap to re-review.
- **No security or USP/JNAP correctness surface, confirmed negative.** No secrets or `flutter_secure_storage`, no network/JNAP/USP transport (so `mapUspErrorToServiceError()` and the `on Exception catch` hazard are out of scope), no `catch` added or removed, no provider/service/model change, no `generated/` import, no reverse dependency, no async-ordering or null-handling surface in shipped code. Constitution Art. III/IV/V/XI/XIII/XV checkpoints are all unreachable from a one-line test-annotation deletion.

</details>

AustinChangLinksys (commented):
🤖 **Automated Review — Round 2** · `ad30c2c..c9ee43a` (incremental)

**Verdict: 💬 Self-review (comment only)** — Austin's own PR; automated approval withheld per self-review policy.

#### Standards

|---|---|---|---|
| ✅ | 🟢High | `dashboard_card_semantics_absorption_test.dart:1` | **C1 resolved** — `@Tags(['ui'])` removed; test now runs under `flutter test --exclude-tags="golden\|\|loc\|\|ui"` |
| ⚠️ | 🟢High | `dashboard_card_semantics_absorption_test.dart` | Tag removed entirely instead of moved to `dashboard-card`; `dart_test.yaml` defines `dashboard-card` as the "must run in PR test command" defensive-gate slot — this test semantically belongs there |
| ⚠️ | 🟡Med | `dashboard_card_semantics_absorption_test.dart:96,131` | (carry-forward W2, pre-existing) `pipelineOwner.semanticsOwner!.rootSemanticsNode!` force-unwrap now visible in CI; crash risk on Flutter SDK change |

#### Spec

|---|---|---|---|
| ✅ | 🟢High | `dashboard_card_semantics_absorption_test.dart:1` | C1 fix satisfies CI-gate requirement — test file now included in standard `run_tests.sh` paths |
| ⚠️ | 🟢High | `dashboard_card_semantics_absorption_test.dart` (whole file) | spec#REQ-4 → non-const breakpoints map: issue #1301 explicitly requires it; harness uses hardcoded `SizedBox(width:420, height:260)` with no `breakpoints` variable — zero results in full grep of head |
| ⚠️ | 🟢High | `dashboard_card_semantics_absorption_test.dart` (whole file) | spec#REQ-6 → test non-vacuity: three hand-written footer fix sites (`usp_device_info_card.dart:162`, `usp_system_status_card.dart:145`, `usp_traffic_analysis_card.dart:159`) remain unguarded — removing their `container: true` passes all current tests |
| ⚠️ | 🟢High | (off-diff) | spec#SW1 → release-build 936×752 click-through: spec lists this as an acceptance criterion; author explicitly deferred |

<details>

**N-W1 — Tag removal strategy: should be `@Tags(['dashboard-card'])` not no-tag** 🟢High  
`dashboard_card_semantics_absorption_test.dart` — The incremental diff removes `@Tags(['ui'])` entirely. `dart_test.yaml` (head version) documents the `dashboard-card` tag as:
```yaml
  # failure here blocks the PR.
  dashboard-card:
```
This test is precisely a defensive widget gate for dashboard card semantics. It semantically fits `dashboard-card` exactly. The author's reason for not using it — "tag taxonomy is under separate review" — is noted, but leaving the test untagged:
- Obscures routing intent: future maintainers see no tag and don't know which CI gate owns it
- Creates fragile implicit membership: if `run_tests.sh` adds new exclusion categories (e.g. `--exclude-tags="slow"`), this test could be silently dropped again with no classification to defend it
- Sets a precedent that the other 11 `ui`-tagged files could also have their tags stripped rather than reclassified

Author explicitly noted the taxonomy is "under separate review" — if that's the case, `dashboard-card` is the right landing zone as-is.  
Fix: replace `@Tags(['ui'])` removal with `@Tags(['dashboard-card'])` assignment. One-line change.

**W2 carry-forward — `pipelineOwner.semanticsOwner!.rootSemanticsNode!` force-unwrap** 🟡Med  
`dashboard_card_semantics_absorption_test.dart:96,131` — Two occurrences of:
```dart
// ignore: deprecated_member_use
walk(tester.binding.pipelineOwner.semanticsOwner!.rootSemanticsNode!);
```
Pre-existing from Round 1 W2; author confirmed "W2 deliberately not addressed." Now that C1 is fixed and this test runs in CI, a future Flutter SDK upgrade that changes semantics tree initialization could cause `Null check operator used on a null value` with no context. Not a blocker, but the risk window is now open.  
Fix: guard with null check + `fail(...)` message, or track as a TODO for the deprecated API migration.

**Spec W1 — Non-const breakpoints map missing** 🟢High  
Issue #1301 "Suggested regression test" explicitly states: *"Note the harness needs `AppLocalizations.localizationsDelegates` (the footer calls `loc(context).viewDetails`) and a **non-`const` breakpoints map**."* The `AppLocalizations` delegate is present (line 76); a breakpoints map is not. `grep -n "breakpoints"` on the full head file returns zero results. The harness substitutes `SizedBox(width: 420, height: 260)` for layout constraints. If `DashboardCardTemplate` reads breakpoints from a provider at footer layout time and the harness falls back to a degenerate default silently, tests may pass vacuously. Spec explicitly called this out as required for the harness.  
Fix: add a non-const breakpoints `Map` to the test harness — the exact form depends on what provider `DashboardCardTemplate` reads (likely an `InheritedWidget` or `Provider`). Needs investigation.

**Spec W2 — Three hand-written footer sites unguarded** 🟢High  
The absorption fix covers 11 cards via `DashboardCardTemplate._buildDetailFooter` plus three hand-written footers. The test's `expectFooterOwnsItsTapAction` oracle only tests `DashboardCardTemplate`'s built-in route. Removing `container: true` from `usp_device_info_card.dart:162`, `usp_system_status_card.dart:145`, or `usp_traffic_analysis_card.dart:159` would not fail any current test. Issue #1301's blast radius section identified all four sites; spec REQ-6 (non-vacuity) applies to all of them.  
Fix: add three `testWidgets` cases — one per hand-written card — using adapted versions of `pumpCardUnderGridBoundary`/`expectFooterOwnsItsTapAction`.

**Spec SW1 — Release-build click-through deferred** 🟢High  
Issue #1301 "Caveats" explicitly: *"A full release-build click-through at the reported 936×752 viewport has not yet been run, and should be part of the fix's acceptance."* Author confirmed: "SW1 still outstanding." This is the acceptance criterion that validates the actual symptom (DOM hit test routing via `<flt-semantics>` overlay) rather than just the widget-tree semantics.  
Noted as scope decision for Austin: accept as follow-up, or require before merge.

</details>

<details>

- **C1 completely resolved.** Removing `@Tags(['ui'])` is a correct and effective fix. Verified: `run_tests.sh` uses `--exclude-tags="golden||loc||ui"` in all three of its `flutter test` invocations; an untagged test file is not matched by any exclude rule and will execute normally. Author's pre/post measurement (`before: No tests ran` → `after: 00:03 +3: All tests passed!`) is consistent with the CI mechanism.
- **Minimal, surgical scope.** The diff is one line — no production code, no test logic changes, no accidental extras. Exactly right for addressing a single CI configuration issue.
- **Author's commentary is detailed and helpful.** The commit message includes a thorough explanation of the CI mechanism, the taxonomy decision, and the cross-PR impact (11 other `ui`-tagged files), making the reasoning auditable.
- **`AppLocalizations.localizationsDelegates` already present** (spec REQ-4 primary requirement) — confirmed at test line 76.
- **Non-vacuity for central path confirmed.** Author re-measured: removing `container: true` from `dashboard_card_template.dart:505` fails 2 of 3 tests with 260px tappable node instead of link-sized. Oracle is sound for the main path.
- **No golden/layout regressions.** One-line tag removal has zero effect on widget rendering or test semantics.

</details>

*Cross-reviewed by two independent agents (security+correctness / architecture+maintainability) plus a spec compliance check. Automated — please sanity-check before merge.*

Issue #1322, opened 2026-08-20
Remote Assistance: activate() frees a UspClient that ~40 services still hold, killing every USP GET until page refresh


AustinChangLinksys:
## Summary

Follow-up from #1317 (part 2 of that triage) and the remaining half of #1309.

`RemoteAssistanceNotifier.activate()` disposes the previously registered `UspClient` as part of the
swap. `dispose()` reaches `free()` on the wasm-bindgen object, but ~40 feature services captured
that instance with `ref.read(uspClientProvider)` in non-autoDispose provider bodies, so they keep
using it after it has been freed. Every subsequent USP GET fails instantly with
`Error: null pointer passed to rust`.

```dart
// lib/core/usp/providers/remote_assistance_provider.dart:97-102
if (getIt.isRegistered<UspClient>()) {
  logger.d('[RA] Unregistering existing UspClient');
  final oldClient = getIt<UspClient>();
  getIt.unregister<UspClient>();
  oldClient.dispose();              // -> UspClientWeb.dispose() -> _client.free()
}
```

`dispose()` → `usp_client.dart:1099` → `usp_client_wasm.dart:375` → `_client.free()`, which zeroes
`__wbg_ptr` on the JS side. Any later method call on that object is rejected by wasm-bindgen before
a request is issued.

## Why the holders don't follow the swap

`uspClientProvider` (`lib/core/usp/providers/usp_client_provider.dart:15`) is a plain
non-autoDispose `Provider`, and **41 call sites read it with `ref.read`** inside provider bodies —
they bind the instance into a service constructor once and never re-read:

```dart
// lib/page/wifi_settings/services/usp_wifi_data_service.dart:23-31 (representative)
final uspWifiDataServiceProvider = Provider<UspWifiDataService>((ref) {
  final usp = ref.read(uspClientProvider);
  if (usp == null) throw const ServiceNotInitializedError(...);
  return UspWifiDataService(usp);
});
```

Same shape in `usp_ethernet_data_service.dart:18`, `usp_wan_data_service.dart:20`,
`usp_lan_data_service.dart:17`, `usp_devices_data_service.dart:24`, and ~36 more. Invalidating
`uspClientProvider` (as #1309 does inside the swap on `dev-2.6.1`) only reaches the 11 `ref.watch`
consumers.

## Evidence

From #1317's attached log — Remote build `2.6.1.600011`, second `activate()` in the same page
lifetime:

| time | event |
|---|---|
[...]

Issue #1323, opened 2026-08-20
Remote Assistance: recovery flow is local-mode only — logout affordance, redirect loop, and a probe that can force logout


AustinChangLinksys:
## Summary

Split out from #1317 (the reporter-facing symptom). The recovery subsystem
(`lib/core/connection/` + `lib/page/_shared/helpers/recovery_dialog_helper.dart`) was designed for
local, on-router sessions and has **no Remote Assistance branch anywhere**. In RA every one of its
three moving parts — the exit affordance, the post-exit routing, and the probe — is either
meaningless or actively harmful.

This is not a fault-tolerance gap; the flow has no correct behaviour defined for RA at all.

## 1. The exit affordance has no meaning in Remote Assistance

Both recovery dialogs offer the same button:

```dart
// lib/page/_shared/helpers/recovery_dialog_helper.dart:61-67
AppButton.text(
  label: loc(context).returnToLoginPage,
  onTap: () {
    logger.d('[Recovery] User tapped Return to login');
    ref.read(appConnectionStateProvider.notifier).exitToLogout();
  },
),
```

```dart
// lib/page/firmware_update/views/dialogs/firmware_update_recovery_dialog.dart:151-156
AppButton.text(
  label: loc(context).returnToLoginPage,
  onTap: () { notifier.exitToLogout(); },
),
```

`exitToLogout()` (`lib/core/connection/providers/app_connection_state_provider.dart:138-149`)
cancels the timers and calls `authProvider.logout()`. That is all.

**A Remote build has no login page.** `autoConfigurationLogic` sends every entry point to
`remoteAssistanceConfirm` when `BuildConfig.isRemote()` (`lib/route/router_provider.dart:205-208`),
and there is no password prompt in the RA flow at all. The only lifecycle verb RA has is *end
session*, and the correct implementation already exists — `_disconnect()` in
`lib/page/_shared/components/remote_session_chip.dart:204-236`, which does **four** things:

1. `service.endSessionForCA(...)` — ends the Guardian session for real
2. `notifier.clearSession()` — clears `remoteAccessProvider` + sessionStorage
3. `router.go('${RoutePath.remoteAssistanceConfirm}?ended=true')`
4. `authNotifier.logout()`

The recovery dialogs do only step 4. From #1317's log, the Guardian session provably survived the
"logout":

```
17:08:41.306 [D][Recovery] User tapped Return to login
17:08:41.307 [D][Auth]: logout: complete
[...]

PR #1331, opened 2026-08-20
merge: promote dev-2.6.1 Remote Assistance fixes into dev-2.7.0 (#1330)

### LINKED ISSUE #1330: Promote dev-2.6.1 Remote Assistance fixes into dev-2.7.0 (merge verification)
## Summary

Promote the three Remote Assistance fixes that shipped only on `dev-2.6.1` into `dev-2.7.0`.

`2.6.1` is a hotfix-only release, so its RA fixes never reached the 2.7.0 line. #1322's proposed fix
builds directly on `6d1aef0a` — without it, `dev-2.7.0` still registers a boot-time
`UspClient(Uri.base.origin)` in a Remote build, which means even the **first** `activate()` frees an
instance that boot-time consumers already captured. So this merge is a prerequisite for #1322, and
it also closes the RA-host and client-cache gaps on 2.7.0 in their own right.

## What is missing on `dev-2.7.0`

`git log --oneline --no-merges origin/dev-2.7.0..origin/dev-2.6.1`:

| commit | change |
|---|---|
| `9cf72b52` | `fix(remote-assistance): send every RA call to the Guardian host` — adds `RemoteAssistanceConfig.guardianOrigin` and a `baseUrl` override on `UspBridgeClient`, so REST/SSE stop deriving the host from `UspClient.baseUrl` |
| `64306778` | `fix(remote-assistance): invalidate the client cache inside the swap` — `ref.invalidate(uspClientProvider)` inside the mutation-lock critical section, in both `activate()` and `deactivate()` |
| `6d1aef0a` | `fix(remote-assistance): stop building a USP client on the app's own origin (#1309)` — adds `canUseAppOriginUspClient()`, skips the boot-time GetIt registration in Remote builds |
| `aadb14ef` | `chore: bump version to 2.6.1` — superseded by 2.7.0's own bump |

Touched surface: 6 files under `lib/`, 2 test files, 169 insertions.

## Merge assessment — merge, do not rewrite

Dry-run performed in a throwaway worktree off `origin/dev-2.7.0`.

**One conflict, trivial:**

```
CONFLICT (content): Merge conflict in pubspec.yaml
  HEAD:             version: 2.7.0+100000
  origin/dev-2.6.1: version: 2.6.1+100000
```

Resolution: keep `2.7.0+100000`. Every other file auto-merges clean — `dev-2.7.0` has **not touched
a single one** of the six `lib/` files since the merge-base (`5beff305`); its 214 own commits all sit
on the dashboard / topology / test / ui_kit side.

```
$ git rev-list --count 5beff305..origin/dev-2.7.0 -- <each of the 6 lib files>
0  lib/core/usp/providers/remote_assistance_provider.dart
0  lib/core/usp/providers/sse_providers.dart
0  lib/core/usp/providers/usp_client_provider.dart
0  lib/core/usp/services/usp_bridge_client_base.dart
0  lib/core/usp/services/usp_bridge_client_web.dart
0  lib/di.dart
```

`test/di_test.dart` is a clean add — 2.7.0 has no file at that path.

**Semantic premises still hold on 2.7.0:**

- `canUseAppOriginUspClient()` reads `BuildConfig.isRemote()`. `lib/constants/build_config.dart` is
  byte-identical between the branches, and 2.7.0's `di.dart:46-55` still registers the boot client
  unconditionally — exactly the state `6d1aef0a` was written against.
- The 41 `ref.read(uspClientProvider)` sites are the same set on both branches (one file's line
  number shifted by 1). `ref.watch` sites: 2.7.0 adds one, `wan_interface_path_provider.dart:63`,
  which uses `ref.watch` and therefore follows the invalidation correctly.
- `uspBridgeClientProvider` has the same shape on both branches, so the `baseUrl:` override applies
  unchanged.

**Verification on the merged tree:**

| check | result |
|---|---|
| `flutter analyze` (whole repo) | 474 issues, **0 error / 0 warning** — all pre-existing `curly_braces_in_flow_control_structures` info |
| `flutter analyze lib/di.dart lib/core/usp/ test/di_test.dart test/core/usp/providers/remote_assistance_provider_test.dart` | No issues found |
| `flutter test test/di_test.dart test/core/usp/` | 459 passed |
| `./run_tests.sh` (full functional suite) | **7118 passed** |

## Known consequence, carried over from 2.6.1

After `6d1aef0a`, `uspClientProvider` returns `null` in a Remote build until `activate()` runs. The
tree has 16 `ref.read(uspClientProvider)!` force-unwraps (`pnp_service.dart:27`,
`usp_admin_service.dart:12`, `usp_wifi_settings_service.dart:20`, …) that would throw if read that
early. This is `6d1aef0a`'s deliberate trade-off — its own comment says the worst a too-early read
can capture is null, "which fails loudly instead of silently talking to the wrong host" — not a
regression introduced by the merge. #1322 removes the whole class by keeping the `UspClient` façade
instance stable and rebinding its transport.

## Plan

- Branch `merge/dev-2.6.1-to-dev-2.7.0` from `origin/dev-2.7.0`, merge `origin/dev-2.6.1`, resolve
  `pubspec.yaml` to `2.7.0+100000`.
- PR base `dev-2.7.0`. Promotion merge of already-reviewed commits — no new code.
- Do **not** cherry-pick: a real merge makes `dev-2.6.1` an ancestor of `dev-2.7.0`, so any further
  2.6.1 hotfix promotes without replaying the same conflicts.

## Acceptance criteria

- [ ] `git merge-base --is-ancestor origin/dev-2.6.1 origin/dev-2.7.0` succeeds after the merge.
- [ ] `pubspec.yaml` on `dev-2.7.0` still reads `version: 2.7.0+100000`.
- [ ] `flutter analyze` reports no new error or warning versus `dev-2.7.0`.
- [ ] `./run_tests.sh` passes, including the added `test/di_test.dart`.
- [ ] No `lib/` file outside the six listed above is modified by the merge.

## Relationship

- Prerequisite for #1322 — that fix's implementation premise assumes `6d1aef0a` is present.
- #1309 — the issue `6d1aef0a` fixed; its AC 2 remains open on 2.7.0 until #1322 lands.
- #1317 / #1323 — the RA symptom ticket and the recovery-flow root cause; unaffected by this merge.

AustinChangLinksys:
Closes #1330

## Summary

Promotion merge — brings the three Remote Assistance fixes that shipped only on the `dev-2.6.1`
hotfix line into `dev-2.7.0`. **No new code**; every commit here was already reviewed and merged on
2.6.1.

| commit | change |
|---|---|
| `6d1aef0a` | `fix(remote-assistance): stop building a USP client on the app's own origin (#1309)` — adds `canUseAppOriginUspClient()`, skips the boot-time GetIt registration in Remote builds |
| `64306778` | `fix(remote-assistance): invalidate the client cache inside the swap` — `ref.invalidate(uspClientProvider)` inside the mutation-lock critical section, in both `activate()` and `deactivate()` |
| `9cf72b52` | `fix(remote-assistance): send every RA call to the Guardian host` — adds `RemoteAssistanceConfig.guardianOrigin` and a `baseUrl` override on `UspBridgeClient`, so REST/SSE stop deriving the host from `UspClient.baseUrl` |

`aadb14ef` (version bump to 2.6.1) is superseded by 2.7.0's own bump.

This is also a prerequisite for #1322: that fix builds on `6d1aef0a`. Without it, `dev-2.7.0` still
registers a boot-time `UspClient(Uri.base.origin)` in a Remote build, so even the **first**
`activate()` frees an instance that boot-time consumers already captured.

## Conflict resolution

One conflict, in `pubspec.yaml` only:

```
HEAD:             version: 2.7.0+100000
origin/dev-2.6.1: version: 2.6.1+100000
```

Resolved to `2.7.0+100000`. Everything else auto-merged — `dev-2.7.0` has not touched a single one
of the six `lib/` files since the merge-base (`5beff305`); its own commits since then all sit on the
dashboard / topology / test / ui_kit side. `test/di_test.dart` is a clean add.

## Why merge rather than cherry-pick

A real merge makes `dev-2.6.1` an ancestor of `dev-2.7.0`, so any further 2.6.1 hotfix promotes
without replaying these same conflicts. Cherry-picking would leave the branches permanently
divergent on the same content.

## Verification (on the merged tree)

| check | result |
|---|---|
| `fvm dart format --set-exit-if-changed` on the 8 changed files | 0 changed |
| `flutter analyze lib/core/usp/ lib/di.dart` + both test files | No issues found |
| `flutter analyze` (whole repo) | 474 issues, **0 error / 0 warning** — all pre-existing `curly_braces_in_flow_control_structures` info |
| affected tests (34 files: `test/core/usp/`, `test/core/connection/`, auth, admin, wifi, local_network, internet_settings, devices, diagnostics, firmware, pnp) | **797 passed** |
| `./run_tests.sh` (full functional suite) | **7118 passed** |
| `git merge-base --is-ancestor origin/dev-2.6.1 HEAD` | passes |

Semantic premises re-checked against 2.7.0:

- `lib/constants/build_config.dart` is byte-identical between the branches, and 2.7.0's `di.dart`
  still registers the boot client unconditionally — exactly the state `6d1aef0a` was written against.
- The 41 `ref.read(uspClientProvider)` sites are the same set on both branches. 2.7.0's one new
  `ref.watch` consumer, `wan_interface_path_provider.dart:63`, follows the invalidation correctly.
- `uspBridgeClientProvider` has the same shape on both branches, so the `baseUrl:` override applies
  unchanged.

## Known consequence, carried over from 2.6.1

After `6d1aef0a`, `uspClientProvider` returns `null` in a Remote build until `activate()` runs, and
the tree has 16 `ref.read(uspClientProvider)!` force-unwraps that would throw on a too-early read.
That is `6d1aef0a`'s deliberate trade-off — a too-early read captures null, which fails loudly
instead of silently talking to the wrong host — not a regression introduced here. #1322 removes the
whole class by keeping the `UspClient` façade instance stable and rebinding its transport.

## Review notes

One INFO from the pre-PR review, pre-existing rather than introduced by this merge: the `baseUrl`
override on `UspBridgeClient` has no dedicated test file. `guardianOrigin` itself is covered
(`remote_assistance_provider_test.dart:43`), but "the override actually reaches the bridge client"
is only covered indirectly.

Refs #1309, #1322

### DIFF
--- a/lib/core/usp/providers/remote_assistance_provider.dart
+++ b/lib/core/usp/providers/remote_assistance_provider.dart
@@ -127,15 +127,15 @@ class RemoteAssistanceNotifier extends Notifier<RemoteAssistanceState> {
       //
       // uspClientProvider caches whatever GetIt held when it was FIRST read,
       // and authProvider.init() reads it during app boot — long before RA
-      // activates. Without this, every USP request and every bridge call keeps
-      // using the disposed boot-time client, whose baseUrl is the web app's own
-      // origin instead of the Guardian API host.
+      // activates. In a Remote build that first read now yields null
+      // ([canUseAppOriginUspClient] keeps the boot slot empty), and this drops
+      // that cached null so watchers pick the session client up.
       //
-      // Invariant: only ref.watch consumers rebuild. Feature services capture
-      // the client via ref.read(uspClientProvider) in the body of a
-      // non-autoDispose provider, so they keep their first instance for the
-      // container's lifetime — the swap MUST happen before any feature service
-      // provider is first read.
+      // Only ref.watch consumers rebuild; a ref.read consumer keeps whatever it
+      // captured. That is no longer a correctness risk: in a Remote build the
+      // only client that can ever exist is this one, so the worst a too-early
+      // read can capture is null — which fails loudly instead of silently
+      // talking to the wrong host.
       ref.invalidate(uspClientProvider);
     });
 
--- a/lib/core/usp/providers/remote_assistance_provider.dart
+++ b/lib/core/usp/providers/remote_assistance_provider.dart
@@ -1,6 +1,7 @@
 import 'package:flutter/foundation.dart';
 import 'package:flutter_riverpod/flutter_riverpod.dart';
 import 'package:privacy_gui/constants/cloud_const.dart';
+import 'package:privacy_gui/core/usp/providers/usp_client_provider.dart';
 import 'package:privacy_gui/core/usp/providers/usp_mutation_lock.dart';
 import 'package:privacy_gui/core/usp/services/usp_client.dart';
 import 'package:privacy_gui/core/utils/logger.dart';
@@ -20,6 +21,13 @@ class RemoteAssistanceConfig {
     required this.temporaryAccessToken,
     this.clientTypeId,
   });
+
+  /// Guardian API origin — the single host for ALL Remote Assistance traffic
+  /// (session REST, USP requests, subscriptions, SSE notifications).
+  ///
+  /// Must NOT be confused with the origin the web app is served from; the
+  /// Guardian API lives on a different host (e.g. `qa.guardian.tools`).
+  String get guardianOrigin => 'https://$guardianBaseUrl';
 
   /// Constructs the Guardian USP endpoint path.
   String get uspEndpoint =>
@@ -75,11 +83,11 @@ class RemoteAssistanceNotifier extends Notifier<RemoteAssistanceState> {
     }
 
     logger.i('[RA] Activating Remote Assistance: ${config.sessionId}');
-    logger.d('[RA] Guardian URL: https://${config.guardianBaseUrl}');
+    logger.d('[RA] Guardian URL: ${config.guardianOrigin}');
     logger.d('[RA] USP Endpoint: ${config.uspEndpoint}');
 
     // Build the client using UspClientBuilder
-    var builder = UspClientBuilderJS('https://${config.guardianBaseUrl}')
+    var builder = UspClientBuilderJS(config.guardianOrigin)
         .endpoint(config.uspEndpoint)
         .authToken(config.temporaryAccessToken);
 
@@ -89,8 +97,8 @@ class RemoteAssistanceNotifier extends Notifier<RemoteAssistanceState> {
     }
 
     final jsClient = builder.build();
-    final guardianUrl = 'https://${config.guardianBaseUrl}';
-    final raClient = UspClient.fromBuilder(jsClient, baseUrl: guardianUrl);
+    final raClient =
+        UspClient.fromBuilder(jsClient, baseUrl: config.guardianOrigin);
 
     // Swap UspClient atomically with mutation lock to prevent races
     await ref.read(uspMutationLockProvider).withLock(() async {
@@ -111,6 +119,13 @@ class RemoteAssistanceNotifier extends Notifier<RemoteAssistanceState> {
       isActive: true,
       config: config,
     );
+
+    // uspClientProvider caches whatever GetIt held when it was FIRST read
+    // (authProvider.init() reads it during app boot, before RA activates).
+    // Without this invalidation, every USP request and every bridge call keeps
+    // using the disposed boot-time client — whose baseUrl is the web app's own
+    // origin, not the Guardian API host.
+    ref.invalidate(uspClientProvider);
   }
 
   /// Deactivates Remote Assistance mode.
@@ -131,6 +146,10 @@ class RemoteAssistanceNotifier extends Notifier<RemoteAssistanceState> {
     });
 
     state = const RemoteAssistanceState();
+
+    // Drop the cached (now disposed) client so downstream providers tear down
+    // instead of calling into freed WASM memory.
+    ref.invalidate(uspClientProvider);
   }
 }
 
--- a/lib/core/usp/providers/remote_assistance_provider.dart
+++ b/lib/core/usp/providers/remote_assistance_provider.dart
@@ -22,12 +22,23 @@ class RemoteAssistanceConfig {
     this.clientTypeId,
   });
 
-  /// Guardian API origin — the single host for ALL Remote Assistance traffic
-  /// (session REST, USP requests, subscriptions, SSE notifications).
+  /// Guardian API origin — the host every USP-over-Guardian call must use:
+  /// USP requests, subscriptions and SSE notifications.
   ///
   /// Must NOT be confused with the origin the web app is served from; the
   /// Guardian API lives on a different host (e.g. `qa.guardian.tools`).
-  String get guardianOrigin => 'https://$guardianBaseUrl';
+  ///
+  /// The session REST API ends up on the same host by construction — both this
+  /// config ([RemoteAssistanceConfig.guardianBaseUrl], set from
+  /// `cloudEnvironmentConfig[kCloudBase]`) and `GuardianApiClient._buildUrl`
+  /// read that same entry — but it builds its URL independently.
+  String get guardianOrigin {
+    // Kept as an assert rather than a constructor assert: the constructor is
+    // const, and const asserts only accept potentially-constant expressions.
+    assert(!guardianBaseUrl.contains('://'),
+        'guardianBaseUrl must be a bare host, without a scheme');
+    return 'https://$guardianBaseUrl';
+  }
 
   /// Constructs the Guardian USP endpoint path.
   String get uspEndpoint =>
@@ -110,6 +121,22 @@ class RemoteAssistanceNotifier extends Notifier<RemoteAssistanceState> {
       }
       getIt.registerSingleton<UspClient>(raClient);
       logger.i('[RA] UspClient replaced with Guardian-proxied client');
+
+      // Invalidated in the same critical section as the swap, so "client
+      // identity changed" and "cache dropped" can never drift apart.
+      //
+      // uspClientProvider caches whatever GetIt held when it was FIRST read,
+      // and authProvider.init() reads it during app boot — long before RA
+      // activates. Without this, every USP request and every bridge call keeps
+      // using the disposed boot-time client, whose baseUrl is the web app's own
+      // origin instead of the Guardian API host.
+      //
+      // Invariant: only ref.watch consumers rebuild. Feature services capture
+      // the client via ref.read(uspClientProvider) in the body of a
+      // non-autoDispose provider, so they keep their first instance for the
+      // container's lifetime — the swap MUST happen before any feature service
+      // provider is first read.
+      ref.invalidate(uspClientProvider);
     });
 
     // Set login type to remote so auth checks pass
@@ -119,13 +146,6 @@ class RemoteAssistanceNotifier extends Notifier<RemoteAssistanceState> {
       isActive: true,
       config: config,
     );
-
-    // uspClientProvider caches whatever GetIt held when it was FIRST read
-    // (authProvider.init() reads it during app boot, before RA activates).
-    // Without this invalidation, every USP request and every bridge call keeps
-    // using the disposed boot-time client — whose baseUrl is the web app's own
-    // origin, not the Guardian API host.
-    ref.invalidate(uspClientProvider);
   }
 
   /// Deactivates Remote Assistance mode.
@@ -143,13 +163,13 @@ class RemoteAssistanceNotifier extends Notifier<RemoteAssistanceState> {
         client.dispose();
         logger.d('[RA] UspClient unregistered and disposed');
       }
+
+      // Drop the cached instance in the same critical section, so watchers
+      // rebuild against an empty GetIt instead of holding a disposed client.
+      ref.invalidate(uspClientProvider);
     });
 
     state = const RemoteAssistanceState();
-
-    // Drop the cached (now disposed) client so downstream providers tear down
-    // instead of calling into freed WASM memory.
-    ref.invalidate(uspClientProvider);
   }
 }
 
--- a/lib/core/usp/providers/sse_providers.dart
+++ b/lib/core/usp/providers/sse_providers.dart
@@ -38,6 +38,8 @@ final uspBridgeClientProvider = Provider<UspBridgeClient?>((ref) {
     bridge = UspBridgeClient(
       usp,
       endpoints: BridgeEndpoints.remote(config.sessionId),
+      // Same host as the Guardian session REST API — NOT the app's own origin.
+      baseUrl: config.guardianOrigin,
       authToken: config.temporaryAccessToken,
       clientTypeId: config.clientTypeId,
       authBehavior: AuthBehavior.remote,
--- a/lib/core/usp/providers/usp_client_provider.dart
+++ b/lib/core/usp/providers/usp_client_provider.dart
@@ -10,6 +10,15 @@ import 'bridge_request_throttler_provider.dart';
 /// Returns null on non-Web platforms since USP is only available
 /// through the WASM transport layer.
 ///
+/// Also null in a Remote Assistance build until
+/// `RemoteAssistanceNotifier.activate()` has registered the session client —
+/// such builds get no boot-time client, because the app's own origin is not the
+/// USP host (see `canUseAppOriginUspClient` in `di.dart`).
+///
+/// Caches for the container's lifetime: a `ref.read` consumer keeps the instance
+/// it first saw. `activate()` invalidates this provider after the swap so
+/// `ref.watch` consumers follow.
+///
 /// Binds the [BridgeRequestThrottler] so all `usp.get()` calls are
 /// automatically throttled to prevent overwhelming the router.
 final uspClientProvider = Provider<UspClient?>((ref) {
--- a/lib/core/usp/services/usp_bridge_client_base.dart
+++ b/lib/core/usp/services/usp_bridge_client_base.dart
@@ -25,6 +25,7 @@ class UspBridgeClient {
   UspBridgeClient(
     UspClient usp, {
     BridgeEndpoints? endpoints,
+    String? baseUrl,
     String? authToken,
     String? clientTypeId,
     AuthBehavior authBehavior = AuthBehavior.local,
--- a/lib/core/usp/services/usp_bridge_client_web.dart
+++ b/lib/core/usp/services/usp_bridge_client_web.dart
@@ -38,6 +38,7 @@ external set _jsSseAbort(JSAny? value);
 class UspBridgeClient {
   final UspClient _usp;
   final BridgeEndpoints _endpoints;
+  final String? _overrideBaseUrl;
   final String? _overrideToken;
   final String? _clientTypeId;
   final AuthBehavior _authBehavior;
@@ -48,10 +49,12 @@ class UspBridgeClient {
   UspBridgeClient(
     this._usp, {
     BridgeEndpoints? endpoints,
+    String? baseUrl,
     String? authToken,
     String? clientTypeId,
     AuthBehavior authBehavior = AuthBehavior.local,
   })  : _endpoints = endpoints ?? BridgeEndpoints.local,
+        _overrideBaseUrl = baseUrl,
         _overrideToken = authToken,
         _clientTypeId = clientTypeId,
         _authBehavior = authBehavior;
@@ -60,7 +63,12 @@ class UspBridgeClient {
   /// synchronously from a `beforeunload` handler.
   web.AbortController? _sseAbortController;
 
-  String get _baseUrl => _usp.baseUrl;
+  /// Host for REST/SSE calls.
+  ///
+  /// Remote mode passes the Guardian API origin explicitly — it must not be
+  /// derived from [UspClient.baseUrl], which may still point at the origin the
+  /// web app was served from.
+  String get _baseUrl => _overrideBaseUrl ?? _usp.baseUrl;
 
   String get _token {
     // Remote mode: use override token (temporaryAccessToken from Guardian)
--- a/lib/di.dart
+++ b/lib/di.dart
@@ -1,6 +1,7 @@
 import 'package:flutter/foundation.dart';
 import 'package:flutter/material.dart';
 import 'package:get_it/get_it.dart';
+import 'package:privacy_gui/constants/build_config.dart';
 import 'package:privacy_gui/core/utils/logger.dart';
 import 'package:privacy_gui/theme/theme_json_config.dart';
 import 'package:privacy_gui/core/usp/services/usp_client.dart';
@@ -44,13 +45,34 @@ void dependencySetup() {
 
   // Register UspClient on Web platform only
   if (kIsWeb && !getIt.isRegistered<UspClient>()) {
-    try {
-      getIt.registerSingleton<UspClient>(UspClient(
-        Uri.base.origin,
-      ));
-      logger.d('[DI]: UspClient registered (Web)');
-    } catch (e) {
-      logger.w('[DI]: UspClient registration failed: $e');
+    if (canUseAppOriginUspClient()) {
+      try {
+        getIt.registerSingleton<UspClient>(UspClient(
+          Uri.base.origin,
+        ));
+        logger.d('[DI]: UspClient registered (Web)');
+      } catch (e) {
+        logger.w('[DI]: UspClient registration failed: $e');
+      }
+    } else {
+      logger.d('[DI]: UspClient boot registration skipped — Remote build, '
+          'the client is built from the Remote Assistance session config');
     }
   }
 }
+
+/// Whether a boot-time [UspClient] may be built against the app's own origin.
+///
+/// True for Local (and unforced) builds: the app is served by the router itself,
+/// so `Uri.base.origin` IS the USP host.
+///
+/// False for Remote Assistance builds: there the app is served from static
+/// hosting and USP is reached through the Guardian API on a different host. A
+/// boot client would therefore point at a host that answers no USP request, and
+/// because [uspClientProvider] caches whatever GetIt held when it was FIRST read
+/// — during `authProvider.init()`, long before a session exists — that wrong
+/// client is what every consumer keeps. Leaving the slot empty makes
+/// `RemoteAssistanceNotifier.activate()` the only thing that can ever create the
+/// client, so no consumer can capture a pre-session instance.
+@visibleForTesting
+bool canUseAppOriginUspClient() => !BuildConfig.isRemote();
--- a/pubspec.yaml
+++ b/pubspec.yaml
@@ -15,7 +15,7 @@ publish_to: "none" # Remove this line if you wish to publish to pub.dev
 # In iOS, build-name is used as CFBundleShortVersionString while build-number used as CFBundleVersion.
 # Read more about iOS versioning at
 # https://developer.apple.com/library/archive/documentation/General/Reference/InfoPlistKeyReference/Articles/CoreFoundationKeys.html
-version: 2.6.0+100000
+version: 2.6.1+100000
 
 environment:
   sdk: ">=3.3.0 <4.0.0"
--- a/test/core/usp/providers/remote_assistance_provider_test.dart
+++ b/test/core/usp/providers/remote_assistance_provider_test.dart
@@ -39,6 +39,16 @@ void main() {
       expect(config.clientTypeId, 'client-type-456');
     });
 
+    test('guardianOrigin prefixes the Guardian API host with https', () {
+      const config = RemoteAssistanceConfig(
+        guardianBaseUrl: 'qa.guardian.tools',
+        sessionId: 'session-123',
+        temporaryAccessToken: 'token-abc',
+      );
+
+      expect(config.guardianOrigin, 'https://qa.guardian.tools');
+    });
+
     test('uspEndpoint returns correct path format', () {
       const config = RemoteAssistanceConfig(
         guardianBaseUrl: 'api.example.com',
--- a/test/core/usp/providers/remote_assistance_provider_test.dart
+++ b/test/core/usp/providers/remote_assistance_provider_test.dart
@@ -2,6 +2,7 @@ import 'package:flutter_riverpod/flutter_riverpod.dart';
 import 'package:flutter_test/flutter_test.dart';
 import 'package:mocktail/mocktail.dart';
 import 'package:privacy_gui/core/usp/providers/remote_assistance_provider.dart';
+import 'package:privacy_gui/core/usp/providers/usp_client_provider.dart';
 import 'package:privacy_gui/providers/auth/auth_provider.dart';
 
 class MockAuthNotifier extends Mock implements AuthNotifier {}
@@ -187,16 +188,40 @@ void main() {
       expect(state.config, isNull);
     });
 
-    test('deactivate resets state to inactive', () {
+    test('deactivate resets state to inactive', () async {
       // First manually set an active state via notifier
       final notifier = container.read(remoteAssistanceProvider.notifier);
 
-      // Call deactivate
-      notifier.deactivate();
+      // Must be awaited — deactivate() resets state only after the GetIt swap
+      // completes inside the mutation lock.
+      await notifier.deactivate();
 
       final state = container.read(remoteAssistanceProvider);
       expect(state.isActive, false);
       expect(state.config, isNull);
+    });
+
+    test('deactivate invalidates uspClientProvider', () async {
+      var builds = 0;
+      final container = ProviderContainer(
+        overrides: [
+          authProvider.overrideWith(() => mockAuthNotifier),
+          uspClientProvider.overrideWith((ref) {
+            builds++;
+            return null;
+          }),
+        ],
+      );
+      addTearDown(container.dispose);
+
+      container.read(uspClientProvider);
+      expect(builds, 1);
+
+      await container.read(remoteAssistanceProvider.notifier).deactivate();
+      container.read(uspClientProvider);
+
+      // Rebuilt after the swap — consumers must not keep the pre-swap client.
+      expect(builds, 2);
     });
 
     // activate() requires web platform and WASM — test the platform check
@@ -284,7 +309,7 @@ void main() {
       expect(identical(state1, state2), isTrue);
     });
 
-    test('notifier updates trigger state changes', () {
+    test('notifier updates trigger state changes', () async {
       final container = ProviderContainer();
       addTearDown(container.dispose);
 
@@ -296,7 +321,7 @@ void main() {
       );
 
       final notifier = container.read(remoteAssistanceProvider.notifier);
-      notifier.deactivate();
+      await notifier.deactivate();
 
       // Initial state + state after deactivate
       expect(states.length, greaterThanOrEqualTo(1));
--- a/test/di_test.dart
+++ b/test/di_test.dart
@@ -1,0 +1,35 @@
+import 'package:flutter_test/flutter_test.dart';
+import 'package:privacy_gui/constants/build_config.dart';
+import 'package:privacy_gui/di.dart';
+
+void main() {
+  group('canUseAppOriginUspClient', () {
+    final original = BuildConfig.forceCommandType;
+
+    tearDown(() {
+      BuildConfig.forceCommandType = original;
+    });
+
+    test('allows the boot client in a Local build', () {
+      BuildConfig.forceCommandType = ForceCommand.local;
+
+      expect(canUseAppOriginUspClient(), isTrue);
+    });
+
+    test('allows the boot client when no mode is forced', () {
+      BuildConfig.forceCommandType = ForceCommand.none;
+
+      expect(canUseAppOriginUspClient(), isTrue);
+    });
+
+    test('rejects the boot client in a Remote Assistance build', () {
+      // The app is served from static hosting, not by the router — a client
+      // built on Uri.base.origin would answer no USP request, and whoever read
+      // uspClientProvider first would cache it. Only activate() may build the
+      // client in this mode.
+      BuildConfig.forceCommandType = ForceCommand.remote;
+
+      expect(canUseAppOriginUspClient(), isFalse);
+    });
+  });
+}


AustinChangLinksys (commented):
🤖 **Automated Review — Round 1** · `71f76e6..5a33a17` (full)

**Verdict: 💬 Self-review (comment only)** — Austin's own PR; automated flow cannot self-approve. Review is for reference / blind-spot check only. All issues below are advisory.

---

#### Standards

|---|---|---|---|
| ⚠️ | 🟢High | `remote_assistance_provider.dart:38` | `guardianOrigin` assert stripped in release mode — scheme injection silently produces broken URL |
| ⚠️ | 🟢High | `remote_assistance_provider.dart:143-149` | `activate()` exception after `ref.invalidate` but before `state =` leaves GetIt swapped with state still `isActive: false` |
| ⚠️ | 🟢High | `test/core/usp/providers/remote_assistance_provider_test.dart` | `activate()` `ref.invalidate` path has no symmetric test — only `deactivate` path is covered |
| 💡 | 🟡Med | `remote_assistance_provider.dart:35-40` | Advisory: `withLock` + `ref.invalidate` rebuild sequencing — see details; no bug, documentation gap |
| 💡 | 🟡Med | `lib/di.dart:67` | [Middleman] `canUseAppOriginUspClient()` is a one-line proxy for `BuildConfig.isRemote()` — acceptable as `@visibleForTesting` seam |
| 💡 | 🟢High | `lib/core/usp/services/usp_bridge_client_base.dart:28` | Stub accepts `baseUrl` without an API-compatibility comment; minor reader confusion risk |
| 💡 | 🟡Med | `remote_assistance_provider.dart:123-138` | `activate()` comment is 8 lines; `deactivate()` is 2 lines for the same invariant — asymmetric, divergence risk |
| 💡 | 🟡Med | `test/core/usp/providers/remote_assistance_provider_test.dart` | `deactivate()` test doesn't verify GetIt unregistration; only Riverpod rebuild-count is checked |

---

#### Spec

|---|---|---|---|
| ✅ | 🟢High | AC1 | `git merge-base --is-ancestor origin/dev-2.6.1 HEAD` — verified pass |
| ✅ | 🟢High | AC2 `pubspec.yaml` | `version: 2.7.0+100000` preserved; pubspec diff is empty |
| ✅ | 🟡Med | AC3 `flutter analyze` | 0 errors/warnings per commit attestation; CI artifact not yet observed |
| ✅ | 🟡Med | AC4 `run_tests.sh` | 7118 passed per commit attestation; CI artifact not yet observed |
| ✅ | 🟢High | AC5 lib/ scope | Exactly the 6 allowed lib/ files modified — no scope creep |
| 💡 | 🟡Med | CI | `flutter analyze` and `run_tests.sh` verified only via commit-message attestation — suggest linking CI artifact before merge |

---

<details>

**W-1 · `guardianOrigin` assert not release-safe** 🟢High *(flagged by both Reviewer A and B)*

`lib/core/usp/providers/remote_assistance_provider.dart:38-40`

```dart
String get guardianOrigin {
  // assert only runs in debug/profile — stripped in --release
  assert(!guardianBaseUrl.contains('://'),
      'guardianBaseUrl must be a bare host, without a scheme');
  return 'https://$guardianBaseUrl';
}
```

In a `--release` build, if `guardianBaseUrl` arrives as `https://qa.guardian.tools` (e.g. if `kCloudBase` values are updated to full URLs), `guardianOrigin` silently returns `https://https://qa.guardian.tools`. All USP requests, SSE subscriptions, and session REST calls for the entire RA session would be directed to a malformed host — request failures would appear as network timeouts, not obvious application errors. The code comment explains the `const` constructor constraint. The current `cloud_const.dart` entries are bare hosts so the bug is latent, not live.

**Fix:** Replace the `assert` with a release-safe guard:
```dart
String get guardianOrigin {
  if (guardianBaseUrl.contains('://')) {
    throw ArgumentError.value(
      guardianBaseUrl, 'guardianBaseUrl',
      'must be a bare host without a scheme',
    );
  }
  return 'https://$guardianBaseUrl';
}
```
Also add a corresponding negative test case (scheme-prefixed input → throws `ArgumentError`).

---

**W-2 · `activate()` exception path leaves GetIt/state inconsistent** 🟢High *(Reviewer A)*

`lib/core/usp/providers/remote_assistance_provider.dart:139-149`

```dart
    // inside withLock — GetIt already swapped, invalidate already fired
    ref.invalidate(uspClientProvider);  // line 139
  });
  // ← withLock returns here; GetIt = raClient, Riverpod consumers rebuilding

  ref.read(authProvider.notifier).setLoginType(LoginType.remote);  // line 143
  state = RemoteAssistanceState(isActive: true, config: config);   // line 145 — may not be reached
```

If `setLoginType()` throws (runtime exception in `AuthNotifier`, mock misconfiguration, or future code change), `state` remains `const RemoteAssistanceState()` (`isActive: false`, `config: null`). At that point:
- GetIt holds the new `raClient` (Guardian-proxied)
- `uspClientProvider` watchers have already rebuilt against the new client
- `state.isActive` is still `false` and `state.config` is still `null`

Consumers that gate on `remoteAssistanceProvider.state.isActive` (including `sse_providers.dart:31`) will see `isActive: false` while network traffic is already being routed through the Guardian client. Subsequent `deactivate()` may or may not clean up correctly depending on whether `isRegistered<UspClient>()` passes.

**Fix:** Wrap the post-lock section in a try-catch with a rollback call to `deactivate()`, or move `state =` into the lock body after the GetIt swap.

---

**W-3 · `activate()` `ref.invalidate` path not covered by test** 🟢High *(Reviewer B)*

`test/core/usp/providers/remote_assistance_provider_test.dart`

The `deactivate()` path has:
```dart
test('deactivate invalidates uspClientProvider', () async { ... expect(builds, 2); });
```

There is no symmetric `'activate invalidates uspClientProvider'` test. The `activate()` invalidation (line 139) is one of the three core fixes in this PR. It cannot be exercised on the Dart VM (requires WASM `UspClientBuilderJS`), but this should be explicitly documented: add a `skip`-tagged test or a `// Platform: Web-only — covered by integration test` comment so future readers know the coverage gap is intentional, not an oversight.

</details>

<details>
<summary>✅ What Looks Good</summary>

- **Three core bugs fixed correctly:** (a) `canUseAppOriginUspClient()` prevents boot-time `UspClient(Uri.base.origin)` in Remote builds; (b) `ref.invalidate(uspClientProvider)` placed inside the `withLock` critical section in both `activate()` and `deactivate()`; (c) `baseUrl` override on `UspBridgeClient` ensures SSE/subscriptions use Guardian host.
- **Merge conflict resolution sound:** Only `pubspec.yaml` version line conflicted; resolved correctly to `2.7.0+100000`. AC5 verified — exactly the 6 allowed `lib/` files modified, no unintended changes.
- **`di_test.dart` covers all three `ForceCommand` branches:** `local`, `none`, `remote` — including `tearDown` that restores `BuildConfig.forceCommandType` (correct isolation).
- **`usp_bridge_client_base.dart` stub in sync:** Signature updated to accept `baseUrl:` parameter, matching the web implementation's constructor shape.
- **`deactivate()` tests upgraded:** Both pre-existing tests correctly `await deactivate()` (fixed async race) and a new build-count test covers the `ref.invalidate` deactivate path.
- **Architecture compliance:** Three-layer dependency direction intact, `remoteAssistanceProvider` non-autoDispose, both `activate()`/`deactivate()` mutations inside `withLock`, no codegen boundary violations.
- **Spec compliance:** All 5 Issue #1330 acceptance criteria met (verified against local repo state at `5a33a17d`).

</details>

---

PeterJhongLinksys (approved):
🤖 **Automated Review — Round 1** · `71f76e6..5a33a17` (full)

**Verdict: ✅ APPROVE** — The diagnosis is the strongest part of this PR: `uspClientProvider` caches whatever GetIt held at its *first* read (`auth_provider.dart:58`, at boot), so not creating the wrong-origin client is the right fix rather than patching consumers. No Critical found by either reviewer. Ten Warnings, none blocking, but **W1 deserves an explicit answer in this thread before merge** — the gate changes `uspClientProvider`'s null from non-Web-only to reachable-in-Remote, and ~15 production `ref.read(...)!` sites were not audited.

|---|---|---|---|
| ⚠️ | 🟢 | `lib/di.dart:48` | Remote builds now leave the GetIt slot empty, so `uspClientProvider` is null-reachable while ~15 production consumers still `ref.read(uspClientProvider)!` |
| ⚠️ | 🟡 | `lib/di.dart:78` | The gate keys off the *compile-time* build flavour, but `router_provider.dart:196-203` lets any build enter Remote Assistance via `?session=` |
| ⚠️ | 🟢 | `remote_assistance_provider.dart:35-41` | `assert` is the sole guard on the "bare host, no scheme" contract and is stripped in release; the value now feeds four URL builders |
| ⚠️ | 🟢 | `remote_assistance_provider.dart:24-34` | [Duplicated Code] Third independent `'https://' + kCloudBase` builder; `cloud_host_resolver.dart:38` already owns this concern |
| ⚠️ | 🟢 | `usp_bridge_client_web.dart:71` | [Speculative Generality] `_overrideBaseUrl ?? _usp.baseUrl` — at the only call site both are the same string, making the override redundant with the `di.dart` half of this PR |
| ⚠️ | 🟡 | `usp_bridge_client_base.dart:28` | Non-Web stub accepts and discards `baseUrl`; stub/web signature drift is unguarded and the web-side behaviour change is untested |
| ⚠️ | 🟡 | `remote_assistance_provider.dart:158-168` | `deactivate()` disposes the client, but `ref.read` consumers keep the disposed instance, and Remote builds have no documented path back to a client |
| ⚠️ | 🟡 | `remote_assistance_provider.dart:139` | `ref.invalidate` fires before `state` is published at `:145`, so `uspBridgeClientProvider` could momentarily see client-present / `config`-null |
| ⚠️ | 🟡 | `lib/di.dart:60-78` | `canUseAppOriginUspClient()` is a production branch predicate misannotated `@visibleForTesting`, and parked in the DI root rather than beside its `GlobalConfig` siblings |
| ⚠️ | 🟢 | `remote_assistance_provider.dart:125-139` | [Comments] ~40 of 185 changed lines are prose carrying cross-file temporal invariants; the "fails loudly" claim is already inaccurate |
| 💡 | 🟢 | `remote_assistance_provider_test.dart:203-224` | Only `deactivate()`'s invalidation is tested; the `activate()` one — the actual fix — and the web `baseUrl` override are uncovered |
| 💡 | 🟢 | `remote_assistance_provider.dart:128-130` | "same critical section … can never drift apart" overstates `ref.invalidate`; only the dirty-marking is under the lock |
| 💡 | 🟡 | `usp_bridge_client_web.dart:41` | `_overrideBaseUrl` does not say what it overrides; `restBaseUrl` would convey the REST/SSE-vs-transport split |
| 💡 | 🟡 | `test/di_test.dart:7-13` | Test mutates the mutable static `BuildConfig.forceCommandType`; constructor injection (cf. `cloud_host_resolver.dart:31-33`) would avoid touching a global |
| 💡 | 🟡 | `usp_client_provider.dart:12-20` | DartDoc explains this `lib/core/usp/` provider's nullability by pointing *up* at `di.dart` — a readability inversion |
| 💡 | ⚪ | `remote_assistance_provider.dart:44-45` | `sessionId`, which originates from a URL query parameter, is interpolated into request paths without `Uri.encodeComponent()` |
| 💡 | ⚪ | `remote_assistance_provider.dart:40` | The `'https://'` literal is now typed out in three cloud-host builders; a single `cloudOriginFor(host)` helper makes the http/https policy greppable |
| 💡 | ⚪ | `remote_assistance_provider.dart:35` | `guardianOrigin` recomputes and re-asserts on every access (4× per activation); `late final` or factory-time normalisation evaluates the contract once |

<sub>Confidence: 🟢 High = mechanism proven from head code · 🟡 Med = plausible, one link unproven · ⚪ Low = defensive/stylistic. Low-confidence items are listed, not dropped.</sub>
<sub>No `[X]`/`[XX]` markers appear because there are no Critical findings. Both reviewers independently reached 0 Critical after explicitly evaluating and downgrading candidates (see their reasoning folded into W1/W3/W4 below).</sub>

<details>
<summary><b>⚠️ Warning details (10)</b> — evidence chains</summary>

### W1 🟢 — Removing the boot client makes `uspClientProvider` null-reachable in Remote builds; ~15 `!` consumers unaudited
`lib/di.dart:45-58` → `lib/core/usp/providers/usp_client_provider.dart:24-30`

```dart
// di.dart
if (kIsWeb && !getIt.isRegistered<UspClient>()) {
  if (canUseAppOriginUspClient()) { /* ... registerSingleton ... */ }
  else { logger.d('[DI]: UspClient boot registration skipped — Remote build, ...'); }
}

// usp_client_provider.dart
final uspClientProvider = Provider<UspClient?>((ref) {
  if (!kIsWeb) return null;
  if (!getIt.isRegistered<UspClient>()) return null;
  ...
});
```

Before this PR `uspClientProvider` could only be null on non-Web. Now it is null for the whole window between boot and `activate()` in a Remote build. Roughly fifteen production consumers assume non-null, e.g. `pnp_service.dart:27`, `usp_firmware_update_service.dart:12`, `usp_admin_service.dart:12`, `usp_firewall_service.dart:11`, `usp_dhcp_service.dart:11`, `usp_system_log_service.dart:9`, `instant_privacy_service.dart:13`, `usp_static_routing_service.dart:13`, `usp_ipv6_port_service_service.dart:12`, `instant_safety_service.dart:11`.

Evidence that this is not hypothetical: `instant_safety_provider.dart:66-69` already carries a hand-written comment describing exactly this failure ("the `uspClientProvider` null assertion in the service provider"). One team hit it and defended locally; the other nine bang-sites have no equivalent guard.

Not Critical because we could not prove a concrete route that reads one of those providers before `activate()` completes — the router sends Remote sessions to `/remoteAssistance` first (`router_provider.dart:136`) — and the trade-off is deliberately documented at `remote_assistance_provider.dart:133-138`. We agree with the direction (fail loud beats wrong host). The caveat: per the `instant_safety_provider` comment, a null assertion inside a Riverpod `build()` manifests as a **hang**, not a diagnosable error, so the "fails loudly" rationale does not hold as written.

**Suggested fix:** funnel the dereference through one shared helper that throws a typed `ServiceError` (e.g. `UspClientUnavailableError`) instead of ten duplicated `!`; or add a regression test asserting no consumer provider is constructed before `activate()`.

### W2 🟡 — The gate is compile-time-only, but Remote Assistance can be entered at runtime in a non-Remote build
`lib/di.dart:78`

```dart
@visibleForTesting
bool canUseAppOriginUspClient() => !BuildConfig.isRemote();
```

`BuildConfig.isRemote()` (`build_config.dart:143-145`) tests `forceCommandType`, which on Web is fixed at compile time from `String.fromEnvironment('force')` (`build_config.dart:44`) and never reassigned on Web — the only runtime write is the `!kIsWeb` bundle-ID branch at `build_config.dart:131`.

But the RA flow has **two** entry points (`router_provider.dart:196-209`):

```dart
final raSession = state.uri.queryParameters['session'];
if (raSession != null && raSession.isNotEmpty) {
  return '${RoutePath.remoteAssistanceConfirm}?session=$raSession&token=$raToken';
}
// Check for Remote build mode (force=remote)
if (BuildConfig.isRemote()) { ... }
```

The `?session=` path is reachable regardless of build flavour. So in a `force=none`/`force=local` build a user still reaches `remote_assistance_confirm_view.dart:207` → `activate(config)` — but `di.dart` **did** register an app-origin boot client (the gate returned true), `authProvider.init()` (`auth_provider.dart:58`) cached it, and the new `ref.invalidate` at `remote_assistance_provider.dart:141` only rebuilds `ref.watch` consumers. Of ~40 consumers, only three use `ref.watch` (`sse_providers.dart:26`, `dashboard_orchestrator.dart:124`, `unified_diagnostics_service.dart:26`); the rest use `ref.read(...)!`.

Warning rather than Critical because whether any of those non-`autoDispose` service providers has actually been *instantiated* before `activate()` depends on route/widget build order we did not trace to a concrete screen. What is proven is that the gate is compile-time-only, which directly contradicts the invariant the new comments assert.

**Suggested fix:** make the gate runtime-aware — also skip the boot client when the launch URL carries `session`, or gate on `GlobalConfig.remote.isActive` computed after URL parsing, or have `activate()` proactively invalidate the derived service providers.

### W3 🟢 — `assert` is the only guard on the scheme contract, and it vanishes in release
`lib/core/usp/providers/remote_assistance_provider.dart:35-41`

```dart
String get guardianOrigin {
  // Kept as an assert rather than a constructor assert: the constructor is
  // const, and const asserts only accept potentially-constant expressions.
  assert(!guardianBaseUrl.contains('://'),
      'guardianBaseUrl must be a bare host, without a scheme');
  return 'https://$guardianBaseUrl';
}
```

Dart strips `assert` in release/profile. If `guardianBaseUrl` ever arrives scheme-qualified, the getter silently yields `https://https://qa.guardian.tools`, and that value now flows into **four** places, three new in this PR: `UspClientBuilderJS(config.guardianOrigin)` (`:101`), `UspClient.fromBuilder(..., baseUrl:)` (`:111`), and `baseUrl: config.guardianOrigin` on the bridge client (`sse_providers.dart:42` → `usp_bridge_client_web.dart:71` → nine `Uri.parse('$_baseUrl...')` sites at web lines 134/202/230/409/428/443/476/493). Failure mode: malformed-URI throw or a request to a nonsense host on every REST/SSE call, surfacing far from the cause.

Not reachable today — the only producer is `remote_assistance_confirm_view.dart:201` (`cloudEnvironmentConfig[kCloudBase]`) and `cloud_const.dart:14/19/24` are all bare hosts. That is precisely the situation in which an assert-only guard rots unnoticed. Note also the reasoning in the comment is sound (a `const` constructor cannot assert on this) but the conclusion — move it to a getter — yields a check that only fires in debug.

**Suggested fix:** normalise instead of assert (strip a leading scheme, or `startsWith('http') ? guardianBaseUrl : 'https://$guardianBaseUrl'`), or add a `RemoteAssistanceConfig.fromHost()` factory that validates once and throws a real `ArgumentError`.

### W4 🟢 — [Duplicated Code / Shotgun Surgery] Three independent builders of the same Guardian origin
`remote_assistance_provider.dart:24-34,40` · `cloud_host_resolver.dart:38` · `linksys_http_client.dart:94`

```dart
// remote_assistance_provider.dart:40
return 'https://$guardianBaseUrl';                                   // value injected from kCloudBase
// cloud_host_resolver.dart:38
String get _cloudBase => 'https://${cloudEnvironmentConfig[kCloudBase]}';
// linksys_http_client.dart:94
_getHost?.call() ?? 'https://${cloudEnvironmentConfig[kCloudBase]}';
```

The new DartDoc at `remote_assistance_provider.dart:29-34` is explicit: *"The session REST API ends up on the same host by construction — both this config … and `GuardianApiClient._buildUrl` read that same entry — but it builds its URL independently."* A comment stating "these must agree but are computed separately" accurately describes the hazard without resolving it. A future change (a port, a path prefix, an http fallback for a local Guardian emulator) must be made in three files — and `cloud_host_resolver.dart` already demonstrates the divergence risk, since it has a `kProxyPrefix` branch for local builds that the RA path does not.

**Suggested fix:** `CloudHostResolver` already owns "turn `kCloudBase` into a URL base". Have `guardianOrigin` delegate to it. That also gives W3's validation a single choke point.

### W5 🟢 — [Speculative Generality / Divergent Change] `baseUrl` is a second, silently-preferred source of truth for the host
`usp_bridge_client_web.dart:41,57,71` · `usp_bridge_client_base.dart:28`

```dart
String get _baseUrl => _overrideBaseUrl ?? _usp.baseUrl;
```

`UspClient` already carries an authoritative `baseUrl` (`usp_client.dart:89-90,124`), and in this PR `activate()` constructs the RA client with exactly the value also passed as the override:

```dart
final raClient = UspClient.fromBuilder(jsClient, baseUrl: config.guardianOrigin);  // :112
baseUrl: config.guardianOrigin,                                                    // sse_providers.dart:80
```

So after this PR the override and `_usp.baseUrl` are **the same string** at the only site that sets it. The DartDoc justifies it defensively ("it must not be derived from `UspClient.baseUrl`, which may still point at the origin the web app was served from") — but the `di.dart` change in this same PR is what guarantees that stale client can no longer exist in a Remote build. The two halves make each other redundant. Costs: the non-Web stub gains an API surface it cannot use (`usp_bridge_client_base.dart:28`), and a reader cannot tell which host wins without reading the null-coalesce.

**Suggested fix:** pick one owner. Either drop the parameter and rely on `_usp.baseUrl`, or keep the defence-in-depth as `assert(baseUrl == null || baseUrl == usp.baseUrl)` so a real divergence is caught rather than silently papered over.

### W6 🟡 — Stub/web signature drift is unguarded, and the actual behavioural change is untested
`lib/core/usp/services/usp_bridge_client_base.dart:28`

```dart
UspBridgeClient(UspClient usp, {BridgeEndpoints? endpoints, String? baseUrl, ...});
```

The stub accepts and discards `baseUrl` (empty body; every method throws `UnsupportedError`) — correct for a conditional-export stub, but the two declarations stay in sync only by hand and the analyzer cannot catch drift because the files are never compiled together. This is the second parameter added this way. Concrete risk: a parameter added to the web class only still compiles and passes unit tests (which resolve to the stub) while the web build diverges. `usp_bridge_client_web.dart:71` is the actual behavioural change in this PR and has no test.

**Suggested fix:** a shared abstract interface or `typedef` both implementations must satisfy, plus a `@TestOn('browser')` test asserting the override wins over `_usp.baseUrl`.

### W7 🟡 — `deactivate()` disposes the client while `ref.read` consumers keep holding it, and Remote builds have no path back
`remote_assistance_provider.dart:153-172`

```dart
await ref.read(uspMutationLockProvider).withLock(() async {
  if (getIt.isRegistered<UspClient>()) {
    final client = getIt<UspClient>();
    getIt.unregister<UspClient>();
    client.dispose();
  }
  ref.invalidate(uspClientProvider);
});
```

The added comment says watchers "rebuild against an empty GetIt instead of holding a disposed client" — true for `ref.watch`, but the `ref.read(uspClientProvider)!` majority from W1 retains the instance `dispose()` was just called on. Any subsequent call hits a disposed transport.

Separately, recovery is now asymmetric: a Local build re-registers the boot client on reload via `dependencySetup()`, whereas a Remote build deliberately does not — so `deactivate()` moves a Remote build into a permanently clientless state with no code path back except a fresh RA session. The new `ref.invalidate` correctly propagates "empty", which makes that dead-end reachable by watchers rather than latent.

Mitigating: `grep` found **zero** callers of `deactivate()` in `lib/` today, so this is currently only reachable from tests — but it also means the path is unexercised and becomes a live use-after-dispose the moment an "end session" button is wired up. The new test (`remote_assistance_provider_test.dart:203-224`) asserts the invalidation fires but not what consumers do afterwards.

**Suggested fix:** drive USP access exclusively through `ref.watch`, or scope `uspClientProvider` `autoDispose` to the session, or have `deactivate()` also invalidate the derived service providers. State the intended post-deactivate contract (hard reload / terminal screen) in code rather than prose; `assert(GlobalConfig.remote.isActive)` at the top would document the Local-build assumption.

### W8 🟡 — `activate()` invalidates `uspClientProvider` before publishing the state its dependent selects on
`remote_assistance_provider.dart:139` vs `:145` · `sse_providers.dart:112-130`

```dart
await ref.read(uspMutationLockProvider).withLock(() async {
  getIt.registerSingleton<UspClient>(raClient);
  ref.invalidate(uspClientProvider);      // line 139
});
ref.read(authProvider.notifier).setLoginType(LoginType.remote);
state = RemoteAssistanceState(isActive: true, config: config);   // line 145
```

`uspBridgeClientProvider` watches **both** inputs and returns null if either is missing:

```dart
final usp = ref.watch(uspClientProvider);
if (usp == null) return null;
if (GlobalConfig.remote.isActive) {
  final config = ref.watch(remoteAssistanceProvider.select((s) => s.config));
  if (config == null) return null;
  bridge = UspBridgeClient(usp, ..., baseUrl: config.guardianOrigin, ...);
}
```

Between `:139` and `:145` the inputs are inconsistent (client present, `config` still null) → the bridge provider would yield null and `sseManagerProvider` (`sse_providers.dart:154-158`) short-circuits with it. In practice Riverpod's invalidation is deferred and there is no `await` between lock exit and `:145`, so the rebuild should observe both changes together — we could not establish that no listener (e.g. a `fireImmediately` subscription, or the `setLoginType` call at `:143` triggering a synchronous chain) forces an earlier flush. Hence Med, not Critical.

**Suggested fix:** assign `state` before the invalidation, or move the invalidation after `:145`. The invariant you want — "config is visible no later than the new client" — costs nothing to express in statement order and removes the need to reason about flush timing at all.

### W9 🟡 — `canUseAppOriginUspClient` is misannotated and misplaced
`lib/di.dart:60-78`

```dart
@visibleForTesting
bool canUseAppOriginUspClient() => !BuildConfig.isRemote();
```

`@visibleForTesting` declares "this exists only so tests can reach it", but this is a production branch predicate called at `di.dart:48`; `test/di_test.dart` imports it from outside the library, which is exactly what the annotation is meant to discourage. The analyzer's intent is inverted here — if the goal is testability, a plain public predicate with the DartDoc it already has is the honest shape.

Placement: `di.dart` is the get_it wiring root, whereas "may this build reach USP on its own origin?" is build-mode policy about the USP transport, whose siblings live in `global_config.dart:126-151` (`RemoteConfig.isActive`, `allowConfigChanges`, `showAdvancedSettings`). As written, `di.dart` now imports `constants/build_config.dart` and owns a piece of USP policy, while `usp_client_provider.dart` reaches back up to `di.dart` in its DartDoc to explain its own nullability.

**Suggested fix:** move it to `GlobalConfig.remote` (e.g. `bool get canUseAppOriginUspClient => !isActive;`), drop `@visibleForTesting`, retarget `test/di_test.dart`.

### W10 🟢 — [Comments as deodorant] Load-bearing invariants live only in prose, and one is already inaccurate
`remote_assistance_provider.dart:24-41,125-139,166-169` · `di.dart:60-76` · `usp_client_provider.dart:12-20`

Counting the diff: `remote_assistance_provider.dart` adds 43 lines of which ~24 are comment; `di.dart` adds 29 of which ~17 are comment; `usp_client_provider.dart` adds 9, all comment. The largest single block (`remote_assistance_provider.dart:125-138`) is 14 lines of prose explaining a one-line `ref.invalidate(uspClientProvider);`.

The prose is high quality and we would rather have it than not — the issue is what it carries. These claims are asserted only in comments:

- "`uspClientProvider` caches whatever GetIt held when it was FIRST read, and `authProvider.init()` reads it during app boot" — verifiable at `auth_provider.dart:58`, but nothing fails if someone adds an earlier reader.
- "in a Remote build the only client that can ever exist is this one" — spread across `di.dart:48`, `usp_client_provider.dart:26` and `activate()`, with no single enforcement point. W2 shows it is conditional on build flavour.
- "the worst a too-early read can capture is null — which fails loudly" — per W1 the failure is a hang, so this sentence is **already inaccurate**.

The three files each explain the other two, so a change in any one leaves two lying comments.

**Suggested fix:** promote the load-bearing claims into tests — one reading `uspClientProvider` before `activate()` in a simulated Remote container asserting `null`, one asserting `activate()` makes a `ref.watch` consumer see the new client — then shorten the prose to a pointer at those tests. Correct the "fails loudly" sentence.

</details>

<details>

- **The root cause is attacked at the right layer.** That `uspClientProvider` caches whatever GetIt held at its first read, that `authProvider.init()` performs that read at boot (`auth_provider.dart:58`), and that consumers therefore keep a wrong-origin client permanently, is a genuinely subtle lifecycle bug. "Don't create the wrong object in the first place" beats patching consumers.
- **The core fix is correct and backwards compatible.** Deriving the bridge REST/SSE host from an explicit override rather than `UspClient.baseUrl` (`usp_bridge_client_web.dart:71`) is the right decoupling: Remote now passes the Guardian origin explicitly across nine call sites, while Local keeps its previous behaviour via `?? _usp.baseUrl`.
- **`guardianOrigin` removes real triplicated interpolation.** Three `'https://${config.guardianBaseUrl}'` sites within `remote_assistance_provider.dart` (logging, `UspClientBuilderJS`, `UspClient.fromBuilder`) collapse to one getter — textbook Extract Method. W4 is about finishing that across files, not about this change.
- **Art. IV Rule 3 respected.** Both `activate()` (`:114`) and `deactivate()` (`:158`) keep the GetIt swap inside `uspMutationLockProvider.withLock(...)`, and the new `ref.invalidate` calls were deliberately placed *inside* the lock rather than bolted on after — correct critical-section reasoning, not convenient placement.
- **Art. IV Rule 1 and Art. V §5.4 clean.** No `.autoDispose` added to `uspClientProvider` or `uspBridgeClientProvider`; `PreservableContract` untouched; no `ref.watch` introduced into an L2 `performFetch()`; no generated-model imports in any touched provider or view.
- **Test bug genuinely fixed, not churn.** `deactivate()` became `Future<void>` and three tests were correctly converted to `async`/`await` (`remote_assistance_provider_test.dart:191,196,324`). Since state now resets only after the lock completes, the old tests were asserting on pre-await state — a real latent flake.
- **The new invalidation test is well shaped.** It uses a build-counter probe rather than asserting on identity, which is a robust way to prove invalidation actually happened. `test/di_test.dart` follows repo conventions (descriptive names, no `TC###`, proper `tearDown` restoration, negative case explains *why*).
- **No secrets, no token logging.** `logger.d` logs only origin and endpoint path (`remote_assistance_provider.dart:97-99`) and `toString()` at `:49` deliberately omits `temporaryAccessToken` — good discipline for a class holding a bearer token. Nothing here persists the token or touches `flutter_secure_storage`.
- **No USP error-handling defects in this diff.** No `on Exception catch` on a USP path, no hand-rolled USP error-string parsing, no new `result`/`_Error` handling. The one `catch (e)` this PR touches (`di.dart:52`) is already the correct broad form.
- **Comment quality is unusually high.** Even where we flag over-reliance on prose (W10), the comments explain *why*, name the specific mechanism, and pre-empt the obvious "why not a constructor assert?" objection (`remote_assistance_provider.dart:36-37`).

</details>

AustinChangLinksys:
## Review response

Verified every finding from both reviews against the merge head (`5a33a17d`). Both CI checks are now
green (`🛡️ Quality & Logic (PR)` runs `flutter analyze` + `run_tests.sh`, 9m46s), which closes the
"attestation only, no CI artifact" note.

**Nothing is being fixed in this PR**, and the reason is the PR's nature rather than the findings'
merit — see Disposition at the bottom.

### Verdict per finding

| # | Finding | Real? | Notes |
|---|---|---|---|
| 1 | `guardianOrigin` assert stripped in release | **yes** | `build_web.sh:47` is `flutter build web`, i.e. release — the assert does not run. Latent only: the sole producer is `cloudEnvironmentConfig[kCloudBase]` and all three env values (`guardian.tools`, `qa.…`, `dev.…`) are const bare hosts. Note the 2.7.0 base had no guard at all, so this is a net improvement. |
| 2 | `activate()` post-lock throw leaves GetIt swapped with `isActive: false` | speculative | `setLoginType` (`auth_provider.dart:174-178`) is a `copyWith` plus a `logger.d` — no throw surface. Its caller `_doConnect` is inside the confirm view's try/catch. |
| 3 | `activate()`'s invalidate has no symmetric test | **yes** | Only `deactivate` has the build-count probe (`remote_assistance_provider_test.dart:204-224`). `UspClientBuilderJS` needs WASM, so the `activate` path is unreachable on the Dart VM. Worth an explicit marker so the gap reads as intentional. |
| 4 | Boot gate makes `uspClientProvider` null-reachable; `ref.read(…)!` consumers unaudited | **yes** | Measured on this head: 16 `ref.read(uspClientProvider)!`, 41 `ref.read`, 11 `ref.watch`. Already documented as a known consequence in #1330 and in this PR's description; owned by #1322. |
| 5 | Gate is compile-time, but `?session=` lets any build enter RA | **yes — most valuable finding of the round** | Expanded below; not covered by #1322 or #1323. |
| 6 | Three independent `'https://' + kCloudBase` builders | **yes** | `remote_assistance_provider.dart:40`, `cloud_host_resolver.dart:38`, `linksys_http_client.dart:94`. `guardianOrigin` collapsed three intra-file interpolations into one; unifying across files is a separate refactor. |
| 7 | `baseUrl` override equals `_usp.baseUrl` at the only call site → redundant | **yes, today** | `activate()` passes `config.guardianOrigin` to both. The decoupling is deliberate. **Not taking the suggested `assert(baseUrl == usp.baseUrl)`** — that would lock in the coupling this change exists to break. |
| 8 | Stub/web signature drift is unguarded | **yes** | Conditional export, never compiled together. Pre-existing pattern (`_overrideToken` arrived the same way). |
| 9 | `deactivate()` disposes while `ref.read` holders keep the instance | **yes** | `grep` confirms **zero** `deactivate()` callers in `lib/` — currently test-only. Becomes live when an end-session button is wired up. Owned by #1322. |
| 10 | `:139` invalidate precedes `:145` state publish | real but self-healing | Correcting the severity: `uspBridgeClientProvider` registers the `config` dependency *during* the null-config rebuild (`sse_providers.dart:34` runs after the `usp == null` early return). So `:145` necessarily triggers a second rebuild and the provider converges: invalidate → rebuild(null) → `state =` → rebuild(real). Not a stuck state, just one extra rebuild. Reordering the statements is free tidiness, not a bug fix. |
| 11 | `@visibleForTesting` misannotated / misplaced | style | Called from within the same library, so the analyzer stays quiet. Arguable either way. |
| 12 | "fails loudly" comment is already inaccurate | **yes** | `instant_safety_provider.dart:66-69` carries a hand-written comment stating that a throw inside a provider body escapes into `build()`'s unawaited microtask and manifests as a **hang**. The prose should be corrected. |
| 13 | analyze / run_tests verified only by commit attestation | resolved | Both CI checks pass on this head. |

### Expanding #5 — the finding worth its own ticket

The `?session=` branch (`router_provider.dart:196-202`) does not consult the build flavour, so a
`force=none` / `force=local` build can reach `activate()`. In that state:

- `GlobalConfig.remote.isActive` is `false`, so `sse_providers.dart:31` takes the **local** branch:
  `BridgeEndpoints.local`, no `baseUrl`, no `authToken`, `AuthBehavior.local`
- but GetIt holds the Guardian-proxied `raClient`

Result: USP requests go to Guardian while REST/SSE go to the app's own origin using on-router
endpoint paths. That hybrid is outside the scope of both #1322 (client lifecycle) and #1323 (RA
recovery flow) — it is new, and it is filed separately.

### Disposition

This PR is a promotion merge of `9cf72b52` / `64306778` / `6d1aef0a` — three commits already
reviewed and merged on `dev-2.6.1`. It contains no new code. Every finding above is about the
*content* of those commits.

Fixing them here would fork the `dev-2.7.0` copy from the `dev-2.6.1` original. The 2.6.1 milestone
is still open (1 open / 3 closed), so the next 2.6.1 hotfix promotion would then re-collide on
exactly these files — the collision this merge exists to eliminate.

So:

| Finding | Destination |
|---|---|
| 4, 9 | #1322 (already owned) — adding a reachability note there |
| 5 (hybrid local-bridge / Guardian-client mode) | new follow-up issue |
| 1 (release-safe guard + negative test), 12 (correct the "fails loudly" prose), 3 (mark the Web-only coverage gap) | same follow-up issue; landing on `dev-2.6.1` so both lines get it without re-diverging |
| 6, 7, 8, 11 | not taking. #7's suggested assert would work against the change's purpose. |

Merging as-is.

PR #1347, opened 2026-08-21
fix(static-routing): validate on focus-loss so the dialog keeps keystrokes (#1332)

### LINKED ISSUE #1332: Static route dialog drops keystrokes: onChanged validation rebuilds the field mid-edit
## Symptom

Typing into the **Add static route** dialog loses characters. The remaining text is a truncated **prefix** of what was typed:

| Field | Typed | Landed |
|---|---|---|
| name | `e2e-route` | `e` |
| Gateway IP octet 1 | `192` | `1` (also seen: `19`) |
| Destination IP | `10.20.30.0` | `10.20.30.0` ✓ |
| Subnet mask | `255.255.255.0` | `255.255.255.0` ✓ |

Because Gateway ends up as `1.168.1.1`, `_isFormValid` (`static_route_dialog.dart:210`) renders **Add** disabled, so the dialog cannot be submitted at all.

This is reachable by a real user: type an IP into this dialog at normal speed and characters go missing. It is not only a test problem — but it is also what currently reddens the daily E2E schedule (`P12-static-routing.spec.ts:155`, [run 32361186378](https://github.com/linksys/PrivacyGUI-USP-E2E/actions/runs/32361186378), 6/6 retries).

## Root cause

`static_route_dialog.dart` validates on **every keystroke**, and that validation assigns the error map:

```dart
// :142, :150, :158, :166 — all four fields
onChanged: (_) => _validate(),

// :96
void _validate() {
  setState(() {
    _errors = _computeErrors();          // ← errorText flips null -> value
  });
}
```

When `_errors` gains an entry, the field's `errorText` changes, the widget rebuilds, and **CanvasKit tears down the semantics `<input>` mid-edit**. Focus is lost and every keystroke after that point is dropped.

Destination IP and Subnet mask survive because their values are valid as typed, so `_computeErrors()` returns nothing new for them and no rebuild is triggered. `192` in the Gateway field is *transiently* invalid (a bare `1` is not in the LAN subnet), so the error appears on the first character and kills the rest.

This mechanism is already documented in this codebase, in a dialog that was fixed for exactly this reason — `port_forwarding_dialog.dart:48-51`:

> One focus node per text field: validation runs on focus-loss, not on every keystroke. Validating in `onChanged` calls `setState` → rebuild → the CanvasKit `<input>` is torn down mid-edit, dropping focus and the value being typed. (Same focus-loss pattern as `usp_local_network_view`.)

## Evidence: a control group, not an inference

I first blamed the E2E helper, then CanvasKit, then the `ui_kit` bump (v2.34.5 → v2.39.0 across this regression window). All three are **wrong**, and a control group rules them out.

Same helper (`fillIpv4ById`), same 3-digit octets, same app sha `71f76e6a`, same `ui_kit` v2.39.0 — the only difference is how `onChanged` is wired:

| | Static routing dialog | Local network page |
|---|---|---|
| `onChanged` | `_validate()` → `setState(() { _errors = ... })` | `notifier.updateSetting(...)` (`usp_local_network_view.dart:233`) |
| validation timing | every keystroke | `onFocusChanged` (`:235`) |
| typed `192` → | **`1` / `19`** | **`192`** ✓ |

Instrumented CI output, static routing:

```
PROBE60 name-after-fill    value="e"    activeElement=INPUT@static-route-name
PROBE60 gateway-octet-1    value="1"       (retry 2: "19", retry 3: "1")
PROBE60 gateway-octet-2    value="168"
PROBE60 snapshot-has-e2e-route=false     ← the controller never got the full value
PROBE60 submit aria-disabled="true"      ← _isFormValid says no
```

Local network, same probe shape:

```
PROBE60-LAN ip-octet-1 value="192"
PROBE60-LAN ip-octet-2 value="168"
PROBE60-LAN ip-octet-3 value="5"
PROBE60-LAN ip-octet-4 value="199"
PROBE60-LAN snapshot-has-199=true
7 passed
```

Two details worth keeping: the loss is present at the **first** read (right after the field is filled), so it is not a later rebuild wiping a complete value; and the truncation length **varies between retries** (`1` / `19` / `1`), which is a race, not deterministic logic.

Note also that `document.activeElement` was on the correct input, and the DOM `.value` was short — so this is not the "DOM has the value, controller doesn't" trap from `wiki/privacygui/testing/e2e-canvaskit-text-input-semantics.md`.

## Scope: one file, four lines

Everything else that renders a text field is already safe. I audited all of `lib/` rather than trusting a grep of handler names — my first pass matched only the handler *name* and produced a 12-file list that was mostly wrong, because what matters is whether the handler **body** assigns an error.

```
onChanged handlers whose body ASSIGNS an error → 4 hit(s)

  AT RISK  page/static_routing/views/dialogs/static_route_dialog.dart:142
  AT RISK  page/static_routing/views/dialogs/static_route_dialog.dart:150
  AT RISK  page/static_routing/views/dialogs/static_route_dialog.dart:158
  AT RISK  page/static_routing/views/dialogs/static_route_dialog.dart:166
```

Verified safe, with the reason (these were asked about specifically — port forwarding and DHCP reservation):

| File | Why it's safe |
|---|---|
| `port_forwarding_dialog.dart` | `_onInputChanged()` is `setState(() {})` — rebuild only, to re-gate the button; never assigns `_errors` |
| `port_range_forwarding_dialog.dart` | same, `:103` |
| `ipv6_port_service_rule_dialog.dart` | same, `:118` |
| `dhcp_reservation_edit_dialog.dart` | text fields have **no** `onChanged`; `_validate()` is driven by `FocusNode` (`:75`, `:79`) |
| `port_triggering_dialog.dart` | text fields have no `onChanged` |
| `pnp_static_ip_view.dart` / `pnp_pppoe_view.dart` | `setState(() {})`, no `errorText` at all |
| `wifi_network_card.dart` / `wifi_quick_setup_card.dart` | flip an `isValid` bool only |

So the answer to "why did this ever work?" is that these dialogs **were already fixed** for this exact defect. Static routing is the one that was missed.

## The audit script

`onchanged_error_rebuild_audit.py`, attached in the first comment below. Run it as:

```bash
LIB=/path/to/PrivacyGUI/lib python3 onchanged_error_rebuild_audit.py
```

It walks handler bodies with brace matching (not a line-level grep) and flags any text-field `onChanged` whose handler assigns an error.

**It self-checks, and that matters here.** An earlier version silently returned **0 hits** — it stripped the leading `_` when capturing the handler name but not when searching for the method, so every private handler failed to resolve. A clean "0 hits" and a broken parser are indistinguishable from the outside, so the script now proves it still discriminates before you trust it:

```
self-check (must pass, else the audit is not discriminating):
  PASS  static_route_dialog.dart            _validate        assigns_error=True  (expected True)
  PASS  port_forwarding_dialog.dart         _onInputChanged  assigns_error=False (expected False)
  PASS  port_range_forwarding_dialog.dart   _onInputChanged  assigns_error=False (expected False)
  PASS  ipv6_port_service_rule_dialog.dart  _onInputChanged  assigns_error=False (expected False)

self-check: PASS — positive and negative cases both discriminate
```

If a future refactor renames those handlers the self-check fails loudly instead of reporting a reassuring zero.

## Suggested fix

Copy the pattern from `port_forwarding_dialog.dart` — it is the closest sibling and already carries the rationale in comments:

1. `onChanged` → a handler that only does `setState(() {})`, so `_isFormValid` is re-evaluated and the **Add** button re-gates, **without** assigning `_errors`.
2. Assign `_errors` from a `FocusNode` listener per field (validation on focus-loss).
3. Keep "empty field reports no error" so errors don't surface mid-fill.

Six in-repo references: `port_forwarding_dialog.dart`, `port_range_forwarding_dialog.dart`, `ipv6_port_service_rule_dialog.dart`, `dhcp_reservation_edit_dialog.dart`, `usp_local_network_view.dart`, `usp_dmz_view.dart`.

## Not in scope here

The E2E side will add a read-back assertion in `focusAndType` so a truncated value fails at the offending field instead of surfacing four steps later as `page-save count 0` — tracked in `linksys/PrivacyGUI-USP-E2E#60`. That is diagnostics, not a fix; this issue is the actual defect.

AustinChangLinksys:
## Summary

Fixes #1332 — the **Add static route** dialog dropped keystrokes, so a typed IP
like `192` landed as `1`/`19`, `_isFormValid` kept **Add** disabled, and the
dialog couldn't be submitted.

## Root cause

`static_route_dialog.dart` validated on **every keystroke** and that validation
assigned the `_errors` map:

```dart
onChanged: (_) => _validate();        // all four fields
void _validate() => setState(() { _errors = _computeErrors(); });
```

When `_errors` gained an entry, the field's `errorText` flipped `null -> value`,
the widget rebuilt, and CanvasKit tore down the semantics `<input>` mid-edit —
focus and every subsequent keystroke were lost. A transiently-invalid Gateway (a
bare `1` isn't in the LAN subnet) triggered the error on the first character.

## Fix — validate on focus-loss (existing pattern)

Matches the pattern already used by `port_forwarding_dialog.dart`,
`usp_local_network_view.dart`, and `dhcp_reservation_edit_dialog.dart`:

| Field | How validation now fires |
|---|---|
| name (`AppTextField`) | `FocusNode` listener → `_validate()` on blur |
| dest / subnet / gateway (`AppIpv4TextField`) | `onFocusChanged` → `_validate()` when focus leaves the whole field (`index == null && !hasFocus`) |

`onChanged` now calls `_onInputChanged()` = `setState(() {})` only — it re-gates
the **Add** button (`_isFormValid` recomputes live from the controllers) and
**never assigns `_errors` mid-edit**.

## Scope

One production file, four handlers rewired (the exact 4 lines called out in the
issue), plus a new widget test. No `.arb` / codegen / l10n changes.

## Verification

```
$ fvm flutter analyze lib/.../static_route_dialog.dart test/.../static_route_dialog_test.dart
No issues found! (ran in 9.4s)

$ fvm flutter test test/page/static_routing/views/dialogs/static_route_dialog_test.dart
00:09 +4: All tests passed!

# no regression on existing static-routing tests
$ fvm flutter test test/page/static_routing/services/... test/page/static_routing/providers/...
00:02 +43: All tests passed!

# 2.x CI Quality gate — dart format (whole repo)
$ fvm dart format --output=none --set-exit-if-changed .
Formatted 1221 files (0 changed) in 9.34s   # EXIT=0
```

The new test guards the regression directly: typing an invalid value **while the
field is still focused** must not surface `errorText`; the error appears only
**after** the field loses focus.

Refs #1332

### DIFF
--- a/lib/page/static_routing/views/dialogs/static_route_dialog.dart
+++ b/lib/page/static_routing/views/dialogs/static_route_dialog.dart
@@ -48,6 +48,13 @@ class _StaticRouteDialogState extends State<StaticRouteDialog> {
   late TextEditingController _destIpController;
   late TextEditingController _subnetMaskController;
   late TextEditingController _gatewayController;
+  // Focus node for the (plain-text) name field. Validation runs on focus-loss,
+  // not on every keystroke: assigning _errors in onChanged calls setState →
+  // rebuild → the CanvasKit <input> is torn down mid-edit, dropping focus and
+  // the value being typed (#1332). The three IPv4 fields use AppIpv4TextField's
+  // own onFocusChanged callback (_onIpv4FocusChanged) for the same purpose.
+  // (Same focus-loss pattern as port_forwarding_dialog / usp_local_network_view.)
+  final _nameFocus = FocusNode();
   late String _interfaceName;
   late bool _enabled;
 
@@ -70,6 +77,11 @@ class _StaticRouteDialogState extends State<StaticRouteDialog> {
     // fields are empty, so computing errors immediately would show "required"
     // errors before the user has typed anything (confusing, non-standard UX).
     _errors = _isEdit ? _computeErrors() : {};
+    // Validate the name field when it loses focus (updates the shown errorText)
+    // instead of on every keystroke.
+    _nameFocus.addListener(() {
+      if (!_nameFocus.hasFocus && mounted) _validate();
+    });
   }
 
   @override
@@ -78,6 +90,7 @@ class _StaticRouteDialogState extends State<StaticRouteDialog> {
     _destIpController.dispose();
     _subnetMaskController.dispose();
     _gatewayController.dispose();
+    _nameFocus.dispose();
     super.dispose();
   }
 
@@ -97,6 +110,20 @@ class _StaticRouteDialogState extends State<StaticRouteDialog> {
     setState(() {
       _errors = _computeErrors();
     });
+  }
+
+  /// Rebuild to re-evaluate the Add-button enable state (_isFormValid) WITHOUT
+  /// surfacing errors mid-edit. Error text is refreshed on focus-loss (via
+  /// _nameFocus / _onIpv4FocusChanged), so it must NOT be assigned here:
+  /// assigning _errors in onChanged → setState → rebuild tears down the
+  /// CanvasKit <input> mid-edit and drops the keystroke (#1332).
+  void _onInputChanged() {
+    setState(() {});
+  }
+
+  /// Validate when focus leaves the entire IPv4 field (not per segment).
+  void _onIpv4FocusChanged(int? index, bool hasFocus) {
+    if (index == null && !hasFocus && mounted) _validate();
   }
 
   /// Convert error key to localized string.
@@ -136,10 +163,11 @@ class _StaticRouteDialogState extends State<StaticRouteDialog> {
           children: [
             AppTextField(
               controller: _nameController,
+              focusNode: _nameFocus,
               identifier: 'static-route-name',
               hintText: loc(context).routeName,
               errorText: _localizeError(_errors['name']),
-              onChanged: (_) => _validate(),
+              onChanged: (_) => _onInputChanged(),
             ),
             AppGap.lg(),
             AppIpv4TextField(
@@ -147,7 +175,8 @@ class _StaticRouteDialogState extends State<StaticRouteDialog> {
               identifier: 'static-route-dest-ip',
               label: loc(context).destinationIp,
               errorText: _localizeError(_errors['destIp']),
-              onChanged: (_) => _validate(),
+              onChanged: (_) => _onInputChanged(),
+              onFocusChanged: _onIpv4FocusChanged,
             ),
             AppGap.lg(),
             AppIpv4TextField(
@@ -155,7 +184,8 @@ class _StaticRouteDialogState extends State<StaticRouteDialog> {
               identifier: 'static-route-subnet-mask',
               label: loc(context).subnetMask,
               errorText: _localizeError(_errors['subnetMask']),
-              onChanged: (_) => _validate(),
+              onChanged: (_) => _onInputChanged(),
+              onFocusChanged: _onIpv4FocusChanged,
             ),
             AppGap.lg(),
             AppIpv4TextField(
@@ -163,7 +193,8 @@ class _StaticRouteDialogState extends State<StaticRouteDialog> {
               identifier: 'static-route-gateway',
               label: loc(context).gatewayIp,
               errorText: _localizeError(_errors['gateway']),
-              onChanged: (_) => _validate(),
+              onChanged: (_) => _onInputChanged(),
+              onFocusChanged: _onIpv4FocusChanged,
             ),
             AppGap.lg(),
             Row(
--- a/test/page/static_routing/views/dialogs/static_route_dialog_test.dart
+++ b/test/page/static_routing/views/dialogs/static_route_dialog_test.dart
@@ -1,0 +1,169 @@
+@Tags(['ui'])
+import 'package:flutter/material.dart';
+import 'package:flutter_test/flutter_test.dart';
+import 'package:go_router/go_router.dart';
+import 'package:ui_kit_library/ui_kit.dart';
+import 'package:privacy_gui/l10n/gen/app_localizations.dart';
+import 'package:privacy_gui/page/static_routing/views/dialogs/static_route_dialog.dart';
+
+/// Widget tests for [StaticRouteDialog].
+///
+/// Covers linksys/PrivacyGUI#1332: the Add static route dialog dropped
+/// keystrokes because validation ran on EVERY keystroke and assigned the
+/// `_errors` map inside `onChanged`. When `_errors` gained an entry the field's
+/// `errorText` flipped null -> value, the widget rebuilt, and CanvasKit tore
+/// down the semantics `<input>` mid-edit, so focus (and every keystroke after)
+/// was lost. The fix moves validation to focus-loss (a FocusNode listener for
+/// the plain-text name field, AppIpv4TextField.onFocusChanged for the three
+/// IPv4 fields), so `onChanged` only rebuilds to re-gate the Add button and
+/// never assigns `_errors` mid-edit — the same focus-loss pattern already used
+/// by port_forwarding_dialog / usp_local_network_view / dhcp_reservation_edit.
+///
+/// The dialog surfaces validation errors via each field's `errorText`. Those
+/// render inline OR as a focus tooltip depending on layout/focus state, so
+/// these tests assert on the widget's `errorText` property (the source of
+/// truth) rather than a rendered string.
+final _testTheme = AppTheme.create(
+  brightness: Brightness.light,
+  seedColor: Colors.blue,
+  designThemeBuilder: (c) => CustomDesignTheme.fromJson({'style': 'flat'}),
+);
+
+/// Pumps the dialog inside a GoRouter (the dialog calls Navigator.pop) and
+/// immediately opens it. Returns the result the dialog pops with (null = still
+/// open / cancelled).
+Future<StaticRouteDialogResult?> _pumpAndOpen(
+  WidgetTester tester, {
+  String? lanIp = '192.168.1.1',
+  String? lanSubnetMask = '255.255.255.0',
+}) async {
+  StaticRouteDialogResult? result;
+  bool popped = false;
+  final router = GoRouter(
+    routes: [
+      GoRoute(
+        path: '/',
+        builder: (context, state) => Scaffold(
+          body: Builder(
+            builder: (context) => Center(
+              child: ElevatedButton(
+                onPressed: () async {
+                  result = await showAppDialog<StaticRouteDialogResult>(
+                    context: context,
+                    builder: (_) => StaticRouteDialog(
+                      lanIp: lanIp,
+                      lanSubnetMask: lanSubnetMask,
+                    ),
+                  );
+                  popped = true;
+                },
+                child: const Text('open'),
+              ),
+            ),
+          ),
+        ),
+      ),
+    ],
+  );
+
+  await tester.pumpWidget(MaterialApp.router(
+    theme: _testTheme,
+    localizationsDelegates: AppLocalizations.localizationsDelegates,
+    supportedLocales: AppLocalizations.supportedLocales,
+    routerConfig: router,
+  ));
+  await tester.tap(find.text('open'));
+  await tester.pumpAndSettle();
+  return popped ? result : null;
+}
+
+/// The dialog's four text fields.
+// AppTextField is the single plain-text field in the dialog (the route name).
+AppTextField _nameField() =>
+    find.byType(AppTextField).evaluate().first.widget as AppTextField;
+
+// The three AppIpv4TextFields are ordered: dest (0), subnet (1), gateway (2).
+AppIpv4TextField _ipv4At(int i) =>
+    find.byType(AppIpv4TextField).evaluate().elementAt(i).widget
+        as AppIpv4TextField;
+
+AppIpv4TextField _destField() => _ipv4At(0);
+AppIpv4TextField _subnetField() => _ipv4At(1);
+AppIpv4TextField _gatewayField() => _ipv4At(2);
+
+/// Enters [text] into the name field. Does NOT drop focus, so this mirrors the
+/// user still typing — used to assert the #1332 no-mid-edit-error property.
+Future<void> _typeName(WidgetTester tester, String text) async {
+  await tester.enterText(_nameTextField(), text);
+  await tester.pumpAndSettle();
+}
+
+/// The raw name TextField (first TextField in tree — the AppIpv4TextFields
+/// render their own TextFields after it).
+Finder _nameTextField() => find.byType(TextField).first;
+
+Future<void> _blur(WidgetTester tester) async {
+  FocusManager.instance.primaryFocus?.unfocus();
+  await tester.pumpAndSettle();
+}
+
+void main() {
+  late String addLabel;
+
+  setUp(() async {
+    final loc = await AppLocalizations.delegate.load(const Locale('en'));
+    addLabel = loc.add;
+  });
+
+  group('StaticRouteDialog #1332 focus-loss validation', () {
+    testWidgets('add-open shows no errors on any field (empty form)',
+        (tester) async {
+      await _pumpAndOpen(tester);
+
+      expect(_nameField().errorText, isNull);
+      expect(_destField().errorText, isNull);
+      expect(_subnetField().errorText, isNull);
+      expect(_gatewayField().errorText, isNull);
+    });
+
+    testWidgets(
+        'typing an invalid name does NOT set errorText mid-edit (#1332): '
+        'onChanged must not assign _errors while still focused',
+        (tester) async {
+      await _pumpAndOpen(tester);
+
+      // A 33+ char name is invalid (nameTooLong), but while the field is still
+      // focused (user typing) the error must NOT appear — assigning it would
+      // rebuild and tear down the CanvasKit <input>, dropping keystrokes.
+      await _typeName(tester, 'x' * 40);
+
+      expect(_nameField().errorText, isNull,
+          reason: 'onChanged must not assign _errors mid-edit (#1332)');
+    });
+
+    testWidgets('name error surfaces only after the field loses focus',
+        (tester) async {
+      await _pumpAndOpen(tester);
+      await _typeName(tester, 'x' * 40); // invalid: too long
+      expect(_nameField().errorText, isNull); // still typing -> no error yet
+
+      await _blur(tester); // tab away -> validation runs
+
+      expect(_nameField().errorText, isNotNull,
+          reason: 'validation runs on focus-loss');
+    });
+
+    testWidgets('Add stays disabled while the form is incomplete/invalid',
+        (tester) async {
+      await _pumpAndOpen(tester);
+      await _typeName(tester, 'e2e-route');
+      await _blur(tester);
+
+      // Dest/subnet/gateway are still empty -> form invalid -> Add disabled ->
+      // tapping does not pop the dialog.
+      await tester.tap(find.widgetWithText(AppButton, addLabel));
+      await tester.pumpAndSettle();
+      expect(find.byType(StaticRouteDialog), findsOneWidget);
+    });
+  });
+}
--- a/test/page/static_routing/views/dialogs/static_route_dialog_test.dart
+++ b/test/page/static_routing/views/dialogs/static_route_dialog_test.dart
@@ -107,6 +107,28 @@ Future<void> _blur(WidgetTester tester) async {
   await tester.pumpAndSettle();
 }
 
+/// Enters [text] into one octet box of the gateway field WITHOUT dropping focus,
+/// mirroring a user mid-typing. The dialog's TextFields are ordered: the name
+/// field first, then each AppIpv4TextField's four octet boxes in field order
+/// (dest, subnet, gateway) — so gateway octet [octet] is at
+/// 1 + (2 * 4) + octet. Asserted against the field count rather than hardcoded
+/// so a layout change fails loudly instead of silently typing into the wrong box.
+Future<void> _typeGatewayOctet(
+  WidgetTester tester,
+  int octet,
+  String text,
+) async {
+  final boxes = find.byType(TextField);
+  expect(
+    boxes.evaluate().length,
+    1 + 3 * 4,
+    reason: 'expected the name field plus three IPv4 fields of four octets; '
+        'if this changed, the octet index below is wrong',
+  );
+  await tester.enterText(boxes.at(1 + 2 * 4 + octet), text);
+  await tester.pumpAndSettle();
+}
+
 void main() {
   late String addLabel;
 
@@ -164,6 +186,36 @@ void main() {
       await tester.tap(find.widgetWithText(AppButton, addLabel));
       await tester.pumpAndSettle();
       expect(find.byType(StaticRouteDialog), findsOneWidget);
+    });
+
+    // Review round 1 flagged that gateway — the field #1332 actually reproduces
+    // on (typed 192, landed 1/19) — had no focus-loss coverage: it was only
+    // asserted null on open. These two cover the AppIpv4TextField
+    // onFocusChanged path that the name field's FocusNode listener does not.
+    testWidgets(
+        'typing an invalid gateway does NOT set errorText mid-edit (#1332): '
+        'the octet path must not assign _errors while still focused',
+        (tester) async {
+      await _pumpAndOpen(tester);
+      // A bare "1" is not a valid address and is outside the LAN subnet, so it
+      // is exactly the transiently-invalid state that used to flip errorText on
+      // the first keystroke and tear the <input> down mid-edit.
+      await _typeGatewayOctet(tester, 0, '1');
+
+      expect(_gatewayField().errorText, isNull,
+          reason: 'still focused mid-edit -> onChanged must not assign _errors');
+    });
+
+    testWidgets('gateway error surfaces only after the IPv4 field loses focus',
+        (tester) async {
+      await _pumpAndOpen(tester);
+      await _typeGatewayOctet(tester, 0, '1');
+      expect(_gatewayField().errorText, isNull); // still typing -> no error yet
+
+      await _blur(tester); // whole IPv4 field blurs -> validation runs
+
+      expect(_gatewayField().errorText, isNotNull,
+          reason: 'validation runs when the entire IPv4 field loses focus');
     });
   });
 }
--- a/test/page/static_routing/views/dialogs/static_route_dialog_test.dart
+++ b/test/page/static_routing/views/dialogs/static_route_dialog_test.dart
@@ -203,7 +203,8 @@ void main() {
       await _typeGatewayOctet(tester, 0, '1');
 
       expect(_gatewayField().errorText, isNull,
-          reason: 'still focused mid-edit -> onChanged must not assign _errors');
+          reason:
+              'still focused mid-edit -> onChanged must not assign _errors');
     });
 
     testWidgets('gateway error surfaces only after the IPv4 field loses focus',


AustinChangLinksys (commented):
🤖 **Automated Review — Round 1** · `71f76e6a..cbd939c0` (full)

**Verdict: ✅ APPROVE (posted as COMMENT — GitHub blocks self-approve via API)** — Fix is correct, matches the established focus-loss pattern, and spec requirements are all met. Warnings are pre-existing or minor; none block merge. A human reviewer can approve.

#### Standards
|---|---|---|---|
| ⚠️ | 🟡Med | `static_route_dialog.dart:210–212` | `onSelectionChanged` still calls `_validate()` directly — breaks the "no `_errors` assignment except on focus-loss" rule if a text field is in mid-edit when Interface is switched (pre-existing from base; not introduced by this PR) |
| ⚠️ | 🟡Med | `static_route_dialog.dart:153` | `_isFormValid` getter calls `_computeErrors()` on every `build()`, triggering full `validateRoute` on each keystroke (pre-existing; PR comment explains the design intent — keeping it for add-mode blank-form gating) |
| 💡 | 🟡Med | `static_route_dialog_test.dart` | No test covers `AppIpv4TextField.onFocusChanged` path — gateway field (the primary bug scenario in #1332) has no focus-loss regression test |
| 💡 | 🟢High | `static_route_dialog_test.dart` | Edit-mode (`route:` param) is entirely untested; `initState` path for edit mode not exercised |
| 💡 | ⚪Low | `static_route_dialog.dart:57` | `_nameFocus` initialized at field-declaration level; other controllers use `late` in `initState` — minor style inconsistency |

#### Spec
|---|---|---|---|
| ✅ | 🟢High | `static_route_dialog.dart:120–122, 170, 178, 187, 196` | S1: All four `onChanged` handlers now call `_onInputChanged()` only — never assign `_errors` |
| ✅ | 🟢High | `static_route_dialog.dart:57, 82–84, 91, 166` | S2a: `_nameFocus` FocusNode listener fires `_validate()` on blur |
| ✅ | 🟢High | `static_route_dialog.dart:125–126, 179, 188, 197` | S2b: `_onIpv4FocusChanged` fires `_validate()` when entire IPv4 field loses focus (`index == null && !hasFocus`) |
| ✅ | 🟢High | `static_route_dialog.dart:72` | S3: `_errors = {}` on open; empty field surfaces no error until blur |
| ✅ | 🟢High | diff (base lines 142, 150, 158, 166) | S4: All four `onChanged` handlers in scope rewired — no omissions |
| ✅ | 🟢High | `static_route_dialog_test.dart:120–157` | S5: Widget test guards regression — typing invalid name while focused does NOT surface `errorText`; error appears only after blur |
| ⚠️ | 🟢High | `static_route_dialog_test.dart` | Spec explicitly names gateway as the primary bug field (`192 → 1/19`); that IPv4 focus-loss path has no test case |

<details>

**W-1 · `onSelectionChanged` calls `_validate()` directly** (🟡Med — pre-existing)
- `static_route_dialog.dart:210–212`
- Code:
  ```dart
  onSelectionChanged: (v) {
    setState(() => _interfaceName = v.first);
    _validate();   // ← assigns _errors immediately
  },
  ```
- This pattern existed in the base version (line 181 of base) — the PR did not introduce it. In the base, every `onChanged` also called `_validate()`, so the `onSelectionChanged` was consistent with the old pattern. Now that `onChanged` is safely redirected to `_onInputChanged()`, this one handler stands out as inconsistent. Practically, clicking a `SegmentedButton` causes the focused text field to blur first (Flutter focus system), which would already fire `_validate()` via the focus-loss listeners — so the double-call is likely harmless. However, in edge cases (e.g., CanvasKit race between blur and selection-change events), this could surface a gateway error mid-edit.
- **Fix:** Replace `_validate()` with `_onInputChanged()` in `onSelectionChanged`, or add a clarifying comment explaining why an immediate `_validate()` call is acceptable for a SegmentedButton (no CanvasKit `<input>` teardown risk). See Reviewer B's suggestion: `// SegmentedButton does not have a mid-edit focus problem; validate immediately because the interface choice affects gateway subnet rules.`
- **Verdict:** Pre-existing, not归本PR. Comment-only, no review-fix card.

**W-2 · `_isFormValid` recomputes `_computeErrors()` on every `build()`** (🟡Med — pre-existing, intentional)
- `static_route_dialog.dart:153`
- Code:
  ```dart
  // Compute validity from the live field values rather than the cached
  // [_errors] map. In add mode [_errors] starts empty (errors are suppressed
  // on open for UX), so relying on it would enable the Save button on a blank
  // form...
  bool get _isFormValid => _computeErrors().isEmpty;
  ```
- The code comment (head version) explicitly documents this design choice — add-mode needs live recomputation because `_errors` starts empty intentionally. This is pre-existing behavior; the PR did not change the getter. Performance cost is `validateRoute()` (subnet math) on each keystroke; acceptable for a dialog-level widget on desktop/tablet.
- **Verdict:** Pre-existing, intentional design, documented in code. No action needed.

</details>

<details>

- **Core fix is correct and complete:** All four `onChanged` handlers rewired to `_onInputChanged()` (no `_errors` assignment); focus-loss via `FocusNode` listener (name) and `onFocusChanged` (IPv4 fields) properly assigned.
- **`mounted` guard on both focus listeners** (`dart:83`, `dart:126`) — prevents `setState` after dispose.
- **`_nameFocus.dispose()` called correctly** (`dart:91`) — no `ChangeNotifier` leak.
- **`_onIpv4FocusChanged` condition `index == null && !hasFocus`** correctly waits for the entire IPv4 field to lose focus (not per-segment blur), matching `AppIpv4TextField` semantics.
- **`_onInputChanged` is clearly commented** explaining why it must never assign `_errors` — good defensive documentation against future regression.
- **Pattern alignment:** Fix matches the established pattern in `port_forwarding_dialog.dart`, `usp_local_network_view.dart`, `dhcp_reservation_edit_dialog.dart` — consistent codebase approach.
- **New widget test covers the regression directly:** "typing an invalid name does NOT set `errorText` mid-edit" and "name error surfaces only after focus loss" — exactly what issue #1332 requires.
- **Scope is tight:** Only `static_route_dialog.dart` (+35/-4) and its new test file (+169). No unrelated files touched.
- **All spec requirements met:** S1–S5 from Issue #1332 fully satisfied with code evidence.

</details>

PeterJhongLinksys (approved):
🤖 **Automated Review — Round 1** · `71f76e6..0ffc1a6` (full)

**Verdict: ✅ APPROVE** — the right fix in the right shape (validation moved off the keystroke path onto focus-loss, following existing in-repo precedent), no Critical on either axis; two items are worth a follow-up commit before merge (the regression test is excluded from the CI gate, and blur-validation still writes the whole error map).

#### Standards

|---|---|---|---|
| ⚠️ | 🟢 | `test/page/static_routing/views/dialogs/static_route_dialog_test.dart:1` | The new regression guard is `@Tags(['ui'])` and `run_tests.sh:80,92,96` excludes `ui`, so it never runs in the PR gate — the repo already solved this with the non-excluded `dashboard-card` tag (`dart_test.yaml:5-8`). |
| ⚠️ | 🟢 | `lib/page/static_routing/views/dialogs/static_route_dialog.dart:109-113` | Focus-loss `_validate()` assigns the **whole** `_errors` map, so blurring `name` paints "required" errors on the three untouched IPv4 fields — reintroducing on first blur the UX the code deliberately avoids on open (`:75-79`). |
| ⚠️ | 🟢 | `lib/page/static_routing/views/dialogs/static_route_dialog.dart:210-212` | `SegmentedButton.onSelectionChanged` still calls `_validate()` directly — the one input left out of the new pattern; tapping the segment on a blank add form paints all four "required" errors. |
| ⚠️ | 🟢 | `lib/page/static_routing/views/dialogs/static_route_dialog.dart:120-122`, `:153` | `_onInputChanged()` → `setState(() {})` → `build()` → `_isFormValid` → `_computeErrors()`: the validator (incl. subnet arithmetic) still runs on **every keystroke**; only the *write* to `_errors` moved. Comments at `:52-53`, `:115-121` overstate this. |
| ⚠️ | 🟢 | `test/.../static_route_dialog_test.dart:33-77`, `:178-190` | No positive submit path anywhere: `_submit()` (`:246-255`) and the enabled branch of `onTap: _isFormValid ? _submit : null` (`:241`) have zero coverage; `_pumpAndOpen`'s `result`/`popped` machinery is dead. |
| ⚠️ | 🟢 | `lib/page/static_routing/views/dialogs/static_route_dialog.dart:79` | Edit mode (`_isEdit ? _computeErrors() : {}`) is never exercised — `_pumpAndOpen` never passes `route:` — yet it is the arm with the *opposite* error-seeding behaviour and drives the Save-vs-Add label (`:240`). |
| ⚠️ | 🟢 | `test/.../static_route_dialog_test.dart:116-130` | Octet targeting uses ordinal arithmetic into `AppIpv4TextField`'s private composition (`boxes.at(1 + 2 * 4 + octet)`); the count guard catches insertions but not reordering. Stable `identifier:` values already exist on the fields. |
| ⚠️ | 🟢 | `lib/page/.../static_route_dialog.dart:57`, `:81-83`, `:115-127` | [Duplicated Code / Shotgun Surgery] The FocusNode + `_onInputChanged` triplet is now hand-copied in ≥4 pages (`port_range_forwarding_dialog.dart:86-103`, `usp_local_network_view.dart`, `usp_dmz_view.dart:234`). #1332 is the 2nd instance of this bug class — durable fix belongs in ui_kit. |
| ⚠️ | 🟡 | `lib/page/static_routing/views/dialogs/static_route_dialog.dart:82-84`, `:87-95` | The `_nameFocus` listener is never `removeListener`'d and `_nameFocus.dispose()` runs *after* the controllers; `mounted` is still `true` during `State.dispose()`, so it is not a valid teardown guard. |
| ⚠️ | 🟡 | `lib/page/static_routing/views/dialogs/static_route_dialog.dart:125-127` | The `index == null` aggregate-blur convention is an undocumented contract of `AppIpv4TextField` from git-pinned `ui_kit_library` (`pubspec.yaml:59-62`, `ref: v2.39.0`); a version bump could silently disable IPv4 error display with no compile or test failure. |
| ⚠️ | 🟡 | `test/.../static_route_dialog_test.dart:35-38` | Art. VIII: LAN context (`lanIp`, `lanSubnetMask`) is inline ad-hoc data; no `test/mocks/test_data/*static*` builder exists, so the three static_routing test files can drift on what "the LAN" is. |
| 💡 | 🟢 | `lib/page/static_routing/views/dialogs/static_route_dialog.dart:129-145` | `_localizeError` collapses `nameRequired` / `nameTooLong` / `subnetMaskRequired` / `invalidSubnetMask` onto `l.invalidInput` (the 32-char limit is undiscoverable), and the `_ => key` fallback at `:144` would leak a raw camelCase identifier into the UI. |
| 💡 | 🟢 | `lib/page/static_routing/views/dialogs/static_route_dialog.dart:170`, `:178`, `:187`, `:196` | [Primitive Obsession] `_errors` keys are bare string literals duplicated between the view and `validateRoute`'s return map; a typo silently shows no error — the exact failure mode of #1332. |
| 💡 | 🟡 | `test/.../static_route_dialog_test.dart` | Every assertion is on `errorText` (the hypothesised proximate cause); none asserts the user-visible symptom (typed N chars → controller holds N). A proxy at best, but it would catch a future re-introduction by a different mechanism. |
| 💡 | 🟡 | `test/.../static_route_dialog_test.dart:116-130` | Make the octet helper field-parametric (`1 + field * 4 + octet`) so dest/subnet blur cases can be added without duplicating the arithmetic — needed by the whole-map fix above. |
| 💡 | 🟡 | `lib/page/static_routing/views/dialogs/static_route_dialog.dart:124-127` | One shared `_onIpv4FocusChanged` is wired to all three IPv4 fields, so it cannot tell *which* field blurred — exactly the information a field-scoped fix needs. Currying the key now avoids re-plumbing three call sites later. |
| 💡 | 🟡 | `test/.../static_route_dialog_test.dart:140` | Group name is `[ClassName] #issue feature`; Art. VIII asks for `[ClassName] - [Feature]` (e.g. `'StaticRouteDialog - focus-loss validation (#1332)'`). Cosmetic. |
| 💡 | ⚪ | `lib/page/static_routing/views/dialogs/static_route_dialog.dart:57`, `:82-84` | A `Focus(onFocusChange: ...)` wrapper would give the same semantics with no `FocusNode` lifecycle to manage — relevant if this dialog gains more plain-text fields. |
| 💡 | ⚪ | `test/page/static_routing/views/dialogs/` | Art. VIII names only `services|providers/` for test layout, but `views/dialogs/` matches existing precedent (`test/page/dhcp/views/dialogs/`, `test/page/dashboard/views/dialogs/`). Read as correct-by-convention — worth blessing `views/` in the Constitution so it stops producing false findings. |

#### Spec

Spec source: #1332

|---|---|---|---|
| ⚠️ | 🟢 | _"3. Keep 'empty field reports no error' so errors don't surface mid-fill."_ | [Omission] Not implemented: `_validate()` (`static_route_dialog.dart:109-113`) assigns the full map and `validateRoute` emits `*Required` for empty `name`/`destIp`/`subnetMask` (`usp_static_routing_service.dart:229-241` — only `gateway` is `isNotEmpty`-guarded), so the first blur during a normal fill paints errors on unreached fields. |
| ⚠️ | 🟡 | _"Gateway IP octet 1: typed `192`, landed `1` … Add disabled, dialog cannot be submitted"_ | [Omission] No evidence of the issue's own acceptance condition — that a full typed value lands intact and Add becomes enabled/submittable. All assertions are on `errorText` or on the dialog *not* closing; the widget harness cannot reproduce CanvasKit teardown at all. |
| 💡 | 🟢 | _"Copy the pattern from `port_forwarding_dialog.dart`"_ | [Wrong] Two mechanisms coexist in one dialog — `_nameFocus.addListener` (`:81-83`) for the text field vs `onFocusChanged` (`:179`, `:188`, `:197`) for IPv4. Defensible (different ui_kit APIs, IPv4 half copied verbatim from `usp_local_network_view.dart:67-72`); flagged so it is a conscious choice. |
| 💡 | 🟡 | _"Copy the pattern … it already carries the rationale in comments"_ | [Scope creep — trivial] `:125-127` adds `&& mounted` where the referenced sibling `usp_local_network_view.dart:69` has only `index == null && !hasFocus`. Harmless, arguably safer for a poppable dialog; noted as the one alteration to the copied pattern. |
| 💡 | 🟢 | _"Not in scope here: the E2E read-back assertion … tracked in PrivacyGUI-USP-E2E#60"_ | [Compliance] Confirmed absent — the change list is exactly the two files, no E2E harness and no other repo touched. Respecting stated non-scope, recorded as a positive. |

<sub>Confidence: 🟢 High = evidenced in head code · 🟡 Med = plausible, one link unproven · ⚪ Low = worth a second look. `[X]` = raised by 1 agent, `[XX]` = corroborated by both — **Standards table only** (Spec has a single reviewer, so agreement marks are meaningless there). No Criticals were raised, hence no marks appear.</sub>

<details>
<summary><b>Standards — Critical evidence</b></summary>

**No Critical findings** on the Standards axis, from either reviewer. Stated positively rather than by omission:

- **Security: clean.** The diff touches only widget-local focus/validation wiring. No credentials, no `flutter_secure_storage`, no new network/USP call site, no interpolation into a query or command, no permission gate added or bypassed. `_submit()` (`static_route_dialog.dart:246-255`) still `trim()`s all four values and forwards them unchanged — no new sink.
- **USP/JNAP error handling: not exercised.** No `catch`, no `on Exception catch`, no USP error-string parsing added or modified. The only service touchpoint is the pure-static validator `UspStaticRoutingService.validateRoute(...)` (`static_route_dialog.dart:96-107`), which returns a `Map<String, String>` of keys and performs no I/O. The service's existing async paths (e.g. `fetch()` at `usp_static_routing_service.dart:52-59`) already use `catch (e) { throw mapUspErrorToServiceError(e); }` and are untouched.
- **Art. V §5.4 tier isolation: clean.** No `generated/` import in the changed view; imports are `models/static_routing_ui_model.dart`, `services/usp_static_routing_service.dart`, `localization_hook.dart`, `ui_kit_library/ui_kit.dart` (`:1-5`). Dependency direction downward only.
- **Art. IV USP provider rules: not applicable.** No provider, no `autoDispose`, no `ref.watch`/`ref.read`, no mutation path, no `PreservableContract` redefinition.
- **Art. XIII ServiceError isolation: clean.** Only error *keys* cross into the view, localized via `_localizeError`. (The `_ => key` fallback is a latent brush with this rule and is filed as a Suggestion, not a violation — no key currently reaches it.)

Two items were considered for Critical and deliberately downgraded rather than dropped:

1. **Whole-map blur validation** is real and reproducible, but it is a UX/intent regression, not a correctness or security break — `_isFormValid` (`:153`) recomputes from live controller values, so an invalid route still cannot be submitted no matter what `errorText` currently displays. → Warning, 🟢High.
2. **The `dispose()` ordering / `mounted` guard hazard** could not be substantiated against the pinned Flutter SDK in this worktree — it is unproven that `FocusNode.dispose()` synchronously dispatches to the attached closure. Per the traceability rule it is stated as a defensive-teardown hazard. → Warning, 🟡Med.

</details>

<details>
<summary><b>Standards — Warning evidence</b></summary>

**W1 — regression guard is excluded from the CI gate** (`test/.../static_route_dialog_test.dart:1`, 🟢High)

```dart
@Tags(['ui'])
```
```bash
# run_tests.sh:80 / :92 / :96
```
All three invocations exclude `ui`, so all 222 lines — including the two tests that encode the #1332 contract — are skipped by the PR test command. The regression can only regress *silently*. The repo already documented the answer:
```yaml
# dart_test.yaml:5-8
  # failure here blocks the PR.
  dashboard-card:
```
**Fix:** drop `@Tags(['ui'])`, or register a non-excluded defensive tag the way `dashboard-card` was. `dhcp_reservation_edit_dialog_test.dart` also uses `ui`, but that precedent is the bug, not the justification — this PR's entire value proposition is a regression guard, and a guard excluded from the gate is documentation.

---

**W2 — blur validation assigns the whole error map** (`static_route_dialog.dart:109-113`, `:96-107`, `:82-84`, 🟢High — raised independently by both Standards reviewers)

```dart
// :82-84 (initState)
_nameFocus.addListener(() {
  if (!_nameFocus.hasFocus && mounted) _validate();
});

// :109-113
void _validate() {
  setState(() {
    _errors = _computeErrors();   // ALL four fields, not just the blurred one
  });
}
```
`_computeErrors()` delegates to `validateRoute`, whose first branches are unconditional emptiness checks (`usp_static_routing_service.dart:229-231`, keys declared at `:21-23`). **Trigger:** open Add → type a name → Tab out → `_errors` gains `destIp` / `subnetMask` required keys → `_localizeError` (`:129-145`) renders them → three untouched fields go red. Symmetrically via `_onIpv4FocusChanged` (`:125-127`): blurring dest IP flags the still-empty subnet mask and gateway.

This contradicts the intent the PR itself documents:
```dart
// :76-79
// Pre-populate validation state only in edit mode. On add-dialog open all
// fields are empty, so computing errors immediately would show "required"
// errors before the user has typed anything (confusing, non-standard UX).
```
The PR suppresses that state on open, then re-introduces it one Tab keypress later. `static_route_dialog_test.dart:166-176` asserts only `_nameField().errorText, isNotNull`, which passes identically under per-field or whole-form validation — the gap is invisible to the added tests.

**Fix:** scope the assignment to the field that blurred, e.g.
```dart
void _validateField(String key) {
  final all = _computeErrors();
  setState(() {
    if (all[key] == null) { _errors.remove(key); } else { _errors[key] = all[key]!; }
  });
}
```
or track a `_touched` set and gate `_localizeError` on it. `_isFormValid` (`:153`) already computes from live values, so button gating is unaffected. Add a test asserting the *other three* fields stay `errorText == null` after `name` blurs.

---

**W3 — `onSelectionChanged` still validates immediately** (`static_route_dialog.dart:210-212`, 🟢High)

```dart
onSelectionChanged: (v) {
  setState(() => _interfaceName = v.first);
  _validate();          // immediate full validation, unlike every text field
},
```
The sole remaining input that violates the PR's own rule ("inputs must not assign `_errors` mid-edit; that happens on focus-loss only"). On a blank add form, tapping the interface segment — a plausible first action, it sits directly under the fields — immediately paints `required` on all four. It is also the path most likely to be reached *while a text field is focused*, i.e. the tear-down condition described at `:51-56`. Not escalated: Flutter may blur the field first, which would make the call redundant rather than harmful. **Fix:** `_onInputChanged()` (the segment's own state is already committed by the preceding `setState`).

---

**W4 — the validator still runs per keystroke; comments say otherwise** (`:120-122`, `:153`, `:52-53`, `:115-121`, 🟢High)

```dart
void _onInputChanged() { setState(() {}); }
bool get _isFormValid => _computeErrors().isEmpty;   // :153, called from build via :241
```
Every keystroke → `setState` → `build()` → `_isFormValid` → `_computeErrors()` → `validateRoute(...)`, including IPv4 parsing and (when `lanIp`/`lanSubnetMask` are supplied) subnet-membership arithmetic. Only the *write to `_errors`* moved to focus-loss.

Two consequences. **(a) Maintainability:** the next engineer reading ":52 validation runs on focus-loss" will assume `validateRoute` is never called per keystroke and may add heavier logic (DNS lookup, route-table overlap scan) inside it. **(b) Possible incompleteness of the fix:** the PR's own analysis (`:115-119`, test header `:11-20`) attributes #1332 to *`setState` → rebuild tearing down the CanvasKit `<input>`*, with the `errorText` flip as trigger. If the true trigger is the rebuild itself rather than specifically the `errorText` property change, `_onInputChanged()` preserves the mechanism. The widget tests cannot distinguish these hypotheses — the Flutter test harness never instantiates the CanvasKit semantics `<input>`, so they stay green either way.

**Fix:** reword the comments to "error *display* refreshes on focus-loss; `_computeErrors()` is still evaluated every build to gate the submit button", and consider holding validity in a `ValueNotifier<bool>` with only the submit button wrapped in a `ValueListenableBuilder`, so a keystroke rebuilds the button subtree and never the text fields. Please also state in the PR description whether the fix was verified on a real CanvasKit web build — that is the only environment where #1332 manifests.

---

**W5 / W6 — no positive submit path, and edit mode is entirely uncovered** (`test/...:33-77`, `:178-190`; `static_route_dialog.dart:79`, 🟢High)

```dart
// test:41 / :58 / :77
bool popped = false; ... popped = true; ... return popped ? result : null;
```
The helper's doc comment (`:33-35`) says `null = still open / cancelled`, so "never closed" and "closed with no result" are indistinguishable — no test in the file can assert *what* the dialog popped with, and the `result`/`popped` machinery is effectively dead. Consequently `_submit()` (`:246-255`), the trimming it performs, and the enabled branch of `onTap: _isFormValid ? _submit : null` (`:241`) have zero coverage. Given that this PR changes exactly *when* `_errors` and validity are recomputed, an unverified submit path is the highest-value missing test here.

Edit mode is untested for the same reason — `_pumpAndOpen` (`:36-39`) never passes `route:`, so `_errors = _isEdit ? _computeErrors() : {};` (`:79`) never takes its `true` arm. That arm has the *opposite* error-seeding behaviour and the least obvious interaction with the new listeners (open an edit dialog on an already-invalid route, focus and blur `name` — the pre-seeded errors must be neither cleared nor duplicated). `_isEdit` also drives the Save-vs-Add label (`:240`), which the tests hardcode as `loc.add`.

**Fix:** distinguish "closed" from "result" in the helper, then add (a) a full happy path asserting the popped `StaticRouteDialogResult` field values, and (b) one edit-mode test asserting errors *are* pre-populated on open per `:76-79`.

---

**W7 — listener leak + `dispose()` ordering** (`static_route_dialog.dart:82-84`, `:87-95`, 🟡Med)

```dart
@override
void dispose() {
  _nameController.dispose();
  _destIpController.dispose();
  _subnetMaskController.dispose();
  _gatewayController.dispose();
  _nameFocus.dispose();          // listener still attached at this point
  super.dispose();
}
```
The closure at `:82-84` is never detached (no `removeListener`, no stored reference). Two hazards: **(1) ordering** — the controllers are disposed *before* `_nameFocus`, so if disposing a focused `FocusNode` notifies its listeners, the closure runs `_validate()` → `_computeErrors()` → reads `_nameController.text` on a disposed controller (`FlutterError: A TextEditingController was used after being disposed`); **(2) wrong guard** — during `State.dispose()`, `mounted` is still `true` until `super.dispose()` completes, so the `mounted` check does not prevent `setState()` during unmount. Realistic trigger, if the notification fires: focus is in the name field → user presses Cancel/Esc, or the parent route pops.

**Fix:** store the closure, `removeListener` it, and dispose `_nameFocus` before the controllers — that makes both hazards unreachable and the `mounted` guard redundant rather than load-bearing.

---

**W8 — the `index == null` contract lives in a git-pinned dependency** (`static_route_dialog.dart:125-127`, 🟡Med)

```dart
void _onIpv4FocusChanged(int? index, bool hasFocus) {
  if (index == null && !hasFocus && mounted) _validate();
}
```
The `index == null` sentinel ("focus left the *entire* field, not just an octet") is an undocumented convention of `AppIpv4TextField`, which lives in `ui_kit_library` pinned by git `ref: v2.39.0` (`pubspec.yaml:59-62`) and is not vendored here, so the widget source could not be read to confirm it. It is consistent with two existing call sites — `usp_local_network_view.dart:67-72` and `usp_dmz_view.dart:45-50`, both identical in shape — which is why this is Med rather than Low. **Risk:** a future ui_kit bump that changes when `onFocusChanged(null, …)` is emitted would silently disable IPv4 error display in this dialog, with no compile error and no test failure beyond the two gateway tests added here. Worth a comment naming the relied-on version, or a fallback so the aggregate-blur case cannot be silently lost.

---

**W9 — ordinal `TextField` arithmetic into ui_kit internals** (`test/...:106-131`, 🟢High)

```dart
Finder _nameTextField() => find.byType(TextField).first;
await tester.enterText(boxes.at(1 + 2 * 4 + octet), text);
```
This encodes `AppIpv4TextField`'s private composition (exactly four `TextField`s per IPv4 field, name field first). Credit where due — the author anticipated this:
```dart
expect(boxes.evaluate().length, 1 + 3 * 4,
    reason: 'expected the name field plus three IPv4 fields of four octets; '
        'if this changed, the octet index below is wrong');
```
which converts the worst case (silently typing into the wrong box) into a loud failure, and is why this is a Warning rather than Critical. It still does not cover *reordering* (count unchanged, indices shifted). **Fix:** the dialog already assigns stable `identifier:` values (`'static-route-name'`, `'static-route-gateway'`, …) — resolve fields through those, or via `find.descendant(of: find.byType(AppIpv4TextField).at(2), matching: find.byType(TextField)).at(octet)`.

---

**W10 — no Test Data Builder for the LAN context** (`test/...:35-38`, 🟡Med)

```dart
Future<StaticRouteDialogResult?> _pumpAndOpen(
  WidgetTester tester, {
  String? lanIp = '192.168.1.1',
  String? lanSubnetMask = '255.255.255.0',
}) async {
```
Art. VIII asks for `test/mocks/test_data/[feature]_test_data.dart` with named params / `copyWith()`; no such file exists for this feature, while `test/page/static_routing/services/` and `.../providers/` tests already carry their own inline fixtures. Med, not High: the rule is written primarily for Service/Provider model fixtures and these are two scalar widget parameters, so a reviewer could reasonably rule it out of scope. Still worth a `StaticRoutingTestData.lanIp` / `.lanSubnetMask` (plus a `route()` builder for the edit-mode case) so the three static_routing test files stop drifting on what "the LAN" is.

---

**W11 — the focus-loss idiom is now hand-copied in ≥4 pages** (🟢High Duplicated Code / 🟡Med Shotgun Surgery)

The identical triplet — a `FocusNode` field, `addListener(() { if (!f.hasFocus && mounted) _validate(); })` in `initState`, `void _onInputChanged() => setState(() {})`, plus manual `dispose()` — now exists in:
```
lib/page/port_forwarding/views/dialogs/port_range_forwarding_dialog.dart:55-59, 86-103, 107-118
lib/page/local_network/views/usp_local_network_view.dart:235,246,314,326,364,375,386
lib/page/dmz/views/usp_dmz_view.dart:234
lib/page/static_routing/views/dialogs/static_route_dialog.dart:57,81-83,115-127   (this PR)
```
`port_range_forwarding_dialog.dart:102-103` is character-for-character the same method as `static_route_dialog.dart:120-121`, with a near-verbatim doc comment. This is **not this PR's fault** — following precedent was the right call, and divergence would have been worse. But #1332 is the second instance of this bug class, so the third is coming, and each recurrence requires editing every one of these files independently. **Durable fix (follow-up ticket, not a blocker):** push `validateOn: ValidationTrigger.focusLoss` into ui_kit's `AppTextField`/`AppIpv4TextField` so no page has to know about CanvasKit `<input>` teardown at all (Art. XV, UI Kit first); a `FocusLossValidationMixin` in this repo is the cheaper interim.

</details>

<details>
<summary><b>Spec — findings vs #1332, and requirements met</b></summary>

**No Critical findings on the Spec axis.** The issue's root cause is fully addressed. All four `onChanged` call sites the issue names (`:142, :150, :158, :166` in the pre-fix file) are rewired at head (`static_route_dialog.dart:169, 178, 187, 196`) to `onChanged: (_) => _onInputChanged()`, and `_onInputChanged` (`:120-122`) does not touch `_errors` — exactly suggested-fix item 1. Error assignment now happens only on focus-loss (`:81-83`, `:125-127`), which is item 2. No requirement is completely unimplemented or inverted, so nothing reaches the Spec Critical bar ("the PR's main purpose was not achieved").

**W1 — suggested-fix item 3 not carried over** (🟢High, Omission)

> Spec: *"3. Keep 'empty field reports no error' so errors don't surface mid-fill."*

```dart
// static_route_dialog.dart:109-113 — focus-loss path assigns the whole map
void _validate() { setState(() { _errors = _computeErrors(); }); }
```
```dart
// usp_static_routing_service.dart:229-241 — and empty DOES produce an error, for 3 of 4 fields
if (name.isEmpty)       { errors['name']       = StaticRoutingErrorKeys.nameRequired; }       else if (name.length > 32) { ... }
if (destIp.isEmpty)     { errors['destIp']     = StaticRoutingErrorKeys.destIpRequired; }     else if (!NetworkUtils.isValidIpAddress(destIp)) { ... }
if (subnetMask.isEmpty) { errors['subnetMask'] = StaticRoutingErrorKeys.subnetMaskRequired; } else if (!NetworkUtils.isValidSubnetMask(subnetMask)) { ... }
```
Only `gateway` is guarded by `gateway.isNotEmpty &&` (`:244-246`). In add mode the dialog suppresses errors *only at open time* (`:79-80`, comment at `:75-79`), but the first blur on any field overwrites `_errors` with the map for all four — so the ordinary fill sequence (type the name, Tab to Destination IP) paints `destIpRequired` / `subnetMaskRequired` on fields the user has not reached. That is precisely the "errors surface mid-fill" condition item 3 exists to prevent.

Not Critical: this is a secondary item of the issue, and the keystroke-dropping defect (the PR's stated purpose) is fixed regardless — these errors appear on *blur*, so no `<input>` is torn down mid-edit. **Fix:** either merge only the blurred field's key (see Standards W2 — the same change satisfies both), or add the `isNotEmpty &&` guard to `name`/`destIp`/`subnetMask` in `validateRoute` the way `gateway` already has it, keeping empty-field gating in `_isFormValid` (`:153`), which recomputes independently so a blank form still cannot be submitted.

**W2 — the issue's own acceptance condition is not demonstrated** (🟡Med, Omission of evidence)

> Spec: *"| name | `e2e-route` | `e` |"*, *"| Gateway IP octet 1 | `192` | `1` (also seen: `19`) |"*, and *"Because Gateway ends up as `1.168.1.1`, `_isFormValid` renders **Add** disabled, so the dialog cannot be submitted at all."*

The new file asserts the *mechanism* (errorText null while focused, non-null after blur) for `name` and `gateway`, plus the negative case at `:178-189` ("Add stays disabled while the form is incomplete/invalid"). It never asserts what the issue defines success as: a full value typed at speed **lands intact** in the controller, Add becomes **enabled**, and the dialog pops with a result. Every `expect` is on `errorText` or on the dialog *not* closing. This is a spec matter rather than mere coverage because the widget-test layer cannot reproduce CanvasKit teardown at all, so the `errorText` assertions are a proxy for the fix, not a demonstration of it. Med confidence: the code change is the exact fix the issue prescribes, so the outcome is very likely correct — it is the evidence that is missing, not (from reading head) the fix.

**Requirements met — 7 of 9**

1. ✅ **All four `onChanged` handlers stop assigning `_errors`** — `:169, 178, 187, 196` now call `_onInputChanged()`, which is `setState(() {})` only (`:120-122`).
2. ✅ **Item 1 — `onChanged` still re-gates Add** — the empty `setState` forces a rebuild and `_isFormValid` (`:153`) recomputes from live controller text.
3. ✅ **Item 2 — errors assigned from focus-loss** — `name` via the `_nameFocus` listener (`:81-83`), IPv4 via `_onIpv4FocusChanged` (`:125-127`, whole-field blur rather than per-octet).
4. ✅ **Scope confined to the one file the issue identified** — `git diff --stat origin/dev-2.7.0...0ffc1a6a` shows only `static_route_dialog.dart` (+35/-4) and the new test (+222). No ui_kit bump, no other dialog, no new dependency, no public API change.
5. ✅ **The copied pattern is one of the six in-repo references the issue names** — `_onIpv4FocusChanged` matches `usp_local_network_view.dart:67-72`; comments at `:50-56` and `:114-119` restate the CanvasKit mechanism the issue documents from `port_forwarding_dialog.dart:48-51`.
6. ✅ **`FocusNode` lifecycle handled** — created `:56`, disposed `:93` (imperfectly — see Standards W7 — but present).
7. ✅ **The explicitly out-of-scope E2E read-back assertion was not added** — no E2E harness, no other repo touched.
8. ⚠️ **Item 3 — "empty field reports no error"** — not met (Spec W1).
9. ⚠️ **Symptom-level acceptance (typed value lands; Add becomes submittable)** — no evidence (Spec W2).

</details>

<details>

- **The root cause is actually addressed, not papered over.** The fix identifies that the bug was `_errors` *assignment* inside `onChanged` (rebuild → CanvasKit `<input>` teardown) and separates the two concerns `onChanged` was conflating: button re-gating (`_onInputChanged`, must be per-keystroke) versus error surfacing (`_validate`, must be per-blur). That is the right seam.
- **Comments explain WHY, at real depth, at the point of change.** `:50-56`, `:76-79` and `:115-119` name the issue, name the mechanism, and state why validation must not be assigned in `onChanged`. Six months from now, nobody will "clean up" the empty-bodied `setState(() {})` as dead code. For a non-obvious framework-behaviour fix this is the single most valuable thing in the diff.
- **Existing in-repo precedent was followed, and the claim is cheap to check.** `_onIpv4FocusChanged`'s shape matches `usp_local_network_view.dart:67-72` and `usp_dmz_view.dart:45-50`; the comments cite `port_forwarding_dialog` and `usp_local_network_view` by name. Consistency beat novelty here.
- **`_isFormValid` recomputing from live values (`:153`) is the correct call**, and `:147-152` spells out the trap it avoids (an empty `_errors` in add mode would otherwise enable Add on a blank form). This is the safety property that keeps everything in this PR out of Critical territory.
- **Both code paths of the fix are covered, as a 2×2.** The name field's `FocusNode` listener and the `AppIpv4TextField.onFocusChanged` path are each exercised with a mid-edit-no-error test *and* an after-blur-error-appears test. That is the right matrix.
- **The test file justifies its assertion strategy.** The header (`:9-25`) explains that `errorText` renders inline *or* as a focus tooltip depending on layout, so asserting the widget property is deliberate rather than lazy — far more robust than `find.text('Invalid input')`.
- **Self-aware brittleness guard.** `expect(boxes.evaluate().length, 1 + 3 * 4, reason: ...)` (`:121-127`) shows the author knew the ordinal indexing was fragile and converted a silent failure into a loud one. Better hygiene than most widget tests in this repo.
- **Round-1 feedback was absorbed and the trail was left in place.** `:190-194` records that gateway focus-loss coverage was requested, adds the two tests, and explains *why* gateway is the field #1332 reproduces on. The follow-up commit also verified the new tests have discriminating power by base-swapping onto the unfixed dialog at `dev-2.7.0` — that is exactly the right way to prove a test guards a regression rather than restating current behaviour.
- **Scope discipline (Art. I §1.3).** Two files, one behaviour, one issue. No drive-by lint fixes, no unrelated refactors, no new dependencies. Exactly the right blast radius.
- **Architecture clean, and verified rather than assumed.** Tier isolation, ServiceError boundary, naming, UI-Kit-first (`AppTextField` / `AppIpv4TextField` / `AppButton` / `AppSwitch` / `AppGap` / `AppText` — nothing hand-rolled). Validation logic stayed in the Service where it belongs; the view only orchestrates *when* it runs.

</details>

AustinChangLinksys (commented):
🤖 **Automated Review — Round 2** · `cbd939c..0ffc1a6` (incremental)

**Verdict: ✅ APPROVE (posted as COMMENT — GitHub blocks self-approve via API)** — Gateway focus-loss tests added; Round 1 suggestion fully resolved. Two new maintainability warnings on the test helper design, both single-reviewer and non-blocking.

#### Standards

|---|---|---|---|
| ⚠️ | 🟢High | `static_route_dialog_test.dart:122` | `_typeGatewayOctet` uses positional `boxes.at(9)` arithmetic instead of the per-octet `identifier` API (`static-route-gateway-octet-1`) already provided by `AppIpv4TextField` |
| ⚠️ | 🟢High | `static_route_dialog_test.dart:94–129` | [Shotgun Surgery] `_typeGatewayOctet` (positional + count assert) and `_nameTextField()` (raw `.first`) use two different locator patterns — a layout change requires updates in 3 places |
| 💡 | 🟢High | `static_route_dialog_test.dart:196` | Test comment says "outside the LAN subnet" but `'1...'` actually triggers format validation (`invalidIpAddress`), not subnet validation — comment-only, test behavior is correct |
| 💡 | 🟡Med | `static_route_dialog_test.dart` | dest/subnet `onFocusChanged` path still has no independent mid-edit / blur-after tests (both share `_onIpv4FocusChanged`; gateway now covered) |
| 💡 | ⚪Low | `static_route_dialog_test.dart:212` | Second gateway test's correctness implicitly depends on single-char input `'1'`; three-digit input would trigger auto-advance, potentially shifting focus before `_blur()` — worth a comment |

#### Spec

|---|---|---|---|
| ✅ | 🟢High | `static_route_dialog_test.dart:197–203` | S5/gateway (Round 1 ⚠️ resolved): `typing invalid gateway mid-edit → errorText isNull` — focused path correctly guarded |
| ✅ | 🟢High | `static_route_dialog_test.dart:205–222` | S5/gateway (Round 1 ⚠️ resolved): `blur → errorText isNotNull` — focus-loss validation confirmed; discriminating power verified by base-swap |
| ✅ | 🟢High | `static_route_dialog.dart:120–126, 179, 188, 197` | S1–S4: All four `onChanged` handlers and all three `onFocusChanged` wires unchanged from Round 1 — still correct |
| 💡 | 🟢High | `static_route_dialog_test.dart` | S2b/dest, S2b/subnet: dest and subnet `onFocusChanged` path still without independent mid-edit / blur tests (same handler; gateway coverage provides indirect protection) |
| 💡 | 🟡Med | `static_route_dialog_test.dart` | edit-mode (`route:` param) remains entirely untested — pre-existing, noted in Round 1, not addressed this round |

<details>

**W-1 · `_typeGatewayOctet` uses positional index instead of identifier** (🟢High — Reviewer B only)
- `static_route_dialog_test.dart:122` (`boxes.at(1 + 2 * 4 + octet)`)
- Code:
  ```dart
  final boxes = find.byType(TextField);
  expect(boxes.evaluate().length, 1 + 3 * 4, reason: '...');
  await tester.enterText(boxes.at(1 + 2 * 4 + octet), text);
  ```
- `AppIpv4TextField._buildSegment()` already emits a per-octet identifier (`'${widget.identifier}-octet-${index+1}'`); the dialog passes `identifier: 'static-route-gateway'` (`static_route_dialog.dart:192`). This means `'static-route-gateway-octet-1'` is available as a semantic anchor. The count assertion makes layout changes loud rather than silent — that's good — but positional arithmetic still couples the test to tree order rather than widget intent.
- **Verdict:** Comment-only (single-reviewer Warning, non-blocking). Consider using `find.semantics(label: 'static-route-gateway-octet-1')` or equivalent in a follow-up.
- **Fix:** Replace `boxes.at(1 + 2 * 4 + octet)` with an identifier-based finder; remove count assertion (now unnecessary).

**W-2 · [Shotgun Surgery] Two locator patterns in one file** (🟢High — Reviewer B only)
- `static_route_dialog_test.dart:94` (`_nameTextField()` uses `find.byType(TextField).first`) and `static_route_dialog_test.dart:110–129` (`_typeGatewayOctet` uses count + positional)
- Code:
  ```dart
  // existing (line 94)
  Finder _nameTextField() => find.byType(TextField).first;
  // new (line 122)
  await tester.enterText(boxes.at(1 + 2 * 4 + octet), text);
  ```
- If the dialog layout changes (e.g., a metric field added before dest), `_nameTextField().first`, the gateway count assert, and the gateway index all need updating independently. Fowler Shotgun Surgery: one change, three files of pain.
- **Verdict:** Comment-only (single-reviewer Warning, non-blocking). A follow-up unifying both to identifier-based finders would fix both W-1 and W-2 together.
- **Fix:** Unify to identifier-based locators for all four field types; remove positional arithmetic entirely.

</details>

<details>

- **Round 1 gateway suggestion fully resolved:** Both new tests — `'typing an invalid gateway does NOT set errorText mid-edit'` and `'gateway error surfaces only after the IPv4 field loses focus'` — exactly cover the primary reproduction scenario from Issue #1332 (octet 0 typed `'1'` → `'1...'` fails format validation, triggering the bug path).
- **Discriminating power base-swap verified:** Author confirmed both new tests fail on unfixed base `c45ee12b` and pass on the fix — this is strong evidence they are guarding the regression rather than rubber-stamping current behavior.
- **`_typeGatewayOctet` count assertion:** `expect(boxes.evaluate().length, 1 + 3 * 4)` makes layout drift loud rather than silent — better than a bare hardcoded index.
- **Focus semantics correct:** `tester.enterText()` calls `showKeyboard()` without unfocusing, so focus remains on the octet box after the call; `pumpAndSettle()` does not trigger `_onIpv4FocusChanged`'s blur path (only fires on `hasFocus=false`). The focus-loss simulation via `FocusManager.instance.primaryFocus?.unfocus()` in `_blur()` correctly chains through `AppIpv4TextField`'s `addPostFrameCallback` to `_onIpv4FocusChanged(null, false)` — verified against ui_kit v2.38.0 source.
- **dart format commit:** Second commit cleanly explains the CI format gate failure and corrects it without any logic change; the commit message makes this clear.
- **Spec S1–S5 still fully met:** Implementation files unchanged; all four `onChanged` rewired handlers and all three `onFocusChanged` wires verified intact from Round 1.
- **Tight scope:** Only one test file changed; zero production code touched in Round 2.

</details>

Issue #1350, opened 2026-08-21
fix(ai): an un-encodable tool result permanently disables the chat session


HankYuLinksys:
## Background

`getDeviceAnalytics` returned `signalLevelDistribution` as a `Map<int, int>`. JSON object keys must be strings, so `jsonEncode` threw on it, surfacing in the UI as:

> An unexpected error occurred: Converting object to an encodable object failed: Instance of 'IdentityMap<int, int>'

**That specific bug is already fixed** (commit f3428c65 on `fix/ai-tool-result-json-encodable`): the levels are now keyed by the labels the dashboard already uses, and tests assert that every command's result encodes with string keys throughout. All 15 read commands were audited; none of the others currently carry an un-encodable value.

**What this issue tracks is the mechanism, which is unchanged.** Removing the one bad value does not stop the next one from doing the same thing.

## The problem

Encoding does not happen when the tool runs. It happens later, while building the outgoing request:

```
tool executes                  → result held as-is in ToolResultPart (no encoding, no error)
_completeBatch()               → ChatMessage.toolResults(...) appended to conversation history
loop rebuilds the request      → ToolResultPart.toMap() → jsonEncode(result) → throws
```

By the time it throws, the offending entry is **already in the conversation history**. Every subsequent request re-serialises that same history and hits the same throw, so the session never replies again — the user sends messages and gets nothing back. Reproduced: after the error, two follow-up messages received no response at all.

The only recovery is the clear-conversation button (`router_assistant_view.dart:575`), which is not discoverable as the remedy for this symptom.

Two properties make this worse than a normal tool failure:

1. **The blast radius is the whole session, not the one answer.** A tool that fails should cost one reply.
2. **It is silent until it is fatal.** Nothing validates a tool result on the way into history, and a tool whose provider has no data returns an empty map that encodes cleanly — so a bad value can sit in a code path for a long time before the day it has data.

Note this is a **different failure mode from #1191** (orphaned `tool_use` when parallel tools mix with confirmation-required commands). That one was about which results reached the batch; this one is about a result that reaches the batch intact and cannot be serialised. #1191's batching fix does not cover it.

## Goal

An un-encodable tool result costs at most one answer, and the session stays usable.

## Proposed direction

Two layers, either of which alone is an improvement:

**A. Validate at the boundary (in this repo, cheap)**
[...]

PR #1352, opened 2026-08-21
fix(ai): key signal levels by label so a tool result stays encodable

HankYuLinksys:
## What broke

Asking the assistant how many devices are on WiFi vs Ethernet produced:

> An unexpected error occurred: Converting object to an encodable object failed: Instance of 'IdentityMap<int, int>'

…and then the chat stopped answering entirely. Two follow-up messages got no response.

`getDeviceAnalytics` passed `signalLevelDistribution` straight through, and that map is keyed by `int`. JSON object keys must be strings, so `jsonEncode` threw on it.

The reason it took the whole session down rather than one answer: encoding happens on the way out to the model, **after** the result is already in the conversation history. Every later request re-serialised the same entry and hit the same throw, leaving the session permanently unable to reply — recoverable only via the clear-conversation button.

## What this PR changes

Signal levels are now keyed by the labels the dashboard and statistics pages already show for this same data (`excellent`/`good`/`fair`/`poor`, matching the documented `3/2/1/0` convention in `usp_device_analytics_card.dart`).

That fixes the encoding failure and settles a second problem at the same time: the bare `0`–`3` never told the model which end of the range was good, so it had to guess. The labels carry it.

Unrecognised levels are kept as `level<n>` rather than dropped — silently losing devices from a distribution would misreport its total.

## Audit

All 15 read commands were checked for values that cannot survive `jsonEncode`. `getDeviceAnalytics` was the only one affected; the rest carry scalars, `List<String>`, `List<int>`, `List<double>` or `Map<String, *>`. Also confirmed `currentLocalTime`, `dnsServers` and `supportedStandards` are `String` (not `DateTime`), and the `*Formatted` accessors all return `String`.

Verified there is no coupling to the old shape: `signalLevelDistribution` appears nowhere in `lib/ai/prompts/`, `lib/ai/registry/` or `lib/ai/components/`, so no prompt or A2UI component documented the `int` keys.

## Tests

Three assertions, all in `usp_command_provider_test.dart`:

1. `getDeviceAnalytics` encodes, and its keys are the labels
2. An unrecognised level is kept, not dropped
3. Every command's result encodes, with string keys at every depth

**Verified by mutation** — reintroducing the raw map makes all three fail, with the same error class seen in the app.

The three timer-backed providers (`uspDeviceAnalyticsProvider`, `uspSystemMonitorProvider`, `uspTrafficAnalysisProvider`) are now stubbed in the shared container. Two reasons: they supply the populated results assertion 3 needs (unstubbed, those tools return "no data" and the assertion passes over them without seeing real data), and they keep a unit test from depending on `appConnectionStateProvider` not being `authenticated` — the only thing currently stopping the real notifiers from opening a `Timer.periodic` that fetches over USP.

## Verification

- Affected tests (direct + dependents): 27 pass. Wider `test/ai/` + `test/page/ai_assistant/`: 156 pass
- `flutter analyze` — no issues in either changed file
- `dart format` / `fvm dart format` — clean
- Verified on hardware: the question that triggered the error now renders correctly

## Follow-up

This removes the one bad value; **the mechanism that let it disable the session is unchanged**, and any future tool returning a non-encodable value would repeat it. Tracked in #1350, including why #1191's batching fix does not cover this failure mode.

### DIFF
--- a/lib/ai/providers/usp_command_provider.dart
+++ b/lib/ai/providers/usp_command_provider.dart
@@ -756,10 +756,43 @@ class UspCommandProvider implements IRouterCommandProvider {
         'totalCount': distribution.totalCount,
       },
       'bandDistribution': distribution.bandDistribution,
-      'signalLevelDistribution': distribution.signalLevelDistribution,
+      'signalLevelDistribution':
+          _labelSignalLevels(distribution.signalLevelDistribution),
       'bandSignalQuality': distribution.bandSignalQuality,
       'hourlyHistoryCount': data.hourlyHistory.length,
     });
+  }
+
+  /// Signal level label for each numeric level, matching the labels the
+  /// dashboard and statistics pages already show for the same data.
+  static const _signalLevelLabels = {
+    3: 'excellent',
+    2: 'good',
+    1: 'fair',
+    0: 'poor',
+  };
+
+  /// Re-keys the signal level distribution from `int` to `String`.
+  ///
+  /// Two reasons, both load-bearing:
+  ///
+  /// 1. **JSON keys must be strings.** A tool result travels to the model as
+  ///    `jsonEncode(result)`, and an `int`-keyed map throws there. That throw
+  ///    lands after the result is already in the conversation history, so every
+  ///    later request re-encodes the same bad entry and fails — one unusable
+  ///    map permanently bricks the session rather than failing one answer.
+  /// 2. **The numbers alone don't say which end is good.** Sending `0`–`3`
+  ///    leaves the model to guess the direction; the labels carry it.
+  ///
+  /// Unrecognised levels are kept under `level<n>` rather than dropped: losing
+  /// devices from a distribution silently would misreport totals.
+  static Map<String, int> _labelSignalLevels(Map<int, int> byLevel) {
+    return byLevel.map(
+      (level, count) => MapEntry(
+        _signalLevelLabels[level] ?? 'level$level',
+        count,
+      ),
+    );
   }
 
   Future<RouterCommandResult> _getWifiStatus() async {
--- a/test/ai/providers/usp_command_provider_test.dart
+++ b/test/ai/providers/usp_command_provider_test.dart
@@ -1,8 +1,16 @@
+import 'dart:convert';
+
 import 'package:flutter_riverpod/flutter_riverpod.dart';
 import 'package:flutter_test/flutter_test.dart';
 import 'package:privacy_gui/ai/abstraction/_abstraction.dart';
 import 'package:privacy_gui/ai/providers/usp_command_provider.dart';
 import 'package:privacy_gui/page/_shared/models/client_device.dart';
+import 'package:privacy_gui/page/_shared/models/device_analytics_state.dart';
+import 'package:privacy_gui/page/_shared/models/system_monitor_state.dart';
+import 'package:privacy_gui/page/_shared/models/traffic_analysis_state.dart';
+import 'package:privacy_gui/page/_shared/providers/usp_device_analytics_notifier.dart';
+import 'package:privacy_gui/page/_shared/providers/usp_system_monitor_notifier.dart';
+import 'package:privacy_gui/page/_shared/providers/usp_traffic_analysis_notifier.dart';
 import 'package:privacy_gui/page/_shared/models/mesh_network.dart';
 import 'package:privacy_gui/page/_shared/models/node_entity.dart';
 import 'package:privacy_gui/page/_shared/models/system_info_ui_model.dart';
@@ -37,6 +45,27 @@ void main() {
           ),
           wanDataProvider.overrideWith(
             () => _FixedWanNotifier(_testWanData),
+          ),
+          // These three are stubbed for two reasons beyond supplying data.
+          //
+          // Their real notifiers open a `Timer.periodic` that fetches over USP,
+          // guarded only by `appConnectionStateProvider` not being
+          // `authenticated` — which happens to hold in tests but is nothing this
+          // file states or controls. Stubbing keeps a unit test from depending
+          // on that, and from starting timers if it ever stops holding.
+          //
+          // They also carry the tools most likely to reach the model with an
+          // un-encodable value: histories and distributions rather than flat
+          // scalars. Left unstubbed they return "no data" and the encodability
+          // test below passes over them without ever seeing a populated result.
+          uspDeviceAnalyticsProvider.overrideWith(
+            () => _FixedDeviceAnalyticsNotifier(_testDeviceAnalyticsState),
+          ),
+          uspSystemMonitorProvider.overrideWith(
+            () => _FixedSystemMonitorNotifier(_testSystemMonitorState),
+          ),
+          uspTrafficAnalysisProvider.overrideWith(
+            () => _FixedTrafficAnalysisNotifier(_testTrafficAnalysisState),
           ),
         ],
       );
@@ -179,6 +208,71 @@ void main() {
           () => provider.execute('unknownCommand', {}),
           throwsA(isA<UnauthorizedCommandException>()),
         );
+      });
+    });
+
+    // Every result here is handed to `jsonEncode` on its way to the model. An
+    // un-encodable value does not merely spoil one answer: the throw happens
+    // after the result is already in the conversation history, so each later
+    // request re-encodes the same entry and fails too, leaving the session
+    // permanently unable to reply. These tests exist to keep that class of
+    // value out of tool results.
+    group('tool results are JSON-encodable', () {
+      test('getDeviceAnalytics keys signal levels by label, not by int',
+          () async {
+        final provider = container.read(_testUspCommandProvider);
+
+        final result = await provider.execute('getDeviceAnalytics', {});
+
+        expect(result.success, isTrue);
+        // The assertion that matters: an int-keyed map throws here.
+        expect(() => jsonEncode(result.data), returnsNormally);
+        expect(result.data['signalLevelDistribution'], {
+          'excellent': 4,
+          'good': 2,
+          'fair': 1,
+        });
+      });
+
+      test('unrecognised signal levels are kept rather than dropped', () async {
+        final analyticsContainer = ProviderContainer(
+          overrides: [
+            uspDeviceAnalyticsProvider.overrideWith(
+              () => _FixedDeviceAnalyticsNotifier(
+                const DeviceAnalyticsState(
+                  current: DeviceDistribution(
+                    signalLevelDistribution: {7: 3},
+                  ),
+                ),
+              ),
+            ),
+          ],
+        );
+        addTearDown(analyticsContainer.dispose);
+        final provider = analyticsContainer.read(_testUspCommandProvider);
+
+        final result = await provider.execute('getDeviceAnalytics', {});
+
+        expect(() => jsonEncode(result.data), returnsNormally);
+        expect(result.data['signalLevelDistribution'], {'level7': 3});
+      });
+
+      // Covers every command at once. Commands whose providers are not stubbed
+      // return a failure result, which is still encoded and sent, so they are
+      // worth asserting on even though they carry no data.
+      test('every command produces an encodable result', () async {
+        final provider = container.read(_testUspCommandProvider);
+        final commands = await provider.listCommands();
+
+        for (final cmd in commands) {
+          final result = await provider.execute(cmd.name, {});
+          expect(
+            () => jsonEncode(result.data),
+            returnsNormally,
+            reason: '${cmd.name} returned a result that cannot be encoded',
+          );
+          _expectStringKeysThroughout(result.data, cmd.name);
+        }
       });
     });
 
@@ -326,8 +420,60 @@ void main() {
 }
 
 // =============================================================================
+// Helpers
+// =============================================================================
+
+/// Fails if any map nested anywhere under [value] is keyed by something other
+/// than a `String`.
+///
+/// `jsonEncode` already rejects those keys, so this adds nothing for a result
+/// that is fully populated. It earns its place on the sparse ones: a tool whose
+/// provider holds no data returns an empty map, which encodes cleanly and would
+/// let a non-string key pass unnoticed until the day that tool has data.
+void _expectStringKeysThroughout(Object? value, String commandName) {
+  if (value is Map) {
+    for (final entry in value.entries) {
+      expect(
+        entry.key,
+        isA<String>(),
+        reason: '$commandName has a non-String map key: ${entry.key}',
+      );
+      _expectStringKeysThroughout(entry.value, commandName);
+    }
+  } else if (value is Iterable) {
+    for (final element in value) {
+      _expectStringKeysThroughout(element, commandName);
+    }
+  }
+}
+
+// =============================================================================
 // Fixed Notifiers (set state immediately in build)
 // =============================================================================
+
+class _FixedDeviceAnalyticsNotifier extends UspDeviceAnalyticsNotifier {
+  final DeviceAnalyticsState _data;
+  _FixedDeviceAnalyticsNotifier(this._data);
+
+  @override
+  DeviceAnalyticsState build() => _data;
+}
+
+class _FixedSystemMonitorNotifier extends UspSystemMonitorNotifier {
+  final SystemMonitorState _data;
+  _FixedSystemMonitorNotifier(this._data);
+
+  @override
+  SystemMonitorState build() => _data;
+}
+
+class _FixedTrafficAnalysisNotifier extends UspTrafficAnalysisNotifier {
+  final TrafficAnalysisState _data;
+  _FixedTrafficAnalysisNotifier(this._data);
+
+  @override
+  TrafficAnalysisState build() => _data;
+}
 
 class _FixedSystemInfoNotifier extends SystemInfoDataNotifier {
   final SystemInfoData _data;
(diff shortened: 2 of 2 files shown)


AustinChangLinksys (approved):
🤖 **Automated Review — Round 1** · `251d2b8..f3428c6` (full)

**Verdict: ✅ APPROVE** — Correct fix, solid test coverage; two warnings worth tracking.

#### Standards

|---|---|---|---|
| ⚠️ | 🟢High | `usp_command_provider.dart:768–773` | [Shotgun Surgery] `_signalLevelLabels` is the 4th int→label mapping in the codebase; next convention change requires 4-site edit |
| ⚠️ | 🟢High | `usp_command_provider_test.dart:238–252` | `analyticsContainer` has no parent/overrides for systemInfo/devices/wifi/wan providers — if provider returns failure the unrecognised-level assertion passes vacuously |
| 💡 | 🟡Med | `usp_command_provider.dart:768` | `_signalLevelLabels` declared after its call site (line 760) and `_labelSignalLevels` (line 789) — inconsistent with codebase convention of declaring statics at the top of the class |
| 💡 | ⚪Low | `usp_command_provider.dart:791–793` | Negative int keys (e.g. from unexpected firmware) would produce `level-1`, which a language model may interpret as "below poor" — defensive annotation only |
| 💡 | ⚪Low | `usp_command_provider_test.dart:263–276` | `_expectStringKeysThroughout` is a no-op on empty maps — commands whose providers are not stubbed yield `{}` and pass without key-depth checks |

#### Spec

|---|---|---|---|
| ⚠️ | 🟡Med | `usp_command_provider_test.dart:237–258` | spec#test-2 → `analyticsContainer` missing parent overrides: if external provider state changes, unrecognised-level test may pass vacuously (failure result, empty `data`, no key to check) |
| 💡 | ⚪Low | `usp_command_provider_test.dart:263–276` | spec#test-3 → `_expectStringKeysThroughout({})` passes without verifying key types for un-stubbed commands; known tradeoff, no blocking concern |

<details>

**W-1 — [Shotgun Surgery] Duplicated int→label mapping** 🟢High  
`lib/ai/providers/usp_command_provider.dart:768–773` — new `_signalLevelLabels` is the fourth place in the codebase encoding the same `{3→excellent, 2→good, 1→fair, 0→poor}` convention:
```dart
static const _signalLevelLabels = {
  3: 'excellent',
  2: 'good',
  1: 'fair',
  0: 'poor',
};
```
Three prior sites: `wifi_connection_info.dart:47–51` (via `NodeSignalLevel`), `usp_device_analytics_card.dart:278–281`, `stats_signal_quality_section.dart:104–107`. This PR already required touching three files to propagate the fix; adding a fourth lookup deepens the shotgun-surgery surface.  
Fix: add a `label` getter to `NodeSignalLevel` (or a `toLabelledMap()` utility on `DeviceDistribution`) so the mapping is maintained in one place. Not required for this PR, but a good candidate for a follow-up cleanup issue.

**W-2 — `analyticsContainer` isolation gap (test robustness)** 🟢High  
`test/ai/providers/usp_command_provider_test.dart:238–252` — the container used by the "unrecognised signal levels" test overrides only `uspDeviceAnalyticsProvider`:
```dart
final analyticsContainer = ProviderContainer(
  overrides: [
    uspDeviceAnalyticsProvider.overrideWith(
      () => _FixedDeviceAnalyticsNotifier(
        const DeviceAnalyticsState(
          current: DeviceDistribution(
            signalLevelDistribution: {7: 3},
          ),
        ),
      ),
    ),
  ],
);
```
No `parent: container` and no overrides for `systemInfoDataProvider`, `devicesDataProvider`, `wifiDataProvider`, `wanDataProvider`. The test currently passes because `_getDeviceAnalytics` only reads `uspDeviceAnalyticsProvider` — but if environmental assumptions shift (e.g. `appConnectionStateProvider` becomes `authenticated`), un-stubbed providers may return failure, `data` becomes `{}`, and `_expectStringKeysThroughout({})` passes without verifying any key. The core spec requirement — "unrecognised level is kept, not dropped" — would be vacuously satisfied.  
Fix: pass `parent: container` to `analyticsContainer`, or explicitly override the other four providers. This is also noted in the Spec axis.

**Spec W-1 — same as Standards W-2 above** (see W-2 for full evidence). The Spec axis flags this because spec#test-2 explicitly requires verifying `level<n>` retention.

</details>

<details>

- **Core fix is correct and complete.** `_labelSignalLevels` converts `Map<int, int> → Map<String, int>` before the result enters `ToolResultPart.toMap()` → `jsonEncode()`. This directly resolves the session-bricking failure mode: no more `JsonUnsupportedObjectError` on subsequent requests.
- **Fallback design is sound.** `'level$level'` for unrecognised values preserves the device count in the distribution rather than silently dropping it, which would misreport totals — the right call.
- **Labels match the existing dashboard convention.** `3→excellent / 2→good / 1→fair / 0→poor` aligns with `usp_device_analytics_card.dart:278–281`, so the AI model receives the same vocabulary the UI already uses.
- **Audit is credible.** Checked: `signalLevelDistribution` is absent from `lib/ai/prompts/`, `lib/ai/registry/`, and `lib/ai/components/` — no coupling to the old int-key shape.
- **`_expectStringKeysThroughout` is a strong defensive guard.** Recursive key-type check across all 15 command results provides a broad regression net for future tool additions.
- **Timer-backed provider stubs are well documented.** The test comment (lines 49–60) explains exactly why `uspDeviceAnalyticsProvider`, `uspSystemMonitorProvider`, and `uspTrafficAnalysisProvider` are stubbed: (a) they supply the populated data the assertions need, and (b) they prevent the test from accidentally opening `Timer.periodic` USP connections.
- **Mutation verification documented.** PR description confirms all three tests fail when the raw `Map<int, int>` is reintroduced — the expected error class was observed, confirming the tests are non-vacuous for the main path.
- **flutter analyze and dart format clean.** No issues introduced in either changed file.

</details>

PeterJhongLinksys (approved):
🤖 **Automated Review — Round 1** · `251d2b8..f3428c6` (full)

**Verdict: ✅ APPROVE** — A small, correct, unusually well-documented fix: the un-encodable `int`-keyed map is removed and the labels also answer "which end is good". Both reviewers found 0 Critical. The recurring theme across every Warning is *placement, not correctness* — the invariant "everything in `RouterCommandResult.data` must be JSON-encodable" is enforced in 1 of ~20 command builders, so this PR fixes the bug but leaves the defect class open (already acknowledged as follow-up #1350).

#### Standards

|---|---|---|---|
| ⚠️ | 🟢 | `lib/ai/abstraction/router_command.dart:48` | [Shotgun Surgery] `data` stays `Map<String, dynamic>` and neither factory checks anything — the next builder passing a domain map straight through reintroduces the same session-bricking bug. |
| ⚠️ | 🟢 | `lib/ai/providers/usp_command_provider.dart:768-773` | [Duplicated Knowledge] `3 → 'excellent'` now exists in three files across three layers; the AI-facing copy is the only one with no UI to reveal a drift. |
| ⚠️ | 🟢 | `test/ai/providers/usp_command_provider_test.dart:270-287` | The `every command produces an encodable result` loop passes trivially for the ~8 unstubbed commands (`data = const {}`), so it reads as whole-surface coverage while guarding almost nothing. |
| ⚠️ | 🟢 | `lib/ai/providers/usp_command_provider.dart:786-795` | `level<n>` fallback is device-driven and reachable, and is indistinguishable from a real quality label from the model's side. |
| ⚠️ | 🟡 | `test/ai/providers/usp_command_provider_test.dart:270-287` | Tests encode `result.data`; production encodes the *wrapper* (`{'status':…,'result':…}` / `{'error':…}`) — a good proxy, but a regression in the wrapping layer is invisible to this suite. |
| ⚠️ | 🟡 | `lib/ai/providers/usp_command_provider.dart:768-773` | Repo already has ≥3 tier vocabularies with these member names (`HealthTier`, `SignalTier`, `DeviceSignalLevel`); a bare `Map<int, String>` is a fourth model with no note on why none fit. |
| ⚠️ | 🟡 | `test/ai/providers/usp_command_provider_test.dart:649-660` | Fixture omits level `0`, so `0 → 'poor'` — the entry most likely to be inverted — is never exercised; an inverted `0`/`3` pair would still pass. |
| ⚠️ | 🟡 | `test/ai/providers/usp_command_provider_test.dart` (`_Fixed*Notifier`) | Stubs `extend` the production notifiers and override `build()` only; per Art. VIII/X convention is Mocktail `implements`. Follows local precedent (`_FixedWanNotifier`). |
| ⚠️ | 🟢 | `test/ai/providers/usp_command_provider_test.dart` (`_test*State`) | Inline fixtures instead of a Test Data Builder (Art. VIII/X); the second test must hand-build a whole second container for lack of `copyWith()`. Pre-existing file debt, extended. |
| ⚠️ | ⚪ | `lib/ai/providers/usp_command_provider.dart:777-784` | Doc comment names `jsonEncode(result)` as the throw site; that call is not in `lib/` — see the correction note below. |
| 💡 | 🟢 | `lib/ai/abstraction/router_command.dart:59-61` | Promote the new test's traversal from *checker* to *enforcer*: a debug `assert(_isJsonEncodable(data))` in `.success`, or one sanitiser at the encode boundary, retires the class. |
| 💡 | 🟢 | `lib/ai/providers/usp_command_provider.dart:790-794` | One `_log` on the unmapped branch turns a silent degradation into a diagnosable one (`_log` is used liberally in this file already). |
| 💡 | 🟡 | `test/ai/providers/usp_command_provider_test.dart:432-448` | `_expectStringKeysThroughout` is reusable infrastructure; a shared `isJsonEncodable` Matcher would pre-empt the copy-paste when a second command provider lands. |
| 💡 | 🟡 | `lib/ai/providers/usp_command_provider.dart:766-798` | Helper + `const` map are spliced mid-way through the otherwise unbroken run of `_get*` handlers. |
| 💡 | 🟡 | `test/ai/providers/usp_command_provider_test.dart:270-287` | Loop never asserts `result.error` is non-empty on failure — the other half of the "what reaches the model" contract. |
| 💡 | ⚪ | `lib/ai/providers/usp_command_provider.dart:131-137` | A clause in the tool *description* ("keyed by quality label, best to worst") removes the last inference and lets the model recognise `level<n>` as out-of-vocabulary. |
| 💡 | ⚪ | `lib/ai/abstraction/router_command.dart:47-48` | The excellent 18-line DartDoc sits on a *private* member; one line on `data` reaches the audience that needs it — whoever writes the next map. |
| 💡 | ⚪ | `lib/ai/providers/usp_command_provider.dart:789-796` | `Map.map` preserves device-iteration order; iterating `_signalLevelLabels` instead would make the payload deterministic and diffable. |

#### Spec

_No linked issue — spec axis skipped this round._

(#1350 is referenced in the body as explicitly **out of scope** — it tracks the unchanged mechanism — and #1191 is a closed, unrelated PR reference. Neither is an acceptance criterion for this PR, so no spec grading was performed.)

<sub>Confidence: 🟢 High · 🟡 Medium · ⚪ Low — low-confidence items are kept, not dropped.</sub>
<sub>`[X]` = raised by 1 agent, `[XX]` = independently raised by 2 (Standards table only).</sub>

---

<details>
<summary>⚠️ Standards Warnings — evidence</summary>

**Both reviewers independently raised the top two findings (W1/W2 below).**

### The invariant has no enforcement point — `router_command.dart:48`

```dart
/// Result data from the command
final Map<String, dynamic> data;
...
factory RouterCommandResult.success(Map<String, dynamic> data) {
  return RouterCommandResult(success: true, data: data);
}
```

`dynamic` values and `Map<int, …>` sub-maps stay type-legal, and neither factory validates. The commit message argues correctly that this is a *class invariant of the boundary type* — encoding happens after the entry is already in conversation history, so one bad value bricks the session — yet it is enforced in one of ~20 builders. Confirmed still-live candidates: `Map<int, int> signalLevelDistribution` (`device_analytics_state.dart:16`) and `Map<TrafficInterface, InterfaceTrafficSnapshot> interfaces`, read at `usp_command_provider.dart:657` and `668` — safe today only because they are indexed rather than spread. Suggested fix in the Suggestions rows above.

### Third hard-coded copy of the same vocabulary — `usp_command_provider.dart:768-773`

```dart
static const _signalLevelLabels = {
  3: 'excellent', 2: 'good', 1: 'fair', 0: 'poor',
};
```

The DartDoc's claim that these "match the labels the dashboard and statistics pages already show" is **true today** — verified against `usp_device_analytics_card.dart:277-283` and `stats_signal_quality_section.dart:104-107` — but it matches by hand, not by construction. The UI sites iterate explicit tuple lists; this one is a lookup with a swallow-all fallback, so a future renamed bucket or extended range breaks only the AI path, silently. Suggested home for a single source of truth: beside `signalLevelDistribution` in `device_analytics_state.dart:16`, sharing the *level→key* pair (the UI still localises).

### The all-commands test is vacuous for most commands — `usp_command_provider_test.dart:270-287`

```dart
for (final cmd in commands) {
  final result = await provider.execute(cmd.name, {});
  expect(() => jsonEncode(result.data), returnsNormally, …);
  _expectStringKeysThroughout(result.data, cmd.name);
}
```

`listCommands()` returns 15 commands (`usp_command_provider.dart:39-138`); the container overrides 7 providers (`…_test.dart:37-70`). For an unstubbed command `RouterCommandResult.failure` leaves `data = const {}` (`router_command.dart:63-65`), so `jsonEncode({})` and a zero-iteration key walk both pass trivially. A future int-keyed map in e.g. `getLanInfo` would sail through green. The test comment is admirably honest about this and the three new stubs were added precisely to widen it — the ask is to make the test *enforce* that honesty (assert a minimum populated count, or an explicit `_knownUnstubbedCommands` set) so coverage cannot decay to zero without a red test.

### `level<n>` fallback reads as a ranked bucket — `usp_command_provider.dart:786-795`

Keeping unrecognised levels rather than dropping them is the right call, and the stated rationale (a dropped device silently misreports the distribution total) is the harder failure to diagnose. Reachability traced to source: `usp_device_analytics_notifier.dart:203-208` bins by `d.signalLevel` with **no clamping**, and `client_device.dart:196` is `int get signalLevel => wifi?.signalLevel ?? 0;` — an unvalidated integer from the router's USP payload, not a bounded enum. The PR's own test exercises `{7: 3}`, so the author agrees the path is live. Residual issue is semantic: given `{'excellent': 4, 'level7': 3}` the model cannot tell whether `level7` beats `excellent`, and will most likely narrate it as a real quality class. Consider `'unknownLevel$level'`, or a single sibling key (`'unknownLevels': {'7': 3}`) that keeps totals intact.

</details>

<details>
<summary>📝 Correction — one reviewer finding was verified down by the synthesizer</summary>

Reviewer A flagged (🟡Med) that the doc comment at `usp_command_provider.dart:777-784` names `jsonEncode(result)` as the throw site, but no such call exists in `lib/ai/` — the only `jsonEncode` hit in that directory is the comment itself — and `grep -rn "class ToolResultPart" .` returns 0 hits, leaving the design premise unverifiable.

**Both observations reproduce, but the conclusion is softened.** `ToolResultPart` is not missing: it comes from the `generative_ui` package, an external git dependency (`pubspec.yaml:63-67`, `linksys/privacyGUI-UI-kit.git @ v2.39.0`, path `generative_ui`), imported at `router_chat_controller.dart:3`. It is not resolvable in this worktree, so the encode site genuinely cannot be confirmed from this repo alone — but the premise is verifiable in principle rather than unfounded, and the symptom quoted in the PR body (`Instance of 'IdentityMap<int, int>'`) is exactly what a `jsonEncode` of an int-keyed map produces. Kept as ⚪Low and reframed as a documentation-precision request: point the comment at the package symbol that performs the encode, so a future reader is not sent looking inside `lib/`.

Re-keying to `String` is correct and desirable regardless of which layer throws.

</details>

<details>

- **Root cause named, not just the symptom** (`usp_command_provider.dart:775-785`). The comment explains why the blast radius is the *whole session* — the throw lands after the entry is in conversation history, so every later request re-fails. That escalation from "one bad answer" to "unrecoverable without clearing the conversation" is the reason the fix matters, and it is written where the next reader will find it.
- **Verified by falsification.** Reintroducing the raw map makes all three new tests fail with the same error class seen in the app. That is the single most valuable line in the commit message.
- **The non-obvious decision is defended, not just made.** Keeping unrecognised levels as `level<n>` *because* dropping them would make a distribution silently misreport its total — exactly the reasoning that normally gets lost.
- **Second problem settled in passing.** The bare `0`–`3` never told the model which end was good; the labels carry it, and they match the words a user already sees on the dashboard, so chat and UI agree.
- **Stubbing rationale documented honestly, including its limits** (`…_test.dart:49-60`). Two independent reasons given, and it names the incidental dependency being removed (`appConnectionStateProvider` being the only thing stopping a real `Timer.periodic` over USP in a unit test). Rare and genuinely useful.
- **Test comments admit what the tests do not cover** — that unstubbed commands carry no data, and that `_expectStringKeysThroughout` adds nothing for a fully populated result and earns its place on the sparse ones. (The Warnings ask the suite to *enforce* that candour, not to supply it.)
- **`_expectStringKeysThroughout` recurses through both `Map` and `Iterable`** (`…_test.dart:434-447`), so nested int keys inside list elements are caught too — the case a top-level check misses.
- **Fixture fidelity + deterministic clocks.** `{3: 4, 2: 2, 1: 1}` mirrors real producer behaviour (the notifier only records levels it observed) rather than inventing a dense map, and all new timestamps are `DateTime.utc(...)`, so the fixtures cannot flake on timezone.
- **Layering and scope discipline.** No `generated/` import enters the provider layer; all four USP Provider Hard Rules and the `ServiceError` boundary are untouched (no new provider, no mutation, no `catch`). 34 production lines, one behavioural change, no drive-by refactors.
- **The 15-command audit in the PR body** — checking every read command for values that cannot survive `jsonEncode`, and confirming no prompt or A2UI component documented the old `int` keys — is the diligence that makes "this was the only one affected" believable.

</details>

PR #1353, opened 2026-08-21
test(static-routing): let the #1332 regression guard actually run in CI

AustinChangLinksys:
#1347 merged the #1332 fix together with six widget tests meant to guard it. The tests have not run once since.

The file was tagged `ui`, and `run_tests.sh` excludes that tag in all three of its `flutter test` invocations (`:80`, `:92`, `:96`) — which is the script `ci.yml:84` runs. So the guard sits in the tree, unexecuted. Delete `_onInputChanged` and put `_validate()` back in the four `onChanged` handlers, and CI stays green.

This is one line: the `@Tags(['ui'])` annotation comes off. No product code, no test logic, no other file.

## Evidence

CI's own exclude expression, before and after, on the merged file:

```
before:  No tests ran.
         No tests match the requested tag selectors:
           exclude: "golden || loc || ui"

after:   00:04 +6: All tests passed!
```

Cross-checked against the merged CI log for #1347 (run `32441744841`): the file appears **0 times** among the 265 test files that executed. The `7113 tests passed` headline on that PR made the path look covered when it was not.

## Why remove the tag instead of moving it to `dashboard-card`

`dart_test.yaml` documents `dashboard-card` as the tag deliberately left in the PR test command:

```yaml
  # failure here blocks the PR.
  dashboard-card:
```

36 files use it, so it would work. But the name describes dashboard card gates, not a static-routing dialog, and the tag taxonomy is under separate review. An untagged widget test falls into the default `flutter test` set — the behaviour wanted here, and it needs no taxonomy decision.

Worth noting: `--exclude-tags` drops a test if **any** of its tags match, so `['ui', 'dashboard-card']` would still have been excluded. The `ui` tag had to come off either way.

## This is not an isolated file

Eleven others are in the same state — **66 tests, zero CI executions between them**:

| file | tests | guards |
|---|---|---|
| `dhcp_reservation_edit_dialog_test.dart` | 7 | #1070 |
| `wifi_channel_dialog_test.dart` | 19 | #1023, #1025 |
| `usp_port_forwarding_tab_identifier_widget_test.dart` | 3 | #1172, #1218, #1246 |
| `usp_advanced_settings_identifier_widget_test.dart` | 2 | #1218 |
| `topology_node_identifier_widget_test.dart` | 2 | #1208 |
| `node_detail_popup_identifier_widget_test.dart` | 6 | #1254 |
| `service_error_view_test.dart` | 6 | — |
| `usp_network_health_card_test.dart` | 4 | #1143 |
| `usp_network_health_card_legend_test.dart` | 3 | #1145 |
| `general_settings_widget_test.dart` | 5 | — |
| `pnp_no_internet_view_test.dart` | 3 | — |

The identifier ones are the sharpest: they guard the semantics identifiers `PrivacyGUI-USP-E2E` locates elements by, referenced in **24 places** across its specs (`P15-topology` 8, `P04a` 6, `S05` 2). Drop an identifier app-side and this repo's CI passes; the breakage surfaces the next day in another repo's scheduled E2E run.

All twelve pass locally when run directly, so the tag is the only thing keeping them out. #1327 is about to add a thirteenth (`dashboard_card_semantics_absorption_test.dart`, 224 lines, 3 tests) — raised there independently as a Critical.

**This PR deliberately changes only one file** so the mechanism is proven on Linux CI before the other eleven follow. Local green on macOS is not evidence that a widget test is stable on the CI runner.

## Check when reviewing

The claim to verify is not "the tests pass" — they passed before this PR too, when run by hand. It is that they now **execute as part of the normal CI test command**. In this PR's run, `static_route_dialog_test.dart` should appear in the executed set, and the total should rise by 6 against #1347's 7113.

### DIFF
--- a/test/page/static_routing/views/dialogs/static_route_dialog_test.dart
+++ b/test/page/static_routing/views/dialogs/static_route_dialog_test.dart
@@ -1,4 +1,3 @@
-@Tags(['ui'])
 import 'package:flutter/material.dart';
 import 'package:flutter_test/flutter_test.dart';
 import 'package:go_router/go_router.dart';


AustinChangLinksys (commented):
🤖 **Automated Review — Round 1** · `251d2b8..9f06030` (full)

**Verdict: 💬 Self-review (comment only)** — GitHub self-approve is not permitted; review posted as reference only. No Critical or Warning issues found; changes are correct and targeted.

#### Standards

|---|---|---|---|
| 💡 | 🟢High | `run_tests.sh:80,92,96` | 4 other test files still carry `@Tags(['ui'])` and are silently excluded from CI (pre-existing, not introduced by this PR) |
| 💡 | 🟡Med | `dart_test.yaml` | `ui` tag semantics undocumented — why is it excluded from CI runs? Rationale absent (pre-existing) |

#### Spec

|---|---|---|---|
| ✅ | 🟢High | `static_route_dialog_test.dart:1` | `@Tags(['ui'])` removed — CI exclusion lifted, 6 tests now in default run set |
| ✅ | 🟢High | `run_tests.sh:80,92,96` | CI exclude expression `golden\|\|loc\|\|ui` confirmed; tag removal is sufficient |
| ✅ | 🟢High | diff (1 file, 1 line) | No product code touched, no test logic modified |
| 💡 | 🟢High | commit msg | 11 remaining `@Tags(['ui'])` test files (66 tests, 0 CI executions) noted in commit; no follow-up issue opened yet |

<details>
<summary>⚠️ Warnings — details</summary>

**S-1 (Standards — both agents): Other `@Tags(['ui'])` test files silently excluded from CI** 🟢High

- `run_tests.sh` lines 80, 92, 96 all use `--exclude-tags="golden||loc||ui"`.
- Reviewer A identified 4 such files still live on head; Reviewer B enumerated 11 (the larger set includes identifier widget tests that guard cross-repo E2E semantics).
- These are **pre-existing issues, not introduced by this PR**. This PR deliberately changes only one file to prove the mechanism on CI before fixing the rest.
- **Fix:** Follow-up issue to track the remaining 11 files — already mentioned in commit message but no ticket opened. Suggested by both agents.

**S-2 (Standards — Reviewer B): `dart_test.yaml` `ui` tag undocumented** 🟡Med

- `dart_test.yaml` documents `dashboard-card` with an explicit rationale comment but `ui` has none.
- The exclusion semantics are counter-intuitive (one might expect a `ui` tag to *include* UI tests, not exclude them from the standard run).
- **Fix:** Add a comment to `dart_test.yaml` explaining that `ui` marks tests needing a real rendering environment and is excluded from headless CI runs; distinguish it from `dashboard-card`.

</details>

<details>

- **Single-purpose, minimal change**: 1 file, 1 line deleted. Exactly matches the stated scope.
- **Mechanism is correct**: `run_tests.sh` exclusion of `ui` tag confirmed at three call sites (lines 80, 92, 96); removing the tag is the right and sufficient fix.
- **Test quality**: 6 `testWidgets` cover the core #1332 regression scenario (mid-edit no error text, focus-loss error surfacing, Add button disabled while invalid). Test names are descriptive and assertions are on observable outcomes.
- **Tag taxonomy decision justified**: Not re-tagging to `dashboard-card` is correct — that tag's documented semantics describe dashboard card gates, not static-routing dialogs; untagged falls into the default `flutter test` set which is exactly the desired behaviour.
- **Evidence-backed PR**: PR body and commit message include before/after CI output and cross-reference against the merged CI log for #1347 (run 32441744841).
- **Spec fully satisfied**: All three spec requirements verified — annotation removed, 6 tests added to CI run set, no product code or test logic changed.

</details>

PeterJhongLinksys (approved):
🤖 **Automated Review — Round 1** · `251d2b8..9f06030` (full)

**Verdict: ✅ APPROVE** — Correct, minimal, well-targeted one-line fix: the `@Tags(['ui'])` removal genuinely moves the six #1332 regression tests from "silently skipped in every CI run" into the blocking PR gate. No Critical on either axis. Everything below concerns the *durability* of the guard now that it blocks, not the correctness of the change.

#### Standards

|---|---|---|---|
| ⚠️ | 🟢 | `dart_test.yaml:1-9`, `README.md:129-135` | The `ui` tag carries two contradictory meanings — "is a widget test" (README's `flutter test --tags ui` suite) vs "excluded from CI" (`run_tests.sh:80,92,96`); untagging fixes the gate but silently drops the file from the documented `--tags ui` suite, and no doc/tag update ships with it. |
| ⚠️ | 🟢 | `test/.../static_route_dialog_test.dart:1` | 10 sibling widget tests remain `@Tags(['ui'])` and therefore still unexecuted, including `dhcp_reservation_edit_dialog_test.dart` — the very dialog this file's docstring (`:17-19`) cites as its pattern precedent. [Shotgun Surgery] |
| ⚠️ | 🟡 | `test/.../static_route_dialog_test.dart:113-128` | Now that the file is PR-blocking, its exact `TextField`-count assertion (`== 1 + 3*4`) and order-dependent `.first` finders (`:79-80`, `:98-101`) will redden this gate on an unrelated `ui_kit_library` bump (git dep, `pubspec.yaml:59-62`). |
| ⚠️ | 🟡 | `test/.../static_route_dialog_test.dart:150-163`, `:193-206` | The guard asserts `errorText` *timing*, not keystroke retention; #1332's actual mechanism (CanvasKit semantics `<input>` teardown) cannot reproduce under headless `flutter test`, so a rebuild-causing refactor that keeps `errorText` null still passes. |
| ⚠️ | 🟡 | `test/.../static_route_dialog_test.dart:103-106` | `_blur()` drives global `FocusManager.instance.primaryFocus?.unfocus()`; if focus is already elsewhere the `?.` makes it a silent no-op and the later assertion fails pointing at validation logic instead of "blur never happened". |
| ⚠️ | 🟡 | `test/.../static_route_dialog_test.dart:177-189` | "Add stays disabled" is asserted with `findsOneWidget` after the tap — satisfied equally by a disabled button, an enabled button whose `onTap` threw, or a failed pop. Read `AppButton.onTap`/`isEnabled` instead. |
| ⚠️ | 🟡 | PR description | A CI-gating behaviour change lands with no recorded evidence: please state why the `ui` tag was originally on this file and paste a green `sh run_tests.sh` run on this branch. If the tag was a flakiness quarantine rather than categorisation, this converts a skipped test into an intermittently-red gate. |
| 💡 | 🟢 | `test/.../static_route_dialog_test.dart:8-23` | Add one line to the existing docstring recording that the file is *intentionally* untagged, so the next contributor doing tag hygiene does not re-mute the guard. |
| 💡 | 🟡 | `run_tests.sh:80,92,96` | `--exclude-tags="golden||loc||ui"` is duplicated verbatim across three branches; extract to one variable and print an "N test files excluded by tag" line — that alone would have surfaced this bug without a dedicated PR. |
| 💡 | 🟡 | `test/.../static_route_dialog_test.dart:127` | `boxes.at(1 + 2*4 + octet)`: the count guard catches field-count changes but not a *reorder* at constant count. `identifier: 'static-route-gateway'` already exists at `lib/.../static_route_dialog.dart:193` — a keyed lookup closes the last silent-green hole. |
| 💡 | ⚪ | `test/.../static_route_dialog_test.dart:131-136` | `setUp` reloads `AppLocalizations` for all 6 cases to read one locale-invariant string; `setUpAll` suffices. |
| 💡 | ⚪ | — | The focus-loss-validation contract now exists in four dialogs (`port_forwarding`, `usp_local_network_view`, `dhcp_reservation_edit`, `static_route`); a shared `expectValidatesOnBlur(...)` helper would make it enforceable repo-wide instead of four hand-rolled copies. |
| 💡 | ⚪ | — | Optional CI step: assert the `ui`-tagged file list matches a checked-in allowlist, making "regression guard accidentally excluded" structurally impossible. |

#### Spec

Spec source: #1332

|---|---|---|---|
| ⚠️ | 🟡 | "reachable by a real user … also what currently reddens the daily E2E schedule" | [Omission] #1332 is still OPEN and nothing in the PR records that `P12-static-routing.spec.ts:155` went green; the unblocked widget tests prove the proximate cause is gone, not the user-visible symptom. |
| ⚠️ | 🟢 | "Keep 'empty field reports no error' so errors don't surface mid-fill" | [Omission] Implemented add-mode-only (`lib/.../static_route_dialog.dart:79`, `_errors = _isEdit ? _computeErrors() : {}`) and the whole guard group covers add mode only (`test:139`), so an edit-path regression would stay green. |
| ⚠️ | 🟡 | `dart_test.yaml`: "Defensive widget gates that must run in the PR test command (#1183)" | [Omission] The repo's documented idiom for "widget test that must block the PR" is a registered-but-not-excluded tag (`dashboard-card`); this file goes bare instead, leaving no marker that its CI-blocking status is deliberate. |
| 💡 | ⚪ | `onchanged_error_rebuild_audit.py` "attached in the first comment below" | The audit script that found this defect class lives in an issue comment attachment; committing it under `test_scripts/` is a natural follow-up (explicitly *not* asked for in this PR). |

<sub>Confidence: 🟢 High / 🟡 Med / ⚪ Low — low-confidence items are reported, not dropped. `[X]` = 1 agent, `[XX]` = 2 agents agree (Standards table only; the Spec axis has a single reviewer, so no such column). The two axes are independent: neither offsets the other.</sub>

<details>
<summary><b>Standards — Warning evidence</b></summary>

**W1 · The `ui` tag means two incompatible things**

CI wiring, read from the head worktree:

- `.github/workflows/ci.yml:84` → `run: sh run_tests.sh`
- `run_tests.sh:80`, `:92`, `:96` → `$FLUTTER test --exclude-tags="golden||loc||ui"`
- `dart_test.yaml` declares `ui` as a first-class tag, alongside:

```yaml
tags:
  golden:
  loc:
  ui:
  # failure here blocks the PR.
  dashboard-card:
```

But `README.md:129-135` documents a *different* entry point for the same tag:

```
### UI Widget Tests
To run tests that are specifically tagged for UI widgets (and are not screenshot tests) …
flutter test --tags ui
```

So untagging is not a no-op: the file leaves that documented suite and dissolves into the ~265-file default set. Whoever runs `flutter test --tags ui` to smoke the widget layer now silently skips the static-route dialog. That conflation — `ui` meaning both "is a widget test" and "do not run in CI" — is the actual defect; this PR patches a symptom of it. Either adopt the `dashboard-card`-style registered-but-not-excluded tag for defensive gates, or stop excluding `ui` and correct `README.md:129-135`.

**W2 · 10 siblings still muted**

After this change, `grep -rl "@Tags(\['ui'\])" test/` returns 10 files (verified in the head worktree):

```
components/styled/general_settings_widget/general_settings_widget_test.dart
components/views/service_error_view_test.dart
page/advanced_settings/usp_advanced_settings_identifier_widget_test.dart
page/dashboard/views/components/usp_network_health_card_legend_test.dart
page/dashboard/views/components/usp_network_health_card_test.dart
page/dashboard/views/dialogs/wifi_channel_dialog_test.dart
page/dhcp/views/dialogs/dhcp_reservation_edit_dialog_test.dart
page/instant_setup/views/pnp_no_internet_view_test.dart
page/port_forwarding/usp_port_forwarding_tab_identifier_widget_test.dart
page/topology/helpers/topology_node_identifier_widget_test.dart
page/topology/views/components/node_detail_popup_identifier_widget_test.dart
```

`dhcp_reservation_edit_dialog_test.dart` and `wifi_channel_dialog_test.dart` are the same shape of artefact as the file in this PR, and this file's own docstring (`:17-19`) names `dhcp_reservation_edit` as the pattern it copies. The reasoning "this regression guard must run in CI" applies to them verbatim. Fixing one instance by hand-editing one annotation means the next dialog guard falls into the same trap. Non-blocking here — the PR body already says one file goes first deliberately to prove the mechanism on Linux CI, which is sound — but the policy should be decided once in a follow-up rather than eleven more times.

**W3 · Brittle assumptions become a shared gate**

```dart
// test/.../static_route_dialog_test.dart:120-127
final boxes = find.byType(TextField);
expect(
  boxes.evaluate().length,
  1 + 3 * 4,
  reason: 'expected the name field plus three IPv4 fields of four octets; '
      'if this changed, the octet index below is wrong',
);
await tester.enterText(boxes.at(1 + 2 * 4 + octet), text);
```

The `reason:` string is good practice, and as an opt-in local test the loud-failure choice is right. As a PR gate the calculus changes: any `ui_kit_library` bump that changes how `AppIpv4TextField` composes its octets (extra hidden field, paste target, wrapper) reddens *static routing* on an unrelated PR, and the failure gets attributed to the wrong change. Same exposure for `_nameTextField() => find.byType(TextField).first` (`:98-101`) and `_nameField() => find.byType(AppTextField).evaluate().first.widget as AppTextField` (`:79-80`), both of which depend on widget-tree order plus the claim (comment at `:78`) that `AppTextField` appears exactly once — unverifiable here, since `ui_kit_library` is an unvendored git dependency (`pubspec.yaml:59-62`, ref `v2.39.0`). Key-based finders would make the gate fail only for real static-routing regressions.

**W4 · The guard is a proxy, not a reproduction**

```dart
// :150-163 / :193-206 (shape)
expect(_nameField().errorText, isNull);   // while focused
await _blur(tester);
expect(_nameField().errorText, isNotNull); // after blur
```

#1332's mechanism is *keystroke loss caused by a rebuild that tears down the CanvasKit semantics `<input>`*. Headless `flutter test` has no CanvasKit and no DOM `<input>`, so these cases verify the timing of `_errors` assignment only. A refactor that keeps `errorText` null during focus but still triggers a disruptive rebuild in `onChanged` (controller reassignment, a `setState` rebuilding the field subtree) passes this gate while the bug regresses on web. Cheap hardening: also assert the field's text still equals what was typed after `onChanged` fires — that is the property users actually lost.

**W5 · Global focus state**

```dart
// :103-106
Future<void> _blur(WidgetTester tester) async {
  FocusManager.instance.primaryFocus?.unfocus();
  await tester.pumpAndSettle();
}
```

`FocusManager.instance` is process-global static state shared by all six cases. If `primaryFocus` is null, or points outside the dialog because an earlier case left it there, `?.` makes this a silent no-op and the following `expect(_nameField().errorText, isNotNull)` (`:172`) fails with a message pointing at validation logic. Assert the precondition, or drive blur through the tree (`tester.tap` on neutral area / `LogicalKeyboardKey.tab`) so the action is observable.

**W6 · Assertion does not prove the claim**

```dart
// :186-188
await tester.tap(find.widgetWithText(AppButton, addLabel));
await tester.pumpAndSettle();
expect(find.byType(StaticRouteDialog), findsOneWidget);
```

`findsOneWidget` after the tap holds for a disabled button (intended), an enabled button whose `onTap` threw, and a pop that failed for an unrelated reason. Reading `AppButton`'s `onTap`/`isEnabled` directly — the same property-assertion style the file already uses well for `errorText` — is order- and hit-test-independent.

</details>

<details>
<summary><b>Spec — evidence and requirements met (vs #1332)</b></summary>

**Framing (important):** the `lib/` fix for #1332 is **already on `dev-2.7.0`** — `cbd939c0 fix(static-routing): validate on focus-loss so the dialog keeps keystrokes (#1332)` — as is the guard test itself (`74704f9b`, `0ffc1a6a`). The three-dot diff `origin/dev-2.7.0...9f06030` is one deleted line. This PR is a follow-up whose only job is to make the already-written guard execute. Reporting "the fix is missing from this PR" would be a false Omission; the implementation was located and verified at head.

**Requirements met** (from #1332's *Suggested fix*, *Scope*, *Symptom*):

1. ✅ `onChanged` only `setState(() {})`, never assigns `_errors` — `lib/.../static_route_dialog.dart:120-122`; all four fields wire `onChanged: (_) => _onInputChanged()` (`:170`, `:178`, `:187`, `:196`). 🟢
2. ✅ Add button still re-gates per keystroke — `_isFormValid => _computeErrors().isEmpty` (`:153`), consumed at `onTap: _isFormValid ? _submit : null` (`:241`). 🟢
3. ✅ `_errors` assigned from per-field focus listeners — name field `:82-84` wired via `focusNode: _nameFocus` (`:166`); IPv4 fields `onFocusChanged: _onIpv4FocusChanged` (`:179`, `:188`, `:197`). 🟢
4. ✅ IPv4 validation fires on whole-field focus-loss, not per octet — `_onIpv4FocusChanged(int? index, bool hasFocus)` guards on `index == null` (`:125-127`), which is exactly what makes a transiently-invalid 3-digit octet safe. 🟢
5. ✅ "Empty field reports no error" — `_errors = _isEdit ? _computeErrors() : {}` (`:79`) with rationale at `:76-78`. Satisfied for add mode, the mode the symptom table describes; edit-mode caveat is Spec W2 above. 🟢
6. ✅ Follows the `port_forwarding_dialog.dart:48-51` pattern and carries the rationale in comments — reproduced at `:51-56`. 🟢
7. ✅ Scope stays at the one file the issue's audit flagged; no file outside static routing touched. 🟢
8. ✅ Gateway can reach a valid `192.168.1.1` so Add enables — verified structurally (no mid-edit `_errors` write ⇒ no `errorText` flip ⇒ no rebuild), **not** end-to-end; see Spec W1. 🟡
9. ✅ This PR's own declared goal — the guard executes in CI: `ci.yml:84` → `run_tests.sh:80/92/96` `--exclude-tags="golden||loc||ui"` → `dart_test.yaml` declares `ui`; the removed line was a **file-level** `@Tags(['ui'])`, so all six `testWidgets` were skipped in every CI run. A revert of `cbd939c0` would have gone green; it no longer will. 🟢

**Coverage: 9 / 9 requirements satisfied.** No Critical, no scope creep, no requirement implemented contrary to spec.

Correctly absent, per the issue's own deferrals — not counted as omissions: the E2E `focusAndType` read-back assertion (tracked as `linksys/PrivacyGUI-USP-E2E#60`, "diagnostics, not a fix"), and the eight "Verified safe" files ("were already fixed for this exact defect").

**Spec W2 evidence** — the guard group covers add mode only:

```dart
// test/.../static_route_dialog_test.dart:139
group('StaticRouteDialog #1332 focus-loss validation', () {
```
Six cases: add-open / name-typing / name-blur / gateway-typing / gateway-blur / Add-disabled. The mode-independent property that actually matters (`_onInputChanged` is `setState(() {})` only, shared by both modes) means edit mode is covered *by construction* — but nothing in the suite pins that. One edit-mode case, or one comment saying so, makes the guard state its own scope.

</details>

<details>

- **Correct diagnosis, minimal blast radius.** A real gap — a regression guard that CI never executed — fixed with a one-line deletion. No production code, no drive-by refactor, no unrelated lint cleanup.
- **The claim in the commit message is true and was verifiable end to end** from the worktree alone: `ci.yml:84` → `run_tests.sh:80/92/96` → `dart_test.yaml`. The measured before/after in the PR body (`No tests ran` → `+6: All tests passed`) matches the mechanism.
- **The PR body is exemplary.** It cross-checks against #1347's merged CI log (file appears 0 times among 265 executed), explains why removing the tag beats moving it to `dashboard-card`, notes that `--exclude-tags` drops a file if *any* tag matches (so `['ui','dashboard-card']` would still have been excluded), and tells the reviewer what to actually verify. The deliberate one-file scope, proving the mechanism on Linux CI before the other eleven follow, is the right sequencing.
- **The underlying test design is worth protecting.** Asserting the widget's `errorText` property rather than a rendered string, with the reason documented at `:20-23` (inline error vs focus tooltip is layout-dependent), is the difference between a durable gate and a flaky one — and it materially strengthens the case that this file belongs in CI.
- **Genuinely good test documentation** at `:8-23`: root cause of #1332, fix shape, sibling precedents, and why the assertions take the form they do.
- **Layering clean** (no `lib/generated/...` import), test path mirrors the lib path, behaviour-describing test names, no `TC001`-style numbering.
- **Commit hygiene:** `cbd939c0` fix → `74704f9b` test → `0ffc1a6a` format → `9f060305` CI-enable, one concern per commit; the format fix was kept separate rather than folded in.

</details>

<sub>Verification caveat, stated for transparency: no Flutter SDK in the review environment and `ui_kit_library` is an unvendored git dependency, so the suite was not executed and `AppTextField` / `AppIpv4TextField` / `AppButton` internals could not be inspected. Findings depending on those internals are marked 🟡/⚪. The CI-wiring claims were read directly from the head worktree.</sub>

PR #1355, opened 2026-08-21
fix(ai): rebuild the router state summary per request

### LINKED ISSUE #1354: fix(ai): router state summary is frozen when the chat opens, so long-lived sessions answer from stale data
## What happens

Open the AI assistant, leave it open, go do something else. Plug in a laptop, take a phone off WiFi, let the WAN drop and come back. Now ask "how many devices are connected?" or "is my internet OK?"

The assistant can answer from the router state summary in its system prompt — and that summary was built **once, when the chat was opened**. It has never been updated since. So the answer describes the network as it was when you opened the window, presented as if it were current.

The heading the model reads is literally `# Current Router State`. Nothing tells it the contents may be minutes or hours old.

## Why it happens

`buildRouterContext()` is called at controller construction and the result is stored as a `String`:

```dart
// router_chat_controller.dart
final String _routerContext;   // assigned in the constructor, never reassigned
```

The two call sites (`router_assistant_view.dart:120` first connect, `:240` after entering credentials) each pass an already-computed string. `_routerContext` is `final` and nothing reassigns it, so every request for the rest of the session re-sends that same snapshot.

The summary carries model, firmware, CPU/memory, WAN status and IP, connected/online device counts, mesh extenders, WiFi bands and SSIDs, LAN/DHCP config, ethernet port status and firewall state — all of it frozen.

Tools are unaffected: they read the L1 provider caches at call time and return current data. So the assistant is *capable* of being right, and whether it is depends on whether it decides a tool call is needed. The system prompt actively discourages that — it instructs the model to check the context first and use tools sparingly, which is the correct instruction for a summary that is actually current.

## Why this is worth fixing rather than living with

The failure is quiet and selective. Most answers stay plausible — the model, the firmware, the SSIDs don't change — so nothing looks broken. What goes wrong is exactly the fast-moving fields a user is most likely to ask about: device counts, whether the internet is up, CPU load. A wrong device count reads as a product that does not know what is connected to the router.

It also gets worse the better the assistant is used. A user who keeps the panel open all day is the one most exposed.

## Proposed fix

Compute the summary per request instead of once per session. Reading it is cheap and does not need to change: `buildRouterContext()` is fully synchronous and only reads Riverpod L1 caches already in memory — no network, no device round trip, no `await` anywhere in it.

```dart
// now
final String _routerContext;

// proposed
final String Function() _buildRouterContext;   // invoked where the prompt is assembled
```

Call sites become `() => buildRouterContext(ref.read)`, and the call moves to where `RouterSystemPrompt.buildParts()` is invoked (`router_chat_controller.dart:191`).

**Prompt caching is not affected.** `buildParts()` already puts the static instructions in their own `SystemPromptPart` with `cache: true` and the summary in a separate part with `cache: false`, static first. Caching matches on prefixes, so a summary that changes every request still leaves the cached prefix intact. Measured: static part 13,478 chars (cached), summary a few hundred (not cached) — a refresh costs nothing the session was not already paying.

## Acceptance criteria

- [ ] The summary sent with a request reflects provider state at that moment, not at controller construction
- [ ] With the chat left open across a change in connected devices, a question about device count is answered with the current number
- [ ] Static instructions still report as cached (`cache: true`), and cache hits still occur on later requests in a session
- [ ] A test asserts two requests separated by a provider state change carry different summaries

## Out of scope

- Whether the model should prefer tools over the summary for volatile fields — a separate question, and less pressing once the summary is current
- Telling the model how fresh the data is (e.g. a timestamp in the summary); worth considering later, unnecessary if the summary is rebuilt per request

HankYuLinksys:
Fixes #1354.

## What's wrong

Leave the assistant open, go do something else, come back and ask "how many devices are connected?" or "is my internet OK?"

The summary of router state in the system prompt was captured **once, in the constructor**, and stored. A session outlives the network it describes — a panel left open sees devices join and leave and the WAN drop — so every request kept re-sending the state found when the panel opened, under a heading that reads `# Current Router State`. Nothing told the model the contents might be hours old.

The failure is quiet and selective. Model, firmware and SSIDs don't change, so answers stay plausible and nothing looks broken. What goes wrong is exactly the fast-moving fields users ask about: device counts, whether the internet is up, CPU load. And it gets worse the more the feature is used — someone who keeps the panel open all day is the most exposed.

Tools were never affected: they read the L1 provider caches at call time. So the assistant was *capable* of being right, and whether it was depended on it choosing to call a tool — which the system prompt correctly discourages, since it tells the model to check the context first.

## Why not just call the existing hook

A `refreshContext()` method already existed for precisely this, and **nothing ever called it**. Someone saw the problem, built the escape hatch, and never wired it up.

That is the argument against fixing it by wiring it up. Correctness that depends on remembering to invalidate is what failed here, and a stored field would have to be invalidated from every path that starts a request. So the stored parts are gone and the prompt is assembled per request instead — there is no longer a field to keep fresh. `refreshContext()` and the now-redundant refresh in `clearConversation()` are removed with it.

## Prompt caching is unaffected

`buildParts()` already puts the static instructions in their own `SystemPromptPart` marked `cache: true` and ordered first, and caching matches on prefixes. The summary sits past that boundary, so it can differ on every request while leaving the cached prefix intact. Measured: static part 13,478 chars (cached), summary a few hundred (not cached).

A test pins that ordering, because moving the summary into the cached part would silently cost the cache on every request — the kind of regression that shows up as a bill, not a failure.

## The builder comes from a provider, not the view

Worth calling out, because the first version of this change got it wrong.

The prompt is now assembled inside the loop in `_processConversation()`, which spans `await`s. A later round can therefore run after the user has navigated away — and `WidgetRef.read` throws once its widget is disposed (`Bad state: Cannot use "ref" after the widget was disposed`, verified). Two things make that reachable: the view's `dispose()` only removes its listener and never disposes the controller, and `_isStale()` tracks conversation generation, not view lifetime.

So the builder comes from `routerContextBuilderProvider`, closing over a ref owned by the container. It must not be `autoDispose` — a disposed provider's ref throws the same way — and the comment says so.

## Tests

Three, each pinning a distinct property, all verified by mutation:

| Test | Mutation that makes it fail |
|---|---|
| Each request carries the summary as it is at that moment | Capture the summary in the constructor again |
| Static instructions stay cacheable and ordered first | Move the summary into the cached part |
| Builder stays callable after its widget is disposed | Close over `WidgetRef` instead of the provider's ref |

The third fails with the exact `StateError` above, so it pins the real failure and not a proxy for it.

## Verification

- Affected tests: 46 pass. Wider `test/ai/` + `test/page/ai_assistant/`: 157 pass
- `flutter analyze` — no issues in any changed file
- `fvm dart format` — clean

## Not covered

`RouterChatController` is never disposed — the view's `dispose()` only removes the listener — and an in-flight `_processConversation()` loop has no cancellation. Pre-existing, and this change makes it less consequential by removing the loop's dependence on widget lifetime, but the leak and the uncancelled loop remain.

### DIFF
--- a/lib/ai/orchestrator/router_chat_controller.dart
+++ b/lib/ai/orchestrator/router_chat_controller.dart
@@ -78,18 +78,22 @@ class PendingConfirmation extends Equatable {
 /// user confirms or cancels, then flushed together with the write result. This
 /// is invisible to the user because tool results are never rendered in the chat.
 class RouterChatController extends ChangeNotifier {
+  /// [routerContextBuilder] is called for every request rather than once, so
+  /// the state the model reads is the state at the time of asking.
+  ///
+  /// It is a callback and not a `String` because a session outlives the network
+  /// it describes: a panel left open sees devices join and leave and the WAN
+  /// drop, and a summary captured at construction would keep asserting the
+  /// network it found on open. Building it per request is close to free — it
+  /// reads Riverpod caches already in memory and touches no device.
   RouterChatController({
     required IConversationGenerator generator,
     required IRouterCommandProvider commandProvider,
-    required String routerContext,
+    required String Function() routerContextBuilder,
   })  : _generator = generator,
         _commandProvider = commandProvider,
-        _routerContext = routerContext {
+        _routerContextBuilder = routerContextBuilder {
     _log('RouterChatController initialized');
-    // The router context is the user's network: model, IP addresses, SSIDs,
-    // device names. Debug builds only.
-    aiLogSensitive(() => 'Router context:\n$routerContext');
-    _initializeSystemPrompt();
   }
 
   static void _log(String message) => aiLog(message);
@@ -103,12 +107,11 @@ class RouterChatController extends ChangeNotifier {
 
   final IConversationGenerator _generator;
   final IRouterCommandProvider _commandProvider;
-  final String _routerContext;
+  final String Function() _routerContextBuilder;
 
   ConversationState _state = ConversationState.initial();
   PendingConfirmation? _pendingConfirmation;
   String? _lastUserMessage;
-  List<SystemPromptPart>? _systemPromptParts;
   List<GenTool>? _routerTools;
   Map<String, RouterCommand>? _commandsByName;
   Future<Map<String, RouterCommand>>? _commandsLoad;
@@ -185,12 +188,24 @@ class RouterChatController extends ChangeNotifier {
   /// Total tokens (input + output) used in this session.
   int get totalTokens => _totalInputTokens + _totalOutputTokens;
 
-  /// Initialize system prompt parts with router context (for caching).
-  void _initializeSystemPrompt() {
-    _systemPromptParts =
-        RouterSystemPrompt.buildParts(routerContext: _routerContext);
-    _log(
-        'System prompt initialized: ${_systemPromptParts!.length} parts, static cached: ${_systemPromptParts!.first.cache}');
+  /// Assemble the system prompt for one request, with a freshly read summary.
+  ///
+  /// Deliberately not stored. Holding the parts in a field is what let the
+  /// summary go stale, and a field would have to be invalidated from every
+  /// path that starts a request to stay correct. Rebuilding removes the
+  /// question.
+  ///
+  /// Prompt caching is unaffected: [RouterSystemPrompt.buildParts] puts the
+  /// static instructions in their own part, marked for caching and ordered
+  /// first, and caching matches on prefixes. The summary sits past that
+  /// boundary, so it can differ on every request and still leave the cached
+  /// prefix intact — cache hits are visible in the per-response token logs.
+  List<SystemPromptPart> _buildSystemPromptParts() {
+    final routerContext = _routerContextBuilder();
+    // The summary is the user's network: model, IP addresses, SSIDs, device
+    // names. Debug builds only.
+    aiLogSensitive(() => 'Router context:\n$routerContext');
+    return RouterSystemPrompt.buildParts(routerContext: routerContext);
   }
 
   /// Load the command registry once per conversation.
@@ -320,7 +335,7 @@ class RouterChatController extends ChangeNotifier {
         final response = await _generator.generateWithHistory(
           _state.messages,
           tools: tools.isNotEmpty ? tools : null,
-          systemPromptParts: _systemPromptParts,
+          systemPromptParts: _buildSystemPromptParts(),
         );
         if (_isStale(batch)) return;
         _log(
@@ -847,14 +862,7 @@ class RouterChatController extends ChangeNotifier {
     // the view sees `isLoading == false` alongside a non-zero round.
     _currentRound = 0;
 
-    // Refresh router context
-    _initializeSystemPrompt();
     notifyListeners();
-  }
-
-  /// Refresh router context (call when dashboard data changes).
-  void refreshContext() {
-    _initializeSystemPrompt();
   }
 
   /// Test seam for [_abandonedToolResult], which is otherwise only reachable
--- a/lib/ai/orchestrator/router_chat_controller.dart
+++ b/lib/ai/orchestrator/router_chat_controller.dart
@@ -200,8 +200,23 @@ class RouterChatController extends ChangeNotifier {
   /// first, and caching matches on prefixes. The summary sits past that
   /// boundary, so it can differ on every request and still leave the cached
   /// prefix intact — cache hits are visible in the per-response token logs.
+  ///
+  /// A summary that cannot be built costs the context, not the answer. The
+  /// model still has its instructions and its tools, so it can fetch what it
+  /// needs; dropping the exchange instead would turn a degraded reply into a
+  /// generic error. This became reachable when the build moved inside the
+  /// request loop — it now runs once per round rather than once per session.
   List<SystemPromptPart> _buildSystemPromptParts() {
-    final routerContext = _routerContextBuilder();
+    String routerContext;
+    try {
+      routerContext = _routerContextBuilder();
+    } catch (e) {
+      // Structural only: the summary is the user's network, so its contents
+      // must not reach a release log.
+      _log('_buildSystemPromptParts: router context unavailable, '
+          'continuing without it (${e.runtimeType})');
+      routerContext = '';
+    }
     // The summary is the user's network: model, IP addresses, SSIDs, device
     // names. Debug builds only.
     aiLogSensitive(() => 'Router context:\n$routerContext');
--- a/lib/page/ai_assistant/providers/router_command_provider.dart
+++ b/lib/page/ai_assistant/providers/router_command_provider.dart
@@ -7,3 +7,17 @@ import 'package:privacy_gui/ai/_ai.dart';
 final routerCommandProviderProvider = Provider<IRouterCommandProvider>((ref) {
   return UspCommandProvider(ref);
 });
+
+/// Builds the router state summary on demand, for the chat's system prompt.
+///
+/// Exists so the summary can be rebuilt per request without the caller holding a
+/// `WidgetRef`. The chat controller assembles its system prompt inside a loop
+/// that spans `await`s, and the view that started the exchange can be gone by a
+/// later round — `WidgetRef.read` throws once its widget is disposed, whereas
+/// this closure captures a ref owned by the container and stays callable.
+///
+/// Must not be `autoDispose`: a disposed provider's ref throws in the same way,
+/// which would reintroduce exactly the failure this avoids.
+final routerContextBuilderProvider = Provider<String Function()>((ref) {
+  return () => buildRouterContext(ref.read);
+});
--- a/lib/page/ai_assistant/views/router_assistant_view.dart
+++ b/lib/page/ai_assistant/views/router_assistant_view.dart
@@ -117,7 +117,7 @@ class _RouterAssistantViewState extends ConsumerState<RouterAssistantView> {
       _controller = RouterChatController(
         generator: AwsContentGenerator(config: awsConfig),
         commandProvider: ref.read(routerCommandProviderProvider),
-        routerContext: buildRouterContext(ref.read),
+        routerContextBuilder: ref.read(routerContextBuilderProvider),
       );
       _controller!.addListener(_onControllerChanged);
       _needsConfig = false;
@@ -237,7 +237,7 @@ class _RouterAssistantViewState extends ConsumerState<RouterAssistantView> {
       _controller = RouterChatController(
         generator: AwsContentGenerator(config: awsConfig),
         commandProvider: commandProvider,
-        routerContext: buildRouterContext(ref.read),
+        routerContextBuilder: ref.read(routerContextBuilderProvider),
       );
       _controller!.addListener(_onControllerChanged);
 
--- a/test/ai/orchestrator/router_chat_controller_test.dart
+++ b/test/ai/orchestrator/router_chat_controller_test.dart
@@ -17,6 +17,10 @@ void main() {
   late MockRouterCommandProvider mockCommandProvider;
   late RouterChatController controller;
 
+  /// Stands in for the live router state. Reassign it mid-test to represent the
+  /// network changing under a session that is already open.
+  late String routerContext;
+
   setUpAll(() {
     registerFallbackValue(<ChatMessage>[]);
     registerFallbackValue(<GenTool>[]);
@@ -26,6 +30,7 @@ void main() {
   setUp(() {
     mockGenerator = MockConversationGenerator();
     mockCommandProvider = MockRouterCommandProvider();
+    routerContext = '# Test Router Context\nModel: TestRouter';
 
     when(() => mockCommandProvider.listCommands()).thenAnswer(
       (_) async => [
@@ -53,7 +58,7 @@ void main() {
     controller = RouterChatController(
       generator: mockGenerator,
       commandProvider: mockCommandProvider,
-      routerContext: '# Test Router Context\nModel: TestRouter',
+      routerContextBuilder: () => routerContext,
     );
   });
 
@@ -79,6 +84,76 @@ void main() {
       test('has no error initially', () {
         expect(controller.hasError, isFalse);
         expect(controller.errorMessage, isNull);
+      });
+    });
+
+    group('router context freshness', () {
+      /// The summary used to be captured once in the constructor, so a session
+      /// left open kept describing the network as it was when the panel opened —
+      /// under a heading that reads `# Current Router State`.
+      test('each request carries the summary as it is at that moment',
+          () async {
+        when(() => mockGenerator.generateWithHistory(
+              any(),
+              tools: any(named: 'tools'),
+              systemPromptParts: any(named: 'systemPromptParts'),
+              forceToolUse: any(named: 'forceToolUse'),
+            )).thenAnswer((_) async => _textResponse('ok'));
+
+        await controller.sendMessage('how many devices?');
+
+        // The network changes while the session stays open.
+        routerContext = '# Current Router State\n- Total connected devices: 9';
+
+        await controller.sendMessage('and now?');
+
+        final sent = verify(() => mockGenerator.generateWithHistory(
+              any(),
+              tools: any(named: 'tools'),
+              systemPromptParts: captureAny(named: 'systemPromptParts'),
+              forceToolUse: any(named: 'forceToolUse'),
+            )).captured.cast<List<SystemPromptPart>?>();
+
+        expect(sent.length, 2);
+        expect(
+          _summaryOf(sent.first),
+          contains('Model: TestRouter'),
+          reason:
+              'first request should carry the summary from before the change',
+        );
+        expect(
+          _summaryOf(sent.last),
+          contains('Total connected devices: 9'),
+          reason: 'second request should carry the changed summary, not the '
+              'one captured when the controller was built',
+        );
+      });
+
+      /// Guards the property the refresh depends on: the static instructions are
+      /// their own cacheable part and come first, so a summary that differs on
+      /// every request still leaves the cached prefix untouched.
+      test('static instructions stay cacheable and ordered first', () async {
+        when(() => mockGenerator.generateWithHistory(
+              any(),
+              tools: any(named: 'tools'),
+              systemPromptParts: any(named: 'systemPromptParts'),
+              forceToolUse: any(named: 'forceToolUse'),
+            )).thenAnswer((_) async => _textResponse('ok'));
+
+        await controller.sendMessage('hi');
+
+        final sent = verify(() => mockGenerator.generateWithHistory(
+              any(),
+              tools: any(named: 'tools'),
+              systemPromptParts: captureAny(named: 'systemPromptParts'),
+              forceToolUse: any(named: 'forceToolUse'),
+            )).captured.cast<List<SystemPromptPart>?>();
+
+        final parts = sent.single!;
+        expect(parts.first.cache, isTrue);
+        expect(parts.first.text, isNot(contains('TestRouter')),
+            reason: 'the cached part must not carry the volatile summary');
+        expect(parts.skip(1).every((p) => p.cache == false), isTrue);
       });
     });
 
@@ -998,6 +1073,10 @@ LLMResponse _textResponse(String text) {
     content: [TextBlock(text: text)],
   );
 }
+
+/// The parts after the cached static instructions, where the summary lives.
+String _summaryOf(List<SystemPromptPart>? parts) =>
+    (parts ?? []).skip(1).map((p) => p.text).join();
 
 LLMResponse _toolUseResponse(String toolName, Map<String, dynamic> input) {
   return LLMResponse(
--- a/test/ai/orchestrator/router_chat_controller_test.dart
+++ b/test/ai/orchestrator/router_chat_controller_test.dart
@@ -21,6 +21,10 @@ void main() {
   /// network changing under a session that is already open.
   late String routerContext;
 
+  /// Set mid-test to make the summary builder throw, standing in for a provider
+  /// read that fails while the exchange is already under way.
+  late bool throwingContext;
+
   setUpAll(() {
     registerFallbackValue(<ChatMessage>[]);
     registerFallbackValue(<GenTool>[]);
@@ -30,7 +34,12 @@ void main() {
   setUp(() {
     mockGenerator = MockConversationGenerator();
     mockCommandProvider = MockRouterCommandProvider();
+    // Deliberately not production's `# Current Router State` heading: the
+    // freshness test reassigns this and then asserts the two requests carried
+    // different summaries, which needs the before and after to be tellable
+    // apart. Aligning it with production would make that assertion vacuous.
     routerContext = '# Test Router Context\nModel: TestRouter';
+    throwingContext = false;
 
     when(() => mockCommandProvider.listCommands()).thenAnswer(
       (_) async => [
@@ -58,7 +67,10 @@ void main() {
     controller = RouterChatController(
       generator: mockGenerator,
       commandProvider: mockCommandProvider,
-      routerContextBuilder: () => routerContext,
+      routerContextBuilder: () {
+        if (throwingContext) throw StateError('provider unavailable');
+        return routerContext;
+      },
     );
   });
 
@@ -127,6 +139,26 @@ void main() {
           reason: 'second request should carry the changed summary, not the '
               'one captured when the controller was built',
         );
+      });
+
+      test('a builder that throws costs the context, not the answer', () async {
+        when(() => mockGenerator.generateWithHistory(
+              any(),
+              tools: any(named: 'tools'),
+              systemPromptParts: any(named: 'systemPromptParts'),
+              forceToolUse: any(named: 'forceToolUse'),
+            )).thenAnswer((_) async => _textResponse('ok'));
+
+        throwingContext = true;
+
+        await controller.sendMessage('hi');
+
+        expect(controller.hasError, isFalse,
+            reason:
+                'a summary that cannot be built must not abort the exchange');
+        final assistant =
+            controller.messages.where((m) => m.role == ChatRole.assistant);
+        expect(assistant, isNotEmpty, reason: 'the model should still answer');
       });
 
       /// Guards the property the refresh depends on: the static instructions are
--- a/test/page/ai_assistant/providers/router_command_provider_test.dart
+++ b/test/page/ai_assistant/providers/router_command_provider_test.dart
@@ -1,7 +1,30 @@
+import 'package:flutter/widgets.dart';
 import 'package:flutter_riverpod/flutter_riverpod.dart';
 import 'package:flutter_test/flutter_test.dart';
 import 'package:privacy_gui/ai/_ai.dart';
 import 'package:privacy_gui/page/ai_assistant/providers/router_command_provider.dart';
+
+/// Hands the builder it obtained during `initState` back to the test, so the
+/// test can keep calling it after this widget is gone.
+class _BuilderCaptor extends ConsumerStatefulWidget {
+  const _BuilderCaptor({required this.onBuilder});
+
+  final void Function(String Function()) onBuilder;
+
+  @override
+  ConsumerState<_BuilderCaptor> createState() => _BuilderCaptorState();
+}
+
+class _BuilderCaptorState extends ConsumerState<_BuilderCaptor> {
+  @override
+  void initState() {
+    super.initState();
+    widget.onBuilder(ref.read(routerContextBuilderProvider));
+  }
+
+  @override
+  Widget build(BuildContext context) => const SizedBox();
+}
 
 void main() {
   group('routerCommandProviderProvider', () {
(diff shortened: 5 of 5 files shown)


AustinChangLinksys (approved):
🤖 **Automated Review — Round 1** · `d60849e..b524d19` (full)

**Verdict: ✅ APPROVE** — Core staleness fix is correct and well-designed; no Critical issues found across both axes. Warnings are non-blocking.

#### Standards

|---|---|---|---|
| ⚠️ | 🟢High | `router_chat_controller.dart:338` | If `_routerContextBuilder()` throws, the outer `catch` surfaces "An unexpected error occurred" with no partial context fallback |
| ⚠️ | 🟢High | `router_chat_controller.dart:204` + `usp_command_provider.dart` | Pre-existing: sensitive fields (WAN IP, SSID, device names) now transmitted on every LLM round (up to 5×/session) instead of once at construction — frequency increase is undocumented |
| ⚠️ | 🟡Med | `router_command_provider.dart:20–22` | [Feature Envy] `routerContextBuilderProvider` lives in `lib/page/ai_assistant/providers/` but is an AI-orchestrator concern; logically belongs in `lib/ai/providers/` |
| ⚠️ | 🟡Med | `router_chat_controller.dart` (`clearConversation`) | Commands are session-scoped (loaded once, reset on clear); router context is now per-request — the two freshness models are asymmetric and undocumented |
| 💡 | 🟡Med | `router_command_provider.dart:20–22` | `routerContextBuilderProvider` with `autoDispose` prohibition: no test guards against accidental `autoDispose` being added |
| 💡 | 🟢High | `router_chat_controller.dart:203` | Per-request rebuild removed the diagnostic log from `_initializeSystemPrompt()`; consider restoring a per-build log for cache diagnostics |
| 💡 | 🟡Med | `router_chat_controller_test.dart:1074` | `_summaryOf()` helper silently skips the first part; no assertion confirms that part is the cached static instructions |
| 💡 | 🟢High | `router_chat_controller_test.dart` (`setUp`) | Fixture initial value `# Test Router Context` differs from production heading `# Current Router State` intentionally — add a comment explaining why, to prevent future "fix" that would break the freshness test |
| 💡 | 🟢High | `router_command_provider_test.dart:6–26` | `void Function(String Function())` callback type in `_BuilderCaptor` could use a `typedef` for readability |

#### Spec

|---|---|---|---|
| ✅ | 🟢High | `router_chat_controller.dart:203–207` → spec#AC1 | AC1: summary rebuilt per request ✅ |
| ⚠️ | 🟡Med | `router_chat_controller_test.dart` (all new tests) → spec#AC2 | AC2: mechanism verified; no widget/integration test validates the user-observable answer with current device count |
| ✅ | 🟢High | `router_system_prompt.dart` + new test → spec#AC3 | AC3: static instructions remain `cache:true`, ordered first, test pins ordering ✅ |
| ✅ | 🟢High | `router_chat_controller_test.dart` (`each request carries…`) → spec#AC4 | AC4: test asserts two requests across provider state change carry different summaries ✅ |
| 💡 | 🟡Med | `router_chat_controller.dart` (removed `refreshContext()`) | `refreshContext()` removed as dead code — grep confirms no call sites in view layer; broader repo scan not done |

<details>

**W1 — Builder exception lacks fallback context** 🟢High  
`lib/ai/orchestrator/router_chat_controller.dart:338`  
```dart
// Inside while loop, inside try block:
final response = await _generator.generateWithHistory(
  _state.messages,
  tools: tools.isNotEmpty ? tools : null,
  systemPromptParts: _buildSystemPromptParts(),  // _routerContextBuilder() called here
);
```
If `_routerContextBuilder()` throws (e.g. Riverpod container in an unexpected state), the outer `catch (e)` at ~line 474 catches it and surfaces `'An unexpected error occurred: $e'` with no fallback. No crash, but the user sees a generic error instead of a degraded-but-working response.  
**Fix:** Wrap in a try-catch inside `_buildSystemPromptParts()` and fallback to `routerContext = ''` on throw, so the LLM call can still proceed without context.

---

**W2 — Sensitive data transmission frequency increase (pre-existing design, escalated)** 🟢High  
`lib/ai/orchestrator/router_chat_controller.dart:204` + `lib/ai/providers/usp_command_provider.dart` (`buildRouterContext`)  
```dart
// buildRouterContext assembles:
buffer.writeln('- IP Address: ${m.ipAddress}');   // WAN IP (plaintext)
buffer.writeln('- ${radio.band}: $status, SSID: "$ssid"');  // SSIDs
// ...and device names, extender models, etc.
```
Pre-existing: these fields were already sent to Bedrock. This PR changes frequency from once-per-session to once-per-LLM-round (up to `maxRounds` times). The per-round overhead is small (a few hundred chars), but the privacy surface in terms of transmission count is multiplied. `aiLogSensitive` correctly guards logs but not network calls.  
**Fix:** Pre-existing; not blocking merge. Recommend documenting this trade-off in the PR or a code comment. Optional: only rebuild context when provider state actually changes (diff-based), avoiding unnecessary re-transmission.

---

**W3 — `routerContextBuilderProvider` layer placement [Feature Envy]** 🟡Med  
`lib/page/ai_assistant/providers/router_command_provider.dart:20–22`  
```dart
final routerContextBuilderProvider = Provider<String Function()>((ref) {
  return () => buildRouterContext(ref.read);
});
```
This provider's sole consumer is `RouterChatController` (AI orchestration layer). Placing it in `lib/page/ai_assistant/providers/` means an AI-layer concern is homed in the Page layer. No hard layer violation (no upward dependency), but creates cohesion debt — if another non-Page consumer needs this provider, it must reach into the Page layer.  
**Fix:** Move to `lib/ai/providers/router_context_provider.dart` or alongside `buildRouterContext` in `usp_command_provider.dart`. Non-blocking.

---

**W4 — Commands vs context freshness asymmetry undocumented** 🟡Med  
`lib/ai/orchestrator/router_chat_controller.dart` (`clearConversation`, `_commandsLoad`)  
Router context is now per-request (always current); command registry is session-scoped (loaded once, reset only on `clearConversation()`). If firmware upgrade changes the command list mid-session, commands could be stale while context is fresh. This is a deliberate design decision, but it's not documented.  
**Fix:** Add a comment in `_getRouterTools()` or `clearConversation()` explaining that commands are intentionally session-scoped while context is per-request, and why that's acceptable.

---

**Spec W — AC2 behavior-level test gap** 🟡Med  
`test/ai/orchestrator/router_chat_controller_test.dart` (new tests)  
AC2 requires "a question about device count is answered with the current number." All new tests verify `systemPromptParts` content (mechanism), but the mock LLM always returns `'ok'` and response content is not asserted. No widget or integration test validates that the user-visible answer reflects current provider state.  
**Fix:** Not blocking. Either add a clarifying comment that AC2 is verified at the mechanism level (matching issue's Out-of-scope policy) or add a higher-level test asserting response content when provider state differs.

</details>

<details>

- **Core fix is exactly right**: replacing `final String _routerContext` with `final String Function() _routerContextBuilder` is the minimal, correct change — eliminates the stale-snapshot problem without over-engineering.
- **`routerContextBuilderProvider` design is sound**: closing over a container-owned `ref` instead of `WidgetRef` correctly solves the `Bad state: Cannot use "ref" after the widget was disposed` throw that would occur when `_processConversation`'s `await` loop outlives the widget.
- **`refreshContext()` and `_initializeSystemPrompt()` removal**: eliminates two dead/misplaced invalidation surfaces. "Correctness that depends on remembering to invalidate" is identified and removed — good design reasoning.
- **Prompt caching unaffected**: static instructions in cacheable part ordered first; summary past the boundary. The ordering test pins this explicitly, preventing a silent regression that would appear only as a billing change.
- **Test quality is high**: tests verify observable outcomes (content of `systemPromptParts`), not call counts. The `testWidgets` in `router_command_provider_test.dart` directly exercises the failure mode (widget disposed, builder still called) and pins the exact `StateError` that motivated the provider-based approach.
- **PR description is exemplary**: explains *what* changed, *why* the naive fix was rejected, *why* prompt caching is unaffected, and *what* each test mutation would break. This level of documentation is rare and valuable.
- **AC3/AC4 fully covered**: both specs are verified at the unit level with targeted, mutation-killable tests.
- **No scope creep**: diff is exactly what the issue proposed — no timestamps, no tool-preference changes, no changes to `buildRouterContext()` output format.

</details>

lib/ai/orchestrator/router_chat_controller.dart
```diff
@@ -320,7 +335,7 @@ class RouterChatController extends ChangeNotifier {
         final response = await _generator.generateWithHistory(
           _state.messages,
           tools: tools.isNotEmpty ? tools : null,
-          systemPromptParts: _systemPromptParts,
+          systemPromptParts: _buildSystemPromptParts(),
```
PeterJhongLinksys, line 353: ⚠️ **Warning (🟢High) — rebuilt per *round*, not per *request*.**

This call sits inside the `while (loopCount < maxLoops)` tool-calling loop, so the summary is rebuilt once per LLM round rather than once per user turn. Every round after the first continues the *same* logical request, replaying the same `_state.messages` — so round 2 can send a fresher summary alongside the round-1 `tool_use`/`tool_result` pair still in history, asking the model to interpret a tool result gathered under a different view of the network.

Hoisting the build to the top of `_processConversation` into a local matches the stated goal ("the state the model reads is the state at the time of asking") exactly, keeps one coherent world-view across the exchange, and costs strictly less than the current code.

Reviewers split here: A would change it before merge, B would approve as-is and treat the loop only as an untested path. Both agree it is not Critical — your call.

PeterJhongLinksys (approved):
🤖 **Automated Review — Round 1** · `d60849e..b524d19` (full)

**Verdict: ✅ APPROVE** — the stale-summary root cause is genuinely removed (field + `refreshContext()` deleted rather than another invalidation call site added), the prompt-cache objection is reasoned about *and* test-pinned, and issue #1354's two core acceptance criteria are met; no Critical on either axis. Nine Warnings remain, and the two reviewers split on one of them — see the note under the Standards table.

#### Standards

|---|---|---|---|
| ⚠️ | 🟢 | `lib/ai/orchestrator/router_chat_controller.dart:338` | Summary is rebuilt per LLM **round**, not per request — the system prompt can change underneath a `tool_use`/`tool_result` pair still in history (A) |
| ⚠️ | 🟡 | `lib/page/ai_assistant/providers/router_command_provider.dart:21-23` | Escaped closure holds a **container**-owned `Ref`; throws if the container (not the widget) is disposed — today's safety is incidental, it only survives because the sole call site sits inside a `try` (A+B) |
| ⚠️ | 🟡 | `lib/ai/orchestrator/router_chat_controller.dart:203-208` | Builder can now throw from *inside* the request loop; the generic `catch` at `:474` turns that into a non-retryable "unexpected error" mid-conversation (A) |
| ⚠️ | 🟢 | `lib/ai/orchestrator/router_chat_controller.dart:86-88` | "close to free … touches no device" is inaccurate for not-yet-initialised `AsyncNotifierProvider`s — and it is the load-bearing justification for rebuilding (A+B) |
| ⚠️ | 🟡 | `lib/ai/providers/usp_command_provider.dart:821-972`, `router_chat_controller.dart:206` | ~150-line `buildRouterContext` + a full sensitive-data dump promoted from once-per-session to once-per-round (debug log volume ×`maxRounds`) (B, A-S1) |
| ⚠️ | 🟡 | `test/ai/orchestrator/router_chat_controller_test.dart:93-131` | Freshness test only exercises single-round turns (both stubs are tool-free), so the per-round path it protects is unguarded (A+B) |
| ⚠️ | 🟢 | `lib/ai/orchestrator/router_chat_controller.dart` (removed `refreshContext()`) | Public method removed from a public class with no deprecation/migration note — zero in-tree callers, verified by grep (B) |
| ⚠️ | 🟡 | `lib/page/ai_assistant/providers/router_command_provider.dart:21-23` | Escaping-`Ref`-via-`Provider` is a new pattern here; the "must not be `autoDispose`" invariant is hand-written and unenforced by any test (B) |
| ⚠️ | ⚪ | `lib/ai/orchestrator/router_chat_controller.dart:81-88`, `:191-202` | Same design rationale told twice (constructor doc + private method doc), two places to keep in sync (B) |
| 💡 | 🟢 | `lib/page/ai_assistant/views/router_assistant_view.dart:117-121`, `:237-241` | [Duplicated Code] One behavioural change required two mechanically identical controller constructions — the PR is its own evidence (B) |
| 💡 | 🟢 | `lib/ai/orchestrator/router_chat_controller.dart:191-194` (removed) | Lost observability: the `parts.length` / `cache` log is gone, leaving only indirect token-log evidence for the PR's central caching claim (A) |
| 💡 | 🟡 | `router_chat_controller.dart:92`, `:110`, `router_command_provider.dart:21` | `String Function()` appears three times unnamed; a `typedef RouterContextBuilder` matches the existing `ProviderReader` precedent (A+B) |
| 💡 | 🟡 | `test/ai/orchestrator/router_chat_controller_test.dart:1077-1079`, `:135-157` | `_summaryOf` hardcodes "summary = everything after index 0" — the same assumption the cache test verifies; assert on `cache == false` instead, and consider testing `buildParts` directly (A+B) |
| 💡 | 🟡 | `lib/ai/providers/usp_command_provider.dart:812`, `lib/core/utils/ip_getter/get_local_ip.dart:16` | Two identically-shaped `ProviderReader` typedefs; this PR routes a new dependency through one of them (B, pre-existing) |
| 💡 | 🟡 | `test/ai/orchestrator/router_chat_controller_test.dart:33`, `:113` | Ad-hoc inline test data where the Testing rules ask for a Test Data Builder (the mutable seam itself is fine and is the point of the test) (B) |
| 💡 | ⚪ | `lib/page/ai_assistant/providers/router_command_provider.dart:20` | Context builder lives in a file named for the *command* provider, and drags a widget harness into a previously pure unit test (A) |
| 💡 | ⚪ | `test/page/ai_assistant/providers/router_command_provider_test.dart:7-27` | `_BuilderCaptor` is a reusable shape ("value must outlive its obtaining widget") — promote to `test/util/` if a second case appears (B) |

> **⚖️ The two Standards reviewers disagree on the first row and this is left to you, not adjudicated here.** Reviewer A would change it before merge (hoisting the build to the top of `_processConversation` is one line, satisfies "fresh at the time of asking" exactly, and costs *less* than the current code). Reviewer B recommends approving as-is and treats the loop only as an untested path. Both agree it is not Critical.

#### Spec

Spec source: #1354

|---|---|---|---|
| ⚠️ | 🟡 | "cache hits still occur on later requests in a session" | [Omission] AC3's second clause is asserted structurally but never behaviourally — nothing checks cache-read tokens across two requests; left to manual token-log observation |
| ⚠️ | 🟡 | "// proposed\nfinal String Function() _buildRouterContext;" | [Scope creep] Public `refreshContext()` also deleted — not authorised by the spec, but a natural consequence of the fix and zero call sites remain |
| ⚠️ | 🟡 | "Call sites become `() => buildRouterContext(ref.read)`" | [Wrong] PR routes through a new `routerContextBuilderProvider` instead of the literal inline form — a justified *correction* to the issue (a `WidgetRef` closure would throw), but the issue text should be amended to match |
| 💡 | 🟢 | "a question about device count is answered with the current number" | [Omission] AC2 is exercised by mutating a stub `String`, not a device provider — mechanism proven, end-to-end chain covered only in two halves |

<sub>Confidence: 🟢 High = evidence chain closed in head code · 🟡 Med = plausible, one link inferred · ⚪ Low = worth a second look. `[X]` = raised by one agent, `[XX]` = by both — **Standards table only** (Spec has a single reviewer, so agreement markers carry no meaning there).</sub>

<details>
<summary>⚠️ Standards — Warning evidence</summary>

**W1 — rebuilt per *round*, not per *request* · `router_chat_controller.dart:338` · 🟢High**

```dart
    while (loopCount < maxLoops) {
      loopCount++;
      ...
      try {
        final response = await _generator.generateWithHistory(
          _state.messages,
          tools: tools.isNotEmpty ? tools : null,
          systemPromptParts: _buildSystemPromptParts(),   // ← line 338, inside the loop
        );
```

`maxLoops = maxRounds`, and every round after the first continues the *same* logical request, replaying the same `_state.messages`. So: round 1 sends summary S1, the model emits a `tool_use` chosen against S1; the `await` yields and an SSE invalidation lands (`lib/page/internet_settings/providers/wan_data_provider.dart:26-27` documents SSE-driven invalidation); round 2 then sends S2 **alongside the round-1 `tool_use`/`tool_result` pair still in history**. The model is asked to interpret a tool result gathered under S1 while its stated ground truth says S2. User-visible shape: a self-contradictory answer inside one turn, or a follow-up tool call with half-stale arguments. The PR's own doc comment (`:81-82`) frames the goal as *"the state the model reads is the state at the time of asking"* — per-round is strictly more aggressive than that. **Fix:** build once at the top of `_processConversation` into a local and pass it at `:338`.

**W2 — closure outlives only the *container*, not any widget · `router_command_provider.dart:21-23` · 🟡Med**

```dart
final routerContextBuilderProvider = Provider<String Function()>((ref) {
  return () => buildRouterContext(ref.read);
});
```

The DartDoc correctly identifies the `WidgetRef`-after-dispose hazard and correctly forbids `autoDispose` — but container lifetime is a *different* lifetime from widget lifetime, and `ref.read` on a disposed container throws exactly the same way. The closure is held by `RouterChatController`, a plain `ChangeNotifier` created imperatively at `router_assistant_view.dart:117` and `:237`, and `_RouterAssistantViewState.dispose()` (`:348-355`) disposes four `TextEditingController`s and never the controller. If the root `ProviderScope` is torn down (logout, scope swap) mid-`await`, the next round throws. The new test (`router_command_provider_test.dart:88-105`) keeps `ProviderScope` alive across the pump, so it proves widget-disposal survival only. Today this degrades gracefully **only because** the sole call site at `:338` sits inside the `try` opened at `:333` — nothing at the provider or at `_buildSystemPromptParts()` records that dependency, so the next caller outside a `try` reintroduces it. **Fix:** state the real contract ("callable for the lifetime of the owning container; throws if disposed") and either guard the closure or make `_buildSystemPromptParts()` fall back to an empty/last-known summary.

**W3 — a provider-init throw becomes a non-retryable dead end · `router_chat_controller.dart:203-208` · 🟡Med**

```dart
  List<SystemPromptPart> _buildSystemPromptParts() {
    final routerContext = _routerContextBuilder();
    aiLogSensitive(() => 'Router context:\n$routerContext');
    return RouterSystemPrompt.buildParts(routerContext: routerContext);
  }
```

`buildRouterContext` performs ~9 `read(...)` calls against `AsyncNotifierProvider`s (`usp_command_provider.dart:833, 850, 868, 896, 915, 931, 944, 962`); a `read` on an uninitialised one triggers a synchronous build. Previously this ran once, in the constructor, on a synchronous UI path. Now a synchronous throw propagates out at `:338` into the generic handler at `:474`, surfacing as `'An unexpected error occurred: $e'` with `retryable` unset. Med rather than High because the code uses `getValue<T>`/`valueOrNull` (tolerant of `AsyncError` *states*) and no concrete synchronously-throwing provider was traced in this diff.

**W4 — the cost claim is load-bearing and inaccurate · `router_chat_controller.dart:86-88` · 🟢High**

> *"Building it per request is close to free — it reads Riverpod caches already in memory and touches no device."*

True only for already-initialised providers. `portForwardingDataProvider`, `firewallDataProvider` and `ethernetDataProvider` are plausibly cold when the user opens the assistant without visiting those pages, and their `build()` is what issues the USP/JNAP fetches. Combined with W1 (per-round), the first conversation of a session can trigger device round-trips it did not before. This comment is the stated justification for rebuilding, so either it or the code should move.

**W5 — hot-path promotion of a long method and a full sensitive dump · `usp_command_provider.dart:821-972`, `router_chat_controller.dart:206` · 🟡Med**

`buildRouterContext` is ~150 lines across 9 sections (systemInfo/WAN/devices/extenders/wifi/LAN/ethernet/firewall/portForwarding) with ~20 `_log(...)` calls per invocation — a pre-existing **Long Method**, now moved from once-per-session to once-per-round. `aiLogSensitive(() => 'Router context:\n$routerContext')` goes with it, so debug builds print the whole network summary (model, WAN IP, SSIDs, every device name) up to `maxRounds` times per turn. Release builds are unaffected: `lib/ai/ai_logging.dart:78-79` gates on `const kDebugMode && _sensitiveLoggingAllowed`, so the branch is tree-shaken out — **not** a security finding, but it will drown AI-log triage. Consider dumping in full on the first build of a turn and a hash/length thereafter.

**W6 — the freshness test cannot distinguish the two behaviours · `router_chat_controller_test.dart:93-131` · 🟡Med**

`'each request carries the summary as it is at that moment'` mutates `routerContext` between two `sendMessage` calls, both stubbed with `_textResponse('ok')` — zero tool use, so each call runs exactly one loop iteration. It therefore verifies per-*turn* freshness and would pass identically under W1's suggested fix *and* under a bug where the builder fires at the wrong point of a multi-round exchange. Stubbing a `tool_use` for round 1, mutating `routerContext`, then asserting what round 2 received would pin the intended semantics — and would force the W1 decision to be made explicitly.

**W7 — public API removal without a note · removed `refreshContext()` · 🟢High**

```dart
-  /// Refresh router context (call when dashboard data changes).
-  void refreshContext() {
-    _initializeSystemPrompt();
-  }
```

`grep -rn "refreshContext" lib test` returns **zero** hits at head (independently re-verified for this review), so nothing breaks in-tree and the removal is correct — per-request rebuild makes manual refresh meaningless. Flagged only because `RouterChatController` is a public class and this is a silent breaking change in a commit labelled `fix(ai)`. One line in the class DartDoc ("context is rebuilt per request; there is nothing to refresh") closes the loop for anyone arriving from the old API.

**W8 — novel pattern, unenforced invariant · `router_command_provider.dart:21-23` · 🟡Med**

A `Provider<String Function()>` returning a closure that captures `ref` is a deliberate lifetime-extension of a Riverpod `Ref` past the framework's normal scoping, and adds no business logic of its own. The "must not be `autoDispose`" line is a hand-written invariant with nothing enforcing it — both new tests use a live container, so adding `.autoDispose` for tidiness breaks it silently. Two lower-risk shapes were not discussed: pass the `ProviderContainer`/`Ref` explicitly to the controller (making the lifetime dependency visible in the signature), or keep `Provider<String>` and `ref.refresh` it per request. If both were rejected, that belongs in the DartDoc.

**W9 — rationale duplicated · `router_chat_controller.dart:81-88` + `:191-202` · ⚪Low**

The 8-line block on the constructor and the 12-line block on `_buildSystemPromptParts()` tell overlapping halves of the same argument, and a reader looking at the *class* doc sees neither. One canonical explanation plus a one-line pointer.

</details>

<details>
<summary>🔎 Spec — findings vs issue #1354</summary>

**S-W1 — AC3's second clause is untested · 🟡Med**

Spec: `- [ ] Static instructions still report as cached (cache: true), and cache hits still occur on later requests in a session`

First clause is covered directly at `test/ai/orchestrator/router_chat_controller_test.dart:135-157`:

```dart
  expect(parts.first.cache, isTrue);
  expect(parts.first.text, isNot(contains('TestRouter')),
      reason: 'the cached part must not carry the volatile summary');
  expect(parts.skip(1).every((p) => p.cache == false), isTrue);
```

Second clause is asserted nowhere — `grep -rn "cache" test/ai/orchestrator/router_chat_controller_test.dart` returns only those lines plus an unrelated command-registry comment at `:882`; nothing inspects `response.usage` cache-read tokens across two requests. The implementation comment concedes it (`router_chat_controller.dart:200-202`: *"cache hits are visible in the per-response token logs"*). Not Critical: cache-hit behaviour is a property of the remote provider, the structural precondition the issue itself calls sufficient ("caching matches on prefixes") *is* tested, and a mocked generator cannot observe a real Bedrock cache hit. Please state whether a cache hit was confirmed manually on a live session, since the issue lists it as an explicit checkbox.

**S-W2 — `refreshContext()` removal is unauthorised but consequential · 🟡Med**

The spec's "Proposed fix" describes exactly one signature change (`final String _routerContext` → `final String Function() _buildRouterContext`). The PR additionally deletes a public method and the `_initializeSystemPrompt()` call from the conversation-reset path. Judged **on the correct side** of the scope-creep line — an eager-refresh hook is meaningless once the summary is built per request, and keeping it would be dead code — and grep confirms zero call sites, so no contract observably in use breaks. Flagged only because it is technically a public-method removal the spec did not ask for.

**S-W3 — deviation from the literal proposed call-site code · 🟡Med**

Spec: *"Call sites become `() => buildRouterContext(ref.read)`, and the call moves to where `RouterSystemPrompt.buildParts()` is invoked."*

The PR instead adds `routerContextBuilderProvider` (`router_command_provider.dart:20-23`) and call sites read it. This is a **necessary correction to the issue, not scope creep**: the literal inline form written in the View would capture a `WidgetRef`, and the builder now runs inside the `await`-spanning loop at `router_chat_controller.dart:321-338`, so a later round can fire after the View is disposed — where `WidgetRef.read` throws. The PR documents the reasoning and pins it with a disposal-survival regression test. Recorded only so the **issue text can be amended**; otherwise future readers see PR and issue contradicting each other.

**S-S1 — AC2 covered at stub level · 🟢High**

Spec: *"With the chat left open across a change in connected devices, a question about device count is answered with the current number."* The standing test reassigns a plain `String` (`router_chat_controller_test.dart:92-131`) rather than mutating the underlying device provider and letting `buildRouterContext` re-read it. Mechanism proven; combined with the existing `buildRouterContext` provider tests (`test/ai/providers/usp_command_provider_test.dart:267+`) the chain is covered in two halves. One container-level test that changes a device provider and asserts the emitted summary's count changes would close AC2 literally.

**Requirements met** (11 fully met / 13 · 2 partial · 0 unmet · 0 genuine scope creep)

1. ✅ Root cause removed — `final String _routerContext` is gone, replaced by `final String Function() _routerContextBuilder` (`router_chat_controller.dart:110`).
2. ✅ Signature changed to the proposed callback shape (`:89-95`).
3. ✅ Call moved to where `RouterSystemPrompt.buildParts()` is invoked (`:203-209`, invoked at `:338`).
4. ✅ Stale prompt field eliminated — `List<SystemPromptPart>? _systemPromptParts` and `_initializeSystemPrompt()` both deleted; nothing holds a snapshot.
5. ✅ Call site 1 converted — `router_assistant_view.dart:120` (first connect).
6. ✅ Call site 2 converted — `router_assistant_view.dart:240` (after entering credentials); grep finds no snapshot-style `routerContext:` call site left in `lib/`.
7. ✅ AC1 — summary reflects provider state at request time, not construction time; asserted by the freshness test.
8. 🟡 AC2 — mechanism satisfied, coverage at summary-string level rather than provider level (S-S1).
9. ✅ AC3 first half — static instructions still `cache: true` and ordered first, and the cached part provably carries no volatile text (`:153-156`).
10. 🟡 AC3 second half — cache hits on later requests not asserted (S-W1).
11. ✅ AC4 — a test asserts two requests separated by a state change carry different summaries (`:92-131`, via `captureAny(named: 'systemPromptParts')`).
12. ✅ Out of scope respected — tools-vs-summary instruction text untouched.
13. ✅ Out of scope respected — no freshness timestamp added; `buildRouterContext` unmodified.

</details>

<details>

- **Root cause removed, not patched.** Deleting the field, `_initializeSystemPrompt()` and `refreshContext()` eliminates a *class* of bug instead of guarding one instance of it. The DartDoc at `:191-197` names exactly why ("a field would have to be invalidated from every path that starts a request") — removing an invalidation obligation beats satisfying it, and it retires a latent Shotgun Surgery hazard. Grep confirms no orphaned callers.
- **The prompt-cache objection was anticipated, verified, and locked down.** `lib/ai/prompts/router_system_prompt.dart:65-83` genuinely puts the static instructions in their own `cache: true` part ordered first with the volatile summary in a `cache: false` part after it, so the per-request rebuild really does preserve the cached prefix — and the new test pins the invariant a future refactor would otherwise break silently. Concern raised, answered, tested; nothing left for a reviewer to do.
- **The disposed-`WidgetRef` hazard was found before review, not during it.** The `_BuilderCaptor` test (`router_command_provider_test.dart:11-32, 88-105`) captures the builder in `initState`, disposes the widget, and asserts the closure still works — a genuinely non-obvious failure mode for a callback that escapes a widget's lifetime, with the "must not be `autoDispose`" reasoning recorded so the next person does not undo it.
- **Layering is cleaner than before.** `views` → `providers` → `ai` only; no generated-code imports in the touched `providers/`/`views/` files (Art. V §5.4 clean). The view no longer knows `buildRouterContext` exists.
- **The type carries the design.** `String routerContext` → `String Function() routerContextBuilder` makes "this is re-read" visible at every call site, so a caller cannot accidentally revert the freshness property. Types over comments.
- **No USP/JNAP error-handling regression.** No `on Exception catch` introduced; the new throw site lands in the pre-existing generic `catch (e)` at `:474`, which is the correct shape here given USP errors arrive as raw `String`s.
- **No new egress surface.** The summary already went to Bedrock on every `generateWithHistory` call before this PR; the change alters *what* is in that part, not *whether* it leaves the device.
- **Tests document the bug they prevent**, names are descriptive with no `TC00x` numbering, Mocktail used correctly with `captureAny(named:)`, and placement mirrors `lib/`.
- **Scope discipline.** One commit, 5 files, one behavioural change, no drive-by fixes; net-negative in the controller (+30/-22) with +136 lines of test.

</details>

<sub>Note on diff scope: `git diff --stat` between the review shas reports 13 paths (topology fixtures, `pubspec.yaml`, `detail_widgets.dart`, three deleted test files) while `git log` shows a single commit. Those extra paths are base-side drift, not PR content; this review covers only the 5 files in the PR's change set.</sub>

